PR #13, opened 2018-10-26
Updated UI to use Sensor Fusion filtered velocity for actual speed

maefromm:


maefromm:
Discussed with Mike and Shuwei, that SF filtered velocity would be the best topic to use for actual speed instead of can/speed. Need to integration test on Monday before assigning someone to review this PR.

maefromm:
Tested. Crusing Plugin and Traffic Signal Plugin now shows actual speed.

PR #23, opened 2018-10-31
Fix/ead plan interpolation timestamps

msmcconnell:
Adds the planning start time and start distance as offsets for nodes when comparing to roadway obstables for collision checks in TrafficSignalPlugin


qswawrq:
Looks fine for me.

PR #26, opened 2018-11-03
Test #3 on acknowledgements file.

VetteDriver:
Another change using default behavior with SourceTree on my remote (no privileges) account.


VetteDriver:
Aborting, as this was still created by TonysCousin account.  I don't like that SourceTree doesn't tell you which account you're working in.

PR #29, opened 2018-11-05
Fix west bound route waypoint locations to be in center of lane

msmcconnell:
As the title says just shifting the route points to be more accurate.


qswawrq:
Let's hold this branch and keep testing on it before we update the develop branch.

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/ead/PlanInterpolator.java
```diff
@@ -68,23 +69,33 @@
       double v_old = v_0;
       double v = Math.sqrt(v_old * v_old + two_a_x);
       double t;
+      double x = x_0 + distanceStep;
```
qswawrq: In line 63, after we added the node directly, should we just go to the next node?

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/ead/PlanInterpolator.java
```diff
@@ -68,23 +69,33 @@
...
-      if (a == 0.0) {
+      if (Math.abs(a) < MIN_ACCELERATION) {
         t = t_0 + distanceStep / v; // With constant speed 
+
+        // Interpolate between current node and previous node
+        while (x < x_f) {
+
+          points.add(new RoutePointStamped(x, 0, t)); // Add previous node and the new interpolated nodes to list
+          v_old = v;
+          v = Math.sqrt(v_old * v_old + two_a_x);
```
qswawrq: If Math.abs(a) < MIN_ACCELERATION, why we do not assume v is equal to v_old?

PR #35, opened 2018-11-08
Configured CARMA Deployment in Docker

Supraluminal:
This adds Dockerfiles to create CARMA images for Docker as well as several shell scripts that are necessary or very useful in the process of utilizing Docker.

**Notable Files:**
* `Dockerfile` - Primary Docker image construction mechanism, specifies all dependencies and build steps to create a fully operational CARMA system from a baseline ROS install
* `docker/Dockerfile.clean_build` - Same as above except instead of copying the local CARMAPlatform files it clones them from Github every time. Use the `--build-arg CARMA_VERSION=<VERSION>` to specify the branch/tag/commit to checkout.
* `engineering_tools/install_docker.sh` - Installation script for docker-ce
* `engineering_tools/run_carma.sh` - Script to start CARMA after the docker image has been created. Primarily responsible for setting the `docker run` command line arguments to remap files and configure networking.
* `docker/entrypoint.sh` - Start script run inside the docker container
* `docker/install.sh` - Analog to remote_install, except local.

**Procedure to Test**
1. Run `install-docker.sh` on local machine
2. Log out and back in to allow new usergroup settings to take effect
3. cd into CARMAPlatform
4. Run: `docker build . -t carma --build-arg SSH_PRIVATE_KEY="$(cat ~/.ssh/id_rsa)" --build-arg EXTRA_PACKAGES=git@github.com:fhwa-saxton/CARMASensitive.git` (id_rsa will not be preserved in final image due to multi-stage build process, we're not publishing these images for now anyway)
5. Run `./engineering_tools/run-carma.sh <RELATIVE_PATH_TO_CONFIGURATION_FOLDER>` where the config folder is cadillac_config, etc.
6. You should see some sort of hash output as a "PID" of sorts for that container instance. To see inside that container run `docker attach carma`. You should see a new TTY containing CARMA running inside a TMUX instance. To detach and leave it running press `Ctrl+P, Ctrl+Q`, to kill it press `Ctrl+C`. To kill it while it's in the background run `docker kill carma`, it will print the name of the killed container if successful.


.dockerignore
```diff
@@ -0,0 +1,44 @@
+.cache
```
msmcconnell: Could maybe have a top level comment for this file

.dockerignore
```diff
@@ -0,0 +1,44 @@
...
+/*.tar.gz
+*.class
+**/bin
+
+# ignore build directories
+**/build
+**/target
+**/devel
+
+# ignore Intellij project files and dirs
```
msmcconnell: Are we going to need to add checks like this for all our Ide's (Intellij, Eclipse, VSCode, etc.)

Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+ENV USERNAME carma
+RUN useradd -m $USERNAME && \
+        echo "$USERNAME:$USERNAME" | chpasswd && \
+        usermod --shell /bin/bash $USERNAME && \
+        usermod -aG sudo $USERNAME && \
+        mkdir -p /etc/sudoers.d && \
+        echo "$USERNAME ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers.d/$USERNAME && \
+        chmod 0440 /etc/sudoers.d/$USERNAME && \
+        usermod  --uid 1000 $USERNAME && \
+        groupmod --gid 1000 $USERNAME
```
msmcconnell: We are going to need to add or modify this user setup if we choose to launch the system using the current remote launch process.

Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+
+# Install necessary packages
+RUN apt-get update
+RUN apt-get install -y sudo ros-kinetic-rosjava ros-kinetic-rosbridge-server ros-kinetic-robot tmux apache2 php7.0 libapache2-mod-php7.0 vim nano less
+
```
msmcconnell: might be worth having a rosdep update call here as you do in line 66 above, since this new container has reinstalled rosjava, rosbridge, and robot

Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+#
+# Performs all necessary tasks related to generation of a generic CARMA docker image
+# suitable for deployment to a vehicle. The generic image will still need to be 
+# configured for each vehicle by means of volume mapping configuration files and
+# networking mapping at run time
+#
+# Build Arguments:
+# SSH_PRIVATE_KEY - If the extra package repository to be used during the build requires
+#                   authentication, please pass in the necessary SSH private key in text
+#                   form here, most likely via "$(cat ~/.ssh/id_rsa)". This data is not
```
msmcconnell: Is it possible to have the user just pass in ~/.ssh/id_rsa and the script would apply the $(cat )

docker/install.sh
```diff
@@ -0,0 +1,44 @@
...
+#  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#  License for the specific language governing permissions and limitations under
+#  the License.
+
+
+# Build the software and its dependencies
+source /opt/ros/kinetic/setup.bash
+cd ~/carma_ws
+rosdep install --from-paths src --ignore-src --rosdistro=kinetic -y
+catkin_make install
```
msmcconnell: We need to scrub the private sensitive launch files for can id's. I believe the launch files will be installed in this step and they may have can id's as parameters.

engineering_tools/deploy-image.sh
```diff
@@ -0,0 +1,34 @@
...
+# 
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#  License for the specific language governing permissions and limitations under
+#  the License.
+
+# Docker deployment script for CARMA
+#
+# Loads a docker image in the local registry to the remote target
```
msmcconnell: Probably should call out the use of pv in the comments

.dockerignore
Supraluminal: We can, yeah, I just copied in our .gitignore as a base so if we have other editors that are worth adding here we can certainly do so.  Nominally one might have those in their personal global gitignore file but redundancy is good in this case. I think this can be grown as needed though, don't need to take care of that under the scope of this PR.

Dockerfile
Supraluminal: That's a good point that we'll need to look into at that phase, we'll need at least another user right? We can set it up to mimic the way things are done currently in the vehicle if that makes sense or we can change the way it's currently run in the remote launch to better fit the docker paradigm, either way.

Dockerfile
Supraluminal: I thought the `rosdep update` was the equivalent of `apt-get update`, eg just fetching a new registry. of available packages in repos. This stage of the build doesn't use `rosdep` to install anything so I'm not sure that's needed.

Dockerfile
Supraluminal: I can give it a quick test, but I don't know if that's possible. The docker build process actually runs as a daemon under it's own user, so it shouldn't have access to `~/.ssh/id_rsa` normally as that's `600` permissions.

Supraluminal: Actually, testing it out, the docker build process only has access to files at or below the Dockerfile in the folder structure, so the build process can't even see the user's home folder. `cat`ing the key is kind of a work around hack I guess

docker/install.sh
Supraluminal: Prior to publication, I agree. We'd probably just want to install a basic non-sensitive version and map our specific launch files in at run time, based on vehicle type.

msmcconnell:
We are tracking the discussion points in Jira, so I will merge this PR and consider remaining changes in the scope of future work.

PR #38, opened 2018-11-09
Fix the Traffic Signal Plugin (TSP) & widget when Stopped

maefromm:
Fix the issue when stopped and TSP plugin asks for Operator if it's safe to continue. The updates to logic are:
- When operator says it's not yet safe to continue, the service call will return false with a message "When it's safe to continue, please press the YES button." Removed the wait times and unecessary variables. 
- Updated UI for the icons to be smaller to fit the tablet screen. Used White Cadillac's tablet to fit properly. 
- Updated UI popup to add message field in the footer to display the service failure message properly.
- Updated TSP widget to remove unnecessary logging.

### DIFF
--- a/carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/TrafficSignalPlugin.java
+++ b/carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/TrafficSignalPlugin.java
@@ -112,8 +112,11 @@ public class TrafficSignalPlugin extends AbstractPlugin implements IStrategicPlu
     private ISubscriber<RoadwayEnvironment> obstacleSub;
     private IPublisher<TrafficSignalInfoList> trafficSignalInfoPub;
     private IPublisher<UIInstructions> uiInstructionsPub;
-    private AtomicBoolean awaitingUserInput = new AtomicBoolean(false);
-    private AtomicBoolean awaitingUserConfirmation = new AtomicBoolean(false);
+
+    private AtomicBoolean awaitingUserInput = new AtomicBoolean(false); //Used to track if the user has provided an input on UI popup.
+    private AtomicBoolean awaitingUserConfirmation = new AtomicBoolean(false); //Used to track when
+    private AtomicBoolean userRequestedToWait = new AtomicBoolean(false); //Used to track when driver said NOT to continue on UI popup.
+    private final long WAIT_INTERVAL = 5000;
     private AtomicLong prevUIRequestTime = new AtomicLong();
     private final long UI_REQUEST_INTERVAL = 100;
     private final String GO_BUTTON_SRVS = "/traffic_signal_plugin/go_button";
@@ -147,6 +150,7 @@ public class TrafficSignalPlugin extends AbstractPlugin implements IStrategicPlu
     private double defaultAccel;
     private GlidepathAppConfig appConfig;
 
+
     public TrafficSignalPlugin(PluginServiceLocator psl) {
         super(psl);
         version.setName("Traffic Signal Plugin");
@@ -177,7 +181,7 @@ public class TrafficSignalPlugin extends AbstractPlugin implements IStrategicPlu
         IManeuverInputs platformInputs = pluginServiceLocator.getManeuverPlanner().getManeuverInputs();
         pluginManeuverInputs = new TrafficSignalManeuverInputs(platformInputs, appConfig.getDoubleValue("ead.response.lag"), appConfig.getDoubleValue("defaultAccel"));
         pluginManeuverPlanner = new ManeuverPlanner(pluginServiceLocator.getManeuverPlanner().getGuidanceCommands(), pluginManeuverInputs);
-        
+
         acceptableStopDistance = appConfig.getDoubleDefaultValue("ead.acceptableStopDistance", 6.0);
 
         twiceAcceptableStopDistance = 2.0 * acceptableStopDistance;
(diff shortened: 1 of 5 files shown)


carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/TrafficSignalPlugin.java
```diff
@@ -192,40 +196,80 @@ public void onInitialize() {
...
+                        response.setSuccess(true);
+                        return;
+
+                    } else {//Else no longer green.
+
+                        log.info("Current Phase is NO LONGER GREEN after user confirmed");
+                        log.info("GO_BUTTON_SRVS 5: awaitingUserConfirmation: " + awaitingUserConfirmation + "; awaitingUserInput:" + awaitingUserInput);
+
+                        response.setMessage("No longer green. Please wait to respond until the next green light."); //TODO: TEST
+                        response.setSuccess(true);  //Need to set to false in order to leave the popup open instead of success which the popup disappears.
```
msmcconnell: Comment says this should be false but it is being set to true

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/TrafficSignalPlugin.java
```diff
@@ -192,40 +196,80 @@ public void onInitialize() {
...
+
+                        //Set true since popup remains open.
+                        awaitingUserInput.set(true);
+                        awaitingUserConfirmation.set(true);
```
msmcconnell: Not sure you need to set these variables since they must be true for you to enter this check.

Also are you confident compareAndSet is not needed to prevent race conditions with the loop()?

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/TrafficSignalPlugin.java
```diff
@@ -192,40 +196,80 @@ public void onInitialize() {
...
+
                         return;
                     }
-                } else { // User indicated it is not safe to continue
+
+                } else { // Else user indicated it is not safe to continue
+
                     log.info("User said it is not safe to continue");
-                    prevUIRequestTime.set(System.currentTimeMillis()); // TODO should this be ROS time
+                    response.setMessage("If not safe to continue, please wait for " + (WAIT_INTERVAL/1000) + " seconds...");
```
msmcconnell: This print will display an incorrect value if WAIT_INTERVAL is ever set to a non full second value (like 2.5 due to integer truncation).

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/TrafficSignalPlugin.java
```diff
@@ -192,40 +196,80 @@ public void onInitialize() {
...
+                    response.setSuccess(true); // If we reach this point the service request has been handled
+
+                    // Update the time since user did respond at this time, instead of expiring sooner.
+                    //TODO: should this be ROS time
+                    prevUIRequestTime.set(System.currentTimeMillis());
```
msmcconnell: Any time which will be compared with timestamps received from external components (pub/sub etc.) needs to use the TimeProvider which can be accessed with the PluginServiceLocator. If it will only be compared internally, it is ok to use System time, but it needs to be consistent do not mix System and TimeProvider.

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/TrafficSignalPlugin.java
```diff
@@ -634,33 +678,82 @@ public void loop() throws InterruptedException {
 
     /**
      * Helper function to evaluate state variables for the stopping condition
+     *
      */
     private void evaluateStatesForStopping() {
```
msmcconnell: Please add a comment both in the function description and in the loop function which describes that this method can result in Thread.Sleep being called

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/TrafficSignalPlugin.java
```diff
@@ -634,33 +678,82 @@ public void loop() throws InterruptedException {
...
+
+            log.info("evaluateStatesForStopping 1: New Request: awaitingUserConfirmation: " + awaitingUserConfirmation + "; awaitingUserInput:" + awaitingUserInput);
+        }
+        // If stopped & light is green/green soon, & still waiting for YES confirmation but NOT user input, & enough time has elapsed
+        else if (stoppedForLight
+          && phaseIsGreenSoon
+          && awaitingUserConfirmation.get()
+          && awaitingUserInput.compareAndSet(false, true)
+          && System.currentTimeMillis() - UI_REQUEST_INTERVAL > prevUIRequestTime.get())
+        {
```
msmcconnell: I'm a bit unclear on what this time check is for compared to the wait time above. It seems like they are both geared towards accomplishing the same thing

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/TrafficSignalPlugin.java
maefromm: please check latest code changes.

PR #39, opened 2018-11-09
Fix/roadway environment velocity

msmcconnell:
This fixes the velocity value on the objects in roadway_environment. Additionally, sensor fusion velocity measurements now account for rotations between the radar and base_link frames. 

A unit test for the object collision checker is also fixed.


carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/ObjectCollisionChecker.java
```diff
@@ -140,7 +140,7 @@ public void updateObjects(List<RoadwayObstacle> obstacles) {
       // Add it to the set of tracked object histories
       // TODO this currently does not handle if the lane index changes between the host and detected object
       if (inLane && frontObjectDistToCenters > -0.0) {
-        
+        obs.getObject().setId(0);
```
qswawrq: Let add a TODO here to explain.

msmcconnell: Good call

msmcconnell:
Addressed comments

PR #40, opened 2018-11-09
Make changes on NCV prediction

qswawrq:


carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/ead/SimpleNCVMotionPredictor.java
```diff
@@ -51,57 +33,37 @@
...
         t += 0.1;
         projection.add(new RoutePointStamped(d, 0, t)); // Assume that the vehicle is stationary
       }
     } else { // Vehicle is in motion so use distance steps
-      while (d < endDistance) {
+      while (d < endDist) {
 
         d += distanceStep;
-        projection.add(new RoutePointStamped(d, 0, d * slope + intercept));
+        projection.add(new RoutePointStamped(d, 0, d + (distanceStep / averageSpeed)));
```
msmcconnell: Shouldn't it be 
t + (distanceStep / averageSpeed)
and t needs to be updated as the loop runs

Also you could factor out distance / averageSpeed into a final variable called timeStep

qswawrq: Ah. Good catch...

PR #43, opened 2018-11-12
Darrell updates

darrelld05:
updates to the Contributing.md file


Contributing.md
```diff
@@ -0,0 +1,97 @@
...
+* This branch contains production-ready code of CARMAPlatform.
+* This branch should be kept stable at all times
+
+### develop
+* This branch contains code for the upcoming release.
+* This branch should be kept stable at all times. This is important because new branches are created off of this branch.
+
+### feature
+* These branches are created from develop; **each feature branch is used to implement a single task.** After completing the task, the feature branch must be merged into develop.
+* To avoid ambiguity, we use a specific naming convention for feature branches − they always begin with feature/ followed by a description based on the functionality implemented by the feature. For example, **feature/sign-up-with-email-and-password.**
```
TonysCousin: Example needs to show an issue # in the name as well.

Contributing.md
```diff
@@ -0,0 +1,97 @@
...
+* Thanks to this naming, team members can easily tell what code each branch contains.
+
+### release
+* This branch is created every iteration (i.e. sprint) from develop, and when the team rolls out a release it’s deployed to the staging server for testing. A stable release is merged first into the master branch and then into develop.
```
TonysCousin: We don't have a staging server.  And we can't guarantee that we will have a release after every sprint.  so I suggest replacing this description with:
A release branch is created whenever a set of functionality is ready for production release. It is normally created from develop and is used to keep the contents stable for final acceptance testing while other development activities may continue on other branches. Once the release is accepted, it will be merged into the master branch and versioned, then merged into develop.

Contributing.md
```diff
@@ -0,0 +1,97 @@
...
+* The naming convention for this branch starts with release/ followed by its version. For example, **release/v1.0.1.**
```
TonysCousin: We had talked about using codenames for release branches, since we don't know how long it will take to release one, which can screw up the numbering convention.  I don't mind leaving this example as is, but I don't want to use version numbers for these branches in real life.

Contributing.md
```diff
@@ -0,0 +1,97 @@
...
+
+### hotfix
+* This branch is created for handling emergency situations – it allows developers to quickly fix something in production. This branch uses master as the parent branch and merges into both master and develop.
+* The name of this branch starts with hotfix/ followed by its version. For example, **hotfix/v0.1.1.**
```
TonysCousin: Again, I don't want to use version numbers in branch names.  Two hotfixes may overlap and we would want to release them in reverse order, but such naming would prevent us from doing that.  Otherwise, I'm fine with this description.

Contributing.md
```diff
@@ -41,22 +41,24 @@ Welcome to the CARMAPlatform contributing guide. There are a few guidelines that
...
 #### Create a New Branch
 **The number one rule is to put each piece of work on its own branch.**
-* Create a branch specific to the issue you are working on. Read the article ["Creating and deleting branches"](https://help.github.com/articles/creating-and-deleting-branches-within-your-repository/) on GitHub. 
-* For enhancements, name the branch according to the feature e.g. **`feature/auto-activate`.**
-* For un-reported bug fixes, add a **`fix-`** prefix e.g. **`fix-admin-notices`.**
-* For code that addresses an existing Issue, add the Issue number as a prefix e.g. **`feature/123-auto-activate`** or **`feature/321-fix-admin-notices`.**
+* Check out the branch you want to make modifications to (normally this is the **`develop branch`**).
+* Create a branch specific to the issue being addressed. Read the article ["Creating and deleting branches"](https://help.github.com/articles/creating-and-deleting-branches-within-your-repository/) on GitHub. 
+* For enhancements, name the branch according to the feature e.g. **`feature/123-auto-activate`.**
+* For un-reported bug fixes, add a **`fix-`** prefix e.g. **`fix/321-admin-notices`.**
```
TonysCousin: Let's avoid having changes to un-reported problems by asking them to write an issue first, then using that issue number in the branch name like the example shows.

Contributing.md
```diff
@@ -0,0 +1,97 @@
...
+In order to make the development process efficient, please comply with the branching model described below. On this model, we mainly use five branches - master, develop, feature, release, and hotfix.
+
+![A successful Git branching model](docs/image/Git_Workflow.png)
+
+### master
+* This branch contains production-ready code of CARMAPlatform.
+* This branch should be kept stable at all times
+
+### develop
+* This branch contains code for the upcoming release.
```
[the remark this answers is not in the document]
TonysCousin: Still looking for a change to line 82 per above comment.

PR #56, opened 2018-11-15
Fix/refactor conflict manager

### LINKED ISSUE #55: Conflict Manager should be refactored to grant plugins more options for tuning
### Descriptive summary
The conflict detector system in Guidance does not allow for plugins to control the parameters such as spatial hash map cell size. This makes it difficult to use the system for purposes other than the original lane change use case it was designed for. The system would better serve users if alternative spatial structures or those with differing dimensions could be injected. To this end IConflictDetector.java and ConflictManager.java could be refactored to support this. In the short term this will directly benefit Traffic Signal Plugin development
### Expected behavior
Conflict detection system in guidance can be used for a variety of collision avoidance use cases
### Actual behavior
The limitations on parameter tuning prevent plugins from modifying the spatial hash map cell size which directly affects collision lookup performance. 
N/A

### Related work
N/A

msmcconnell:
Fixes #55  ; refs #54

Addresses Issue 55 by refactoring the conflict manager system in guidance to support dependency injection of different collision checking structure objects. This allows a plugin or other guidance component to use it's own parameters for collision checking while still levering the logic in ConflictManager

* Create an ISpatialStructureFactory for making different collision checking structures
* Refactor ConflictManager.java to use the ISpatialStructureFactory and ISpatialStructure interface instead of the NSpatialHashMap directly. Different ISpatialStructure objects can then be injected
* Rename spatialhashmap package to spatialstructure package to better reflect its purpose.
* Fix all unit tests based on api changes
* Update toString method in CartesianObject to be more verbose
* Update Traffic Signal Plugin to support this change

@qswawrq

### DIFF
--- a/carmajava/geometry/src/main/java/gov/dot/fhwa/saxton/carma/geometry/cartesian/spatialstructure/NSpatialHashMapFactory.java
+++ b/carmajava/geometry/src/main/java/gov/dot/fhwa/saxton/carma/geometry/cartesian/spatialstructure/NSpatialHashMapFactory.java
@@ -21,21 +21,22 @@ import java.util.HashMap;
 import gov.dot.fhwa.saxton.carma.geometry.cartesian.AxisAlignedBoundingBox;
 
 /**
- * Factory/builder pattern class for the construction of NSpatialHashMap instances.
+ * Factory pattern class for the construction of NSpatialHashMap instances.
  * <p>
  * Assigns the strategies needed by the NSpatialHashMap instance.
  */
-public class NSpatialHashMapFactory {
+public class NSpatialHashMapFactory implements ISpatialStructureFactory{
+    public final double[] cellDims;
     /**
-     * Build a NSpatialHashMap with the default configuration.
-     * <p>
-     * {@link SimpleHashStrategy} is used to hash objects
-     * {@link AxisAlignedBoundingBox} is used for collision detection
-     * {@link HashMap} is used for the underlying data structure
+     * Constructor
      * 
-     * The returned map is not thread safe on its own
+     * @param cellDims The dimensions of a cell which this will map points to
      */
-    public static NSpatialHashMap buildSpatialHashMap(double[] cellDims) {
+    public NSpatialHashMapFactory(double[] cellDims) {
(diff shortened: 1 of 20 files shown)


carmajava/geometry/src/main/java/gov/dot/fhwa/saxton/carma/geometry/cartesian/spatialstructure/NSpatialHashMapFactory.java
```diff
@@ -14,28 +14,29 @@
...
 
 /**
- * Factory/builder pattern class for the construction of NSpatialHashMap instances.
+ * Factory pattern class for the construction of NSpatialHashMap instances.
  * <p>
  * Assigns the strategies needed by the NSpatialHashMap instance.
  */
-public class NSpatialHashMapFactory {
+public class NSpatialHashMapFactory implements ISpatialStructureFactory{
+    public final double[] cellDims;
```
qswawrq: I think the goal for factory pattern is to create different classes with the same interface with one factory. So it should make more sense for me to have buildSpatialStructure taking in the dimension and returning a NSpatialHashMap object.

msmcconnell: The issue is that not every spatial structure would have the same constructor. If we had an octree there wouldn't be a cell size array to pass in. A kdtree would similarly be a different set of parameters.

msmcconnell: You are right this isn't pure factory pattern (maybe it qualifies as builder). I want dependency injection of the specific type of spatial structure but I need to be able to create multiple structures, thus the interface for the factory instead of just the object

qswawrq: Got you.

PR #57, opened 2018-11-15
Updated version ID to 2.8.1 w/description in release notes.

TonysCousin:
Fixes #issuenumber ; refs #issuenumber

Present tense short summary (50 characters or less)

More detailed description, if necessary. Try to be as descriptive as you can: even if you think that the PR content is obvious, it may not be obvious to others. Include tracebacks if helpful, and be sure to call out any bits of the PR that may be work-in-progress.

Description can have multiple paragraphs and you can use code examples inside:

``` ruby
class PostsController
  def index
    respond_with Post.limit(10)
  end
end
```

* 
* 
* 

@projecthydra/sufia-code-reviewers


darrelld05:
push to master version Version-2.8.1.

darrelld05:
pushed to master

PR #60, opened 2018-11-15
Resolve NaN values coming out of the plan interpolator

### LINKED ISSUE #59: NaN present in objects returned by PlanInterpolator used by Traffic Signal Plugin
### Descriptive summary
The PlanInterpolator.java class can generate RoutePointStamped objects with fields containing NaN values if the is a speed inflection caused by the acceleration. For example if the velocity is heading towards zero, then the vehicle should theoretically start moving backwards. This results in the square root of a negative number being taken causing the NaN
### Expected behavior
No NaN values present
### Actual behavior
NaN value found in time field of some RoutePointStamped objects

1. Do this
Create a unit test like this
```
  /**
   * Tests if the interpolation between two nodes can generate an NaN value
   * @throws Exception
   */
  @Test
  public void testNaNEdgeCase() throws Exception {
    PlanInterpolator planInterpolator = new PlanInterpolator();


    // Empty list will return empty list
    List<RoutePointStamped> points = planInterpolator.interpolateMotion(new ArrayList<Node>(), 0.2, 0, 0);
    assertTrue(points.isEmpty());

    // A single node will return a single point
    Node n1 = new Node(0.0, 0.0, 2.0);
    Node n2 = new Node(84.0, 64.0, 0.0);
    points = planInterpolator.interpolateMotion(Arrays.asList(n1,n2), 2.5, 0, 0);

    for (RoutePointStamped p: points) {
      assertFalse(Double.isNaN(p.getDowntrack()));
      assertFalse(Double.isNaN(p.getCrosstrack()));
      assertFalse(Double.isNaN(p.getStamp()));
    }
  }
```
This will fail to pass on commit c2662c3 on the develop branch, which was eventually rolled into release 2.9.0.

msmcconnell:
Fixes #59 ; 
 This resolves Issue 59 by checking if the new velocity value crosses 0. If it does the value will be capped at 0 rather than the square root being taken

The following check is added in two places to prevent the issue
```
        double v_f_sqr = v_old * v_old + two_a_x;
        if (v_f_sqr <= 0.0) {
          v = 0.0;
        } else {
          v = Math.sqrt(v_f_sqr);
        }
```

Additionally, a unit test was created to replicate the issue and prove out the fix. 

* Prevent negative velocities resulting in NaN values

@qswawrq

### DIFF
--- a/carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/ead/PlanInterpolator.java
+++ b/carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/ead/PlanInterpolator.java
@@ -84,15 +84,26 @@ public class PlanInterpolator implements IMotionInterpolator {
         }
 
       } else {
-        v = Math.sqrt(v_old * v_old + two_a_x);
+        double v_f_sqr = v_old * v_old + two_a_x;
+        if (v_f_sqr <= 0.0) {
+          v = 0.0;
+        } else {
+          v = Math.sqrt(v_f_sqr);
+        }
         t = t_0 + (v - v_old) / a; // With constant acceleration
 
         // Interpolate between current node and previous node
         while (x < x_f) {
 
           points.add(new RoutePointStamped(x, 0, t)); // Add previous node and the new interpolated nodes to list
+
           v_old = v;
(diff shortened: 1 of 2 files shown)


msmcconnell:
This PR will be closed pending further work on Traffic Signal Plugin.

Issue #66, opened 2018-11-20
TrafficSignalPlugin planning time is very long if a NCV present


qswawrq:
### Descriptive summary
During developing TrafficSignalPlugin's new feature to handling NCV, we notices extremely long planning time from the EAD A* algorithm when it tries to handle a detected NCV. The longest record is 16s, which is not acceptable.
### Expected behavior
The EAD algorithm running time should be reasonable. Under 1 second ideally.
### Actual behavior
The EAD algorithm running time varies from 5 seconds to 16 seconds if it tries to avoid the NCV prediction.
### Steps to reproduce the behavior
Run TrafficSignalPlugin with a NCV in front and check the planning time by search "EADAStar"
### Related work
N/A

msmcconnell:
Planner was updated in PR #92 to use Anytime Non-Parametric A* (ANA*) as a means of limiting planning time to a fixed value. This works but performance is still an issue and further improvements are needed.

PR #92, opened 2018-12-05
Update Traffic Signal Plugin with Performance Improvements

msmcconnell:
# PR Details
Adds current improvements to the Traffic Signal Plugin for NCV handling. This should also help address Issue #66  . 
## Description

The old two step A* planner used by the Traffic Signal Plugin has two major issues. First the time based course plan must generate achievable goals for fine fuel plan to achieve. This becomes very difficult in the presence of an NCV. Secondly the planner is very slow when collision checks must be run between each node. This PR attempts to resolve these issues. Testing has shown substantial improvement in both regards, but real world tests have yet to produce smooth and reliable execution.

Sub-Issue 1: Achievable goals
In order for a two phase planner to work all constrains imposed on search 2 must also be imposed search 1. In regards to the NCV this means that collision checks must be run in the time based plan. However, when the host vehicle must plan to stop at the red light it becomes challenging to apply the NCV check because there no defined stopping behavior in the course plan . The exact trajectory is not defined until the fine plan. But our course plan must know that we will stop behind an NCV otherwise the total distance which we must cover after accelerating from a stop will be different. To know we must stop behind the NCV, collision checks must be run. This means that a trajectory must be defined. This becomes a chicken and egg problem. It might be possible to resolve it through sampling of the red light timing, but the relationship between both plans has continually caused problems and many extra collision checks incurs performance penalties. For these reasons we elected to move away from a two plan system. Instead a single planner has been developed which combines fuel and time cost. 

The new cost calculations are as follows

codecogseqn 5

These equations were implemented in the MovesFuelCostModel.java class with some additional edge case checks. 

The goal check has been modified to only check distance and speed as well. Since time is now being minimized in the search there is no need to check it in the goal. 
The initial implementation also allows a weight to be applied to the heuristic. A weight greater than 1.0 makes the search Weighted A* and will execute much faster. However, doing this breaks the heuristic consistency contract and paths will become less optimal. However, there is still no way to know if the search will complete in time. This leads into Sub-Issue 2.

Sub-Issue 2:
Adding NCV collision checks to the A* search substantially impacts planning times. Using large heuristic weights will make the search mode greedy and resolve faster, but how to pick a weight that provides a good trade off between optimality and fuel is unclear. Additionally, there is no way to assign a time constraint to the planner. 

To address this the A* planner has been switched out for an ANA* planner. Anytime Non-Parametric A* is a search algorithm which performs iterative searches by starting from a maximally greedy search and rapidly descending towards and optimal solution by reusing the previous iteration's search results. Given enough time the final path will be optimal and identical to A*. ANA* uses the same g(n) and h(n) as A*. Since the search is done iteratively the time spent searching can be evaluated and a sub-optimal path returned if needed. Additionally, ANA* provides a sub-optimality bound for each new path which can be evaluated to analyze how optimal the path was. 
Algorithm
http://goldberg.berkeley.edu/pubs/ana-aaai-2011-vdberg-shah-goldberg.pdf

Using the ANA* planner the timestep size could be reduced to 1s instead of 2s. However, to get it smaller the Node units need to change which leads to substantial performance penalties. This is something that should be investigated further. 

Additional Changes:

- Parameter tuning
- Add flag for turning off NCV handling in TSP
- Planning is done on a background thread. A new trajectory is requested only after it completes
- Fixed a bug in the SignalViolation check
- Unit tests for neighbor calculator and MovesCostModel were added. 

Until the sensor fusion changes are merged from PR #89 you should avoid reviewing those files if there are any changes to them. The EadAStarPlanTest.java file contains many tests for debugging and can be skipped. (We should address it before ci integration is done as it will run for a long time). There are many commented out System.out.println statements used for debugging. Until the NCV work is done, I would like to keep these in. 

## Related Issue

#66 

## Motivation and Context

Traffic Signal Plugin is supposed to handle NCVs in the host vehicle's lane

## How Has This Been Tested?

This has been tested over the past two weeks. The most recent tests on Monday should good planning behavior but still ride comfort and ncv gap distance issues. Below are some graphs from those tests

image

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist:


carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/NoCollisionChecker.java
```diff
@@ -0,0 +1,29 @@
+package gov.dot.fhwa.saxton.carma.signal_plugin;
```
Supraluminal: Missing copyright notice

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/ObjectCollisionChecker.java
```diff
@@ -1,6 +1,7 @@
 package gov.dot.fhwa.saxton.carma.signal_plugin;
```
Supraluminal: Missing copyright notice

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/TrafficSignalPlugin.java
```diff
@@ -381,6 +400,8 @@ static protected SpatMessage convertSpatMessage(IntersectionData data) {
                         //log.warn("Unsupported signal phase: " + phase);
                         break;
                 }
+                // TODO Uncomment to disable lights
```
Supraluminal: Is this (and line 382 above) something likely to be useful again? Maybe worth adding an `ignore_lights` param?

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/TrafficSignalPlugin.java
```diff
@@ -390,15 +411,103 @@ static protected SpatMessage convertSpatMessage(IntersectionData data) {
...
+                @Override
+                public void run() {
+                    log.info("Attempting Ead Plan");
+                    // Planning
+                    double currentDowntrack = pluginServiceLocator.getRouteService().getCurrentDowntrackDistance();
+                    double speedLimit = pluginServiceLocator.getRouteService().getSpeedLimitAtLocation(currentDowntrack).getLimit();
+                    DataElementHolder state = null;
+                    List<Node> eadResult = null;
+                    
+                    for (int i = 0; i < 3; i++) {
```
Supraluminal: This is probably worth adding a config value for, or at least refactoring into a constant `MAX_PLAN_RETRIES` variable.

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/TrafficSignalPlugin.java
```diff
@@ -390,15 +411,103 @@ static protected SpatMessage convertSpatMessage(IntersectionData data) {
...
+            
+        }
+    }
+
+    private List<Node> generatePlan(DataElementHolder state) {
+        // Try 3 times to plan. With updated state data each time if needed. We will accept the first completed plan
+        log.info("Requesting AStar plan with intersections: " + intersections.toString());
+
+        try {
+            // TODO remove this print after NCV handling is stable
```
Supraluminal: Maybe just put this at `DEBUG` level, so if we need it we have it but it won't be output most of the time

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/ead/trajectorytree/ANAStarSolver.java
```diff
@@ -0,0 +1,270 @@
...
+    // Each node's actual cost to reach from the start node
+    final Map<Node, Double> gScore = new HashMap<>(1500);
+
+    // Each node's heuristic cost to reach the goal
+    final Map<Node, Double> hScore = new HashMap<>(1500);
+
+    // Expected e-score cost to reach goal from start through each node
+    // eScore determines order of node expansion
+    // eScore = (G - gScore) / hScore
+    final Map<Node, Double> eScore = new HashMap<>(1500);
```
Supraluminal: Is this 1500 initial size based on observed data?

carmajava/signal_plugin/src/test/java/gov/dot/fhwa/saxton/carma/signal_plugin/AStarSolverTest.java
```diff
@@ -0,0 +1,82 @@
+package gov.dot.fhwa.saxton.carma.signal_plugin;
```
Supraluminal: Missing copyright notice

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/TrafficSignalPlugin.java
```diff
@@ -915,23 +1027,29 @@ public TrajectoryPlanningResponse planTrajectory(Trajectory traj, double expecte
...
                 } 
 
+                prevManeuverEndDist = maneuverEndDist;
                 prev = cur;
             }
         }
-
+        
         log.info("Planning complete");
+        log.info("Added Maneuvers: " + traj.getLongitudinalManeuvers().toString());
```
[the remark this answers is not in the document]
msmcconnell: This works as expected. Each maneuver's toString function is called.

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/ead/trajectorytree/ANAStarSolver.java
msmcconnell: Yes. I was seeing around 1000 elements in eScore so I went with 1500 as a buffer.

carmajava/launch/params/SignalParams.yaml
```diff
@@ -201,6 +202,9 @@ ead/useIdleMin: false
 
 ## NCV Handling ##
 
+# Boolean: True if NCV handling will be enabled when operating the plugin. False otherwise
+ead/handleNCV: true
```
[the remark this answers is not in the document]
msmcconnell: Changed to false

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/NoCollisionChecker.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.List;
+
+import cav_msgs.RoadwayObstacle;
+
+import gov.dot.fhwa.saxton.carma.signal_plugin.ead.trajectorytree.Node;
+
+/**
+ * No-Op Collision Checker for the Traffic Signal Plugin. Allows NCV handling to be effectively turned off by injecting this as the primary ITrafficSignalPluginCollisionChecker
+ */
+public class NoCollisionChecker implements ITrafficSignalPluginCollisionChecker {
```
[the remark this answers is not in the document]
msmcconnell: Done

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/ObjectCollisionChecker.java
```diff
@@ -76,6 +80,8 @@
   private static final double MS_PER_S = 1000.0; // ms
   private Long ncvDetectionTime = null; // ms
```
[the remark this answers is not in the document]
msmcconnell: I'll leave it for now. But I'll keep that in mind when we pick up work on this again later.

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/ead/FuelCostModel.java
```diff
@@ -22,6 +22,9 @@
 import gov.dot.fhwa.saxton.carma.signal_plugin.logger.LoggerManager;
 
 /**
+ * 
+ * TODO: This model has been deprecated for MOVES and has therefore not been updated for new algorithm changes as of 12/5/18
```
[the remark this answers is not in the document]
msmcconnell: Removed

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/NoCollisionChecker.java
msmcconnell: Added copyrights to all files

carmajava/signal_plugin/src/main/java/gov/dot/fhwa/saxton/carma/signal_plugin/TrafficSignalPlugin.java
msmcconnell: This log is extremely useful for testing NCVs, but activating debug logs for guidance slows everything down too much. I really wish I could control debug level on a per-class basis. Maybe there is a way but I don't know how. I would rather leave this for now.

carmajava/signal_plugin/src/test/java/gov/dot/fhwa/saxton/carma/signal_plugin/ead/MovesCostModelTest.java
```diff
@@ -0,0 +1,536 @@
...
+   */
+  @Test
+  public void testToJPerSec() throws Exception {
+    MovesFuelCostModel costModel = new MovesFuelCostModel(
+      0.22112,  // rollingTermA
+      0.002838, // rotatingTermB
+      0.000698, // dragTermC
+      1.86686,  // vehicleMassInTons
+      1.86686,  // fixedMassFactor
+      "/opt/carma/src/CARMAPlatform/carmajava/launch/params/BaseRateForPassengerTruck.csv", // baseRateTablePath
```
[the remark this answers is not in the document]
msmcconnell: Looks like I was able to make it relative

msmcconnell:
Addressed comments

Issue #177, opened 2019-02-01
Remove Interface Mgr Driver Discovery Capability


qswawrq:
### Descriptive summary
As a Carma developer, I depend on the code being as lean an maintainable as possible. Modify legacy Carma consumer nodes to use the modified CARMA3 API instead of using Interface Manager for driver discovery.
### Expected behavior
CARMA nodes connect to drivers directly without going through Interface Manager.
### Actual behavior
Some CARMA nodes are using Interface Manager get_driver_capabilities service.
### Steps to reproduce the behavior
N/A
### Related work
N/A

msmcconnell:
@qswawrq Is this still an issue?

qswawrq:
Yup. We did not start working on it.

TonysCousin:
This issue is now OBE due to new structure and API provided by Health Monitor, which replaces Interface Manager.

PR #178, opened 2019-02-01
Remove interactions with interface mgr

qswawrq:
# PR Details
Modify legacy CARMA consumer nodes to use the launch file mapping instead of using Interface Manager for driver discovery.
## Description
Originally, the CARMA consumer nodes used Interface Manager to connect to different drivers. However, this feature is removed from CARMA3 design document. This PR is about removing driver discovery service call from consumer nodes. There will be a follow-up PR for modifying Interface Manager.

The interactions are removed from: GuidanceCommands, GuidanceStateHandler, LightBarManager, MessageConsumer and Platooning Plugin. The launch file is modified for mapping to correct topic names.
## Related Issue
#177
## Motivation and Context
Update CARMA legacy code to match new API requirement.
## How Has This Been Tested?
N/A
## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/carmajava/launch/saxton_cav_src.launch
+++ b/carmajava/launch/saxton_cav_src.launch
@@ -326,9 +326,9 @@
         <remap from="outgoing_mobility_operation"
                to="/$(arg TOP_NS)/$(arg GUIDE_NS)/outgoing_mobility_operation"/>
         <remap from="inbound_binary_msg"
-               to="/$(arg TOP_NS)/$(arg DRIVER_NS)/dsrc/comms/inbound_binary_msg"/>
+               to="/comms/inbound_binary_msg"/>
         <remap from="outbound_binary_msg"
-               to="/$(arg TOP_NS)/$(arg DRIVER_NS)/dsrc/comms/outbound_binary_msg"/>
+               to="/comms/outbound_binary_msg"/>
 
       </node>
     </group>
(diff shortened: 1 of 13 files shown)


carmajava/launch/saxton_cav_src.launch
```diff
@@ -325,8 +325,11 @@
                to="/$(arg TOP_NS)/$(arg GUIDE_NS)/outgoing_mobility_response"/>
         <remap from="outgoing_mobility_operation"
                to="/$(arg TOP_NS)/$(arg GUIDE_NS)/outgoing_mobility_operation"/>
-        <remap from="get_drivers_with_capabilities"
-               to="/$(arg TOP_NS)/$(arg INTR_NS)/get_drivers_with_capabilities"/>
+        <remap from="inbound_binary_msg"
+               to="/$(arg TOP_NS)/$(arg DRIVER_NS)/dsrc/comms/inbound_binary_msg"/>
+        <remap from="outbound_binary_msg"
+               to="/$(arg TOP_NS)/$(arg DRIVER_NS)/dsrc/comms/outbound_binary_msg"/>
```
msmcconnell: I don't think we should remap to private driver name spaces. It would be better to update every driver to publish to public topics. Similar to how it will be in CARMA3

msmcconnell: Same for all below

PR #208, opened 2019-04-12
Feature/docker dev tooling

Supraluminal:
# PR Details

## Description

Improve the docker user experience for local development and testing of CARMA

## Motivation and Context

Previously dockerized testing on a developer's PC was not possible due to no clear way to run the mock drivers and other configuration details.

## How Has This Been Tested?
This has been tested locally by building the development docker configuration and then running the docker compose system via the CARMA script (`carma start all`). From the command line all systems function as expected. The remote start functionality does not seem to function, as the timeout before connecting to the ROSBridge server is not long enough, though I believe this fix is out of scope for this task. The proper fix to this is to wrap the initial ROS connection from the web UI to the ROS bridge server in a timed retry loop so that it can properly connect when the ROSbridge server takes longer to start.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] I have updated the documentation accordingly. (n/a)
- [ ] I have added tests to cover my changes. (n/a)
- [ ] All new and existing tests passed. (n/a)


configurations/cadillac_srx_2013/docker-compose.yml
```diff
@@ -16,12 +16,26 @@
 version: '2'
 
 services:
+  roscore:
```
qswawrq: What is the reason for taking roscore out of docker-compose-background?

configurations/development/docker-compose.yml
```diff
@@ -16,16 +16,106 @@
 version: '2'
 
 services:
-  carma:
-    image: usdot-carma/carma
+  roscore:
+    image: usdotfhwastol/carma-base:2.8.3
```
qswawrq: Similarly, can we use "usdotfhwastol/carma:latest" here?

configurations/development/docker-compose-background.yml
```diff
@@ -0,0 +1,29 @@
...
+#  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#  License for the specific language governing permissions and limitations under
+#  the License.
+
+# Docker Compose Spec Version
+version: '2'
+
+services:
+  web-ui:
+    image: usdotfhwastol/carma-web-ui:2.8.2
```
qswawrq: If later on, we have a newer version, do I need to update this to much the version I have?

configurations/cadillac_srx_2013/docker-compose.yml
Supraluminal: If we have it in the compose-background file, the rosmaster instance always stays up and we get one big rosout log. It's unclear how parameter loading/unloading also works in that instance, particularly if you're trying to tweak things. This is part of why we run into the new issue with the UI not being able to connect, since the roscore service now takes time to start up.

configurations/development/docker-compose.yml
Supraluminal: Yes, and the same below.

configurations/cadillac_srx_2013/docker-compose.yml
qswawrq: Got you.

PR #223, opened 2019-05-06
Add C++ and Java Unit Test Code Coverage Metrics to Sonar Cloud

msmcconnell:
This is a branch being used for CI testing


carmajava/build.gradle
```diff
@@ -90,12 +115,45 @@ subprojects {
...
+    }
+}
+
 task testReport(type: TestReport) {
     destinationDir = file("$buildDir/reports/allTests")
     // Include the results from the `test` task in all subprojects
     reportOn subprojects*.test
 }
 
+task testCoverageReport() {
```
Supraluminal: Is this supposed to be empty? Unless it matters that it's defined, can we just delete it?

engineering_tools/code_coverage/make_with_coverage.bash
```diff
@@ -0,0 +1,72 @@
+
```
Supraluminal: 1. I think the `#!/bin/bash` shebang line needs to be the first in the file, not sure if that's why you're explicitly invoking `bash make_with_coverage.bash` elsewhere in this PR, but with that you should just be able to run it with `./make_with_coverage.bash`.
2 . Right after that line I'd drop in the standard copyright notice on this and the other shell script.

engineering_tools/code_coverage/collect_gcovr.bash
```diff
@@ -0,0 +1,70 @@
+
```
Supraluminal: Same as below.

Issue #253, opened 2019-06-06
Invalid XML in ndt_pose_tf_wrapper package.xml; JIRA Issue#: CAR-819


Supraluminal:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
`ndt_pose_tf_wrapper` package.xml contains invalid XML causing a failing build.

### Expected behavior
Successful docker build.
### Actual behavior
Catkin make errors out at line 1 character 1 of `ndt_pose_tf_wrapper/package.xml` because of comments with "#" instead of "".

### Steps to reproduce the behavior

1. Checkout CARMAPlatform repository
1. Run `./docker/build-image.sh`

qswawrq:
Fixed by #258

Issue #252, opened 2019-06-06
Autoware message package dependencies missing; JIRA Issue#: CAR-820


Supraluminal:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
```-- Could not find the required component 'autoware_msgs'. The following CMake error indicates that you either need to install the package with the same name or change your environment so that it can be found.
CMake Error at /opt/ros/kinetic/share/catkin/cmake/catkinConfig.cmake:83 (find_package):
  Could not find a package configuration file provided by "autoware_msgs"
  with any of the following names:

    autoware_msgsConfig.cmake
    autoware_msgs-config.cmake

  Add the installation prefix of "autoware_msgs" to CMAKE_PREFIX_PATH or set
  "autoware_msgs_DIR" to a directory containing one of the above files.  If
  "autoware_msgs" provides a separate development package or SDK, be sure it
  has been installed.
Call Stack (most recent call first):
  CARMAPlatform/pure_pursuit_wrapper/CMakeLists.txt:13 (find_package)
```

during Docker build. Suspect that Autoware message packages need to be baked into a new CARMABase release and the Docker build needs to be updated accordingly.

### Expected behavior
Docker build completes successfully.
### Actual behavior
Docker build errors out with the above error message.
### Steps to reproduce the behavior

1. Checkout CARMAPlatform repository
1. Run `./docker/build-image.sh`

msmcconnell:
@Supraluminal This is because you need to use the carma_build script now to build the platform. I haven't had a chance to send out an update about it yet.

qswawrq:
This is fixed by #258 .

PR #254, opened 2019-06-06
Feature/docker refactor

Supraluminal:
# PR Details
Refactoring the docker versioning and image dependencies to avoid the kinds of circular loops we've been struggling with.
## Description
Moved the entire `configurations` folder --> https://github.com/usdot-fhwa-stol/CARMAConfig
Moved the entire `carma_base` folder --> https://github.com/usdot-fhwa-stol/CARMABase

Updated the automated version generation to use `git describe` instead of our previous `gradle` based method so that we can uniformly apply it across all repositories. Docker image versioning is now dependent on git tags. As described in the release process writeup component versions start with "<REPO_NAME>" and system versions start with "CARMASystem".

`docker/get-component-version.sh` and `docker/get-system-version.sh` correspond to these `git describe` commands and some basic string processing.
## How Has This Been Tested?

Locally tagged repo temporarily to verify that the parsing of the version numbers is done correctly, docker build scripts find correct versions based on the scripts but docker build fails based on two other known issues #252 and #253 .

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly. (See new docker release process document)
- [ ] I have added tests to cover my changes. (N/a)
- [ ] All new and existing tests passed. (N/a


carma-platform.repos
```diff
@@ -0,0 +1,81 @@
+repositories:
+  CARMAAvtVimbaDriver:
+    type: git
+    url: git@github.com:usdot-fhwa-stol/CARMAAvtVimbaDriver.git
+    version: develop
```
TonysCousin: You may have described to me, but I am not remembering what does version: develop mean?  I see that Autoware.ai and the Novatel driver are set to master, but everything else is develop.  Are the rules for this value documented somewhere?

docker/build-image.sh
```diff
@@ -23,25 +23,58 @@ echo ""
...
+docker tag $USERNAME/$IMAGE:$COMPONENT_VERSION_STRING $USERNAME/$IMAGE:latest
+TAGS+=("$USERNAME/$IMAGE:latest")
 
-echo "Tagged $USERNAME/$IMAGE:$FULL_VERSION_STRING as $USERNAME/$IMAGE:latest"
+echo "Tagged $USERNAME/$IMAGE:$COMPONENT_VERSION_STRING as $USERNAME/$IMAGE:latest"
+
+if [ "$SYSTEM_RELEASE" = true ]; then
+    SYSTEM_VERSION_STRING=$("./get-system-version.sh")
+    docker tag $USERNAME/$IMAGE:$COMPONENT_VERSION_STRING $USERNAME/$IMAGE:$SYSTEM_VERSION_STRING
+    echo "Tagged $USERNAME/$IMAGE:$COMPONENT_VERSION_STRING as $USERNAME/$IMAGE:$SYSTEM_VERSION_STRING"
```
TonysCousin: Does this redundantly apply the component tag? It seems this was already handled on line 61 above.

carma-platform.repos
Supraluminal: In this case, when you use `vcstool` to checkout all these repos at once, this one is checked out to develop by default. I think for all these repos it's just using the default branch as defined in github, the one that default PRs are against, so this behavior makes sense to me.

docker/build-image.sh
Supraluminal: The tag command is `docker tag <CURRENT_NAME> <NEW_ADDITIONAL_NAME>` so in this case it references the image by the component tag, then also assigns it the `$SYSTEM_VERSION_STRING` tag too.

carma-platform.repos
TonysCousin: Then we probably need to fix the Novatel driver repo, since it's showing master here.

TonysCousin: Yep, it was set with master as default.  I have fixed that.  Please make the corresponding change in this file.

TonysCousin: I guess I should do the same thing on autoware.ai fork as well, since we don't want to be branching/pushing directly from master as a rule.  So please change both of those here.

TonysCousin: I actually updated the autoware.ai repo to default to the carma-develop branch.

PR #268, opened 2019-06-25
Feature/trajectory executor node

Supraluminal:
# PR Details

## Description

Add `trajectory_executor` package to CARMA. Trajectory Executor is a new node that processes the output of the CARMA planning process and delegates it between one or more Control Plugins for execution.

## Motivation and Context

This new package is part of the functionality required for CARMA 3 beta release.

## How Has This Been Tested?

This package has been tested by unit tests contained in the test subdirectory of this new package.

## Types of changes

## Checklist:


trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+namespace trajectory_executor 
+{
+    cav_msgs::TrajectoryPlan trimFirstPoint(const cav_msgs::TrajectoryPlan plan) {
+        cav_msgs::TrajectoryPlan out(plan);
+        out.trajectory_points.erase(out.trajectory_points.begin());
+        return out;
+    }
+
+    TrajectoryExecutor::TrajectoryExecutor(int traj_frequency) :
```
qswawrq: It does not look like this constructor will ever get called. Because it is likely that you will get traj_frequency from a ROS parameter, which is only from a ROS NodeHandle. But NodeHandle will generally created in the init() step.

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+    bool TrajectoryExecutor::init()
+    {
+        ROS_DEBUG("Initializing TrajectoryExecutor node...");
+    
+        _default_nh = std::unique_ptr<ros::CARMANodeHandle>(new ros::CARMANodeHandle);
+        _timer_nh = std::unique_ptr<ros::CARMANodeHandle>(new ros::CARMANodeHandle);
+        _timer_nh->setCallbackQueue(&_timer_callbacks);
+        _msg_nh = std::unique_ptr<ros::CARMANodeHandle>(new ros::CARMANodeHandle);
+        _msg_nh->setCallbackQueue(&_msg_callbacks);
```
qswawrq: Usually, we are using one general NodeHandle and a private NodeHandle and using their global default callback queue. What is the reason to have three here?

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
```diff
@@ -0,0 +1,188 @@
...
+        ROS_DEBUG("Initialized all node handles");
+
+        _default_nh->param("spin_rate", _default_spin_rate, 10);
+        _default_nh->param("trajectory_publish_rate", _min_traj_publish_tickrate_hz, 10);
```
qswawrq: Parameters should be read from a private nh to avoid namespace conflict.

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+        ROS_DEBUG_STREAM("Initalized params with default_spin_rate " << _default_spin_rate 
+            << " and trajectory_publish_rate " << _min_traj_publish_tickrate_hz);
+
+        this->_plan_sub = this->_msg_nh->subscribe<cav_msgs::TrajectoryPlan>("trajectory", 1000, &TrajectoryExecutor::onNewTrajectoryPlan, this);
```
qswawrq: I think we can lower the queue size. Because there is a server delay, we do not want to publish a trajectory from few seconds ago. It should be a low risk to just drop it. Do you agree?

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
```diff
@@ -0,0 +1,188 @@
...
+        msg_spinner.start();
+
+        ROS_DEBUG("TrajectoryExecutor component started succesfully! Starting to spin.");
+
+        ros::Rate r(_default_spin_rate);
+        while (ros::ok())
+        {
+            ros::spinOnce();
+            r.sleep();
+        }
```
qswawrq: There is a spin() wrapper function in CARMANodeHandle that you are use easily.

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
```diff
@@ -0,0 +1,188 @@
...
+            ros::Duration(ros::Rate(this->_min_traj_publish_tickrate_hz)),
+            &TrajectoryExecutor::onTrajEmitTick, 
+            this);
+
+        ros::AsyncSpinner timer_spinner(1, &_timer_callbacks);
+        timer_spinner.start();
+        ROS_DEBUG("Timer thread started!");
+
+        ros::AsyncSpinner msg_spinner(1, &_msg_callbacks);
+        msg_spinner.start();
```
qswawrq: Not sure if we need extra spinner as the same reason for avoiding multiple NodeHandles.

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
```diff
@@ -0,0 +1,188 @@
...
+        _default_nh->param<std::string>("default_control_plugin", default_control_plugin, "NULL");
+
+        std::string default_control_plugin_topic;
+        _default_nh->param<std::string>("default_control_plugin_topic", default_control_plugin_topic, "NULL");
+
+        out[default_control_plugin] = default_control_plugin_topic;
+        return out;
+    }
+    
+    void TrajectoryExecutor::onNewTrajectoryPlan(cav_msgs::TrajectoryPlan msg)
```
qswawrq: What is the reason for passing by value in this function?

trajectory_executor/config/trajectory_executor.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+# String: Name of default control plugin. Should match field in TrajectoryPlan message
+# Due to lack of existing plugin discovery mechanism this is the only control plugin
+# which will be available to TrajectoryExecutor
+# Units: N/a
+default_control_plugin: pure_pursuit
+
+# String: Full path to default control plugin's trajectory input topic
+# Units: N/a
+default_control_plugin_topic: /carma/guidance/control_plugins/pure_pursuit/trajectory
```
qswawrq: My memory is pretty vague. I forgot if we need each topic for each plugin or just one topic for all of them.

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
```diff
@@ -0,0 +1,188 @@
...
+        std::unique_lock<std::mutex> lock(_cur_traj_mutex);
+        ROS_DEBUG("TrajectoryExecutor tick start!");
+        
+
+        if (_cur_traj != nullptr) {
+            if (_timesteps_since_last_traj > 0) {
+                _cur_traj = std::unique_ptr<cav_msgs::TrajectoryPlan>(new cav_msgs::TrajectoryPlan(trimFirstPoint(*_cur_traj)));
+            }
+            if (!_cur_traj->trajectory_points.empty()) {
+                std::string control_plugin = _cur_traj->trajectory_points[0].controller_plugin_name;
```
qswawrq: I think this assume each trajectory will only be sent to one control plugin. You might want to add a note to call it out.

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
```diff
@@ -0,0 +1,188 @@
...
+                } else {
+                    std::ostringstream description_builder;
+                    description_builder << "No match found for control plugin " 
+                        << control_plugin << " at point " 
+                        << _timesteps_since_last_traj << " in current trajectory!";
+                    shutdown(description_builder.str());
+                }
+                _timesteps_since_last_traj++;
+            } else {
+                shutdown("Ran out of trajectory data to consume!");
```
qswawrq: I am not 100% sure. But I think those shutdown function can be replaced by the exception handling function in CARMANodeHandle.

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
Supraluminal: The idea is to allow the timer thread (which uses a TimerEvent in the callback queue) to run at whatever rate it needs to without depending on the rate of the message processing queue and to allow the message queue to spin as fast as it receives messages (through use of an async spinner). I have made the `_default_nh` private so that we can read params from it as you mentioned, though.

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
Supraluminal: Agreed

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
Supraluminal: Was just the easiest way to get an object i could pass into a copy constructor, wasn't having much luck with their `cav_msgs::TrajectoryPlan::ConstPtr&` interface here because getting the actual object out of the `ConstPtr&` (I think under the hood it's a `boost::shared_ptr`) was being a real pain.

trajectory_executor/config/trajectory_executor.yaml
Supraluminal: I think the plan is one topic per each plugin, but for now it only supports 1 plugin since we don't have an implementation for our new plugin manager component.

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
Supraluminal: Yeah, at any given timestep the current trajectory is only sent to a single control plugin.

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
Supraluminal: Fixed!

Supraluminal:
I believe I've added fixes for all your comments and all the unit tests still pass.

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
qswawrq: Yep. I was using ```boost::shared_ptr``` but I do not think it is a big deal for this small message.

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+        ROS_DEBUG("Setting up publishers for control plugin topics...");
+
+        std::map<std::string, ros::Publisher> control_plugin_topics;
+        auto discovered_control_plugins = queryControlPlugins();
+
+        for (auto it = discovered_control_plugins.begin(); it != discovered_control_plugins.end(); it++)
+        {
+            ROS_DEBUG("Trajectory executor discovered control plugin %s listening on topic %s.", it->first.c_str(), it->second.c_str());
+            ros::Publisher control_plugin_pub = _public_nh->advertise<cav_msgs::TrajectoryPlan>(it->second, 1000);
```
qswawrq: It seems you did not initialize this public nh?

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
```diff
@@ -0,0 +1,156 @@
...
+        ROS_DEBUG("Starting operations for TrajectoryExecutor component...");
+        _timer = _private_nh->createTimer(
+            ros::Duration(ros::Rate(this->_min_traj_publish_tickrate_hz)),
+            &TrajectoryExecutor::onTrajEmitTick, 
+            this);
+
+        ROS_DEBUG("TrajectoryExecutor component started succesfully! Starting to spin.");
+
+        _private_nh->setSpinRate(_default_spin_rate);
+        _private_nh->spin();
```
qswawrq: Try to use CARMANodeHandle::spin() for this.

Issue #273, opened 2019-07-01
Initialize localization with GNSS


msmcconnell:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
It is tedious initializing autoware ndt_matching with Rviz it would be better if it could be initialized with GNSS data.

msmcconnell:
Resolved by PR #274

PR #274, opened 2019-07-01
Add GNSS initialization to NDT

msmcconnell:
# PR Details

## Description

This PR adds GNSS initialization of NDT to carma. This is primarily done using the new gnss_to_map convertor node. This node takes in lat/lon fix and GNSS heading and converts them first to the ecef frame then to the map frame. The map frame pose is published to the gnss_pose topic which is used by NDT and points_map_loader. The ecef frame pose is published to the ecef_pose_with_cov topic which is not currently used but is available if needed in the future. NDT uses the gnss_pose to initialize. 

Notes:
- The interface manager had some bad logic in it that prevented the system from starting. After discussing with Kyle I updated it. 

- The SWRI robotics novatel driver NavSatFix messages do not include fused IMU data, so for now the gnss_to_map_convertor node does no obey the driver API. It is using Novatel messages directly which do include fused IMU information (at least that's the claim). 

- I use the message_filter packages for synchronizing fix and heading. I have to wrap this callback in a try/catch as it does not support the CARMANodeHandle wrappers. To make this easier I made the CARMANodeHandle::handleExecption function public so that it can be easily called if we need to do this again. 

## Related Issue

#273

## How Has This Been Tested?

Unit tests as well as integration tests on the Lexus

## Types of changes

## Checklist:


carmajava/interfacemgr/src/main/java/gov/dot/fhwa/saxton/carma/interfacemgr/InterfaceMgr.java
```diff
@@ -204,9 +206,9 @@ public void onNewMessage(cav_msgs.RobotEnabled msg) {
                         if (robotEnabled_) {
 
                             //if robotic control is no longer active then warn and alert all nodes to shut down
-                            if (!msg.getRobotActive()) {
-                                log_.warn("SHUTDOWN", "InterfaceMgr.robotListener senses robot is no longer active at the hardware level.");
-                                publishSystemAlert(AlertSeverity.FATAL, "Robotic control has been disengaged.", null);
+                            if (Math.abs(System.currentTimeMillis() - lastRobotStatusMsgTime_) > robotStatusTimeout_ && !isShutdownUnderway()) {
```
qswawrq: It does not matter in this case, But it is still better to use ROS time whenever is possible.

gnss_to_map_convertor/include/gnss_to_map_convertor/GNSSToMapNode.h
```diff
@@ -0,0 +1,89 @@
...
+#include <carma_utils/CARMAUtils.h>
+
+/**
+ * \class GNSSToMapNode
+ * \brief ROS Node which maintains a tf2 transform tree which can be accessed by other nodes.
+ *
+ * The get_transform service can be used to obtain coordinate transformations between two frames.
+ * Only transforms published on the /tf or /tf_static topics are recorded by this node.
+ */
+class GNSSToMapNode {
```
qswawrq: Do we need to put this class in its own namespace?

gnss_to_map_convertor/src/gnss_to_map_convertor/GNSSToMapNode.cpp
```diff
@@ -0,0 +1,106 @@
...
+  sensor_msgs::NavSatFix nav_fix_msg;
+  nav_fix_msg.latitude = fix_msg->lat;
+  nav_fix_msg.longitude = fix_msg->lon;
+  nav_fix_msg.altitude = fix_msg->height;
+  sensor_msgs::NavSatFixConstPtr fix_ptr(new sensor_msgs::NavSatFix(nav_fix_msg));
+
+  geometry_msgs::PoseWithCovarianceStamped ecef_pose = gnss_to_map_convertor::poseFromGnss(
+    baselink_in_sensor_, sensor_in_ned_, fix_ptr, heading_msg
+  ); 
+  ecef_pose.header.frame_id = earth_frame_; // Set correct frame id
```
qswawrq: Do we need to set header for ecef_pose?

gnss_to_map_convertor/include/gnss_to_map_convertor/GNSSToMapNode.h
msmcconnell: I don't think so. This class would never be included in a library for usage outside this package.

gnss_to_map_convertor/src/gnss_to_map_convertor/GNSSToMapNode.cpp
msmcconnell: The header is assigned inside the poseFromGnss function

qswawrq: Got you.

PR #325, opened 2019-08-16
Initial Guidance Node

qswawrq:
# PR Details
This PR includes the initial implementation of the CARMA 3 guidance node.
## Description
The current guidance node has two functionalities. One is plugin discovery and the other is guidance state machine management.
## Related Issue
N/A

## Motivation and Context
N/A

## How Has This Been Tested?
Unit test passes for state machine logic.
Currently, it does not have unit tests for plugin discovery logic because it will be soon modified and merged into C++ interface manager node.

## Types of changes

## Checklist:


guidance/src/guidance/guidance_worker.cpp
```diff
@@ -0,0 +1,183 @@
...
+                res.plugins.push_back(plugins[i]);
+            }
+        }
+    }
+
+    bool GuidanceWorker::is_required_plugin(std::string plugin_name, std::string version)
+    {
+        for(int i = 0; i < required_plugins.size(); i++)
+        {
+            if(plugin_name.append(" " + version).compare(required_plugins[i]) == 0)
```
msmcconnell: Rather than adding a single space with the version number it would be more robust to strip spaces from the required_plugin list at startup and compare without spaces. Then if someone accidentally types "Autoware Plugin   1.0.0" (extra spaces). It will not fail to find a match

guidance/src/guidance/guidance_worker.cpp
```diff
@@ -0,0 +1,183 @@
...
+        populate_active_plugin_list_response(res);
+        return true;
+    }
+
+    bool GuidanceWorker::activate_plugin_cb(cav_srvs::PluginActivationRequest& req, cav_srvs::PluginActivationResponse& res)
+    {
+        for(int i = 0; i < plugins.size(); ++i)
+        {
+            // Go through the plugin list to find the corresponding plugin that the user wants to activate
+            if(req.pluginName.compare(plugins[i].name) && req.pluginName.compare(plugins[i].versionId))
```
msmcconnell: I would expect string.compare() here to return 0 if the string is equal which should evaluate to false. I think you want !req... or use ==0

guidance/src/guidance/guidance_worker.cpp
```diff
@@ -0,0 +1,183 @@
...
+            if(plugins[index].name.compare(msg.name) == 0 && plugins[index].versionId.compare(msg.versionId) == 0)
+            {
+                // only availability can be changed by plugin itself
+                plugins[index].available = msg.available;
+                // this break will make sure that "index" variable stays at the location of matching entry if it exists
+                break;
+            }
+            ++index;
+        }
+        // index points to the end of pllugin list without a matching entry
```
msmcconnell: typo on pllugin extra l

guidance/src/guidance/guidance_worker.cpp
```diff
@@ -0,0 +1,183 @@
...
+        if(index == plugins.size())
```
msmcconnell: seems like you could just return rather break then this check wouldn't be nessasary

guidance/src/guidance/guidance_worker.cpp
```diff
@@ -0,0 +1,183 @@
...
+        cav_msgs::GuidanceState state;
+        state.state = gsm.getCurrentState();
+        state_publisher_.publish(state);
+        return true;
+    }
+
+    int GuidanceWorker::run()
+    {
+        ROS_INFO("Initalizing guidance node...");
+        using std::placeholders::_1;
```
msmcconnell: I could be wrong but I don't think you actually need to define this. You can just say _1 in bind and this should resolve

guidance/src/guidance/guidance_worker.cpp
```diff
@@ -0,0 +1,183 @@
...
+        std::function<void(const cav_msgs::SystemAlertConstPtr&)> system_alert_cb_function = std::bind(&GuidanceWorker::system_alert_cb, this, _1);
+        ros::CARMANodeHandle::setSystemAlertCallback(system_alert_cb_function);
```
msmcconnell: you should be able to place you bind call inside this, but that is a stylistic choice and up to you

guidance/src/guidance/guidance_worker.cpp
```diff
@@ -0,0 +1,183 @@
...
+        // Load the spin rate param to determine how fast to process messages
+        // Default rate 10.0 Hz
+        double spin_rate = pnh_.param<double>("spin_rate_hz", 10.0);
+
+        pnh_.getParam("required_plugins", required_plugins);
+
+        // Spin until system shutdown
+        ROS_INFO_STREAM("Guidance node initialized, spinning at " << spin_rate << "hz...");
+        std::function<bool(void)> spin_cb_function = std::bind(&GuidanceWorker::spin_cb, this);
+        ros::CARMANodeHandle::setSpinCallback(spin_cb_function);
```
msmcconnell: same note about bind

guidance/src/guidance/guidance_worker.cpp
qswawrq: Sounds good to me.

guidance/src/guidance/guidance_worker.cpp
qswawrq: Good catch.

guidance/src/guidance/guidance_worker.cpp
qswawrq: Yes. I do not have to use "using" keyword here. But in the function, I still need to specify the full namespace as ```std::placeholders::_1```.
But I updated it anyway because using "using" here is bad practice anyway.

PR #362, opened 2019-09-20
Fix/state machine

itabrl:
# PR Details
## Description

state machine factory is added to support different type of vehicles.
## Related Issue
349 - Guidance state machine does not transition properly

## How Has This Been Tested?
this has been tested on the car.

## Types of changes

- [`x`] Defect fix (non-breaking change that fixes an issue)

## Checklist:

- [`x`] I have added any new packages to the sonar-scanner.properties file
- [`x`] I have read the **CONTRIBUTING** document.
- [`x`] All new and existing tests passed.


msmcconnell:
Can you address the 4 code smells detected by sonar cloud. They seem pretty minor
https://sonarcloud.io/project/issues?id=usdot-fhwa-stol_CARMAPlatform&pullRequest=362&resolved=false&types=CODE_SMELL

guidance/config/params.yaml
```diff
@@ -20,3 +20,5 @@ spin_rate_hz: 10
 required_plugins:
   - Autoware Plugin 1.0.0
   - Pure Pursuit Plugin 1.0.0
+
+vehicle_type: cadilac # lexus or cadilac
```
msmcconnell: This parameter is an aspect of the vehicle configuration and should therefore be stored in the CARMAConfig repo. I would suggest renaming it to something like vehicle_state_machine_type or something similar and adding it to the 
VehicleConfigParams.yaml file located in the appropriate sub directories of that repo.

guidance/include/guidance/guidance_worker.hpp
```diff
@@ -82,15 +82,20 @@ namespace guidance
...
+
+            std::string vehicle_type;
+
+            void create_guidance_state_machine();
         
         private:
+            GuidanceStateMachineFactory guidance_state_machine_factory;
             // Guidance state machine
-            GuidanceStateMachine gsm;
+            GuidanceStateMachine* gsm;
```
msmcconnell: Please use smart pointers instead of raw pointers unless there is a justification for it.

guidance/include/guidance/guidance_state_machine.hpp
```diff
@@ -71,15 +71,57 @@ namespace guidance
...
             uint8_t getCurrentState();
 
+            virtual void StartUpState(Signal signal) = 0;
+            virtual void DriversReadyState(Signal signal) = 0;
+            virtual void ActiveState(Signal signal) = 0;
+            virtual void EngagedState(Signal signal) = 0;
+            virtual void InactiveState(Signal signal) = 0;
+            virtual void OffState(Signal signal) = 0;
+            // a local variable keeps the current state machine state
+            State currentGuidanceState;
```
msmcconnell: According to the ROS Style guide member variables should be underscored and have trailing underscores. So this would be current_guidance_state_

guidance/src/guidance/guidance_worker.cpp
```diff
@@ -157,6 +157,17 @@ namespace guidance
...
     }
 
+    void GuidanceWorker::create_guidance_state_machine()
+    {
+        if(vehicle_type == "cadilac") {
+            gsm = guidance_state_machine_factory.createCadilacInstance();
+        }
+        else if (vehicle_type == "lexus") {
+            gsm = guidance_state_machine_factory.createLexusInstance();
+        }
```
msmcconnell: Not a required change, but I think generally with factory pattern the standard practice is to put this logic inside the factory such that you would call 
guidance_state_machine_factory.createInstance("lexus").

guidance/src/guidance/guidance_worker.cpp
```diff
@@ -180,6 +191,9 @@ namespace guidance
         pnh_.getParam("required_plugins", required_plugins);
         process_required_plugin_list(required_plugins);
 
+        pnh_.getParam("vehicle_type", vehicle_type);
```
msmcconnell: After moving this parameter to the CARMAConfig you will need to access it using the regular nh_

guidance/include/guidance/guidance_state_machine.hpp
itabrl: currentGuidanceState is legacy, I think then its needed to be update though out all the files.

msmcconnell: That's fine you can go ahead and update it.

guidance/config/params.yaml
itabrl: I do not know where is the yaml file you are mentioning is located can you send the link to that file ?

guidance/include/guidance/guidance_state_machine.hpp
```diff
@@ -71,15 +71,57 @@ namespace guidance
              */
             uint8_t getCurrentState();
 
+            virtual void StartUpState(Signal signal) = 0;
```
msmcconnell: Please add a comment describing the use of these functions.

guidance/include/guidance/guidance_state_machine.hpp
```diff
@@ -71,15 +71,57 @@ namespace guidance
...
+            virtual GuidanceStateMachine* createCadilacInstance() = 0;
+            virtual GuidanceStateMachine* createLexusInstance() = 0;
+    };
+
+    class GuidanceStateMachineFactory : public Factory {
+        public:
+            GuidanceStateMachine* createCadilacInstance() {
+                return new Cadilac;
+            }
+            GuidanceStateMachine* createLexusInstance() {
```
msmcconnell: Public functions should have doxygen style comments on them

guidance/include/guidance/guidance_state_machine.hpp
```diff
@@ -71,15 +71,57 @@ namespace guidance
...
+            void OffState(Signal signal);
+    };
+
+    class Factory {
+        public:
+            virtual GuidanceStateMachine* createCadilacInstance() = 0;
+            virtual GuidanceStateMachine* createLexusInstance() = 0;
+    };
+
+    class GuidanceStateMachineFactory : public Factory {
```
msmcconnell: Please keep the implementation of classes. In cpp files since that is the standard we have been using. It also is probably worth seperating the Factory classes into a seperate header file so that we have guidance_state_machine.h and guidance_state_machine_factory.h

guidance/include/guidance/guidance_state_machine.hpp
```diff
@@ -71,15 +71,57 @@ namespace guidance
...
+        public:
+            void StartUpState(Signal signal);
+            void DriversReadyState(Signal signal);
+            void ActiveState(Signal signal);
+            void EngagedState(Signal signal);
+            void InactiveState(Signal signal);
+            void OffState(Signal signal);
+    };
+
+    class Factory {
```
msmcconnell: This class name is too generic. I'm also not quite sure what the point of having this factory interface is when there is only a single type of GuidanceStateMachineFactory. It is a nice abstraction but it might be overkill in this case. How likely is it that we have different kinds of factories here.

guidance/include/guidance/guidance_worker.hpp
```diff
@@ -90,7 +90,7 @@ namespace guidance
         private:
             GuidanceStateMachineFactory guidance_state_machine_factory;
             // Guidance state machine
-            GuidanceStateMachine* gsm;
+            std::auto_ptr<GuidanceStateMachine> gsm;
```
msmcconnell: don't use std::auto_ptr. In c++ 11 that is a deprecated class. Instead you should use std::shared_ptr, std::unique_ptr, or std::weak_ptr depending on the use case. I think std::unique_ptr is probably sufficient here, but if you are passing gsm into functions then std::shared_ptr might be better.

guidance/config/params.yaml
msmcconnell: There is one in nearly every directory of the repo just under the top level folder. You need to add it to all of them

itabrl: we need to discuss this further.

msmcconnell:
I think some unit tests are needed for the instance creation logic and the new lexus state machine

guidance/include/guidance/guidance_state_machine_factory.hpp
```diff
@@ -0,0 +1,40 @@
...
+             */
+            std::unique_ptr<GuidanceStateMachine> createHardwareEngagedAutomationInstance();
+
+            /*!
+             * \brief Create Guidance state machine for SoftwareOnlyEngagedStateMachine
+             */
+            std::unique_ptr<GuidanceStateMachine> createSoftwareOnlyEngagedStateMachineInstance();
+
+            /*!
+             * \brief Create Guidance state machine for SoftwareOnlyEngagedStateMachine
```
msmcconnell: Looks like a duplicated comment

guidance/src/guidance/guidance_worker.cpp
```diff
@@ -157,6 +157,16 @@ namespace guidance
         }
     }
 
+    void GuidanceWorker::create_guidance_state_machine()
+    {
+        try {
+            guidance_state_machine_factory.createStateMachineInstance(vehicle_state_machine_type);
```
msmcconnell: Looks like you forgot to assign the returned value here. 
Also I guess since you are returning the nullptr this try catch won't notice it. It would be better to have an if statment that checks for the nullptr and if it occurs then it calls nh_.handleException() and passes in an std::invalid_argument exception containing a log describing that an invalid parameter was set. This will cause the log to made and the node to shutdown cleanly.

PR #390, opened 2019-10-11
Feature/minimal arbitrator

Supraluminal:
# PR Details
## Description

Implementation of the minimal Arbitrator and mock Plan Delegator nodes. 

Adds the `arbitrator` package which contains both the arbitrator library (for unit testing) and arbitrator node proper. Arbitrator consists of a skeleton main function delegating execution flow to `arbitrator.cpp` which is primarily structural code that uses a generic `PlanningStrategy` to generate plans. `TreePlanner` is the only existing implementation of `PlanningStrategy` at present, and is itself generalized over `CostFunction`, `NeighborGenerator`, and `SearchStrategy` for testability and flexibility. The only implementations of these interfaces at present serve to implement a kind of Beam Search that uses the plugins to generate new states, and a list of fixed priorities to derive cost for each state. The search proceeds until a plan reaches the target length or if no new states can be expanded and the length is not reached, the longest plan is returned.

The Plan Delegator implementation is extremely minimalistic as this is not the focus of this task. It simply implements enough functionality to receive a maneuver plan and call the `plan_trajectory` service on the associated plugin, finally forwarding the new trajectory on to the Trajectory Executor node.

## Motivation and Context

These nodes form 2 of the remaining elements of the core CARMA3 planning flow and thus are needed for all future plugin-based planning operation.

## How Has This Been Tested?

25 unit tests have been written and passed for all algorithmically relevant components of the Arbitrator (i.e., non-structural, I/O, etc. code). The Plan Delegator node has no unit tests since most of it will be scrapped within the next sprint or two.

## Types of changes

## Checklist:

- [ ] All new and existing tests passed. (KR: Existing tests which failed prior to this change may still fail)


Supraluminal:
Note: The dependency on usdot-fhwa-stol/CARMAMsgs#19 causes unit tests to break when run against develop branch CARMAMsgs. Unit tests for Arbitrator pass locally when run against the correct CARMAMsgs:

```-- run_tests.py: execute commands
  /home/rushk/carma-ws/devel/lib/arbitrator/arbitrator-test --gtest_output=xml:/home/rushk/carma-ws/build/test_results/arbitrator/gtest-arbitrator-test.xml
[==========] Running 25 tests from 5 test cases.
[----------] Global test environment set-up.
[----------] 8 tests from ArbitratorStateMachineTest
[ RUN      ] ArbitratorStateMachineTest.testInitalize
[       OK ] ArbitratorStateMachineTest.testInitalize (0 ms)
[ RUN      ] ArbitratorStateMachineTest.testPlanningCycle
[       OK ] ArbitratorStateMachineTest.testPlanningCycle (0 ms)
[ RUN      ] ArbitratorStateMachineTest.testPauseResumeFromPlanning
[       OK ] ArbitratorStateMachineTest.testPauseResumeFromPlanning (0 ms)
[ RUN      ] ArbitratorStateMachineTest.testPauseResumeFromWaiting
[       OK ] ArbitratorStateMachineTest.testPauseResumeFromWaiting (0 ms)
[ RUN      ] ArbitratorStateMachineTest.testShutdown1
[       OK ] ArbitratorStateMachineTest.testShutdown1 (0 ms)
[ RUN      ] ArbitratorStateMachineTest.testShutdown2
[       OK ] ArbitratorStateMachineTest.testShutdown2 (0 ms)
[ RUN      ] ArbitratorStateMachineTest.testShutdown3
[       OK ] ArbitratorStateMachineTest.testShutdown3 (0 ms)
[ RUN      ] ArbitratorStateMachineTest.testShutdown4
[       OK ] ArbitratorStateMachineTest.testShutdown4 (0 ms)
[----------] 8 tests from ArbitratorStateMachineTest (0 ms total)

[----------] 3 tests from PluginNeighborGeneratorTest
[ RUN      ] PluginNeighborGeneratorTest.testInitalize
[       OK ] PluginNeighborGeneratorTest.testInitalize (0 ms)
[ RUN      ] PluginNeighborGeneratorTest.testGetNeighbors1
[       OK ] PluginNeighborGeneratorTest.testGetNeighbors1 (1 ms)
[ RUN      ] PluginNeighborGeneratorTest.testGetNeighbors2
[       OK ] PluginNeighborGeneratorTest.testGetNeighbors2 (0 ms)
[----------] 3 tests from PluginNeighborGeneratorTest (1 ms total)

[----------] 4 tests from FixedPriorityCostFunctionTest
[ RUN      ] FixedPriorityCostFunctionTest.testSingleManeuver
[       OK ] FixedPriorityCostFunctionTest.testSingleManeuver (0 ms)
[ RUN      ] FixedPriorityCostFunctionTest.testMixedPlanners
[       OK ] FixedPriorityCostFunctionTest.testMixedPlanners (0 ms)
[ RUN      ] FixedPriorityCostFunctionTest.testSingleManeuverUnitDistance
[       OK ] FixedPriorityCostFunctionTest.testSingleManeuverUnitDistance (1 ms)
[ RUN      ] FixedPriorityCostFunctionTest.testUnitDistanceCost
[       OK ] FixedPriorityCostFunctionTest.testUnitDistanceCost (0 ms)
[----------] 4 tests from FixedPriorityCostFunctionTest (1 ms total)

[----------] 7 tests from BeamSearchStrategyTest
[ RUN      ] BeamSearchStrategyTest.testEmptyList
[       OK ] BeamSearchStrategyTest.testEmptyList (0 ms)
[ RUN      ] BeamSearchStrategyTest.testSort1
[       OK ] BeamSearchStrategyTest.testSort1 (0 ms)
[ RUN      ] BeamSearchStrategyTest.testSort2
[       OK ] BeamSearchStrategyTest.testSort2 (0 ms)
[ RUN      ] BeamSearchStrategyTest.testSort3
[       OK ] BeamSearchStrategyTest.testSort3 (0 ms)
[ RUN      ] BeamSearchStrategyTest.testSortAndPrune1
[       OK ] BeamSearchStrategyTest.testSortAndPrune1 (0 ms)
[ RUN      ] BeamSearchStrategyTest.testSortAndPrune2
[       OK ] BeamSearchStrategyTest.testSortAndPrune2 (0 ms)
[ RUN      ] BeamSearchStrategyTest.testSortAndPrune3
[       OK ] BeamSearchStrategyTest.testSortAndPrune3 (0 ms)
[----------] 7 tests from BeamSearchStrategyTest (0 ms total)

[----------] 3 tests from TreePlannerTest
[ RUN      ] TreePlannerTest.testGeneratePlan1
[       OK ] TreePlannerTest.testGeneratePlan1 (1 ms)
[ RUN      ] TreePlannerTest.testGeneratePlan2
[       OK ] TreePlannerTest.testGeneratePlan2 (1 ms)
[ RUN      ] TreePlannerTest.testGeneratePlan3
[       OK ] TreePlannerTest.testGeneratePlan3 (1 ms)
[----------] 3 tests from TreePlannerTest (3 ms total)

[----------] Global test environment tear-down
[==========] 25 tests from 5 test cases ran. (5 ms total)
[  PASSED  ] 25 tests.```

File: arbitrator/include/arbitrator_state_machine.hpp [hunk 1]
```diff
@@ -0,0 +1,119 @@
...
+{
+    /**
+     * Possible states for the Arbitrator to be in:
+     * INITIAL- Startup state while the rest of the system is initalizing
+     * PLANNING - Actiively generating a maneuver plan
+     * WAITING - After the completion of a plan, waiting for a timer trigger to replan
+     * PAUSED - Similar to waiting, but will ignore a timer trigger. Must be resumed externally.
+     */
+    enum ArbitratorState {
+        INITIAL,
```
qswawrq: It would be more clear if you mark the first ENUM element as zero ```INITIAL = 0,```.

arbitrator/include/arbitrator_utils.hpp
```diff
@@ -0,0 +1,104 @@
...
+#include <ros/ros.h>
+#include <cav_msgs/ManeuverPlan.h>
+
+/**
+ * \brief Macro definition to enable easier access to fields shared across the maneuver typees
+ * \param mvr The maneuver object to invoke the accessors on
+ * \param property The name of the field to access on the specific maneuver types. Must be shared by all extant maneuver types
+ * \return Expands to an expression (in the form of chained ternary operators) that evalutes to the desired field
+ */
+#define GET_MANEUVER_PROPERTY(mvr, property)\
```
qswawrq: Maybe you can make it as a inline function instead of a Macro? In C++, the style guide says: "Avoid defining macros, especially in headers": https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros

arbitrator/include/capabilities_interface.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include <carma_utils/CARMAUtils.h>
+#include <vector>
+#include <map>
+#include <string>
+
+namespace arbitrator
+{
+    const std::string STRATEGIC_PLAN_CAPABILITY = "strategic_plan/plan_maneuvers";
+    const std::string TACTICAL_PLAN_CAPABILITY = "tactical_plan";
+    const std::string CONTROL_CAPABILITY = "control";
```
qswawrq: Does it makes sense to define those as const static in side CapabilitiesInterface class? Otherwise, it will be global to the entire arbitrator namespace.

arbitrator/include/capabilities_interface.tpp
```diff
@@ -0,0 +1,53 @@
...
+
+    /*template<typename Mmsg>
+    void CapabilitiesInterface::multiplex_publication_for_capability(std::string query_string, Mmsg msg) const
+    {
+        std::vector<std::string> topics = get_topics_for_capability(query_string);
+        for (auto i = topics.begin(); i != topics.end(); i++) 
+        {
+            publishers_[*i].publish(msg);
+        }
+    }*/
```
qswawrq: Do you want to remove those?

arbitrator/include/cost_function.hpp
```diff
@@ -0,0 +1,53 @@
...
+             * \brief Compute the unit cost over distance of a given maneuver plan
+             * \param plan The plan to evaluate
+             * \return double The total cost divided by the total distance of the plan
+             */
+            virtual double compute_cost_per_unit_distance(cav_msgs::ManeuverPlan plan) const = 0;
+
+            /**
+             * \brief Virtual destructor provided for memory safety
+             */
+            virtual ~CostFunction(){};
```
qswawrq: You can mark it as pure virtual as well to make sure it will never be called by any child class. This apply to all destructors you have in other abstract class.

arbitrator/include/capabilities_interface.tpp
```diff
@@ -0,0 +1,53 @@
...
+#define __CAPABILITIES_INTERFACE_TPP__
+
+#include <vector>
+#include <map>
+#include <string>
+#include <functional>
+
+namespace arbitrator 
+{
+    template<typename MSrv>
```
qswawrq: Not sure I get the reason why you template this class.

arbitrator/src/capabilities_interface.cpp
```diff
@@ -0,0 +1,89 @@
...
+                            break;
+                        case cav_msgs::Plugin::TACTICAL:
+                            capability = TACTICAL_PLAN_CAPABILITY;
+                            break;
+                        case cav_msgs::Plugin::CONTROL:
+                            capability = CONTROL_CAPABILITY;
+                            break;
+                        case cav_msgs::Plugin::UNKNOWN:
+                        default:
+                            throw new std::invalid_argument("Activated plugin of unknown type discovered: " + it->name);
```
qswawrq: Same comment for ```new```

arbitrator/src/fixed_priority_cost_function.cpp
```diff
@@ -0,0 +1,61 @@
...
+            if (it->second > max_priority)
+            {
+                max_priority = it->second;
+            }
+        }
+
+        // Normalize the list and invert into costs
+        for (auto it = plugin_priorities.begin(); it != plugin_priorities.end(); it++)
+        {
+            plugin_costs_[it->first] = 1.0 - (it->second / max_priority);
```
qswawrq: Will be a potential "Divide by Zero" error here if all priorities are not set and default to zero? This also apply to the following ```compute_cost_per_unit_distance``` function.

arbitrator/include/arbitrator_utils.hpp
Supraluminal: For this one in particular I don't believe it can be anything other than a macro, due to the fact that it's actually doing something syntactically invalid for normal C++ code (using that property parameter to determine which field to access on an object by name). 

I can define it only in the CPP file if it makes a difference, but I'll have to refactor the other uses of it to call function that uses the macro instead.

arbitrator/include/capabilities_interface.tpp
Supraluminal: It's just this function that's templated, but it's so it could be used for multiple different kinds of services calls (each needing a different kind of MSrv template parameter for their message type).

arbitrator/src/fixed_priority_cost_function.cpp
Supraluminal: Yeah, it would be. The alternative to throwing a divide by zero exception here would be to catch the exception (or check if it would be thrown) and then throw another exception though, so I'm not clear if that's any better than just throwing a divide by zero.

arbitrator/include/arbitrator_utils.hpp
qswawrq: After discussion, we will keep this but add a TODO note for future enhancement.

arbitrator/src/arbitrator.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    void Arbitrator::planning_state()
+    {
+        ros::Time planning_process_start = ros::Time::now();
+        cav_msgs::ManeuverPlan plan = planning_strategy_.generate_plan();
+        ros::Time plan_end_time = get_plan_end_time(plan);
+        ros::Time plan_start_time = get_plan_start_time(plan);
+        ros::Duration plan_duration = plan_end_time - plan_start_time;
+
+        if (plan_duration < min_plan_duration_)
```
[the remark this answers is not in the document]
Supraluminal: Good catch, I'll change this to warn for now. There's not a mechanism in place right now for Arbitrator to be aware of the end of the route, that's something we'd need to build out in the future as part of a proper Route system.

arbitrator/src/arbitrator_state_machine.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+    ArbitratorState ArbitratorStateMachine::get_state()
+    {
+        return current_state;        
+    }
+
+
+    ArbitratorState ArbitratorStateMachine::submit_event(ArbitratorEvent event)
+    {
+        for (auto iter = ARBITRATOR_TRANSITIONS.begin(); iter != ARBITRATOR_TRANSITIONS.end(); iter++)
```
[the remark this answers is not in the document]
Supraluminal: That's good to know, I wasn't aware of that bit of syntax!

arbitrator/src/beam_search_strategy.cpp
```diff
@@ -0,0 +1,38 @@
...
+            plans.end(), 
+            [] (std::pair<cav_msgs::ManeuverPlan, double> a, std::pair<cav_msgs::ManeuverPlan, double> b) 
+            {
+                return a.second < b.second;
+            }
+        );
+
+        if (plans.size() > beam_width_)
+        {
+            plans.resize(beam_width_);
```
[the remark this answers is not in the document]
Supraluminal: I don't think it should, since I'm only resizing if we're above the specified size. It should just truncate the list, destroying anything in positions `beam_width_` to `plans.size() - 1`. The unit tests seem to confirm this behavior as well.

arbitrator/include/arbitrator_state_machine.hpp
Supraluminal: I'm pretty sure this is implied, but I'll go ahead and specify it anyway.

arbitrator/include/cost_function.hpp
Supraluminal: Hmm, defining this as pure virtual seems to remove the ability for my child classes to use the implicit destructor (which seems cleaner to me). I think it'd rather leave this as is, if it's not a big deal.

arbitrator/include/capabilities_interface.tpp
Supraluminal: Not in the present, thought in the future I believe it probably will. For example it'll need to use a different service to interact with the cost plugins to get their costs for a given plan.

arbitrator/src/arbitrator.cpp
```diff
@@ -128,6 +128,6 @@ namespace arbitrator
     {
         ROS_INFO_STREAM("Arbitrator shutting down...");
         ros::shutdown();
-        exit(0);
+        std::exit(0);
```
qswawrq: Actually, I think if you get rid of this line directly, ros::shutdown() would stop the spin and the program will end because main function ended.

arbitrator/src/beam_search_strategy.cpp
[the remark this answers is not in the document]
qswawrq: Ah. Sorry, I missed your check above. You are right,

arbitrator/src/fixed_priority_cost_function.cpp
qswawrq: I think the easy way to handle it is to check if max_priority == 0 and assign its cost to infinity/maximum.

arbitrator/include/arbitrator.hpp
```diff
@@ -14,8 +14,8 @@
  * the License.
  */
 
-#ifndef __ARBITRATOR_HPP__
-#define __ARBITRATOR_HPP__
+#ifndef __ARBITRATOR_INCLUDE_ARBITRATOR_HPP__ARBITRATOR_INCLUDE_
+#define __ARBITRATOR_INCLUDE_ARBITRATOR_HPP__ARBITRATOR_INCLUDE_
```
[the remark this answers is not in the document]
Supraluminal: Yeah, I messed up my initial find and replace, but I thought I'd fixed it everywhere. Must have missed this spot.

arbitrator/src/arbitrator.cpp
Supraluminal: Good point, I'll go ahead and remove it.

maefromm:
@qswawrq feel free to merge to develop since freeze has been lifted now. whenever it's ready. thanks.

PR #409, opened 2019-11-01
Add re-engage capability to guidance state machine

### LINKED ISSUE #378: Fix the Operator Override state for SSC Wrapper and Guidance
### Types of Issue

### Descriptive summary
During the https://usdot-carma.atlassian.net/browse/CAR-1029 investigation, found that the SSC wrapper does not process the status="failure" for Operator Override after CARMA is engaged. 

Requested documentation from AStuff regarding the veh_controller state machine to ensure we fully understand the flow, before updating the SSC wrapper to evaluate this accordingly. 
### Expected behavior
1. SSC Wrapper should have the condition to evaluate the /hardware_interface/veh_contoller state = "failure" (Operator Override) and set the robot_active = False.  
2. Once that's set, the Guidance should then set the guidance/state to INACTIVE during Operator Override. 

### Actual behavior
robot_active remains True during Operator Override. 

1. Engage CARMA
2. Override using throttle/brake/steering.
3. CARMA guidance/state remains Engaged, instead of changing to INACTIVE.

### Related work

qswawrq:
# PR Details
## Description
This PR mainly corrects the transition from inactive state to engage in the software only engaged state machine. With https://github.com/usdot-fhwa-stol/CARMASscInterfaceWrapper/pull/36, it makes re-engage functional after manual override. This PR fixes #378 
Besides the above, this PR also cleans some unused comments in CMake file and package.xml and removes a unused function implementation in state machine.

## Related Issue
#378 

## Motivation and Context
Bug fix.

## How Has This Been Tested?
Integration tested on Ford Fusion.

## Types of changes

## Checklist:


msmcconnell:
Looks good. For the sonar cloud check you need to added a comment in the unimplemented functions
{//State machine does not implement this state}
Something like that.

guidance/src/guidance/guidance_worker.cpp
```diff
@@ -18,62 +18,65 @@
...
+        if(!req.guidance_active)
+        {
+            cav_srvs::SetEnableRobotic srv;
             srv.request.set = cav_srvs::SetEnableRobotic::Request::DISABLE;
-            res.guidance_status = false;
+            enable_client_.call(srv);
         }
-        enable_client_.call(srv);
+        gsm_.onSetGuidanceActive(req.guidance_active);
+        res.guidance_status = req.guidance_active;
```
msmcconnell: Shouldn't we set the guidance_status based on the result from onSetGuidanceActive

msmcconnell:
For SonarCloud. It looks like you need a unit test for the GuidanceWorker class. I think the spin_cb is your big culprit here.

guidance/src/guidance/guidance_worker.cpp
qswawrq: Good catch!

qswawrq:
> For SonarCloud. It looks like you need a unit test for the GuidanceWorker class. I think the spin_cb is your big culprit here.

Well. The worker class here is actually not a worker class. I just keep its name from ui_integration node. It is actually the ROS interface and guidance_node.cpp is the usual main.cpp file we used.
I guess I can move those logic code into the state machine to make it testable.

msmcconnell:
It passed! Will merge now

PR #429, opened 2019-11-20
Refactor plan delegator

qswawrq:
# PR Details
## Description

This PR refactors plan delegator into an well structured node. It keeps all its functionalities existed in the previous temporary version. This PR also adds rostest to it.
## Related Issue
N/A

## Motivation and Context
Upgrade plan delegator node into a structured CARMA node.

## How Has This Been Tested?
Unit test in progress

## Types of changes

## Checklist:


Supraluminal:
Is this PR for CAR-1218? If so, the big thing I'm not seeing is the iterative spinning over a plan. It looks like in current implementation (as was with the temporary one) it just generates 1 trajectory per 1 maneuver plan. I think we'd want the trajectory generation to happen at a configurable frequency so that the plugins planning the trajectory have opportunities to revise their plans over time. I could be missing where this is handled, though.

plan_delegator/CMakeLists.txt
```diff
@@ -1,9 +1,23 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
 cmake_minimum_required(VERSION 2.8.3)
 project(plan_delegator)
 
 ## Compile as C++11, supported in ROS Kinetic and newer
 add_compile_options(-std=c++11)
-set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
```
Supraluminal: Interesting that you need -pthread here, does ROS not automatically compile with pthread support? Could you describe a bit the high-level threading structure if you have one in here?

plan_delegator/src/plan_delegator.cpp
```diff
@@ -14,65 +14,169 @@
...
+    bool PlanDelegator::IsTrajectoryLongEnough(const cav_msgs::TrajectoryPlan& plan) const noexcept
+    {
+        return (plan.trajectory_points.back().target_time - plan.trajectory_points.begin()->target_time) * MILLISECOND_TO_SECOND >= max_trajectory_duration_;
+    }
+
+    cav_msgs::TrajectoryPlan PlanDelegator::PlanTrajectory()
+    {
+        cav_msgs::TrajectoryPlan latest_trajectory_plan;
+        // iterate through maneuver list to make service call
+        for(const auto& maneuver : latest_maneuver_plan_.maneuvers)
```
Supraluminal: I don't see any explicit multithreading so far, so I'm not super concerned, but if there's any that I've missed it would be important to ensure that there's a lock on the vector this for-each loop is iterating.

Only adding this since the `-pthread` compile option has be a bit nervous.

plan_delegator/include/plan_delegator.hpp
```diff
@@ -33,18 +43,105 @@
...
             /**
-             * \brief Initialize the plan delegator and run it
-             * \return Standard unix program exit code
+             * \brief Initialize the plan delegator
              */
-            int run();
+            void init();
+
+            /**
+             * \brief Run the main thread of plan delegator
```
Supraluminal: If there's no multithreading here (i.e. this method just runs in the main process) then I'd change the way this is phrased to reflect that, maybe `Run the main loop of the plan delegator` or something like that.

plan_delegator/include/plan_delegator.hpp
```diff
@@ -33,18 +43,102 @@
...
+            std::unordered_map<std::string, ros::ServiceClient> trajectory_planners_;
+            // local storage of incoming messages
+            cav_msgs::ManeuverPlan latest_maneuver_plan_;
+            geometry_msgs::PoseStamped latest_pose_;
+            geometry_msgs::TwistStamped latest_twist_;
+
+            /**
+             * \brief Callback function of maneuver plan subscriber
+             */
+            void ManeuverPlanCallback(const cav_msgs::ManeuverPlanConstPtr& plan);
```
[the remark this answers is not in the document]
qswawrq: Sure.

plan_delegator/src/plan_delegator.cpp
```diff
@@ -14,65 +14,168 @@
...
+        }
+        return plan_req;
+    }
 
-                cav_srvs::PlanTrajectory plan_req;
-                plan_req.request.maneuver_plan = *plan;
-                if (trajectory_planners[first_maneuver_planner].call(plan_req))
+    bool PlanDelegator::IsTrajectoryLongEnough(const cav_msgs::TrajectoryPlan& plan) const noexcept
+    {
+        return (plan.trajectory_points.back().target_time - plan.trajectory_points.begin()->target_time) * MILLISECOND_TO_SECOND >= max_trajectory_duration_;
```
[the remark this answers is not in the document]
qswawrq: Updated.

plan_delegator/include/plan_delegator.hpp
```diff
@@ -33,18 +43,105 @@
...
+
+            /**
+             * \brief Run the main thread of plan delegator
+             */
+            void run();
+
+            /**
+             * \brief Callback function of maneuver plan subscriber
+             */
+            void ManeuverPlanCallback(const cav_msgs::ManeuverPlanConstPtr& plan);
```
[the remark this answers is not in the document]
qswawrq: Updated.

plan_delegator/CMakeLists.txt
qswawrq: Ops. It is something I used to use but I do not think we still need it to be here.

plan_delegator/src/plan_delegator.cpp
qswawrq: Yup. It is a single thread node. I just removed the pthread option.

plan_delegator/include/plan_delegator.hpp
qswawrq: Sure. Updated.

qswawrq:
I pushed a commit to address your comments. It seems the CircleCI does not count my unit test properly (because my test is for a child class of PlanDelegator class). I will do some research on how to address that.

qswawrq:
This PR also fixes #450.

Issue #441, opened 2019-11-27
Implement interface for working with roadway information


msmcconnell:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
To  develop valid plans an understanding of the road geometry is needed. An interface should be created for this

PR #442, opened 2019-11-27
Feature/carma wm

msmcconnell:
# PR Details
## Description

Initial implementation of the carma_wm library for working with road networks (ie. lanelet2). 
This PR closely matches the original with some design tweaks based on the lanelet2 api. 

These are

- In the WorldModel classes Downtrack/crosstrack functions have been combined into 3 functions which return a TrackPos stuct. This contains the downtrack/crosstrack values and is easier to implement using existing lanelet2 functionality. 
- In the WMListener classes the getMutex function has been replaced with a getLock(bool) function that returns an std::unique_lock object which I believe will provide better user exception safety and simplicity of use. 
- The getLocalCurvatures function has been simplified to only calculate curvatures for every set of 3 points. To calculate an average curvature would require performing a polynomial fit which would take a larger amount of time to implement and might not provide major benefits. 

## Related Issue

#441 

## Motivation and Context

Implement world model interface design

## How Has This Been Tested?

Unit tests
## Types of changes

## Checklist:


msmcconnell:
The check is passing because I am still working on the unit tests and haven't update the sonar.properties file. Do not merge yet

carma_wm/include/carma_wm/WorldModel.h
```diff
@@ -0,0 +1,282 @@
...
+        * \return lanelet::ArcCoordinates with values copied from this object
+        */
+      inline lanelet::ArcCoordinates toArcCoordinates() {
+        return lanelet::ArcCoordinates{downtrack, -crosstrack};
+      }
+      
+      // Overload == and != operators
+      bool operator==(const TrackPos &other) const {
+        return this == &other 
+          || (this->downtrack == other.downtrack
```
Supraluminal: Does this really make sense as a raw double comparison? Should we consider adding an epsilon factor here to ensure some sanity?

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -0,0 +1,61 @@
...
+  void WMListenerWorker::mapCallback(autoware_lanelet2_msgs::MapBin map_msg) {
+
+    lanelet::LaneletMapPtr new_map(new lanelet::LaneletMap);
+
+    lanelet::utils::conversion::fromBinMsg(map_msg, new_map);
+
+    world_model_->setMap(new_map);
+
+    // Call user defined map callback
+    if (map_callback_) {
```
Supraluminal: Does an empty `std::function` object evaluate to false? Not sure what the initialization behaviors are for it, but it shouldn't be nullptr considering it's a class member, and would thus have it's default constructor invoked.

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -0,0 +1,61 @@
...
+      map_callback_();
+    }
+  }
+
+  void WMListenerWorker::routeCallback() {
+    // TODO Implement when route message has been defined
+    // world_model_->setRoute(route_obj);
+    // Call route_callback_;
+    if (route_callback_) {
```
Supraluminal: Same as above.

carma_wm/include/carma_wm/WMListener.h
```diff
@@ -0,0 +1,106 @@
+#pragma once
```
Supraluminal: Do we have a preference for `#pragma once` vs `#ifndef __...___` guards?

carma_wm/README.md
```diff
@@ -0,0 +1 @@
+TODO this readme and lanelet2 license notice
```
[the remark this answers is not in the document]
msmcconnell: Plan to after unit tests are completed

carma_wm/package.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>carma_wm</name>
+  <version>1.0.0</version>
+  <description>The carma_wm package</description>
+  <maintainer email="carma@todo.todo">carma</maintainer>
```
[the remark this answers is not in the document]
msmcconnell: Will do

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -0,0 +1,333 @@
...
+  }
+
+  LaneletRouteConstPtr CARMAWorldModel::getRoute() const {
+    return std::static_pointer_cast< const lanelet::routing::Route >(route_); // Cast pointer to const variant
+  }
+
+  void CARMAWorldModel::setMap(lanelet::LaneletMapPtr map) {
+    semantic_map_ = map;
+    // Build routing graph from map
+    lanelet::traffic_rules::TrafficRulesUPtr traffic_rules = lanelet::traffic_rules::TrafficRulesFactory::create(lanelet::Locations::Germany, lanelet::Participants::VehicleCar);
```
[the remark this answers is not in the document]
msmcconnell: It is the only one that currently exists, and I think it will work for our current use case. I will look into making a user story to flesh out a US one.

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+
+namespace carma_wm
+{
+
+  CARMAWorldModel::CARMAWorldModel() {}
+ 
+  CARMAWorldModel::~CARMAWorldModel() {}
+
+  std::tuple<TrackPos, TrackPos> CARMAWorldModel::routeTrackPos(lanelet::ConstArea& area) const {
```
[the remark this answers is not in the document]
msmcconnell: Hmm seems you are right. If I can find an easy way to auto fix it I will. Otherwise I'll leave it as is (since it is everywhere...)

carma_wm/include/carma_wm/WorldModel.h
msmcconnell: I believe in this case it should be a direct comparison for ==. This way a copy constructed object will evaluate as equal but a computed one will need the user to account for epsilons themselves and therefore be aware of the desired range.

carma_wm/src/WMListenerWorker.cpp
msmcconnell: It does

carma_wm/include/carma_wm/WMListener.h
msmcconnell: Shuwei and I have been using ```#pragma once``` but I don't have a big preference either way. Technically if we tries to compile this for something other than ubuntu it could cause issues if the compiler did not support pragma, but that seems unlikely at the moment.

msmcconnell:
@Supraluminal I have added unit tests to cover my changes and made the following adjustments based on your comments. 

- Added README content. Package.xml uses carma@dot.gov content
- Applied the ROS C++ clang-format to all source code. 
- After discovering problematic edge cases, I reworked the point to segment matching logic in the following way. 
  - Created matchSegment function which is an expensive but more robust algorithm for matching a point to a segment on a linestring. 
  - Since the matchSegment function is too slow to be used for the whole route I created the IndexedDistanceMap class which is used on route load to pre-compute distances along the route. This allows the routeTrackPos functions to be far more efficient. 
- I have also removed the getAreasBetween function as I discovered it is unlikely to be of much use as Areas in lanelet to are mostly for parking spots and similar non-road space. It had extremely similar logic to getLaneletsBetween so we can add it back in easily if needed in the future. 

I have squashed the previous set of WIP commits given the fairly substantial changes I had to make.

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -0,0 +1,644 @@
...
+
+  for (auto basic_p : line.basicLineString())
+  {
+    coppied_points.push_back(lanelet::Point3d(lanelet::utils::getId(), basic_p));
+  }
+
+  return lanelet::LineString3d(lanelet::utils::getId(), coppied_points);
+}
+
+void CARMAWorldModel::computeDowntrackReferenceLine()
```
Supraluminal: Just noting our discussion about the discontinuity in crosstrack whenever a lane change happens along the reference line, worth thinking of ways we can keep that in mind but I think your documentation is sufficient for now.

carma_wm/src/CARMAWorldModel.h
```diff
@@ -0,0 +1,150 @@
...
+  void setRoute(LaneletRoutePtr route);
+
+  ////
+  // Overrides
+  ////
+  std::pair<TrackPos, TrackPos> routeTrackPos(const lanelet::ConstArea& area) const override;
+
+  TrackPos routeTrackPos(const lanelet::ConstLanelet& lanelet) const override;
+
+  TrackPos routeTrackPos(const lanelet::BasicPoint2d& point) const override;
```
Supraluminal: Maybe add a note in the doc comment here about the potential for a discontinuous crosstrack

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -0,0 +1,644 @@
...
+  shortest_path_filtered_centerline_view_ = lanelet::utils::createMap(shortest_path_centerlines_);
+}
+
+LaneletRoutingGraphConstPtr CARMAWorldModel::getMapRoutingGraph() const
+{
+  return std::static_pointer_cast<const lanelet::routing::RoutingGraph>(map_routing_graph_);  // Cast pointer to const
+                                                                                              // variant
+}
+
+// NOTE: See header file for details on source of logic in this function
```
Supraluminal: I might just have missed it but I don't see the note in the header file that's mentionedh ere.

carma_wm/src/IndexedDistanceMap.h
```diff
@@ -0,0 +1,195 @@
...
+public:
+  /*!
+   * \brief Add a linestring to this structure. This function will iterate over the line string to compute distances
+   * between each segment.
+   *
+   * \throws std::invalid_argument If the provided linstring was already added to this structure
+   *
+   * \param ls The linestring to add
+   */
+  void pushBack(const lanelet::LineString2d& ls)
```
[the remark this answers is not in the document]
msmcconnell: Given the minimal functionality in 90% of the functions I figured a header file was easier since it was internal to the library. I can change it if you want.

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -0,0 +1,644 @@
...
+    if (tp_next.downtrack >= 0 || ls_i == 0)
+    {
+      bestRouteSegId = lineString_1.id();
+      tp = tp_next;
+      // If downtrack is positive then we are on the correct segment
+    }
+    else
+    {
+      // If downtrack is negative then iterate over route segments to find preceeding segment
+      const size_t prev_ls_i = ls_i - 1;
```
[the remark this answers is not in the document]
msmcconnell: prev_ls_i will never be negative because of the ls_i == 0 check above.

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -0,0 +1,644 @@
...
+    {
+      // If downtrack is less then seg length then we are on the correct segment
+      bestRouteSegId = lineString_1.id();
+      tp = tp_prev;
+      tp.downtrack += shortest_path_distance_map_.distanceToPointAlongElement(ls_i, lineString_1.size() - 2);
+    }
+    else
+    {
+      // If downtrack is greater then seg length then we need to find the succeeding segment
+      auto next_centerline = lanelet::utils::to2D(shortest_path_centerlines_[ls_i + 1]);  // Get prev centerline
```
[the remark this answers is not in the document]
msmcconnell: ls_i == shortest_path_centerlines_.size() - 1 check above will prevent that

carma_wm/src/CARMAWorldModel.cpp
msmcconnell: It's in the WorldModel.h file but I can see how this note is unclear. I'll make it more explicit.

carma_wm/src/CARMAWorldModel.h
msmcconnell: Will do

carma_wm/src/IndexedDistanceMap.h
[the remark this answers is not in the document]
Supraluminal: Since it's internal, it's probably fine. Was just curious what the reason was.

carma_wm/src/CARMAWorldModel.cpp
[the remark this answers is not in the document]
Supraluminal: Ah, I follow now. I missed that in my read through but that seems to cover the case I'm thinking of.

PR #449, opened 2019-12-02
DO NOT MERGE

jtbaird:
This branch is to resolve the current issues in sonarqube. Merging this with another branch will cause that branch's tests to fail. This PR is simply to trigger to a CircleCI build.

### DIFF
--- a/bsm_generator/test/test_bsm_generator_worker.cpp
+++ b/bsm_generator/test/test_bsm_generator_worker.cpp
@@ -43,6 +43,12 @@ TEST(BSMWorkerTest, testMsgId)
     EXPECT_TRUE(msgId2 != msgId3);
 }
 
+/* added as part of sonarqube bug fix */
+TEST(BSMWorkerTest, testSecMark)
+{
+    EXIT_FAILURE;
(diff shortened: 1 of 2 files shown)


bsm_generator/test/test_bsm_generator_worker.cpp
```diff
@@ -43,6 +43,12 @@ TEST(BSMWorkerTest, testMsgId)
     EXPECT_TRUE(msgId2 != msgId3);
 }
 
+/* added as part of sonarqube bug fix */
+TEST(BSMWorkerTest, testSecMark)
```
msmcconnell: You might get compile errors because this Test has the same name as the next function. I suggest naming it something else. 
fakeTest instead of testSecMark

jtbaird: Okay, I'll go ahead and change that.

maefromm:
@jtbaird what's the status of this? Can this now be merged, or should this just be closed?

jtbaird:
we can close this now and circle back to it later

PR #458, opened 2019-12-11
Bugfix/arbitrator planning rate

### LINKED ISSUE #451: Arbitrator node publishes maneuver plan every 10 seconds
### Types of Issue

### Descriptive summary
Arbitrator node publishes maneuver plan every 10 seconds in guidance ENGAGE state, which is significantly lower than expect. It also causes the plan delegator deems some maneuver plan as expired (because the end time of the plan is later than the current time).
### Expected behavior
Arbitrator node publishes maneuver plan at 1 Hz after entering planning state.
### Software version this applies to

3.1.0
### Actual behavior
Arbitrator node publishes maneuver plan every 10 seconds in guidance ENGAGE state.
Launch CARMAPlatform in docker and use rostopic hz on "guidance/final_maneuver_plan" topic.
### Related work
N/A

Supraluminal:
# PR Details
## Related Issue

Fixes #451 

## Motivation and Context

As per issue #451 the Arbitrator <--> Autoware Plugin interaction was not occurring at the correct frequency (1hz) and was instead occuring at a much lower (0.1hz) frequency. This issue was the result of a confluence of factors:

1. Arbitrator internal timing loop configuration caused it to loop far tighter than anticipated, even though service calls were failing. This issue was masked by the two other factors.
2. A comparison of plan duration ended up comparing two doubles that were supposed to be the same (`ros::Duration(15)` vs the duration of a plan from `ros::Time::now()` to `ros::Time::now() + ros::Duration(15)` but depending on the value of ros::Time::now(), this comparison failed several times before it passed.
3. Plugin discovery was not clearing already discovered plugins from the records when it received new plugin data, thus effectively registering a plugin multiple times and causing extra services calls.

## How Has This Been Tested?

Tested locally via an integration test with the `Arbitrator`, `Health Monitor`, and `Autoware Plugin` nodes.

```rostopic hz /guidance/final_maneuver_plan
subscribed to [/guidance/final_maneuver_plan]
no new messages
no new messages
average rate: 0.953
	min: 1.049s max: 1.049s std dev: 0.00000s window: 2
average rate: 0.952
	min: 1.049s max: 1.051s std dev: 0.00057s window: 3
average rate: 0.952
	min: 1.049s max: 1.051s std dev: 0.00056s window: 4
average rate: 0.953
	min: 1.049s max: 1.051s std dev: 0.00076s window: 5
average rate: 0.952
	min: 1.049s max: 1.051s std dev: 0.00070s window: 6
average rate: 0.952
	min: 1.049s max: 1.051s std dev: 0.00064s window: 7
average rate: 0.953
	min: 1.049s max: 1.051s std dev: 0.00062s window: 8
average rate: 0.952
	min: 1.049s max: 1.051s std dev: 0.00073s window: 9
average rate: 0.952
	min: 1.049s max: 1.051s std dev: 0.00070s window: 10
average rate: 0.952
	min: 1.049s max: 1.051s std dev: 0.00067s window: 11
average rate: 0.952
	min: 1.049s max: 1.051s std dev: 0.00070s window: 12
average rate: 0.952
	min: 1.049s max: 1.051s std dev: 0.00069s window: 13
average rate: 0.952
	min: 1.049s max: 1.051s std dev: 0.00071s window: 14
average rate: 0.952
	min: 1.049s max: 1.051s std dev: 0.00070s window: 15
```

## Types of changes

## Checklist:

### DIFF
--- a/arbitrator/src/arbitrator.cpp
+++ b/arbitrator/src/arbitrator.cpp
@@ -26,21 +26,26 @@ namespace arbitrator
 {
     void Arbitrator::run()
     {
+        ROS_INFO("Aribtrator started, beginning arbitrator state machine.");
         while (!ros::isShuttingDown())
         {
             ros::spinOnce();
             switch (sm_->get_state()) 
             {
                 case INITIAL:
+                    ROS_INFO("Aribtrator spinning in INITIAL state.");
                     initial_state();
                     break;
                 case PLANNING:
+                    ROS_INFO("Aribtrator spinning in PLANNING state.");
                     planning_state();
                     break;
                 case WAITING:
+                    ROS_INFO("Aribtrator spinning in WAITING state.");
                     waiting_state();
                     break;
                 case PAUSED:
+                    ROS_INFO("Aribtrator spinning in PAUSED state.");
                     paused_state();
                     break;
                 default:
@@ -63,6 +68,7 @@ namespace arbitrator
                 // NO-OP
                 break;
             case cav_msgs::GuidanceState::ENGAGED:
+                ROS_INFO("Received notiace that guidance has been engaged!");
                 if (sm_->get_state() == ArbitratorState::INITIAL) {
                     sm_->submit_event(ArbitratorEvent::SYSTEM_STARTUP_COMPLETE);
                 } else if (sm_->get_state() == ArbitratorState::PAUSED) {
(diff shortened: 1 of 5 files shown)


arbitrator/src/arbitrator.cpp
```diff
@@ -94,23 +103,33 @@ namespace arbitrator
...
+        else
         {
-            ROS_INFO_STREAM("Arbitrator is publishing plan " << plan.maneuver_plan_id << " of duration " << plan_duration << " as current maneuver plan");
+            ROS_WARN("Arbitrator was unable to generate a plan!");
         }
-        final_plan_pub_.publish(plan);
 
         next_planning_process_start_ = planning_process_start + time_between_plans_;
+
+        sm_->submit_event(ArbitratorEvent::PLANNING_COMPLETE);
```
qswawrq: What happens after you submit this PLANNING_COMPLETE event? I do not think we want state machine to transit to another state?

Supraluminal: After we submit `PLANNING_COMPLETE` the state machine transitions to `WAITING` state, where it stays for the remainder of the time until the next planning cycle. It'll sleep for that much time, the transition back to `PLANNING` state.

qswawrq: Cool!

PR #560, opened 2020-02-12
Route Following Plugin

### LINKED ISSUE #559: Build Route Following Strategic Plugin
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)

### Descriptive summary
To utilize lanelet2 route and replace autoware_plugin, we need to build a route following feature. The top level strategic plugin is called route following plugin. The current implementation only supports lane keeping functionality and will log a warning when a lane-change maneuver is required on the route.
### Expected behavior
N/A
### Software version this applies to

3.2.0
### Actual behavior
N/A
N/A

### Related work

qswawrq:
# PR Details
## Description
To resolve #559, this is the initial implementation of route following plugin.

## Related Issue
#559 

## Motivation and Context
Adding a new feature.

## How Has This Been Tested?
All unit tests passed.

## Types of changes

## Checklist:

### DIFF
--- a/route_following_plugin/include/route_following_plugin.h
+++ b/route_following_plugin/include/route_following_plugin.h
@@ -33,22 +33,50 @@ namespace route_following_plugin
 
     public:
 
+        // constant speed limit for test
+        // TODO Once world_model and vector map is ready, it should be removed
         static constexpr double TWENTY_FIVE_MPH_IN_MS = 11.176;
 
+        /**
+         * \brief Default constructor for RouteFollowingPlugin class
+         */
         RouteFollowingPlugin();
 
-        // general starting point of this node
+        /**
+         * \brief General entry point to begin the operation of this class
+         */
         void run();
 
+        /**
+         * \brief Given a LaneletPath object, find index of the lanelet which has target_id as its lanelet ID
+         * \param target_id The laenlet ID this function is looking for
+         * \param path A list of lanelet with different lanelet IDs
+         * \return Index of the target lanelet in the list
+         */
         int findLaneletIndexFromPath(int target_id, lanelet::routing::LaneletPath& path);
 
+        /**
+         * \brief Compose a lane keeping maneuver message based on input params
+         * \param current_dist Start downtrack distance of the current maneuver
+         * \param end_dist End downtrack distance of the current maneuver
+         * \param current_speed Start speed of the current maneuver
+         * \param lane_id Lanelet ID of the current maneuver
+         * \param current_time Start time of the current maneuver
+         * \return A lane keeping maneuver message which is ready to be published
+         */
         cav_msgs::Maneuver composeManeuverMessage(double current_dist, double end_dist, double current_speed, int lane_id, ros::Time current_time);
 
+        /**
+         * \brief Given a LaneletRelations and ID of the next lanelet in the shortest path
+         * \param relations LaneletRelations relative to the previous lanelet
+         * \param target_id ID of the next lanelet in the shortest path
+         * \return Whether we need a lanechange to reach to the next lanelet in the shortest path
+         */
         bool identifyLaneChange(lanelet::routing::LaneletRelations relations, int target_id);
 
     private:
 
-        // node handles
+        // CARMA ROS node handles
         std::shared_ptr<ros::CARMANodeHandle> nh_, pnh_;
 
         // ROS publishers and subscribers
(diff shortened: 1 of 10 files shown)


route_following_plugin/include/route_following_plugin.h
```diff
@@ -0,0 +1,134 @@
...
+namespace route_following_plugin
+{
+
+    class RouteFollowingPlugin
+    {
+
+    public:
+
+        // constant speed limit for test
+        // TODO Once world_model and vector map is ready, it should be removed
```
Supraluminal: We might want to think about whether this data will be available by the time this integration branch gets merged into develop. I think this world model and vector map data is a requirement for the tactical plugin to work, and thus we'll need that to function before we can merge this. Since the interface is already there, is it possible to start using the speed limits from that in this strategic plugin?

route_following_plugin/package.xml
```diff
@@ -0,0 +1,33 @@
...
+ WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ License for the specific language governing permissions and limitations under
+ the License.
+-->
+
+<package format="2">
+  <name>route_following_plugin</name>
+  <version>3.2.0</version>
+  <description>The strategic plugin generates maneuvers to make the vehicle follow the route. </description>
+  <maintainer email="carma@todo.todo">carma</maintainer>
```
Supraluminal: I'd update both of these emails to `carma@dot.gov` now that we have that email operational.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -0,0 +1,176 @@
...
+{
+    RouteFollowingPlugin::RouteFollowingPlugin() : mvr_duration_(16.0), current_speed_(0.0) { }
+
+    void RouteFollowingPlugin::initialize()
+    {
+        
+        nh_.reset(new ros::CARMANodeHandle());
+        pnh_.reset(new ros::CARMANodeHandle("~"));
+        
+        plan_maneuver_srv_ = nh_->advertiseService("plugins/RouteFollowingPlugin/plan_maneuvers", &RouteFollowingPlugin::plan_maneuver_cb, this);
```
Supraluminal: Should this just advertise `strategic_plan/plan_maneuvers` or `~/strategic_plan/plan_maneuvers` (I can never recall which one is which) so that it can be remapped in the launch file to the appropriate place?

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -0,0 +1,176 @@
...
+        
+        pnh_->param<double>("minimal_maneuver_duration", mvr_duration_, 16.0);
+
+        wml_.reset(new carma_wm::WMListener());
+        // set world model point form wm listener
+        wm_ = wml_->getWorldModel();
+
+        ros::CARMANodeHandle::setSpinCallback([this]()
+        {
+            plugin_discovery_pub_.publish(plugin_discovery_msg_);
```
Supraluminal: Is publishing this discovery message with `activated = false` every timestep going to cause issues? If the user activates it from the UI, would publishing this message with `activated = false` deactivate it with the plugin manager?

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -0,0 +1,176 @@
...
+        }
+        auto current_lanelet = current_lanelets[0];
+        auto shortest_path = wm_->getRoute()->shortestPath();
+        double current_progress = wm_->routeTrackPos(current_loc).downtrack;
+        double speed_progress = current_speed_;
+        double total_maneuver_length = current_progress + mvr_duration_ * RouteFollowingPlugin::TWENTY_FIVE_MPH_IN_MS;
+        int last_lanelet_index = findLaneletIndexFromPath(current_lanelet.second.id(), shortest_path);
+        if(last_lanelet_index == -1)
+        {
+            ROS_WARN_STREAM("Cannot find current lanelet in shortest path!");
```
Supraluminal: This is the case where we'd expect `RouteManager` or someone else to throw a left route error, correct? Just making sure I understand.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -0,0 +1,176 @@
...
+        {
+            if(path[i].id() == target_id)
+            {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    cav_msgs::Maneuver RouteFollowingPlugin::composeManeuverMessage(double current_dist, double end_dist, double current_speed, int lane_id, ros::Time current_time)
```
Supraluminal: Assuming we can eventually get the speed limits (or other target speed) from the World Model, you'd need to parameterize this method for target speed too, correct? Might be a better idea to add target speed as a parameter here, and then if we do need to use the placeholder `RouteFollowingPlugin::TWENTY_FIVE_MPH_IN_MS` value for now we can do that wherever we'd be invoking this method.

route_following_plugin/include/route_following_plugin.h
qswawrq: On one hand, there is no interface in the world_model to expose traffic rule. I believe it is still in progress. One the other hand, I am not sure if Samir finished his task on adding speed limit into vector map. Those two things are what I am waiting on. I agree it should not be merged into develop before we have those. Right now it targets to the routing integration branch.

route_following_plugin/src/route_following_plugin.cpp
qswawrq: Sure.

route_following_plugin/package.xml
qswawrq: Sure.

route_following_plugin/src/route_following_plugin.cpp
qswawrq: I do not think so. Health Manager only updates a plugin's activity based on user selection (service call from UI). This field will be ignored.

route_following_plugin/src/route_following_plugin.cpp
qswawrq: If we stay on the route, it will not be trigger. Otherwise, we expect route manager to flag left_route first.

route_following_plugin/src/route_following_plugin.cpp
qswawrq: Sure. Make sense to me.

qswawrq:
Last commit should address all your comments.

route_following_plugin/include/route_following_plugin.h
Supraluminal: I'm not sure if we have a separate story for integrating that, or how the best way to track that is then. I just want to make sure we don't lose track of that somewhere as we'll really need to use that functionality once that data is available. Do you think that fits within the scope of another story, I don't want to hold this one open for a single issue if that's avoidable.

qswawrq: Does it make sense to do it under that wrap-all integration test? We probably are ready at that time.

Supraluminal:
We'll plan to create a new story to implement the speed limit integration with the CARMA world model once that data is exposed, for now we can close out this PR and story.

Issue #603, opened 2020-03-17
Add MPC Follower Wrapper


saina-ramyar:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
The MPC Wrapper node is developed to enable CARMA to use Autoware MPC follower node.
### Expected behavior
Code builds
### Software version this applies to

3.3.0
### Related work

Link to related tickets or prior related work here.

msmcconnell:
Resolved by above PRs

PR #604, opened 2020-03-17
Feature/mpc wrapper

saina-ramyar:
# PR Details
## Description
Added MPC Follower Wrapper Node to enable CARMA to use Autoware MPC.

created new mpc follower wrapper node
adjusted default control plugins in trajectory executor
update guidance launch file

## Related Issue
#603 603

## Motivation and Context
Model Predictive Control (MPC) has a better path tracking performance compared to the previous algorithm (pure pursuit).

Improve tracking performance.

## How Has This Been Tested?
Integration tested on all three types of passenger vehicles.

## Types of changes

## Checklist:


Dockerfile
```diff
@@ -54,7 +54,7 @@ RUN ~/carma_ws/src/CARMAPlatform/docker/install.sh
 # Stage 3 - Finalize deployment
 # /////////////////////////////////////////////////////////////////////////////
 
-FROM usdotfhwastol/autoware.ai:3.7.0
+FROM usdotfhwastol/autoware.ai:latest
```
msmcconnell: Don't commit latest tags. They don't correspond to existing docker images.

mpc_follower_wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,216 @@
...
+#   FILES
+#   Action1.action
+#   Action2.action
+# )
+
+## Generate added messages and services with any dependencies listed here
+# generate_messages(
+#   DEPENDENCIES
+#   cav_msgs#   std_msgs
+# )
```
msmcconnell: You can remove all these unused commend blocks

mpc_follower_wrapper/include/mpc_follower_wrapper/mpc_follower_wrapper.hpp
```diff
@@ -0,0 +1,117 @@
...
+        
+        // SyncPolicy
+        typedef message_filters::sync_policies::ApproximateTime<geometry_msgs::PoseStamped, cav_msgs::TrajectoryPlan> SyncPolicy;
+
+        void TrajectoryPlanPoseHandler(const geometry_msgs::PoseStamped::ConstPtr& pose, const cav_msgs::TrajectoryPlan::ConstPtr& tp);
+
+    private:
+
+        //@brief ROS node handle.
+        ros::NodeHandle& nh_;
```
msmcconnell: All new nodes in the carma platform repo should use CARMANodeHandle from carma_utils.

pure_pursuit_wrapper/launch/pure_pursuit_wrapper.launch
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <launch>
-    <remap from="final_waypoints" to="carma_final_waypoints"/>
+    <remap from="final_waypoints" to="/guidance/carma_final_waypoints"/>
```
msmcconnell: Why was the pure_pursuit wrapper changed? This shouldn't be necessary

mpc_follower_wrapper/src/mpc_follower_wrapper.cpp
```diff
@@ -0,0 +1,122 @@
...
+      lane.waypoints = waypoints;
+      PublisherForWayPoints(lane);
+    }
+    catch(const std::exception& e) {
+      HandleException(e);
+    }
+
+};
+
+void MPCFollowerWrapper::SystemAlertHandler(const cav_msgs::SystemAlert::ConstPtr& msg) {
```
msmcconnell: If you use CARMANodeHandle this function, HandleException, and the Shutdown functions don't need to be implemented.

mpc_follower_wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,216 @@
...
+## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)
+## is used, also find other catkin packages
+find_package(catkin REQUIRED COMPONENTS
+  cav_msgs
+  roscpp
+  rospy
+  std_msgs
+  geometry_msgs
+  autoware_msgs
+  autoware_config_msgs
```
msmcconnell: This message set is never used in this package and can be removed from here the code and package.xml

mpc_follower_wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,216 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(mpc_follower_wrapper)
+
+## Compile as C++11, supported in ROS Kinetic and newer
+add_compile_options(-std=c++11)
+set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
+set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall")
+add_definitions(-DBOOST_LOG_DYN_LINK=1)
+set(CMAKE_BUILD_TYPE Debug)
+set(CMAKE_CXX_FLAGS "-lboost_system")
```
msmcconnell: No sure why these lines are nessasary
```
add_definitions(-DBOOST_LOG_DYN_LINK=1)
set(CMAKE_BUILD_TYPE Debug)
set(CMAKE_CXX_FLAGS "-lboost_system") 
```

None of our other boost packages use them

mpc_follower_wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,216 @@
...
+set(CMAKE_CXX_FLAGS "-lboost_system") 
+
+## Find catkin macros and libraries
+## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)
+## is used, also find other catkin packages
+find_package(catkin REQUIRED COMPONENTS
+  cav_msgs
+  roscpp
+  rospy
+  std_msgs
```
msmcconnell: This package is never used and can be removed in all places.

mpc_follower_wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,216 @@
...
+  geometry_msgs
+  autoware_msgs
+  autoware_config_msgs
+  message_filters
+)
+
+## System dependencies are found with CMake's conventions
+find_package(Boost REQUIRED COMPONENTS system)
```
msmcconnell: I think this can probably just be 
find_package(Boost REQUIRED)

mpc_follower_wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,216 @@
...
+## The catkin_package macro generates cmake config files for your package
+## Declare things to be passed to dependent projects
+## INCLUDE_DIRS: uncomment this if your package contains header files
+## LIBRARIES: libraries you create in this project that dependent projects also need
+## CATKIN_DEPENDS: catkin_packages dependent projects also need
+## DEPENDS: system dependencies of this project that dependent projects also need
+catkin_package(
+ INCLUDE_DIRS include
+ LIBRARIES mpc_follower_wrapper_worker_library
+ CATKIN_DEPENDS cav_msgs roscpp rospy std_msgs geometry_msgs autoware_msgs
```
msmcconnell: Clean this line up to match the find_package(catkin REQUIRED COMPONENTS) line

mpc_follower_wrapper/src/mpc_follower_wrapper.cpp
```diff
@@ -0,0 +1,122 @@
...
+  // Pose Subscriber
+  pose_sub.subscribe(nh_, "current_pose", 1);
+  // Trajectory Plan Subscriber
+  trajectory_plan_sub.subscribe(nh_, "trajectory_plan", 1);
+
+  // WayPoints Publisher
+  way_points_pub_ = nh_.advertise<autoware_msgs::Lane>("final_waypoints", 10, true);
+}
+
+bool MPCFollowerWrapper::ReadParameters() {
```
msmcconnell: I don't see the point of this. It always returns true and the one time it is used is before Initialize is called so if it did what the comment said it really should have returned false

mpc_follower_wrapper/launch/mpc_follower_wrapper.launch
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
+<launch>
+    <remap from="final_waypoints" to="/guidance/carma_final_waypoints"/>
```
msmcconnell: Since this launch file is included in guidance.launch this can just be 
<remap from="final_waypoints" to="carma_final_waypoints"/>

The /guidance namespace is applied at runtime

mpc_follower_wrapper/launch/mpc_follower_wrapper.launch
```diff
@@ -0,0 +1,18 @@
...
+    <!-- Pure Pursuit Node -->
```
msmcconnell: Typo comment

mpc_follower_wrapper/launch/mpc_follower_wrapper.launch
```diff
@@ -0,0 +1,18 @@
...
+    <group>
+        <!--<remap from="vehicle_info/wheel_base" to="/vehicle_wheel_base"/>-->
```
msmcconnell: I think this is required

mpc_follower_wrapper/src/mpc_follower_wrapper.cpp
```diff
@@ -0,0 +1,122 @@
...
+        cav_msgs::TrajectoryPlanPoint t2 = tp->trajectory_points[i + 1];
+        autoware_msgs::Waypoint waypoint = mpcww.TrajectoryPlanPointToWaypointConverter(current_time, *pose,t1, t2);
+        waypoints.push_back(waypoint);
+      }
+
+      lane.waypoints = waypoints;
+      PublisherForWayPoints(lane);
+    }
+    catch(const std::exception& e) {
+      HandleException(e);
```
msmcconnell: After using CARMANodeHandle you can change this to 
CARMANodeHandle::handleException(e);

Normally CARMANodeHandle setsup the try/catch for you but using message_filters is a special case so they are still required.

carmajava/launch/guidance.launch
```diff
@@ -92,8 +92,10 @@
   <!-- Trajectory Executor -->
   <include file="$(find trajectory_executor)/launch/trajectory_executor.launch" />
 
-  <!-- Pure Pursuit Wrapper -->
-  <include file="$(find pure_pursuit_wrapper)/launch/pure_pursuit_wrapper.launch" />
+
+  <rosparam command="load" file="$(arg vehicle_calibration_dir)/mpc_follower/params.yaml"/>
+  <!-- MPC Follower Wrapper -->
+  <include file="$(find mpc_follower_wrapper)/launch/mpc_follower_wrapper.launch" />
```
msmcconnell: Probably would be better to add and argument to mpc_follower_wrapper.launch that takes in the vehicle_calibration_dir. Then you can move the <rosparam> line inside that launch file. Keeps the launch file self consistent.

mpc_follower_wrapper/src/mpc_follower_wrapper.cpp
```diff
@@ -0,0 +1,122 @@
...
+};
+
+void MPCFollowerWrapper::PublisherForWayPoints(autoware_msgs::Lane& msg){
+  try {
+    ROS_DEBUG_STREAM("Sending WayPoints message.");
+    way_points_pub_.publish(msg);
+  }
+  catch(const std::exception& e) {
+    HandleException(e);
+  }
```
msmcconnell: By using CARMANodeHandle you can remove this try/catch

mpc_follower_wrapper/src/mpc_follower_wrapper_node.cpp
```diff
@@ -0,0 +1,42 @@
...
+  ros::init(argc, argv, "mpc_follower_wrapper_node");
+  ros::NodeHandle nh("");
+
+  mpc_follower_wrapper::MPCFollowerWrapper MPCFollowerWrapper(nh);
+
+  // Approximate time of 100ms used because NDT outputs at 10Hz
+  message_filters::Synchronizer<mpc_follower_wrapper::MPCFollowerWrapper::SyncPolicy> sync(mpc_follower_wrapper::MPCFollowerWrapper::SyncPolicy(100), MPCFollowerWrapper.pose_sub, MPCFollowerWrapper.trajectory_plan_sub);
+  sync.registerCallback(boost::bind(&mpc_follower_wrapper::MPCFollowerWrapper::TrajectoryPlanPoseHandler, &MPCFollowerWrapper, _1, _2));
+
+  while (ros::ok() && !MPCFollowerWrapper.shutting_down_) {
```
msmcconnell: After using CARMANodeHandle this loop can be replaced with 
CARMANodeHandle::setSpinRate(rate);
CARMANodeHandle::spin();

Also the rate variable you have at the moment seems to be unused/unset.

mpc_follower_wrapper/src/mpc_follower_wrapper_node.cpp
```diff
@@ -0,0 +1,42 @@
...
+    ros::spinOnce();
+  }
+
+  ROS_INFO("Successfully launched node.");
```
msmcconnell: This log doesn't make sense as the node would exit as soon as it reached this point. 
"Node exiting" would be more appropriate.

mpc_follower_wrapper/include/mpc_follower_wrapper/mpc_follower_wrapper.hpp
msmcconnell: https://github.com/usdot-fhwa-stol/CARMAPlatform/blob/develop/health_monitor/src/health_monitor.cpp

Has an example of using it

mpc_follower_wrapper/launch/mpc_follower_wrapper.launch
saina-ramyar: updated it with the mpc wheelbase parameter name. However, I tried it before during the integration test and did not update the parameter.

mpc_follower_wrapper/src/mpc_follower_wrapper.cpp
saina-ramyar: I wrote this node based on pure pursuit wrapper and this was part of that, so how fo you suggest I modify it?

mpc_follower_wrapper/launch/mpc_follower_wrapper.launch
```diff
@@ -1,17 +1,21 @@
 <?xml version="1.0"?>
 <launch>
     <remap from="final_waypoints" to="/guidance/carma_final_waypoints"/>
```
msmcconnell: You can remove the leading /guidance/ from this remapping

carmajava/launch/guidance.launch
```diff
@@ -93,7 +93,9 @@
   <include file="$(find trajectory_executor)/launch/trajectory_executor.launch" />
 
 
-  <rosparam command="load" file="$(arg vehicle_calibration_dir)/mpc_follower/params.yaml"/>
+  <remap from="/vehicle_status" to="/hardware_interface/vehicle_status"/>
+  <remap from="/current_pose" to="/localization/current_pose"/>
```
msmcconnell: This should be 
```
<remap from="/current_pose" to="$(optenv CARMA_LOCZ_NS)/current_pose"/>
```

carmajava/launch/guidance.launch
```diff
@@ -93,7 +93,9 @@
   <include file="$(find trajectory_executor)/launch/trajectory_executor.launch" />
 
 
-  <rosparam command="load" file="$(arg vehicle_calibration_dir)/mpc_follower/params.yaml"/>
+  <remap from="/vehicle_status" to="/hardware_interface/vehicle_status"/>
```
msmcconnell: This should be
```
<remap from="/vehicle_status" to="$(optenv CARMA_INTR_NS)/vehicle_status"/>
```

mpc_follower_wrapper/src/mpc_follower_wrapper.cpp
msmcconnell: Seems you could just get rid of this function entirely

mpc_follower_wrapper/test/mpc_follower_wrapper_test.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+    tpp.x = 0;
+    tpp.y = 0;
+    tpp.target_time = 0;
+
+    tpp2.x = 1;
+    tpp2.y = -1;
+    tpp2.target_time = 1e8;
+
+    waypoint = ppww.TrajectoryPlanPointToWaypointConverter(current_time, current_pose, tpp, tpp2);
```
msmcconnell: ppww is not defined so this unit test fails.

mpc_follower_wrapper/src/mpc_follower_wrapper_worker.cpp
```diff
@@ -0,0 +1,43 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+#include "mpc_follower_wrapper/mpc_follower_wrapper_worker.hpp"
+
+namespace mpc_follower_wrapper {
+
+autoware_msgs::Waypoint MPCFollowerWrapperWorker::TrajectoryPlanPointToWaypointConverter(double current_time, geometry_msgs::PoseStamped pose, cav_msgs::TrajectoryPlanPoint tpp, cav_msgs::TrajectoryPlanPoint tpp2) {
```
msmcconnell: Something caught by sonar cloud is that this function does not use the current pose or current time. You should remove these arguments to avoid confusion.

mpc_follower_wrapper/src/mpc_follower_wrapper_node.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+int main(int argc, char** argv) {
+
+  ros::init(argc, argv, "mpc_follower_wrapper_node");
+  ros::CARMANodeHandle nh("");
+
+  mpc_follower_wrapper::MPCFollowerWrapper MPCFollowerWrapper(nh);
+
+  // Approximate time of 100ms used because NDT outputs at 10Hz
+  message_filters::Synchronizer<mpc_follower_wrapper::MPCFollowerWrapper::SyncPolicy> sync(mpc_follower_wrapper::MPCFollowerWrapper::SyncPolicy(100), MPCFollowerWrapper.pose_sub, MPCFollowerWrapper.trajectory_plan_sub);
```
msmcconnell: Seems like if the pose and current time are not used in your conversion function you could get rid of the message_filter entirely and just have the trajectory subscriber.

Issue #613, opened 2020-03-20
Implement Truck Inspection Client to Handle Incoming Request


qswawrq:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
For the ACE Demo, a new system called CARMAMessenger was created with a new Truck Inspection plugin and new CARMAPlatform truck inspection plugin client. 

### Expected behavior
N/A
### Software version this applies to
N/A
### Actual behavior
N/A
### Steps to reproduce the behavior
N/A
### Related work

CAR-1778: CARMA messenger truck safety log request
CAR-1779:  Broadcast Truck safety message log
CAR-1781: CARMA messenger should receive a truck safety message log.

Link to related tickets or prior related work here.

maefromm:
closed by the PR 614.

PR #614, opened 2020-03-20
Truck Inspection Plugin Client

qswawrq:
# PR Details
## Description
Address #613. This PR is still in progress until we have a complete definition of truck safety message.

## Related Issue
#613

## Motivation and Context
Implement new feature.

## How Has This Been Tested?
Not tested.

## Types of changes

## Checklist:


maefromm:
@qswawrq not sure if you got the chance to ask John, but I mentioned about potentially adding the logic that we can pull from the platform, instead of just the yaml file. @TonysCousin should add the logic to pull in the values dynamically instead of hard-coding in a yaml file? I think some of these values could be added/pulled from the https://github.com/usdot-fhwa-stol/CARMAConfig/blob/develop/freightliner_cascadia_2012/VehicleConfigParams.yaml dynamically.

truck_inspection_client/config/parameters.yaml
```diff
@@ -0,0 +1,33 @@
...
+# Date of Last ADS Calibration: YYYY-MM-DD
+date_of_last_ads_calibration: YYYY-MM-DD
+
+#ISS Score: ### (number format + font color)
+#Color coding on the # only. 
+#1-49 (Green font) (Pass - Inspection Not Warranted)
+#50-74 (Yellow font) (Optional- Inspection Warranted – Next Level Priority)
+#75-100 (Red font) - (Inspect – Inspection warranted – top priority)
+iss_score: 49
+
```
maefromm: missing 2 fields ads_health_status and ads_pretrip_check. NOTE I will send you the latest on naming convention so more to come. Awaiting client input.

truck_inspection_client/src/truck_inspection_client.cpp
```diff
@@ -0,0 +1,61 @@
...
+        pnh_->getParam("license_plate", license_plate_);
+        pnh_->getParam("carrier_name", carrier_name_);
+        pnh_->getParam("carrier_id", carrier_id_);
+        pnh_->getParam("weight", weight_);
+        pnh_->getParam("ads_software_version", ads_software_version_);
+        pnh_->getParam("date_of_last_state_inspection", date_of_last_state_inspection_);
+        pnh_->getParam("date_of_last_ads_calibration", date_of_last_ads_calibration_);
+        pnh_->getParam("iss_score", iss_score_);
+        pnh_->getParam("permit_required", permit_required_);
+        mo_pub_ = nh_->advertise<cav_msgs::MobilityOperation>("mobility_operation_outbound", 5);
```
maefromm: missing 2 more fields, see comments on yaml.

TonysCousin:
I'd like to see ads_software_version dynamically assigned from the actual operating version ID (from the docker image).  ADS health status should be dynamically derived from health monitor info to reflect actual status of each driver (if not all green we need to come up with some logic to report what is going on).  Otherwise, I'm okay having all these come from the param file as is; if we pull them from the truck config, then it seems we need some logic to determine if it's a truck or not, because some of these params will not apply to passenger cars; not worth that effort right now.

maefromm:
@TonysCousin I'm good with that. But if this becomes another future demo, we should make it pull from truck config and include the weight in there. 

 I think for the UI screen, i'm surprised we didn't include the vehicle year, make and model "Freightliner" "Cascadia" "2012", which is from the truck config. Anyways, I haven't heard any additional inputs, so I think we're good for now. I will followup today.

qswawrq:
> @TonysCousin I'm good with that. But if this becomes another future demo, we should make it pull from truck config and include the weight in there.
> 
> I think for the UI screen, i'm surprised we didn't include the vehicle year, make and model "Freightliner" "Cascadia" "2012", which is from the truck config. Anyways, I haven't heard any additional inputs, so I think we're good for now. I will followup today.

The main reason I did not put those params in CARMAConfig file is the difficulty on deciding which should stay here and which should go there. The idea of CARMAConfig file is to hold vehicle specific  params. Following that definition, most of parameters in truck inspection plugin are vehicle specific and should be put in CARMAConfig. But they are also used by this single plugin in the near future. So I decide to keep them here.

maefromm:
@qswawrq yes agree. And with the different truck weight and trailer weight, not sure if that should be in the one CARMAConfig that we currently have for cascadia. So please proceed with the changes John suggested above and missing fields i mentioned.

TonysCousin:
Yes, weight, at least, would better come from vehicle config.  If that field doesn't already exist and you were to add that field, would that require changes in other code?  I doubt it, as it's just a param they don't have to read, right?  I'd still prefer truck-specific info to stay where it is for this iteration, just for simplicity.

maefromm:
@TonysCousin Shuwei's asking if the health status (besides health monitor and/or system_alert), should that also include the automated status of the CARMA vehicle (engage/disengaged)? I think we discussed and we said that could be lumped into health status. Right now our UI, we remove "ENGAGED" on the description or Safety Log.

TonysCousin:
My feeling is that ENGAGED is implied, since these messages won't be available if the vehicle is in manual operation mode, right?  If that's not true, then we definitely need to include the engagement state in the health status somehow.

qswawrq:
> My feeling is that ENGAGED is implied, since these messages won't be available if the vehicle is in manual operation mode, right? If that's not true, then we definitely need to include the engagement state in the health status somehow.

Are you saying to only send out this message when the platform is engaged?

maefromm:
@TonysCousin we can always listen to guidance state topic, but not sure if you are more interested if it's ENABLED vs. ENGAGED/ACTIVE/INACTIVE/and other guidance states. Are you just looking to see if the ADS system has been enabled but not actually driving fully automated?

TonysCousin:
Hmm, I'd have to refresh my understanding of that state model.  I'm thinking from a simplistic end user point of view for now.  That user will only care if the truck is actively being controlled by the software or by a human.  Subtle differences of engaged vs enabled vs active, etc, will be totally confusing.  We need it to be VERY SIMPLE:  currently AUTOMATED or HUMAN OPERATOR only.  If this AUTOMATED "state" can be determined simply by the receipt of messages, so much the better (i.e. if it is under human control, will the Message node still interact with the police car's request?)

maefromm:
@qswawrq can answer your message node question. 
If his answer is no, then the only way we know vehicle is AUTOMATED (not HUMAN), is to look at guidance state = ENGAGE i think.

qswawrq:
> Hmm, I'd have to refresh my understanding of that state model. I'm thinking from a simplistic end user point of view for now. That user will only care if the truck is actively being controlled by the software or by a human. Subtle differences of engaged vs enabled vs active, etc, will be totally confusing. We need it to be VERY SIMPLE: currently AUTOMATED or HUMAN OPERATOR only. If this AUTOMATED "state" can be determined simply by the receipt of messages, so much the better (i.e. if it is under human control, will the Message node still interact with the police car's request?)

Message node will keep running despite the vehicle state. For now, I used GREEN state to represent ENGAGE, and RED represents other states.

qswawrq:
@maefromm I think I addressed most of your comments. I intentionally kept the weight info here because I think it is a trip specific info (depends on the load of truck each trip?), which is not quite belong to CARMAConfig.

truck_inspection_client/config/parameters.yaml
```diff
@@ -0,0 +1,37 @@
+# Vehicle VIN number - 17 alphanumeric char
+vin_number: 1FUJGHDV0CLBP8834
+
+# License plate number - 8 alphanumeric char
+license_plate: DOT-10003
+
+# Carrier Name - 20 alphanumeric char
```
maefromm: Please update to 25 characters, since what they have now is more than 20.

truck_inspection_client/config/parameters.yaml
```diff
@@ -0,0 +1,37 @@
+# Vehicle VIN number - 17 alphanumeric char
+vin_number: 1FUJGHDV0CLBP8834
+
+# License plate number - 8 alphanumeric char
```
maefromm: Please update to 9 characters based on what they have now.

truck_inspection_client/config/parameters.yaml
```diff
@@ -0,0 +1,37 @@
...
+carrier_name: Silver Truck FHWA TFHRC
+
+# Carrier ID - (USDOT + space + 7digits)
+carrier_id: USDOT 0000001
+
+# Weight: ###,### lbs. (6 number format)
+weight: 100000
+
+# ADS Software Version: CARMA Platform v#.#.# (25characters)
+ads_software_version: CARMA Platform v3.3.0
```
maefromm: Per John, he wants this from the CARMAConfig version like the UI

qswawrq: Looks like that is something not doable from C++ node.

qswawrq: Fixed as discussed.

truck_inspection_client/config/parameters.yaml
```diff
@@ -0,0 +1,34 @@
+# Vehicle VIN number - 17 alphanumeric char
+vin_number: 1FUJGHDV0CLBP8834
```
maefromm, line 2: Please update to "1FUJGHDV0CLBP8896" for Red Truck. The one above was for White Truck by mistake.

maefromm:
@qswawrq as discussed with @TonysCousin :

1. All of the headers of the topics you created for carma-platform and carma-messenger topics (if they have it), should have a timestamp.
2. The CARMAPlatform mobilityoperation topics shall have the header's sender_ bsm_id.
3. The remaining fields on the headers can be left blank.
4. With the timestamp on the header, the CarmaPlatform truck inspection client does not need to have it on the strategy_params. However carma-messenger truck_safety_info will still need that field, so carma-messenger truck inspection plugin will need to digest it and add it back.

Let me know if I had missed anything.

maefromm:
@qswawrq as discussed, yaml changes requested below to match red truck specs that will be used as the primary demo truck. I am awaiting monica's input on the weight and confirmation of the plate #s.

truck_inspection_client/config/parameters.yaml
```diff
@@ -0,0 +1,34 @@
...
+license_plate: DOT-10003
+
+# Carrier Name - 25 alphanumeric char
+carrier_name: FMCSA Tech Division
+
+# Carrier ID - (USDOT + space + 7digits)
+carrier_id: USDOT 1
+
+# Weight: ###,### lbs. (6 number format)
+weight: 100000
```
maefromm, line 14: 16639

PR #657, opened 2020-04-20
Add carma exec option to carma script

msmcconnell:
# PR Details
## Description

This PR adds a new option to the carma script called exec
This option will create a new container from carma-platform or carma-messenger then allow you to run any command with full GUI support. 

For example
```
carma exec rviz
```
This will launch rvzi with hardware acceleration support (at least in my VM)

```
carma exec
```
This will open a new bash session that already has /opt/carma/install sourced so that the user could just immediately type the desired command.

## Related Issue

#658
## Motivation and Context

Better developer support and dependency management. 

## How Has This Been Tested?

Tested in local VM. But requires carma-base with these changes https://github.com/usdot-fhwa-stol/carma-base/pull/43

## Types of changes

## Checklist:

### DIFF
--- a/engineering_tools/carma
+++ b/engineering_tools/carma
@@ -290,6 +290,13 @@ carma__exec() {
     
     fi
 
+    # Verify target image exists
+    if [ -z "$(docker images -q $TARGET_IMAGE 2> /dev/null)" ]; then
+        echo "Specified image could not be found. carma exec attempted to use image $TARGET_IMAGE but it was not found."
+        echo "If an alternative image is available, please specify it using the -i option."
+        exit -1
+    fi
+
     # If a command was specified then assign it
     if [ -n "$1" ]; then
         local COMMAND="$@"
--- a/engineering_tools/carma
+++ b/engineering_tools/carma
@@ -252,6 +252,70 @@ carma-config__status() {
     fi
 }
 
+carma__exec() {
+    local TARGET_IMAGE="usdotfhwastol/carma-platform:latest"
+    local COMMAND="bash"
+
+    # Check if the -i argument was provided
+    # If the image was specified then use that
+    if [ "$1" = "-i" ]; then 
+
+        if [ -z "$2" ]; then
+            echo " When -i is specified you must enter a carma docker image and tag."
+            exit -1
+        fi
+
+        local TARGET_IMAGE="$2"
+
+        shift
+        shift
+
+    # If the image was not specified use what was in the carma config
+    elif [ docker container inspect carma-config > /dev/null 2>&1 ]; then
+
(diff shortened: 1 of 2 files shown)


engineering_tools/carma
```diff
@@ -252,6 +252,70 @@ carma-config__status() {
     fi
 }
 
+carma__exec() {
+    local TARGET_IMAGE="usdotfhwastol/carma-platform:latest"
```
Supraluminal, line 256: Is there a way to check if the `latest` image exists? Might not be bad to print an error if it's not found and there wasn't anything else specified.

engineering_tools/carma
```diff
@@ -252,6 +252,70 @@ carma-config__status() {
...
+    # If the image was not specified use what was in the carma config
+    elif [ docker container inspect carma-config > /dev/null 2>&1 ]; then
+
+        local IMAGE_BACKUP=$TARGET_IMAGE
+        local SUPPORTED_ORANIZATIONS="usdotfhwastol\|usdotfhwastoldev"
+        local SUPPORTED_IMAGES="carma-platform:\|carma-messenger:"
+
+        # sed command finds the first instance of an image which is part of the set defined in SUPPORTED_ORGANIZATIONS and SUPPORTED_IMAGES.
+        # then returns the name of the image.  
+        local TARGET_IMAGE=$(docker run --rm --volumes-from carma-config:ro --entrypoint sh busybox:latest -c \
```
Supraluminal, line 282: At some point this might be worth making into it's own mode, just to get the current configured image name for a service in the docker-compose.

engineering_tools/carma
```diff
@@ -252,6 +252,70 @@ carma-config__status() {
...
+        fi
+    
+    fi
+
+    # If a command was specified then assign it
+    if [ -n "$1" ]; then
+        local COMMAND="$@"
+    fi
+
+    # Docker setup here is based on Pierre Killy's docker-ros-box https://github.com/pierrekilly/docker-ros-box located here.
```
Supraluminal, line 298: Is this dependent again on the USERIDs lining up? Does the CARMA user in the container have to match the USERID for the current running X11 session?

engineering_tools/carma
msmcconnell: Added

engineering_tools/carma
msmcconnell: Yeah I agree. Though it might take a bit more though. This one is looking for the image name rather than service name so the regex would probably need to be reworked.

engineering_tools/carma
msmcconnell: As far as I could tell its not. Worked both before and after I set the user id. But probably need more testing to be sure of the limitations.

Issue #708, opened 2020-05-28
Development build scripts - update to support usdotstoldev and develop image


maefromm:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
1)
The carma config script does not support usdotstoldev. When you call carma config remote list, install, and others, it points to usdotfhwastol/master. Need to have a –d option as well like build-image.sh to point to usdotstoldev.

 https://github.com/usdot-fhwa-stol/carma-platform/blob/develop/engineering_tools/carma

2) in addition, the carma-platform build-sh needs to be updated to support development configs already pointing to usdotstoldev and develop image. 

Getting the error below: 
![image](https://user-images.githubusercontent.com/27976519/83141272-60b59f00-a0bd-11ea-9795-20a396a5007f.png)

### Expected behavior
Should support the development builds and configuration without errors. 

### Software version this applies to

develop

### Actual behavior
explained above. 

### Steps to reproduce the behavior
Run through all the carma config options like remote list, install, set, start, etc. 
Install carma-config develop image, set and install. 

### Related work

Link to related tickets or prior related work here.

msmcconnell:
Resolved by above PRs

PR #717, opened 2020-06-03
Bugfix/carma config

jtbaird:
# PR Details
## Description
Added -d and -c options to make engineering script compatible with new docker hub orgs.

## Related Issue
#708 

## Motivation and Context
Needed in order to use carma engineering script with the new docker hub organizations

## How Has This Been Tested?
Tested by running carma set with images from usdotfhwastoldev and usdotfhwastolcandidate and then running rest of functions with -d and -c options

## Types of changes

## Checklist:

### DIFF
--- a/docker/build-image.sh
+++ b/docker/build-image.sh
@@ -56,7 +56,7 @@ echo "Final image name: $USERNAME/$IMAGE:$COMPONENT_VERSION_STRING"
 
 cd ..
 if [[ $COMPONENT_VERSION_STRING = "develop" ]]; then
-    sed "s|usdotfhwastol|$USERNAME|g; s|:[0-9]*\.[0-9]*\.[0-9]*|:$COMPONENT_VERSION_STRING|g; s|checkout.bash|checkout.bash -d|g" \
+    sed "s|usdotfhwastol/|$USERNAME/|g; s|:[0-9]*\.[0-9]*\.[0-9]*|:$COMPONENT_VERSION_STRING|g; s|checkout.bash|checkout.bash -d|g" \
         Dockerfile | docker build -f - --no-cache -t $USERNAME/$IMAGE:$COMPONENT_VERSION_STRING \
         --build-arg VERSION="$COMPONENT_VERSION_STRING" \
         --build-arg VCS_REF=`git rev-parse --short HEAD` \
--- a/engineering_tools/carma
+++ b/engineering_tools/carma
@@ -25,26 +25,50 @@
 
 
 __pull_newest_carma_base() {
-    local remote_image=$(wget -q https://registry.hub.docker.com/v1/repositories/usdotfhwastol/carma-base/tags -O -  | \
+    if [ "$1" = "-d" ]; then
+        local remote_image=$(wget -q https://registry.hub.docker.com/v1/repositories/usdotfhwastoldev/carma-base/tags -O -  | \
+        sed -e 's/[][]//g' -e 's/"//g' -e 's/ //g' | tr '}' '\n'  | \
+        awk -F: '{print "usdotfhwastoldev/carma-base:"$3}' | tail -n 1)
+    elif [ "$1" = "-c" ]; then
+        local remote_image=$(wget -q https://registry.hub.docker.com/v1/repositories/usdotfhwastolcandidate/carma-base/tags -O -  | \
+        sed -e 's/[][]//g' -e 's/"//g' -e 's/ //g' | tr '}' '\n'  | \
+        awk -F: '{print "usdotfhwastolcandidate/carma-base:"$3}' | tail -n 1)
+    else
+        local remote_image=$(wget -q https://registry.hub.docker.com/v1/repositories/usdotfhwastol/carma-base/tags -O -  | \
         sed -e 's/[][]//g' -e 's/"//g' -e 's/ //g' | tr '}' '\n'  | \
         awk -F: '{print "usdotfhwastol/carma-base:"$3}' | tail -n 1)
+    fi
     
     echo "No local carma-base image found, pulling down the most recent from Dockerhub..."
     docker pull $remote_image
 }
 
 __get_most_recent_carma_base() {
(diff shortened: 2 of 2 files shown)


docker/build-image.sh
```diff
@@ -56,7 +56,7 @@ echo "Final image name: $USERNAME/$IMAGE:$COMPONENT_VERSION_STRING"
 
 cd ..
 if [[ $COMPONENT_VERSION_STRING = "develop" ]]; then
-    sed "s|usdotfhwastol|$USERNAME|g; s|:[0-9]*\.[0-9]*\.[0-9]*|:$COMPONENT_VERSION_STRING|g; s|checkout.bash|checkout.bash -d|g" \
+    sed "s|usdotfhwastol/|$USERNAME/|g; s|:[0-9]*\.[0-9]*\.[0-9]*|:$COMPONENT_VERSION_STRING|g; s|checkout.bash|checkout.bash -d|g" \
```
kjrush, line 59: This seems good to keep, I assume this is part of a general dockerfile update that's applied across the repos?

engineering_tools/carma
```diff
@@ -116,22 +151,44 @@ carma-config__reset() {
 }
 
 carma-config__list_local() {
+    if [ "$1" = "-d" ]; then
```
kjrush, line 154: This'd probably be best not as an option but as a sequential thing, i.e. run `docker images usdotfhwastol/carma-config` then run `docker images usdotfhwastoldev/carma-config` then `docker images usdotfhwastolcandidate/carma-config` whenever this gets called. That way it really is listing all local images.

engineering_tools/carma
```diff
@@ -116,22 +151,44 @@ carma-config__reset() {
...
 }
 
 carma-config__list_remote() {
-    echo "Remotely available images from usdotfhwastol Dockerhub: "
-    echo ""
-    echo "IMAGE                           TAG"
-    wget -q https://registry.hub.docker.com/v1/repositories/usdotfhwastol/carma-config/tags -O -  | \
-    sed -e 's/[][]//g' -e 's/"//g' -e 's/ //g' | tr '}' '\n'  | \
-    awk -F: '{print "usdotfhwastol/carma-config\t"$3}'
+    if [ "$1" = "-d" ]; then
```
kjrush, line 165: This one I'm less sure of, I do like the flag here so people don't get confused as to what configs should be installed (the released ones vs the dev/candidate images).

engineering_tools/carma
```diff
@@ -116,22 +151,44 @@ carma-config__reset() {
...
+    else
+        wget -q https://registry.hub.docker.com/v1/repositories/usdotfhwastol/carma-config/tags -O -  | \
+        sed -e 's/[][]//g' -e 's/"//g' -e 's/ //g' | tr '}' '\n'  | \
+        awk -F: 'BEGIN {print "Remotely available images from usdotfhwastol Dockerhub:\nIMAGE\t\t\t\tTAG"} {print "usdotfhwastol/carma-config\t"$3}'    
+    fi
 }
 
 carma-config__install() {
-    local IMAGE_NAME="usdotfhwastol/carma-config:$1"
+    if [ "$1" = "-d" ]; then
```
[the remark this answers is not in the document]
kjrush: I think the above comment is OBE given the fix in place for the dev images (dockerfile build updates)

engineering_tools/carma
```diff
@@ -253,7 +310,20 @@ carma-config__status() {
 }
 
 carma__exec() {
-    local TARGET_IMAGE="usdotfhwastol/carma-platform:latest"
+    if [ "$1" = "-d" ]; then
```
kjrush, line 313: I think this functionality already exists in the form of the `-i` implementation below.

engineering_tools/carma
```diff
@@ -94,10 +123,16 @@ carma-config__inspect() {
...
+        local OPTION="-d"
+    elif [ "$1" = "-c" ]; then
+        local OPTION="-c"
+    fi
+
+    local carma_base=$(__get_most_recent_carma_base $OPTION)
     if [[ -z $carma_base ]]; then
-        __pull_newest_carma_base
-        carma_base=$(__get_most_recent_carma_base)
+        __pull_newest_carma_base $OPTION
```
kjrush, line 134: I'm uncertain on this. This and the edit command aren't really related to the precise version/config of CARMA-base, since they're just for opening a quick bash shell to maybe edit one of the CARMA configuration files. Not sure how much it matters what CARMA base is being used, as long as it has ROS and other basic tools installed.

jtbaird: Oh okay got it. Yeah i'll remove the option then.

maefromm:
@jtbaird discussed with @kjrush that I will test without this fix. If it works then we can close without merging this changes.

engineering_tools/carma
kjrush: I'd still suggest we make this list **all** locally available images, so basically just list the usdotfhwastol usdotfhwastoldev and usdotfhwastolcandidate images in sequence. This avoids confusion about which images are actually installed locally and available. I think leaving the flags for the list remote is fine (since it makes sense to kind of hide the pre-release images), but they should just always apply here.

engineering_tools/carma
kjrush: I just hesitate to have redundant options for accomplishing the same outcome for clarity's sake, and this is a case (specifically overriding the currently installed carma version to exec a program) that's advanced enough the explicitness of -i fits better I feel.

engineering_tools/carma
```diff
@@ -75,11 +99,16 @@ carma-config__edit() {
     fi
 
     echo "Opening shell inside carma-config container with read/write privileges..."
+    if [ "$1" = "-d" ]; then
```
kjrush, line 102: I'm not sure the flags here (and on inspect) make sense either, these functions are just for using non-carma software to look at and/or edit the CARMA config contents. All that these functions really require is that there is **a** version of carma-base installed so that we can guarantee vim/emacs/sed/awk/etc are available for manipulating the config files. It should never really matter which version of carma is being used for that.

PR #721, opened 2020-06-09
Feature/geofence speed support

MishkaMN:
# PR Details
## Description

This PR is to update the carma_wm_ctrl local map copy to support updates to speed limit regulatory elements from geofences. This PR assumes all speed limit changes will be at a lanelet boundary. Local map is being published as of now although it can be handled in a future story. Integration testing focused on the carma_wm_ctrl node remaining operational when sent a speed limit geofence. 

General flow of how this process works is that:
- map_param_loader scans map's georereference and publishes so that carma_wm_ctrl can use it to compare with geofence's georeference. The broadcaster saves it in the worker, as it won't be changed.
- when geofence receives new msg, it populates Geofence object with relevant information. The callback function should determine, in the future, whether it is speed_change, or lanelet_change etc type of geofence.  
- affectedLaneletOrArea(geofence_msg) function is called to get affected lanelets (areas will be added in the future). The returned lanelets are stored in the geofence object. This is part of the reason why all functions are changed from `const Geofence&` to `Geofence&`
- then this Geofence object is added to the scheduler for call.
- when scheduler calls addGeofence(), it removes the old speed limit and stores those regems (can be general instead of speed_limit) in the geofence object itself, as this object is called again in removeGeofence() by the scheduler. Then it will add the new ones.
- when schedulre calls removeGeofence, the reverse happens. Old regems are read from the geofence and put back in the map.

## Related Issue

https://github.com/usdot-fhwa-stol/carma-platform/issues/722
## Motivation and Context
This is the first of its features in line where geofence requests to make changes to the map and carma should be able to accommodate that.

## How Has This Been Tested?

Existing unit tests passed and new ones too.
Regression tested locally by sending manual geofence and the node doesn't crash.

## Types of changes

## Checklist:


carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcaster.h
```diff
@@ -53,26 +56,40 @@ class WMBroadcaster
...
    */
-  void removeGeofence(const Geofence& gf);
+  void removeGeofence(Geofence& gf);
+
+  /*!
+   * \brief Gets the affected lanelet or areas based on the geofence_msg
+   * NOTE: Currently only checks lanelets
+   * TODO add throws and whatnot
+   */
+  lanelet::Lanelets getAffectedLaneletOrAreas(const cav_msgs::ControlMessage& geofence_msg);
```
msmcconnell, line 92: If you intend to have this function work with areas as well then you should use the lanelet::LaneletOrArea data type as the return value.

carma_wm_ctrl/launch/carma_wm_broadcaster.launch
```diff
@@ -12,5 +12,8 @@
 -->
 
 <launch>
-  <node name="carma_wm_broadcaster" pkg="carma_wm_ctrl" type="carma_wm_ctrl_node"/>
+  <arg name="georeference_topic" default="georeference" doc="The topic name of the georeference that is published by map_param_loader form the vector_map.osm"/>
```
msmcconnell, line 15: Don't pass topic names as arguments. This is a ros anti-pattern and makes remapping the topics later very difficult.

msmcconnell: Instead you want to give the topic a clear name and remap it as needed.

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -129,26 +141,57 @@ lanelet::ConstLaneletOrAreas WMBroadcaster::getAffectedLaneletOrAreas(const cav_
...
     {
-      if (boost::geometry::within(pt.basicPoint2d(), llt.polygon2d()))
+    
+      lanelet::BasicLineString2d gf_dir_line({gf_pts[idx].basicPoint2d(), gf_pts[idx+1].basicPoint2d()});
+      lanelet::BasicLineString2d llt_boundary({(llt.leftBound2d().end() -1)->basicPoint2d(), (llt.rightBound2d().end() - 1)->basicPoint2d()});
+      if (boost::geometry::intersects(llt_boundary, gf_dir_line))
+      {
+        // record the llts that are on the same dir
         affected_lanelets.insert(llt);
+      }
```
msmcconnell, line 194: This looks like a good implementation of your original algorithm, but I think there is an edge case for crossing lanelets. Lets discuss directly.

carma_wm_ctrl/test/WMBroadcasterTest.cpp
```diff
@@ -82,9 +82,8 @@ TEST(WMBroadcaster, baseMapCallback)
   ASSERT_EQ(1, base_map_call_count);
 }
 
-
 // here test the proj string transform test
-TEST(WMBroadcaster, getAffectedLaneletOrAreasWithTransform)
+TEST(WMBroadcaster, getAffectedLaneletOrAreasFrameTransform)
```
msmcconnell, line 86: Typo, should be From instead of Frame I think

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -151,32 +152,47 @@ lanelet::ConstLaneletOrAreas WMBroadcaster::getAffectedLaneletOrAreas(const cav_
...
     };
-   
+
+    // this call updates possible_lanelets
     current_map_->laneletLayer.nearestUntil(gf_pts[idx], searchFunc);
 
+    // we utilize routes to filter llts that are overlapping but not connected
+    lanelet::traffic_rules::TrafficRulesUPtr traffic_rules_car = lanelet::traffic_rules::TrafficRulesFactory::create(
+      lanelet::traffic_rules::CarmaUSTrafficRules::Location, lanelet::Participants::VehicleCar);
+    lanelet::routing::RoutingGraphUPtr map_graph = lanelet::routing::RoutingGraph::build(*current_map_, *traffic_rules_car);
```
msmcconnell, line 178: Can you confirm this routing graph will only be built once per geofence? I don't want to have it happen more than that. In the future we may want to find a way to optimize it further but we will leave that for future work.

carma_wm_ctrl/src/WMBroadcasterNode.cpp
```diff
@@ -40,6 +40,11 @@ int WMBroadcasterNode::run()
   georef_sub_ = cnh_.subscribe("georeference", 1, &WMBroadcaster::geoReferenceCallback, &wmb_);
   // Geofence Sub
   geofence_sub_ = cnh_.subscribe("geofence", 1, &WMBroadcaster::geofenceCallback, &wmb_);
+  
+  double lane_max_width;
+  cnh_.getParam("lane_max_width", lane_max_width);
```
msmcconnell, line 45: You should make a private CARMANodeHandle and use that for node specific parameters. CARMANodeHandle("~")

carma_wm_ctrl/src/WMBroadcaster.cpp
MishkaMN: Yeah. This logic is in getAffectedLaneletOrAreas function which is only called when geofence_msg is being translated into Geofence object that happens once in GeofenceCallback.

msmcconnell:
Looks good. But we need to wait for CI. I am working with Kyle on getting CI pointed at the correct image.

PR #725, opened 2020-06-17
Updated unit test to use the lanelet2_extension libraries

afrye51:
Updated unit test to use the lanelet2_extension libraries for the DigitalSpeedLimit regulatory element, and changed the test to use lanelet IDs instead of starting/ending points because that was not working correctly

# PR Details
## Description

The route unit test was updated to support the lanelet2_extension libraries, specifically the DigitalSpeedLimit regulatory element in osm files. 

After this was complete, the original method of route planning (grabbing a start and end point, and the code finds the closest lanelets to use) was just not working. Possibly a geo-reference location issue, bit the test would always grab the same lanelet for the starting and ending, so no routing test would occur.

This was changed so that the user puts the lanelet IDs directly into the file. There is a check that fails the unit test if either of the starting/ending lanelet IDs don't match a lanelet in the osm file.

This required a new function, routing_from_lanelets, to be created (copied from the original routing function). This then calls a pre-built function to generate a route between the two lanelets. 

## Related Issue

This change was made to support testing CAR-2671: https://usdot-carma.atlassian.net/browse/CAR-2671. Though not strictly necessary, without these updates, any files from the opendrive2lanelet converter need to have every digital_speed_limit regulatory element manually deleted to run with the original unit test. 

## Motivation and Context

This change allows osm files from the opendrive2lanelet converter to be run through the unit test, and also makes it easier to test routing on a new map. 

## How Has This Been Tested?

This was tested by testing 2 seperate files - Crossing8Course and Town02 osm files converted with the opendrive2lanelet converter and confieming that the route's shortest path was correct. This test was run in VS code in a carma docker container in a carma development VM.

    1) Libraries you should have: carma-base, carma-config, carma-msgs (integration/routing branch), 
        carma-platform (integration/routing branch), carma-utils, caram-web-ui, opendrive2lanelet.
    2) Then the test can be built with the command: 
        ./carma_build -c /workspaces/carma_ws/carma/ -a /workspaces/carma_ws/autoware.ai/ -x -m "--only-pkg-with-deps route"
    3) Update the osm file location and starting/ending IDs to match the file you want to test
    4) Run the unit test with
        catkin_make run_tests_route
    5) Confirm that the test passed and that the list of lanelet IDs does traverse from the start to the end

These changes shouldn't affect any other code, since the original functions are still there to be called. If there is a script calling this unit test, that should have the same shortcoming of having to manually edit the osm file location and starting/ending locations beforehand. 

## Types of changes

## Checklist:

I was planning to update the How to Verify HDMaps from Continental Mapping (https://vector.leidos.com/sites/STOL/STOL2/_layouts/15/WopiFrame.aspx?sourcedoc={8C47CB7D-EB7D-4AEA-9444-7587CA015A65}&file=Verify%20HD%20Maps%20from%20Continental%20Mapping.docx&action=default&DefaultItemOpen=1) documentation with the specifics of this unit test change once this pull request goes through.

### DIFF
--- a/route/test/test_route_generator.cpp
+++ b/route/test/test_route_generator.cpp
@@ -25,29 +25,79 @@
 #include <lanelet2_core/Attribute.h>
 #include <lanelet2_core/geometry/LineString.h>
 #include <lanelet2_core/primitives/Traits.h>
+#include <lanelet2_extension/traffic_rules/CarmaUSTrafficRules.h>
+#include <lanelet2_extension/utility/query.h>
+#include <lanelet2_extension/utility/utilities.h>
 #include <lanelet2_extension/projection/local_frame_projector.h>
+#include <lanelet2_extension/io/autoware_osm_parser.h>
+
 #include <gtest/gtest.h>
 #include <ros/ros.h>
+
+/*
+Using this file:
+    1) Libraries you should have: carma-base, carma-config, carma-msgs (integration/routing branch), 
+        carma-platform (integration/routing branch), carma-utils, caram-web-ui, opendrive2lanelet.
+    2) Then the test can be built with the command: 
+        ./carma_build -c /workspaces/carma_ws/carma/ -a /workspaces/carma_ws/autoware.ai/ -x -m "--only-pkg-with-deps route"
+    3) Update the osm file location and starting/ending IDs to match the file you want to test
+    4) Run the unit test with
+        catkin_make run_tests_route
+    5) Confirm that the test passed and that the list of lanelet IDs does traverse from the start to the end
+*/
 
 TEST(RouteGeneratorTest, testLaneletRouting)
 {
     tf2_ros::Buffer tf_buffer;
     carma_wm::WorldModelConstPtr wm;
     route::RouteGeneratorWorker worker(tf_buffer);
-    //lanelet::LaneletMapPtr map = lanelet::load("/home/qswawrq/Desktop/ATEF_pretty.osm",
-    //    lanelet::projection::LocalFrameProjector("+proj=tmerc +lat_0=39.46636844371259 +lon_0=-76.16919523566943 +k=1 +x_0=0 +y_0=0 +units=m +vunits=m"));
-    lanelet::LaneletMapPtr map = lanelet::load("/home/qswawrq/Desktop/ATEF_pretty.osm", lanelet::Origin({0, 0}));
+
+    int projector_type = 0;
+    std::string target_frame;
+    lanelet::ErrorMessages load_errors;
+
+    // If the output is an error about the geoReference field in the osm file, here is a correct example. If the lat/lon coordinates are already correct, simply add:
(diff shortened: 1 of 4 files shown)


kjrush:
Tagging myself as one of the reviewers here for the route generation logic specifically.

route/test/test_route_generator.cpp
```diff
@@ -25,36 +25,83 @@
...
+
+    int projector_type = 0;
+    std::string target_frame;
+    lanelet::ErrorMessages load_errors;
+
+    // If the output is an error about the geoReference field in the osm file, here is a correct example. If the lat/lon coordinates are already correct, simply add:
+    // <geoReference>+proj=tmerc +lat_0=0 +lon_0=0 +k=1 +x_0=0 +y_0=0 +datum=WGS84 +units=m +geoidgrids=egm96_15.gtx +vunits=m +no_defs</geoReference>
+
+    // File location of osm file
+    std::string file = "/workspaces/carma_ws/carma/src/Town02_pretty.osm";
```
kjrush, line 63: Is this file going to exist in CI and on people's development machines? If it's an open-source test file (sounds like it's form CARLA so that should be the case) then we can add it to the test directory here if it's not super large. Should make the tests consistent.

route/test/test_route_generator.cpp
```diff
@@ -25,36 +25,83 @@
...
-    std::vector<lanelet::BasicPoint2d> via;
-    // via.emplace_back(lanelet::BasicPoint2d(-158.0, 533.5));
-    //via.emplace_back(lanelet::BasicPoint2d(-17.5, 322.0));
-    lanelet::BasicPoint2d end(1052.35, 771.24);
     lanelet::traffic_rules::TrafficRulesUPtr traffic_rules = lanelet::traffic_rules::TrafficRulesFactory::create(lanelet::Locations::Germany, lanelet::Participants::VehicleCar);
     lanelet::routing::RoutingGraphUPtr map_graph = lanelet::routing::RoutingGraph::build(*map, *traffic_rules);
     // Output graph for debugging
-    map_graph->exportGraphViz("/home/qswawrq/Desktop/routing.txt");
-    auto route = worker.routing(start, via, end, const_map, std::move(map_graph));
+    map_graph->exportGraphViz("/workspaces/carma_ws/carma/src/routing.txt");
```
kjrush, line 93: Same here with the filepath, I'd suggest making this a relative path so that it works no matter the home folder/user.

route/test/test_route_generator.cpp
afrye51: There is already a test_vector_map.osm file that I could change it to instead. My understanding is that the point of this file is for testing maps, so a user has their own map and changes the path and lanelet IDs to match, to confirm that that map is routable, rather than running it with an example file.

afrye51: I will change both links to be relative and to valid files just in case

PR #730, opened 2020-06-22
Feature/collision checking

itabrl:
# PR Details
## Description

Basic collision checking for CARMA based on Motion detection implemented in CARMAUtiles.

## Related Issue
#734 

## How Has This Been Tested?

Unit tests implemented.

## Types of changes

## Checklist:


carma_wm/include/carma_wm/collision_detection.h
```diff
@@ -0,0 +1,93 @@
...
+        };
+
+        /*!
+        * Main class for the CollisionChecking interfacing.
+        */
+
+        class CollisionChecking {
+            public:
+                /*! \brief Main collision detection function to be called when needed to check for collision detection of the vehicle with 
+                * the current trajectory plan and the current world objects
```
msmcconnell, line 57: Please provide lines describing each parameter and the return value in all your public function comments such as
```
\param size The size of the host vehicle defined in meters
\return A list of obstacles the provided trajectory plan collides with
```
Otherwise there is no way for the user to know what those parameters mean.

carma_wm/include/carma_wm/collision_detection.h
```diff
@@ -0,0 +1,93 @@
...
+            polygon_t object_polygon;
+            geometry_msgs::Vector3 linear_velocity;
+            std::vector<polygon_t> future_polygons;
+        };
+
+        /*!
+        * Main class for the CollisionChecking interfacing.
+        */
+
+        class CollisionChecking {
```
msmcconnell, line 54: It looks like this class is stateless and has no private members. So it would be better to just implement the functions as free functions inside the collision_detection namespace.

carma_wm/src/collision_detection.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+    namespace collision_detection {
+
+        std::vector<cav_msgs::RoadwayObstacle> CollisionChecking::WorldCollisionDetection(cav_msgs::RoadwayObstacleList& rwol, cav_msgs::TrajectoryPlan& tp, geometry_msgs::Vector3& size, geometry_msgs::Twist& veloctiy, int target_time){
+
+            std::vector<cav_msgs::RoadwayObstacle> rwo_collison;
+
+            collision_detection::MovingObject vehicle_object = CollisionChecking::ConvertVehicleToMovingObject(tp, size, veloctiy);
+
+            for (int i = 0; i < rwol.roadway_obstacles.size(); i++) {
```
msmcconnell, line 13: best practice is to use size_t as the iterator type in for loops since it is unsigned.

carma_wm/src/collision_detection.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+            geometry_msgs::Pose pose;
+            pose.position.x = tp.trajectory_points[0].x;
+            pose.position.y = tp.trajectory_points[0].y;
+
+            Eigen::Vector2d vehicle_vector = {tp.trajectory_points[0].x - tp.trajectory_points[1].x , tp.trajectory_points[0].y - tp.trajectory_points[1].y};
+
+            Eigen::Vector2d x_axis = {1, 0};
+
+            double yaw = std::acos(vehicle_vector.dot(x_axis)/(vehicle_vector.norm() * x_axis.norm()));
```
msmcconnell, line 54: vehicle_vector.dot([1, 0]) is always going to give you 
vehicle_vector.x

similarly x_axis.norm() is always going to equal 1.

So this line could be made much simpler and and faster as
```
std::acos(vehicle_vector.x / vehicle_vector.norm())
```

carma_wm/src/collision_detection.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+        collision_detection::MovingObject CollisionChecking::ConvertVehicleToMovingObject(cav_msgs::TrajectoryPlan& tp, geometry_msgs::Vector3& size, geometry_msgs::Twist& veloctiy){
+            
+            collision_detection::MovingObject v;
+
+            geometry_msgs::Pose pose;
+            pose.position.x = tp.trajectory_points[0].x;
+            pose.position.y = tp.trajectory_points[0].y;
+
+            Eigen::Vector2d vehicle_vector = {tp.trajectory_points[0].x - tp.trajectory_points[1].x , tp.trajectory_points[0].y - tp.trajectory_points[1].y};
```
msmcconnell, line 50: Won't this give you an inverted vector since you are subtracting the second point from the first.

carma_wm/src/collision_detection.cpp
```diff
@@ -0,0 +1,179 @@
...
+            pose.orientation.z = vehicle_orientation.getZ();
+            pose.orientation.w = vehicle_orientation.getW();
+
+            v.object_polygon = CollisionChecking::ObjectToBoostPolygon<polygon_t>(pose, size);
+            v.linear_velocity = veloctiy.linear;
+
+            for(int i=0; i < tp.trajectory_points.size() - 1; i++){
+
+                vehicle_vector = {tp.trajectory_points[i + 1].x - tp.trajectory_points[i].x , tp.trajectory_points[i+1].y - tp.trajectory_points[i].y};
+                yaw = std::acos(vehicle_vector.dot(x_axis)/(vehicle_vector.norm() * x_axis.norm()));
```
msmcconnell, line 70: same note as above.

carma_wm/src/collision_detection.cpp
```diff
@@ -0,0 +1,179 @@
...
+            pose.orientation.y = vehicle_orientation.getY();
+            pose.orientation.z = vehicle_orientation.getZ();
+            pose.orientation.w = vehicle_orientation.getW();
+
+            v.object_polygon = CollisionChecking::ObjectToBoostPolygon<polygon_t>(pose, size);
+            v.linear_velocity = veloctiy.linear;
+
+            for(int i=0; i < tp.trajectory_points.size() - 1; i++){
+
+                vehicle_vector = {tp.trajectory_points[i + 1].x - tp.trajectory_points[i].x , tp.trajectory_points[i+1].y - tp.trajectory_points[i].y};
```
msmcconnell, line 69: This is the inverse subtraction order from what you had above and look correct.

carma_wm/src/collision_detection.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+            pose.orientation.x = vehicle_orientation.getX();
+            pose.orientation.y = vehicle_orientation.getY();
+            pose.orientation.z = vehicle_orientation.getZ();
+            pose.orientation.w = vehicle_orientation.getW();
+
+            v.object_polygon = CollisionChecking::ObjectToBoostPolygon<polygon_t>(pose, size);
+            v.linear_velocity = veloctiy.linear;
+
+            for(int i=0; i < tp.trajectory_points.size() - 1; i++){
```
msmcconnell, line 67: I think by having this loop start from 0 you are causing a prediction to be placed on top of the current vehicle. Seems like it should start from 1. 
Also use size_t

carma_wm/src/collision_detection.cpp
```diff
@@ -0,0 +1,179 @@
...
+            return false;
+        };
+
+        bool CollisionChecking::CheckPolygonIntersection(collision_detection::MovingObject ob_1, collision_detection::MovingObject ob_2) {    
+
+                std::deque<polygon_t> output;
+
+                boost::geometry::intersection(ob_1.object_polygon, ob_2.object_polygon, output); 
+
+                BOOST_FOREACH(polygon_t const& p, output)
```
msmcconnell, line 108: seems like you could just use a regular C++ 11 for each loop here instead of the boost macro

```
for (auto p : output) {
// logic
}
```

carma_wm/src/collision_detection.cpp
```diff
@@ -0,0 +1,179 @@
...
+            polygon_t union_polygon;  
+            boost::geometry::assign_points(union_polygon, unioin_polygon_points);
+
+            polygon_t hull_polygon;
+            boost::geometry::convex_hull(union_polygon, hull_polygon);
+
+            collision_detection::MovingObject output_object = {hull_polygon, op.linear_velocity};
+            
+            return output_object;
+        };
```
msmcconnell, line 142: This function needs some comments describing what is happening. Is it creating a convex hull around all the predictions?

itabrl: yes i added comments in header file

carma_wm/src/collision_detection.cpp
itabrl: It makes scene but  your suggested code and has compile errors.
```
error: invalid operands of types ‘<unresolved overloaded function type>’ and ‘Eigen::GenericNumTraits<double>::Real {aka double}’ to binary ‘operator/’
             std::acos(vehicle_vector.x / vehicle_vector.norm());
```
I think it is better we keep it this way.

carma_wm/include/carma_wm/collision_detection.h
```diff
@@ -55,20 +55,33 @@ namespace carma_wm {
             public:
                 /*! \brief Main collision detection function to be called when needed to check for collision detection of the vehicle with 
                 * the current trajectory plan and the current world objects
+                * \param size The size of the host vehicle defined in meters
+                * \param rwol The list of Roadway Obstacle
+                * \param tp The TrajectoryPlan of the host vehicle
+                * \param size The size of the host vehicle defined in meters
+                * \param veloctiy of the host vehicle
+                * \param target_time amount of unit of time in future to look for collision
```
msmcconnell: What are the units? The argument is an int which makes me think they are not seconds. If they are miliseconds or something then an unsigned int would be a better type since negative time does not make sense.

carma_wm/include/carma_wm/collision_detection.h
```diff
@@ -55,20 +55,33 @@ namespace carma_wm {
...
                 /*! \brief Convert RodwayObstable object to the collision_detection::MovingObject 
+                 * \param rwo A RoadwayObstacle
                 */
  
                 collision_detection::MovingObject ConvertRoadwayObstacleToMovingObject(cav_msgs::RoadwayObstacle& rwo);
                 
                 /*! \brief Creates collision_detection::MovingObject for the host vehicle using the veloctiy, size, TrajectoryPlan.
+                * \param tp The TrajectoryPlan of the host vehicle
+                * \param size The size of the host vehicle defined in meters
+                * \param veloctiy of the host vehicle
```
msmcconnell: units. I assume (m/s)

carma_wm/include/carma_wm/collision_detection.h
```diff
@@ -55,20 +55,33 @@ namespace carma_wm {
...
                 */
 
                 collision_detection::MovingObject ConvertVehicleToMovingObject(cav_msgs::TrajectoryPlan& tp, geometry_msgs::Vector3& size, geometry_msgs::Twist& veloctiy);                
 
                 /*! \brief .function is to create a monving object with a polygon that represents area that object is going to allocated until a given time
+                * by creating a convex hull around the future polygons
+                * \param op a MovingObject
```
msmcconnell: Missing target time param

msmcconnell: I like the idea of using this convex polygon but doesn't this mean that the trajectory would be considered collided even if there was a big time gap between locations. Seems like we would get a lot of false positives as the trajectories can be up to 6s long at the moment. 

Got any thoughts on how to minimize that?

carma_wm/src/collision_detection.cpp
msmcconnell: Try this
```
std::acos(vehicle_vector.x / (double)vehicle_vector.norm())
```
If that doesn't work you can go back to the other method.

carma_wm/include/carma_wm/collision_detection.h
itabrl: I suggested future enhancement in issue https://github.com/usdot-fhwa-stol/carma-platform/issues/740  we can add functionality that would break the prediction in to smaller chunks and check for collision checking.

carma_wm/src/collision_detection.cpp
itabrl: I added this and again got error. 

error: invalid operands of types ‘<unresolved overloaded function type>’ and ‘Eigen::GenericNumTraits<double>::Real {aka double}’ to binary ‘operator/’
             double yaw = std::acos(vehicle_vector.x / (double)vehicle_vector.norm());

msmcconnell:
@itabrl 
Please resolve merge conflicts. 
In addition, a number of my comments are still not addressed. Please add the units for the time parameters and address some of the code smells. Most of the code smells looks easy to resolve. 

With regards to the convex hull, please add a note to the function header comments that describes the risk of the false positive and mentions that users will need to be aware of this when using the library.

msmcconnell:
Based on the added comments the target_time parameter is in seconds. Given that it needs to be changed to a double in all cases, since integer seconds does not provide the resolution required for planning which occurs at 10 Hz

msmcconnell:
When checking the double type change, sonar cloud pointed out that there were comparisons with target_time in the for loop, so now I'm seeing some logical issues with this.
In you function to convert object predictions to moving objects you have the following loop.
```
            for (auto i : rwo.object.predictions){
                ro.future_polygons.push_back(ObjectToBoostPolygon<polygon_t>(i.predicted_position, rwo.object.size)); }
```
This populates the moving object with the polygons matching each prediction, which makes sense.
The problem is that in your PredictObjectPosition function  you are then running this loop
```
            for(size_t i = 0; i< target_time; i++){
                size = size + op.future_polygons[i].outer().size(); }
 ```
This means that your are looking for the 1s polygon at position [1] but the polygon at that position will actually correspond to whatever time is specified in the header of the PredictedState.msg message.
So it would be better if you used a while loop to compare current timestamp of object at index to the target time

File: carma_wm/src/collision_detection.cpp [hunk 14]
```diff
@@ -25,17 +26,20 @@ namespace carma_wm {
...
-            collision_detection::MovingObject ro;
+            collision_detection::MovingObject mo;
             
-            ro.object_polygon = ObjectToBoostPolygon<polygon_t>(rwo.object.pose.pose, rwo.object.size);
+            mo.object_polygon = ObjectToBoostPolygon<polygon_t>(rwo.object.pose.pose, rwo.object.size);
 
+            // Add future polygons for roadway obstacle
             for (auto i : rwo.object.predictions){
-                ro.future_polygons.push_back(ObjectToBoostPolygon<polygon_t>(i.predicted_position, rwo.object.size));
+                std::tuple <__uint32_t,polygon_t> future_object(i.header.stamp.nsec/1000,ObjectToBoostPolygon<polygon_t>(i.predicted_position, rwo.object.size));
```
msmcconnell, line 35: I'm fairly certain the .nsec field is only the number of nsec past the current sec mark. So this will generate errors when the seconds roll over. You will need to combine the seconds value for an accurate absolute reference time. 

In addition you can't use a uint32_t to store microseconds since the current epoch of 1970 since the uint32 has a max value of 4,294,967,295
and the current unix time is 1.594228e+15 microseconds. 

If you want to use a singular time value I would suggest either using milliseconds or taking the first time stamp and subtracting it from all others before doing the conversion to micro seconds. If you use milliseconds please document that this is the max resolution.

carma_wm/include/carma_wm/collision_detection.h
```diff
@@ -58,7 +58,7 @@ namespace carma_wm {
         * \param tp The TrajectoryPlan of the host vehicle
         * \param size The size of the host vehicle defined in meters
         * \param veloctiy of the host vehicle m/s
-        * \param target_time amount of unit of time in future to look for collision in seconds
+        * \param target_time amount of unit of time in future to look for collision in milisecounds
```
msmcconnell, line 61: Isn't this in microseconds since you are dividing nano seconds by 1000?

msmcconnell:
Need @kjrush To merge since the sonar results haven't propagated back.

Issue #770, opened 2020-07-27
CARMA requires ControlRequest logic to be implemented in carma_wm_ctrl


lewisid:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
CARMA requires logic to allow the vehicle to request traffic rules from CARMA Cloud based on the route using a ControlRequest message that is populated with route information once it has been received.
### Expected behavior
CARMA should be able to receive route information from CARMA Cloud and use it to populate a ControlRequest message containing 
### Software version this applies to

3.3.0
### Actual behavior
CARMA requires the user to manually publish a value to the `route_lanelet_ids_array` on the `/route` topic in order to acquire the information necessary to populate the ControlRequest message. 
### Steps to reproduce the behavior

N/A

### Related work

N/A

msmcconnell:
Resolved by above PR

PR #771, opened 2020-07-27
Feature/control request

lewisid:
# PR Details
## Description

Added logic which enables the vehicle to create a ControlRequest message that is populated with route information once it has been received. To complete this, two new functions `routeCallbackMessage()` and `routeCallbackMessageLogic()` were added. The former would receive a route message containing lanelet information. This function then calls `routeCallbackMessageLogic`, sending the route message to populate and return the ControlRequest message.  Once `routeCallbackMessageLogic` receives the route message, the local xyz coordinates will be converted to latitude/longitude values, with three additional offset values provided. These values would then populate the respective fields in a ControlBounds message which will in turn populate the ControlRequest message to be published to the /outgoing_geofence_request topic. 

## Related Issue

[#770 ](https://github.com/usdot-fhwa-stol/carma-platform/issues/770)
## Motivation and Context

This will allow the vehicle to request traffic rules from CARMA Cloud based on the route using the ControlRequest message populated with the appropriate route information.
## How Has This Been Tested?

All testing took place in VMware WorkStation Player 15. Unit tests were performed in gtest and integration testing was performed by creating Docker images to be used by running CARMA in the virtual machine.
## Types of changes

## Checklist:


carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcaster.h
```diff
@@ -89,6 +106,20 @@ class WMBroadcaster
...
+  * \param route_msg The message containing route information
+  */
+  void routeCallbackMessage(const cav_msgs::Route& route_msg);
+
+ /*!
+  * \brief Pulls vehicle information from CARMA Cloud at startup by providing its selected route in a ControlRequest message that is published after a route is selected.
+  * During operation at ~10s intervals the vehicle will make another control request for the remainder of its route.
+  * \param route_msg The message containing route information pulled from routeCallbackMessage()
+  */
+  cav_msgs::ControlRequest routeCallbackMessageLogic(const cav_msgs::Route& route_msg);
```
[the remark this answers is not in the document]
msmcconnell: If this function is just a helper function it should also be declared private.

carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcasterNode.h
```diff
@@ -53,6 +54,13 @@ class WMBroadcasterNode
    * @param map_msg The map message to publish
    */
   void publishMap(const autoware_lanelet2_msgs::MapBin& map_msg);
+  
+  /**
+   * @brief Callback to publish route messages
+   *
+   * @param route_msg The route message to publish
+   */
+  void publishRouteMsg(const cav_msgs::ControlRequest& route_msg);
```
msmcconnell, line 63: Same note about naming here. This doesn't operate on route messages.

carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcasterNode.h
```diff
@@ -67,8 +75,12 @@ class WMBroadcasterNode
 
   ros::Publisher map_pub_;
   ros::Publisher map_update_pub_;
+  ros::Publisher route_callmsg_pub_;
```
msmcconnell, line 78: Should be control_request_pub_

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -39,11 +42,13 @@ namespace carma_wm_ctrl
...
 
-WMBroadcaster::WMBroadcaster(const PublishMapCallback& map_pub, const PublishMapUpdateCallback& map_update_pub, std::unique_ptr<TimerFactory> timer_factory)
-  : map_pub_(map_pub), map_update_pub_(map_update_pub), scheduler_(std::move(timer_factory))
+WMBroadcaster::WMBroadcaster(const PublishMapCallback& map_pub, const PublishMapUpdateCallback& map_update_pub, const RouteMsgCallback& route_callmsg_pub,
+std::unique_ptr<TimerFactory> timer_factory)
+  : map_pub_(map_pub), map_update_pub_(map_update_pub), route_callmsg_pub_(route_callmsg_pub), scheduler_(std::move(timer_factory))
 {
   scheduler_.onGeofenceActive(std::bind(&WMBroadcaster::addGeofence, this, _1));
   scheduler_.onGeofenceInactive(std::bind(&WMBroadcaster::removeGeofence, this, _1));
+  std::bind(&WMBroadcaster::routeCallbackMessage, this, _1);
```
msmcconnell, line 51: This line is doing nothing because the bound object is not stored and can be deleted.

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -369,6 +374,115 @@ void WMBroadcaster::removeGeofence(std::shared_ptr<Geofence> gf_ptr)
...
+ route_callmsg_pub_(cR);
+
+}
+
+
+cav_msgs::ControlRequest WMBroadcaster::routeCallbackMessageLogic(const cav_msgs::Route& route_msg) 
+{
+
+
+  auto path = lanelet::ConstLanelets();
```
msmcconnell, line 392: You could just declare this object using the data type shown. No need for an assignment with the constructor. 
```
lanelet::ConstLanelets path;
```

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -369,6 +374,115 @@ void WMBroadcaster::removeGeofence(std::shared_ptr<Geofence> gf_ptr)
...
+cav_msgs::ControlRequest WMBroadcaster::routeCallbackMessageLogic(const cav_msgs::Route& route_msg) 
+{
+
+
+  auto path = lanelet::ConstLanelets(); 
+
+
+  if (!current_map_) {
+// Return / log warning etc.
+  ROS_INFO_STREAM("Value 'current_map_' does not exist.");
```
msmcconnell, line 397: use ROS_WARN_STREAM here as this should be a warning that we got a route but don't have a map. Also list what function the message is coming from in the error message.

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -369,6 +374,115 @@ void WMBroadcaster::removeGeofence(std::shared_ptr<Geofence> gf_ptr)
...
+  throw lanelet::InvalidObjectStateError(std::string("Base lanelet map is not loaded to the WMBroadcaster"));
+
+  exit(0);
```
msmcconnell, line 400: Do not call ever call exit(0) after throwing an exception. 
Just throw the exception.

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -369,6 +374,115 @@ void WMBroadcaster::removeGeofence(std::shared_ptr<Geofence> gf_ptr)
...
+}
+
+  for(auto id : route_msg.route_path_lanelet_ids) 
+  {
+    auto laneLayer = current_map_->laneletLayer.get(id);
```
msmcconnell, line 405: This variable is not a Layer but rather a lanelet, so the variable name could be clearer.

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -369,6 +374,115 @@ void WMBroadcaster::removeGeofence(std::shared_ptr<Geofence> gf_ptr)
...
+    path.push_back(laneLayer);
+  }
+ 
+  
+  if(path.size() == 0) exit(0);
```
msmcconnell, line 410: If this occurs you should just log a warning and return.

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -369,6 +374,115 @@ void WMBroadcaster::removeGeofence(std::shared_ptr<Geofence> gf_ptr)
...
+  ROS_WARN_STREAM("Got here safely");
```
msmcconnell, line 411: Looks like a debug log and can be removed.

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -369,6 +374,115 @@ void WMBroadcaster::removeGeofence(std::shared_ptr<Geofence> gf_ptr)
...
+   /*logic to determine route bounds*/
+  std::vector<lanelet::ConstLanelet> llt; 
+  std::vector<lanelet::BoundingBox3d> pathBox; 
+  float minX = 99999;
+  float minY = 99999;
+  float minZ = 99999;
+  float maxX = -99999;
+  float maxY = -99999;
+  float maxZ = -99999;
```
msmcconnell, line 420: Use std::numeric_limits to get these values.
Also use doubles by default instead of floats. Stack memory is not at a premium in our system so better to prefer doubles by default.

carma_wm_ctrl/src/WMBroadcaster.cpp
msmcconnell: Same for the others like this below

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -369,6 +374,115 @@ void WMBroadcaster::removeGeofence(std::shared_ptr<Geofence> gf_ptr)
...
+
+      if (pathBox.back().corner(lanelet::BoundingBox3d::BottomLeft).y() < minY)
+        minY = pathBox.back().corner(lanelet::BoundingBox3d::BottomLeft).y(); //minimum y-value
+
+    if (pathBox.back().corner(lanelet::BoundingBox3d::BottomLeft).z() < minZ)
+        minZ = pathBox.back().corner(lanelet::BoundingBox3d::BottomLeft).z(); //minimum z-value
+
+
+     if (pathBox.back().corner(lanelet::BoundingBox3d::TopRight).x() > maxX)
+        maxX = pathBox.back().corner(lanelet::BoundingBox3d::TopRight).x(); //maximum x-value
```
msmcconnell, line 443: I think for all these checks we could be working in 2d no need for the Z axis or 3d bounding boxes.

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -369,6 +374,115 @@ void WMBroadcaster::removeGeofence(std::shared_ptr<Geofence> gf_ptr)
...
+
+  std::string target_frame = base_map_georef_;
+  lanelet::projection::LocalFrameProjector local_projector(target_frame.c_str());
+  lanelet::BasicPoint3d localRoute;
+
+  localRoute.x()= minX;
+  localRoute.y()= minY;
+  localRoute.z()= minZ; 
+
+  lanelet::GPSPoint gpsRoute = local_projector.reverse(localRoute); //If the appropriate library is included, the reverse() function can be used to convert from local xyz to lat/lon
```
msmcconnell, line 468: I think the use of Route in these variable names is pretty confusing. These are just points on the route bounds.

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -369,6 +374,115 @@ void WMBroadcaster::removeGeofence(std::shared_ptr<Geofence> gf_ptr)
...
+
+  cav_msgs::ControlRequest cR; /*Fill the latitude value in message cB with the value of lat */
+  cav_msgs::ControlBounds cB; /*Fill the longitude value in message cB with the value of lon*/
+  cB.latitude = gpsRoute.lat;
+  cB.longitude = gpsRoute.lon;
+
+  cB.offsets[0] = maxX - minX;
+  cB.offsets[1] = maxY - minY;
+  cB.offsets[2] = maxZ - minZ;
+
```
msmcconnell, line 478: Add a TODO comment here to update this for new geofence spec. The spec you had for offsets was old, but we can update once the message updates are completed.

carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcaster.h
```diff
@@ -25,7 +25,22 @@
...
+#include <lanelet2_extension/projection/local_frame_projector.h>
 #include <carma_wm_ctrl/GeofenceScheduler.h>
+#include <lanelet2_core/geometry/BoundingBox.h>
+#include <lanelet2_core/primitives/BoundingBox.h>
+#include <carma_wm/WMListener.h>
+#include <carma_wm/WorldModel.h>
+#include <cav_msgs/Route.h>
+#include <cav_msgs/ControlRequest.h>
+#include <cav_msgs/ControlBounds.h>
+#include <lanelet2_extension/projection/mgrs_projector.h>
```
MishkaMN, line 42: Let's remove unused headers like mgrs and worldmodel

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -369,6 +374,115 @@ void WMBroadcaster::removeGeofence(std::shared_ptr<Geofence> gf_ptr)
...
+
+      path.pop_back(); //remove the added lanelet from path an reduce pack.size() by 1
+  } //end of while loop
+
+  ROS_WARN_STREAM("Successfully exited");
+
+  
+
+  std::string target_frame = base_map_georef_;
```
MishkaMN, line 460: I think we should check if it has been loaded similar to the way we are checking current_map_. Let's at least check if it is "" (emptry string). if so, throw, as other carma_wm_ctrl does.

docker/checkout.bash
```diff
@@ -38,12 +38,12 @@ done
...
 
 if [[ "$BRANCH" = "develop" ]]; then
-      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch $BRANCH
+      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch integration/routing
       git clone --depth=1 https://github.com/usdot-fhwa-stol/novatel_gps_driver.git --branch $BRANCH
       git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-utils.git --branch $BRANCH
       git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-messenger.git --branch $BRANCH
 else
-      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch integration/routing
+      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch develop
```
msmcconnell, line 46: This should be integration/routing

docker/checkout.bash
```diff
@@ -38,12 +38,12 @@ done
 cd ${dir}/src
 
 if [[ "$BRANCH" = "develop" ]]; then
-      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch $BRANCH
+      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch integration/routing
```
msmcconnell, line 41: This change was correct for integration testing, but I don't want it commited since it breaks the expected behavior of the -d option please change it back to $BRANCH

Issue #814, opened 2020-08-19
Add MobilityPath Publisher


saina-ramyar:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
A Mobility Path Publisher Node added that broadcasts the host vehicle's trajectory in ECEF frame.

### Expected behavior
Trajectory plan points converted to ECEF frame and published

### Software version this applies to

3.3.0
### Steps to reproduce the behavior

1. Do this
1. Then do this...

### Related work

Link to related tickets or prior related work here.

msmcconnell:
Added by above PRs

PR #815, opened 2020-08-19
Feature/mobilitypath publisher

saina-ramyar:
# PR Details
## Description
A Mobility Path Publisher Node is added that broadcasts the host vehicle's trajectory in ECEF frame.

## Related Issue
#814 

## Motivation and Context

Mobility Path message broadcasting periodically to announce its newest trajectory plan. It can be used for safety monitoring and plan cooperation.

## How Has This Been Tested?

Unit tested and Local integration tested.

## Types of changes

## Checklist:

### DIFF
--- a/mobilitypath_publisher/CMakeLists.txt
+++ b/mobilitypath_publisher/CMakeLists.txt
@@ -1,0 +1,98 @@
+#
+# Copyright (C) 2018-2020 LEIDOS.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+#
+
+cmake_minimum_required(VERSION 2.8.3)
+project(mobilitypath_publisher)
+
+## Compile as C++11, supported in ROS Kinetic and newer
+add_compile_options(-std=c++11)
+set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
+set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall")
+
+## Find catkin macros and libraries
+find_package(catkin REQUIRED COMPONENTS
+  roscpp
+  std_msgs
+  cav_msgs
+  carma_utils
+  tf2_ros
+)
+
+## System dependencies are found with CMake's conventions
+find_package(Boost REQUIRED COMPONENTS system)
+
+###################################
+## catkin specific configuration ##
+###################################
+
+catkin_package(
+  CATKIN_DEPENDS roscpp std_msgs cav_msgs carma_utils tf2_ros
+)
+
+###########
+## Build ##
+###########
+
+include_directories(
+  ${catkin_INCLUDE_DIRS}
+  ${Boost_INCLUDE_DIRS}
+  include
+)
+
+file(GLOB_RECURSE headers */*.hpp */*.h)
+
+add_executable( ${PROJECT_NAME}
+  ${headers}
+  src/mobilitypath_publisher.cpp
+  src/main.cpp)
+
+target_link_libraries(${PROJECT_NAME} mobilitypath_publisher_lib ${catkin_LIBRARIES} ${Boost_LIBRARIES})
+
+add_library(mobilitypath_publisher_lib src/mobilitypath_publisher.cpp)
+
+add_dependencies(mobilitypath_publisher_lib ${catkin_EXPORTED_TARGETS})
+
+
+#############
+## Install ##
+#############
+
+install(DIRECTORY include/${PROJECT_NAME}/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+  FILES_MATCHING PATTERN "*.hpp"
+  PATTERN ".svn" EXCLUDE
+)
+
+## Install C++
+install(TARGETS ${PROJECT_NAME}
+  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
+
+## Install Other Resources
+install(DIRECTORY launch config
+  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
+)
+
+#############
+## Testing ##
+#############
+catkin_add_gmock(${PROJECT_NAME}-test
+  test/test_mobility_path_publisher.cpp
+  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
+target_link_libraries(${PROJECT_NAME}-test mobilitypath_publisher_lib ${catkin_LIBRARIES})
+
--- a/mobilitypath_publisher/config/parameters.yaml
+++ b/mobilitypath_publisher/config/parameters.yaml
@@ -1,0 +1,3 @@
+# uncertainty threshould for updating the mobility path trajectory
+x_threshold: 20.0; #m
+y_threshold: 3; #m (width of a lane)
--- a/mobilitypath_publisher/include/mobilitypath_publisher.h
+++ b/mobilitypath_publisher/include/mobilitypath_publisher.h
@@ -1,0 +1,109 @@
+#pragma once
+
+/*
+ * Copyright (C) 2019-2020 LEIDOS.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+#include <string>
+#include <vector>
+#include <boost/uuid/uuid_generators.hpp>
+#include <boost/uuid/uuid_io.hpp>
+#include <tf2_ros/transform_listener.h>
+#include <boost/shared_ptr.hpp>
+#include <carma_utils/CARMAUtils.h>
+#include <cav_msgs/TrajectoryPlan.h>
+#include <cav_msgs/MobilityPath.h>
+#include <geometry_msgs/PoseStamped.h>
+
+namespace mobilitypath_publisher
+{
+
+    class MobilityPathPublication
+    {
+
+    public:
+
+        MobilityPathPublication();
+
+        // general starting point of this node
+        void run();
+
+        // local copy of pose
+        boost::shared_ptr<geometry_msgs::PoseStamped const> current_pose_;
+
+        cav_msgs::MobilityPath mobilityPathMessageGenerator(cav_msgs::TrajectoryPlan trajectory_plan);
+        
+
+    private:
+
+        // node handles
+        std::shared_ptr<ros::CARMANodeHandle> nh_, pnh_;
+
+
+        // ROS publisher
+        ros::Publisher  path_pub_;
+
+        // ROS subscribers
+        ros::Subscriber traj_sub_;
+        ros::Subscriber pose_sub_;
+        ros::Subscriber accel_sub_;
+
+        // ROS publishers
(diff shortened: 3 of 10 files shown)


mobilitypath_publisher/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+## Build ##
+###########
+
+include_directories(
+  ${catkin_INCLUDE_DIRS}
+  ${Boost_INCLUDE_DIRS}
+  include
+)
+
+file(GLOB_RECURSE headers */*.hpp */*.h)
```
msmcconnell, line 55: This is not needed. It keeps getting copied around in new packages, but it is not required as the include_directories covers this.

mobilitypath_publisher/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+
+add_executable( ${PROJECT_NAME}
+  ${headers}
```
msmcconnell, line 58: remove. You should only have .cpp files in your executable/library targets.

mobilitypath_publisher/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+
+
+#############
+## Install ##
+#############
+
+install(DIRECTORY include/${PROJECT_NAME}/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+  FILES_MATCHING PATTERN "*.hpp"
+  PATTERN ".svn" EXCLUDE
```
msmcconnell, line 76: You don't need 
```
 FILES_MATCHING PATTERN "*.hpp"
  PATTERN ".svn" EXCLUDE
```
Please remove. Another thing that keeps ending up in new packages.

mobilitypath_publisher/config/parameters.yaml
```diff
@@ -0,0 +1,3 @@
+# uncertainty threshould for updating the mobility path trajectory
+x_threshold: 20.0; #m
+y_threshold: 3; #m (width of a lane)
```
msmcconnell, line 3: Remove ";" from both lines. Does this actually load? I'm concerned this node hasn't actually been run. I think semi colons in the yaml would cause a failure.

mobilitypath_publisher/include/mobilitypath_publisher.h
```diff
@@ -0,0 +1,109 @@
...
+
+        // ROS subscribers
+        ros::Subscriber traj_sub_;
+        ros::Subscriber pose_sub_;
+        ros::Subscriber accel_sub_;
+
+        // ROS publishers
+        ros::Publisher mob_path_pub_;
+
+        cav_msgs::TrajectoryPlan latest_trajectory;
```
msmcconnell, line 65: Should add an ending "_" to be consistent with other member variables in this file.

mobilitypath_publisher/include/mobilitypath_publisher.h
```diff
@@ -0,0 +1,109 @@
...
+        double x_threshold , y_threshold; //m (width of a lane)
+
+        // initialize this node
+        void initialize();
+
+        // callbacks for the subscribers
+        void currentpose_cb(const geometry_msgs::PoseStampedConstPtr& msg);
+        void trajectory_cb(const cav_msgs::TrajectoryPlanConstPtr& msg);
+
+        // Compise Mobility Header
```
msmcconnell, line 84: typo should be "Compose"

mobilitypath_publisher/launch/mobilitypath_publisher.launch
```diff
@@ -0,0 +1,19 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+  License for the specific language governing permissions and limitations under
+  the License.
+-->
+<!--
+  This file is used to launch a the Mobility Path Publication Node
+-->
+<launch>
+    <rosparam command="load" file="$(find mobilitypath_publisher)/config/parameters.yaml"/>
```
msmcconnell, line 17: This should be nested inside the <node> tag so that it loads the parameters as private parameters.

mobilitypath_publisher/src/mobilitypath_publisher.cpp
```diff
@@ -0,0 +1,138 @@
...
+    MobilityPathPublication::MobilityPathPublication():
+                            x_threshold(20.0),
+                            y_threshold(3.0) {}
+
+    void MobilityPathPublication::initialize()
+    {
+        nh_.reset(new ros::CARMANodeHandle());
+        pnh_.reset(new ros::CARMANodeHandle("~"));
+        pnh_->param<double>("trajectory_threshold_x", x_threshold, 20.0);
+        pnh_->param<double>("trajectory_threshold_y", y_threshold, 3.0);
```
msmcconnell, line 32: These parameter names do not match the names in your config file.

mobilitypath_publisher/src/mobilitypath_publisher.cpp
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    cav_msgs::MobilityPath MobilityPathPublication::mobilityPathMessageGenerator(cav_msgs::TrajectoryPlan trajectory_plan)
+    {
+        cav_msgs::MobilityPath mobility_path_msg;
+        mobility_path_msg.header = composeMobilityHeader();
+        
+
+        std::vector<cav_msgs::TrajectoryPlanPoint> traj_points;
+        traj_points = trajectory_plan.trajectory_points;
```
msmcconnell, line 65: why is this intermediate variable needed? Seems like you could just pass trajectory_plan.trajectory_points into the function.

mobilitypath_publisher/src/mobilitypath_publisher.cpp
```diff
@@ -0,0 +1,138 @@
...
+        mobility_path_msg.trajectory = mob_path_traj;
+
+        return mobility_path_msg;
+    }
+
+    cav_msgs::MobilityHeader MobilityPathPublication::composeMobilityHeader(){
+        cav_msgs::MobilityHeader header;
+        header.sender_id = sender_id;
+        header.recipient_id = recipient_id;
+        header.sender_bsm_id = sender_bsm_id;
```
msmcconnell, line 76: sender_id should be loaded from the global ros parameter /vehicle_id
sender_bsm_id needs to match the current bsm id of this vehicle. So you need to have a subscriber for the outgoing bsm message which tracks the current id (changes every 5min) and uses it to populate this field.

mobilitypath_publisher/src/mobilitypath_publisher.cpp
```diff
@@ -0,0 +1,138 @@
...
+    cav_msgs::Trajectory MobilityPathPublication::TrajectoryPlantoTrajectory(std::vector<cav_msgs::TrajectoryPlanPoint> traj_points){
+        cav_msgs::Trajectory traj;
+        cav_msgs::LocationECEF ecef_location = TrajectoryPointtoECEF(traj_points[0]);
+
+        if (traj_points.size()<2){
+            ROS_WARN("Received Trajectory Plan is too small");
+            traj.offsets = {};
+        }
+        else{
+            for (int i=1; i<traj_points.size(); i++){
```
msmcconnell, line 93: Don't use signed types for iterators. Sonar cloud will complain as its bad form.
For for-loops like this you should default to the type size_t
```
for(size_t i = 1; ...)
```

mobilitypath_publisher/src/mobilitypath_publisher.cpp
```diff
@@ -0,0 +1,138 @@
...
+        traj.location = ecef_location;
+        return traj;
+    }
+
+    bool MobilityPathPublication::trajectoryPoseCheck(cav_msgs::TrajectoryPlanPoint point){
+        if (fabs(current_pose_->pose.position.x - point.x) > x_threshold || fabs(current_pose_->pose.position.y - point.y) > y_threshold)
+        {
+            return false;
+        }
+        return true;
```
msmcconnell, line 119: I'm not sure the check in this function makes sense. You are trying to evaluate which points to include in the path, but the x,y positions are in the map frame and could be any value. So you could reject large chunks of the path based on the orientation of the vehicle in the map. 

I think the main goal of this function is to only include points which are still to be executed. So it would make more sense to have a function that tells you which is the nearest point in the trajectory in both space and time. Select that as the current point, then keep all future points in the path but drop the previous points.

mobilitypath_publisher/src/mobilitypath_publisher.cpp
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    cav_msgs::LocationECEF MobilityPathPublication::TrajectoryPointtoECEF(cav_msgs::TrajectoryPlanPoint traj_point){
+        cav_msgs::LocationECEF ecef_point;    
+        try
+        {
+            geometry_msgs::TransformStamped tf = tf2_buffer_.lookupTransform("earth", "map", ros::Time(0));
```
msmcconnell, line 126: I would not recommend doing this lookup in a loop as it will be pretty slow and earth->map should not be allowed to change between points on a single mobility path message. Perform the lookup just once at the top of your callback before you begin processing the path and pass the transform into this function.

mobilitypath_publisher/src/mobilitypath_publisher.cpp
msmcconnell: After reviewing the design doc, my initial comment above isn't entirely accurate but there is still a fundamental issue.
The function "trajectoryPoseCheck" doesn't actually perform the way described in the highlevel design doc which used downtrack/crosstrack instead of x,y,z to perform the comparison.
What you are doing is comparing vehicle position in the map frame, to trajectory points in the map frame. This is less useful because all your are evaluating is the distance from each individual point rather than the distance from the curve you wish to follow or your error in time. Additionally, the uncertainty values in your config file will not map correctly as the map frame X-axis is rarely going to be aligned with your roadway.
I would recommend that you use the trackPos function in carma_wm/Geometry.h to compute the downtrack/crosstrack position of your vehicle relative to the entire path. Then have a crosstrack threshold, time threshold, and downtrack threshold that can be used to evaluate the vehicle's performance against the planned trajectory

msmcconnell:
Per our discussion with @kjrush, it would be acceptable to remove the error tracking logic given that the mobility paths are published at the same rate as the trajectories. Any attempt to do something more complex can be handled once an actual bandwidth limitation has been identified. Further discussion will be needed with UI team to identify if they actually require this progress report on the trajectory or if the maneuver plan would be more appropriate.

mobilitypath_publisher/CMakeLists.txt
saina-ramyar: removed

mobilitypath_publisher/CMakeLists.txt
saina-ramyar: fixed

mobilitypath_publisher/CMakeLists.txt
saina-ramyar: Removed second line, but the FILES_MATCHING PATTERN "*.hpp" seems to be needed for install

mobilitypath_publisher/config/parameters.yaml
saina-ramyar: changed

mobilitypath_publisher/include/mobilitypath_publisher.h
saina-ramyar: fixed

mobilitypath_publisher/launch/mobilitypath_publisher.launch
saina-ramyar: fixed

mobilitypath_publisher/src/mobilitypath_publisher.cpp
saina-ramyar: fixed

mobilitypath_publisher/src/mobilitypath_publisher.cpp
saina-ramyar: Modified

mobilitypath_publisher/src/mobilitypath_publisher.cpp
saina-ramyar: removed the check

mobilitypath_publisher/src/mobilitypath_publisher.cpp
saina-ramyar: fixed

mobilitypath_publisher/src/mobilitypath_publisher.cpp
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+
+
+    cav_msgs::LocationECEF MobilityPathPublication::TrajectoryPointtoECEF(cav_msgs::TrajectoryPlanPoint traj_point, geometry_msgs::TransformStamped tf){
+        cav_msgs::LocationECEF ecef_point;    
+
+        ecef_point.ecef_x = traj_point.x * tf.transform.translation.x;
+        ecef_point.ecef_y = traj_point.x * tf.transform.translation.y;
+        ecef_point.ecef_z = traj_point.x * tf.transform.translation.z;
```
MishkaMN, line 139: I am not sure how traj_point works but should these be traj_point.x traj_point.y traj_point.z?

Issue #800, opened 2020-08-05
Add localization status report


msmcconnell:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
To allow the user and other system components to better understand the status of localization, a report should be generated at the localization frequency 10Hz and provides a simple description of status that can be used by the UI to report to the user.

### Expected behavior
Localization status reported using single message and topic that can be used by the UI

### Software version this applies to

develop after 3.3.0
### Actual behavior
No unified status reporting exists. 
### Steps to reproduce the behavior
N/A

### Related work
Jira issue CAR-3109
Link to related tickets or prior related work here.

MishkaMN:
Resolved with above PRs

PR #816, opened 2020-08-19
CAR-3109: Feature/localization status report

msmcconnell:
# PR Details
## Description

This PR refactors/overhauls the gnss_ndt_selector package into a new localization_manager package. This package contains new logic to implement a more complex localization state machine and status reporting mechanism. 

The PR contains the following core classes
Localizer: This is the ROS node.
LocalizationManager: This is the worker class for Localizer and contains the entry/exit/do/signaling logic for the state machine
LocalizationTransitionTable: This class contains the actual transition rules used to determine the current state based on the provided signals.
LocalizationState: Enum of current states
LocalizationSignal: Enum of possible signals which could result in state changes
LocalizerMode: Enum of possible operational modes which this node can support. 

This is the state diagram that has been implemented
LocalizationStatusStatesComplex

Initial discussion on the draft design for this state machine can be found here
https://github.com/usdot-fhwa-stol/carma-msgs/pull/72
As development progressed the need to keep support for the localizer mode selection became apparent, so the state machine acquired some additional complexity. 

I'm still working on updating the design docs. 
## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/800

## Motivation and Context
Support UI display of localization status, more robust sensor failure behavior, future auto initialization behavior

## How Has This Been Tested?
Unit tests and basic local integration testing

## Types of changes

## Checklist:


localization_manager/src/LocalizationTransitionTable.cpp
```diff
@@ -0,0 +1,230 @@
...
+  {
+    // How to handle the combined conditions?
+    case LocalizationSignal::GOOD_NDT_FREQ_AND_FITNESS_SCORE:
+      setAndLogState(LocalizationState::OPERATIONAL, signal);
+      break;
+    case LocalizationSignal::POOR_NDT_FREQ_OR_FITNESS_SCORE:
+      setAndLogState(LocalizationState::DEGRADED, signal);
+      break;
+    case LocalizationSignal::LIDAR_SENSOR_FAILURE:
+      setAndLogState(LocalizationState::DEGRADED_NO_LIDAR_FIX, signal);
```
MishkaMN, line 85: On the state diagram, it seems like we need to check mode here?

localization_manager/src/LocalizationTransitionTable.cpp
```diff
@@ -0,0 +1,230 @@
...
+      logDebugSignal(signal);
+      break;
+  }
+}
+void LocalizationTransitionTable::signalWhenAWAIT_MANUAL_INITIALIZATION(LocalizationSignal signal)
+{
+  switch (signal)
+  {
+    case LocalizationSignal::INITIAL_POSE:
+      setAndLogState(LocalizationState::INITIALIZING, signal);
```
MishkaMN, line 191: On the state diagram, it seems like we need to check the mode here?

localization_manager/src/LocalizationTransitionTable.cpp
```diff
@@ -0,0 +1,230 @@
...
+      }
+      break;
+    case LocalizationSignal::UNUSABLE_NDT_FREQ_OR_FITNESS_SCORE:
+      if (mode_ == LocalizerMode::AUTO)
+      {
+        setAndLogState(LocalizationState::DEGRADED_NO_LIDAR_FIX, signal);
+      }
+      else
+      {
+        setAndLogState(LocalizationState::AWAIT_MANUAL_INITIALIZATION, signal);
```
MishkaMN, line 122: LocalizationState::AWAIT_MANUAL_INITIALIZATION,
LocalizationSignal::LIDAR_SENSOR_FAILURE
seem to have same operation, should we combine them to make fewer lines for testing, or is this way more clear?

localization_manager/src/LocalizationTransitionTable.cpp
msmcconnell: It's not actually possible to enter AWAIT_MANUAL_INITIALIZATION while in GNSS mode, so the check is implicit.

localization_manager/src/LocalizationTransitionTable.cpp
msmcconnell: I'll make it a fall through case

localization_manager/src/LocalizationTransitionTable.cpp
msmcconnell: Good catch. A check is needed, but the diagram showed the wrong check. I have updated the diagram and code.

msmcconnell:
Updated state diagram 
LocalizationStatusStatesComplex

PR #829, opened 2020-08-24
Feature/update route following lanechange

adev4a:
# PR Details
The branch adds creation of a lane change maneuver message to the route following plugin when a lane change is identified.

## Description
As identified in CAR-2785: Create High Level Design for Basic TravelDone, route following and lane change plugins are not communicating yet. This story is to design and implement the communication between route following and lane change tactical plugins to compose a lane change maneuver, and lane change request. 

When a lane change is identified a lane change maneuver messaged is push to the maneuver response vector.

## How Has This Been Tested?
A unit test was created to test the features added to the route following plugin.

The unit test uses a relation vector to identify a lane change and passes arguments to a compose_maneuver_lanechange function for creating a lane change maneuver message. The arguments of the message are then checked to see if they are accurate.
The relations vector is given two lanelet relation inputs, one without a relationType and the other specified as "Successor", using the logic from identifyLaneChange function, a "Successor" relation should call for a lane change. The test checks whether the correct vector element is being used for lane changing and then the parameters of the message composed.

## Types of changes

## Checklist:


route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -109,13 +97,36 @@ namespace route_following_plugin
...
+                //calculate start distance
+                double longl_acceleration=(RouteFollowingPlugin::TWENTY_FIVE_MPH_IN_MS-speed_progress)/LANE_CHANGE_TIME_MAX;
+                double lane_change_dist=(speed_progress*LANE_CHANGE_TIME_MAX)+0.5*(longl_acceleration)*pow(LANE_CHANGE_TIME_MAX,2);
+                double start_dist=end_dist-current_progress-lane_change_dist;
+                double start_time=start_dist/speed_progress;
+                //lane following till start_distance
+                resp.new_plan.maneuvers.push_back(
+                composeManeuverMessage(current_progress, start_dist, 
+                                       speed_progress, speed_progress, 
+                                       shortest_path[last_lanelet_index].id(), ros::Time::now()));
```
kjrush, line 109: Looking at how `composeManeuverMessage` works, I think this `ros::Time::Now()` needs to update based on the end time of the previous maneuver. The parameter might be named `current_time` but it's used to set the start time of each maneuver, and currently they all start at the beginning of the maneuver plan, overlapping instead of being in sequence.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -84,6 +84,7 @@ namespace route_following_plugin
             ROS_WARN_STREAM("Cannot find current lanelet in shortest path!");
             return true;
         }
+        ros::Time current_time=ros::Time::now();
```
kjrush, line 87: One last item, can't recall if I've mentioned this prior: I'd recommend at the start of the function checking to see if the service request has any maneuvers in it already. If it doesn't, everything here looks good. If it does, you should use the end time and distance of the last maneuver in that request as the `current_time` and `current_loc` for this planning, with the same target of 15s total plan length. Essentially just add on to the end of the plan if there's already maneuvers there.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -84,23 +85,64 @@ namespace route_following_plugin
...
+       if(req.prior_plan.maneuvers.size()!=0){
+            current_time=req.prior_plan.planning_completion_time;
+            double dist_covered=0.0;
+            // find current progress, current_lanelet and total maneuver length
+            for(auto i=0;i<req.prior_plan.maneuvers.size();i++){
+               auto type= req.prior_plan.maneuvers[i].type;
+               cav_msgs::Maneuver maneuver_msg=req.prior_plan.maneuvers[i];
+                switch(type)
+                {
+                    case 0:
```
kjrush, line 98: I'd recommend using the enum value names rather than integer literals `cav_msgs::Maneuver::LANE_FOLLOWING` rather than `0` for readability

kjrush (changes requested):

kjrush (approved):

kjrush:
Approved, holding off on merge until further testing on integration/routing branch to not introduce instability during testing.

msmcconnell:
Closing as changes from this PR have already been applied in other PRs and this PR is therefore no longer valid.

PR #837, opened 2020-08-31
Feature/port drayage plugin

itabrl:
# PR Details
## Description

add functions regarding communication between PortDrayagePlugin and StopAndWait tactical plugin.

## Motivation and Context

this is part of Automated CMV arrives at the Loading Point for the container in the facility (Port or Staging Area)
## How Has This Been Tested?

Unit tests
## Types of changes

## Checklist:


port_drayage_plugin/src/port_drayage_plugin.cpp
```diff
@@ -66,5 +77,99 @@ namespace port_drayage_plugin
...
+        //     return NULL;
+        // } else{
+        //     StopRule::Ptr stopRegelem = stop_rules.front();
+        //     // TODO get stopRegelem position and return the object with downtrack
+        //     stop_loc = geometry_msgs::PoseStampedConstPtr pose;
+        // }
+
+        // TODO
+        double stop_loc_downtrack = 0;
+        double current_loc_downtrack = wm_->routeTrackPos(current_loc).downtrack;
```
kjrush, line 107: You should probably use the downtrack `end_dist` value from the last maneuver in the maneuver list in the `req` or the vehicle's current position if there are no maneuvers in the request list. We're basically trying to plan what we'd like to do after the maneuvers in the request list have executed, so imagine we're at the end location of the last maneuver at the specified time, what would we do from there?

port_drayage_plugin/src/port_drayage_plugin.cpp
```diff
@@ -66,5 +77,99 @@ namespace port_drayage_plugin
...
+                    break;
+                case cav_msgs::Maneuver::INTERSECTION_TRANSIT_LEFT_TURN : start_time = req.prior_plan.maneuvers.back().intersection_transit_left_turn_maneuver.end_time;
+                    break;
+                case cav_msgs::Maneuver::INTERSECTION_TRANSIT_RIGHT_TURN : start_time = req.prior_plan.maneuvers.back().intersection_transit_right_turn_maneuver.end_time;
+                    break;
+                case cav_msgs::Maneuver::STOP_AND_WAIT : start_time = req.prior_plan.maneuvers.back().stop_and_wait_maneuver.end_time;
+                    break;
+            }
+
+            resp.new_plan.maneuvers.push_back(
```
kjrush, line 134: If - for example - we have a long lanelet and a stop at the end of it I think we'd want to add a lane-keeping maneuver prior to the stop, if that makes sense. Should be able to calculate how long it'll take to stop, then put in a lane keeping maneuver for the distance prior.

kjrush (changes requested):

kjrush (approved):

PR #853, opened 2020-09-10
Add an argument "non_simulation_mode" for CARMA simulation support

fangzhou1227:
# PR Details
## Description
CARMA simulation efforts require the ability to alter the way localization and object perception launch in order to replace those components with the simulated interfaces. As such these elements of the CARMA launch process should be refactored such that they can be configured via roslauch args which the CARMA Config for simulation can set.

Add an argument "non_simulation_mode", false if CARMA Plaform is launched in CARMA simulation mode with CARLA simulator. This argument is passed to environement.launch and localization.launch to configure related nodes.
## Motivation and Context

CARMA simulation interface publishes vehicle localization and external objects topics that are existing in CARMA platform. To make sure the right message is published, the launched nodes should be configured for different modes: simulation mode/normal mode. This refactoring will solve the duplicated topics issue.
## How Has This Been Tested?
Tested with CARMA simulation launching and standard mock drivers launching.

## Types of changes

## Checklist:


kjrush (changes requested):
Is it possible to invert this? So instead of `non_simulation_mode` defaulting to true, we use `simulation_mode` defaulted to false? Just a minor clarity thing.

kjrush (approved):

Issue #859, opened 2020-09-16
Integration Test - Integration Testing of new Guidance plugins, route node.


MishkaMN:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [X] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

In preparation for Vanden Plas, we need to integration test the following together to make sure they are working as expected without any major issues;

Route Following strategic plugin
In-lane cruising plugin/Route Following
Route node (generator and manager)
Guidance state updates to work with the new Route node
Guidance Re-Start feature is working.
waypoint_generator is working
 
### Carma version where this issue was discovered

3.3.0 integration/routing develop+
### Expected behavior
N/A
### Actual behavior

N/A
### Steps to reproduce the actual behavior

N/A
### Related work

Jira issue CAR-2029

MishkaMN:
This has been resolved with above PRs and as part of series of integration/routing merges.

PR #873, opened 2020-09-25
Fix/routing integration

MishkaMN:
# PR Details
## Description

Integration tested the route_following, inlanecruising, and route packages.

Testing is not fully completed yet as the car stops suddenly just after turning left from C-HUB before security post. 
So the car at least moves now. 
Current issue is that the plan_delegator stops receiving maneuver plans at some point which should be coming from inlanecruising_plugin. 

Troubleshooting page should include about how to correctly setup route file.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/859

## Motivation and Context
Need to merge integration/routing into develop soon for Vanden Plas.

## How Has This Been Tested?
Integration tested on Silver Lexus.
Car moves but stops after turning left. Sometimes goes a bit longer after turning, sometimes almost right away after finishing turn.

## Types of changes

## Checklist:

- [X] I have added any new packages to the sonar-scanner.properties file | No Need

### DIFF
--- a/arbitrator/config/arbitrator_params.yaml
+++ b/arbitrator/config/arbitrator_params.yaml
@@ -6,7 +6,7 @@ min_plan_duration: 6.0
 # Float: The nominal amount of time that an arbitrated plan should cover for the 
 # system to operate at best performance
 # Unit: s
-target_plan_duration: 15.0
+target_plan_duration: 13.0
 
 # Float: The planning frequency for generation for arbitrated plans
 # Unit: Hz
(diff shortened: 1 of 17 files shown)


arbitrator/config/arbitrator_params.yaml
```diff
@@ -6,7 +6,7 @@ min_plan_duration: 6.0
 # Float: The nominal amount of time that an arbitrated plan should cover for the 
 # system to operate at best performance
 # Unit: s
-target_plan_duration: 15.0
+target_plan_duration: 13.0
```
msmcconnell, line 9: Why is this 13 now? Seems like 15 was just fine and I noticed another 15 parameter below which might need to match

msmcconnell (changes requested):

arbitrator/config/arbitrator_params.yaml
msmcconnell: Accepted per our discussion on Slack.

PR #887, opened 2020-10-02
add logic for successing lanelets

saina-ramyar:
# PR Details
## Description

Added logic to find and eliminate adjacent lanelets in shortest path.
## Types of changes

## Checklist:

### DIFF
--- a/waypoint_generator/src/waypoint_generator/waypoint_generator.cpp
+++ b/waypoint_generator/src/waypoint_generator/waypoint_generator.cpp
@@ -355,13 +355,13 @@ std::vector<double> WaypointGenerator::get_speed_limits(std::vector<lanelet::Con
   return out;
 }
 
-std::vector<lanelet::ConstLanelet> WaypointGenerator::findSuccessingLanelets(){
+std::vector<lanelet::ConstLanelet> WaypointGenerator::findSuccessingLanelets() const
+{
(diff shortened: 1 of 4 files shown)


waypoint_generator/src/waypoint_generator/waypoint_generator.cpp
```diff
@@ -355,17 +355,37 @@ std::vector<double> WaypointGenerator::get_speed_limits(std::vector<lanelet::Con
...
+  out.push_back(shortest_path[0]);
+  for (size_t i=1; i<shortest_path.size(); i++)
   {
-    tmp.push_back(l);
+    lanelet::ConstLanelet l = shortest_path[i];
+    auto following = _wm->getRoute()->followingRelations(shortest_path[i-1]);
+    if (following[0].lanelet.id()==l.id() && following[0].relationType == lanelet::routing::RelationType::Successor){
+      out.push_back(l);
+    }
+    else{
```
msmcconnell, line 372: Rather than logging an error. I think we should probably throw an exception. The curvature calculation will not work in this case, so we shouldn't let the vehicle drive this route. If you throw an exception you can also use ASSERT_THROW in your unit test to validate the event.

msmcconnell (changes requested):

msmcconnell (approved):

PR #900, opened 2020-10-08
Fix/add lane array to lane converter

MishkaMN:
# PR Details
## Description
Lane_select node in the autoware does more than we originally expected. Therefore, we created our own node with a simpler logic.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/859

## Motivation and Context

lane_select in autoware was not working as expected. 
So we need to reduce the source of error.

## How Has This Been Tested?
Regression tested locally and saw if the topics match.

## Types of changes

## Checklist:


msmcconnell:
Let's not name this node lane_select since it doesn't actually "select" and that will cause confusion with the autoware component. Instead lets call it lane_array_to_lane_converter

lane_select/src/lane_select.cpp
```diff
@@ -0,0 +1,96 @@
...
+        if (current_lane_.waypoints.empty())
+        {
+            ROS_WARN_STREAM("lane_select: First lane in the Lane Array has no waypoints inside! Returning");
+            return;
+        }
+        publishVisualizer();
+        lane_pub_.publish(current_lane_);
+    }
+
+    void LaneSelect::publishVisualizer()
```
msmcconnell, line 60: Doesn't the waypoint_marker_publisher you remapped already contain this logic? Is is necessary to publish visualization waypoints from this node?

msmcconnell (approved):

PR #910, opened 2020-10-14
calc orientation and curvature

saina-ramyar:
# PR Details
## Description
Fit a spline to each trajectory and calculate points orientation and curvature.

## Types of changes

## Checklist:

### DIFF
--- a/inlanecruising_plugin/src/inlanecruising_plugin.cpp
+++ b/inlanecruising_plugin/src/inlanecruising_plugin.cpp
@@ -273,21 +273,15 @@ namespace inlanecruising_plugin
             throw std::invalid_argument("Insufficient Spline Points");
         }
         
-        int first_idx = 0;
-        int mid_idx = basic_points.size()/2;
-        int last_idx = basic_points.size() - 1;
 
         tk::spline spl;
-
         std::vector<double> points_x;
-        points_x.push_back(basic_points[first_idx].x());
-        points_x.push_back(basic_points[mid_idx].x());
-        points_x.push_back(basic_points[last_idx].x());
-
         std::vector<double> points_y;
-        points_y.push_back(basic_points[first_idx].y());
-        points_y.push_back(basic_points[mid_idx].y());
-        points_y.push_back(basic_points[last_idx].y());
+
+        for (size_t i=0; i<basic_points.size(); i++){
+            points_x.push_back(basic_points[i].x());
+            points_y.push_back(basic_points[i].y());
+        }
 
         spl.set_points(points_x, points_y);
 
--- a/inlanecruising_plugin/src/inlanecruising_plugin.cpp
+++ b/inlanecruising_plugin/src/inlanecruising_plugin.cpp
@@ -343,7 +343,7 @@ namespace inlanecruising_plugin
         double r = 0.5*(dist/std::sin(angle));
 
         double max_curvature = 100000;
-        double curvature = (1/r, max_curvature);
+        double curvature = std::min(1/r, max_curvature);
 
         return curvature;
     }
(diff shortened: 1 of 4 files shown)


inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -267,4 +267,89 @@ namespace inlanecruising_plugin
...
+
+    tk::spline InLaneCruisingPlugin::compute_fit(std::vector<lanelet::BasicPoint2d> basic_points){
+        int n = int(basic_points.size())/3;
+
+        tk::spline spl;
+
+        std::vector<double> points_x;
+        points_x.push_back(basic_points[n].x());
+        points_x.push_back(basic_points[2*n].x());
+        points_x.push_back(basic_points[3*n].x());
```
msmcconnell, line 279: This will seg fault for even numbers
If size = 12 for a vector with indexes 0-11
12 / 3 = 4
n = 4, n*2 = 8, n*3 = 12

Maybe wrap this in a min call
[std::min(basic_points.size() - 1, n*3)]

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -267,4 +267,89 @@ namespace inlanecruising_plugin
 
         return trajectory;
     }
+
+    tk::spline InLaneCruisingPlugin::compute_fit(std::vector<lanelet::BasicPoint2d> basic_points){
+        int n = int(basic_points.size())/3;
```
msmcconnell, line 272: The int() here is not needed as all types here are already integers and I don't think will compile. A cast is normally shown as (int) not int()

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -267,4 +267,89 @@ namespace inlanecruising_plugin
...
+        std::vector<double> cur_point{0.0, 0.0};
+        std::vector<double> next_point{0.0, 0.0};
+        double lookahead = 0.3;
+        for (size_t i=0; i<sampling_points.size(); i++){
+            cur_point[0] = sampling_points[i];
+            cur_point[1] = curve(cur_point[0]);
+            next_point[0] = cur_point[0] + lookahead;
+            next_point[1] = curve(next_point[0]);
+            double res = calculate_yaw(cur_point, next_point);
+            orientations.push_back(res);
```
msmcconnell, line 307: What's the conclusion on the reference frame of sampling_points here? Is it still X in the map frame? Won't that fail when there is multiple possible x values on the curve like we discussed in slack yesterday?

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -267,4 +267,89 @@ namespace inlanecruising_plugin
...
+        double lookahead = 0.3;
+        for (size_t i=0; i<sampling_points.size(); i++){
+            cur_point[0] = sampling_points[i];
+            cur_point[1] = curve(cur_point[0]);
+            next_point[0] = cur_point[0] + lookahead;
+            next_point[1] = curve(next_point[0]);
+            double cur = calculate_curvature(cur_point, next_point);
+            curvatures.push_back(cur);
+
+        }
```
msmcconnell, line 326: Same question as in the orientation part

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -267,4 +267,89 @@ namespace inlanecruising_plugin
...
+
+    double InLaneCruisingPlugin::calculate_curvature(std::vector<double> cur_point, std::vector<double> next_point){
+        double dist = sqrt(pow(cur_point[0] - next_point[0], 2) + pow(cur_point[1] - next_point[0], 2));
+
+        double angle = calculate_yaw(cur_point, next_point);
+
+        double r = 0.5*(dist/std::sin(angle));
+
+        double max_curvature = 100000;
+        double curvature = (1/r, max_curvature);
```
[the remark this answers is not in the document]
msmcconnell: Looks like you added std::min so it's fine now

msmcconnell (approved):

PR #926, opened 2020-10-26
Fix/testing no regions

msmcconnell:
# PR Details
## Description
* Refactors inlane-cruising to use maneuvers instead of the waypoint generator as a basis for trajectory generation. 
* Uses every 8th point at the basis for the spline fit (2m increments on tfhrc map)
* Resamples the spline at 1m increments
* Use a 8 waypoint look ahead distance to determine the speed of the current waypoint
* cleans up some of the accel limiting logic
* applys really simple moving average filter to both speeds and curvatures 

I was able to drive the full route without unsmooth speeds. 

## Types of changes

## Checklist:

### DIFF
--- a/carma_wm/src/CARMAWorldModel.cpp
+++ b/carma_wm/src/CARMAWorldModel.cpp
@@ -239,7 +239,7 @@ std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double st
   auto lanelet_map = route_->laneletMap();
   for (lanelet::ConstLanelet lanelet : lanelet_map->laneletLayer)
   {
-    if (shortest_path_only && shortest_path_view_->laneletLayer.exists(lanelet.id())) {
+    if (shortest_path_only && !shortest_path_view_->laneletLayer.exists(lanelet.id())) {
       continue; // Continue if we are only evaluating the shortest path and this lanelet is not part of it
     }
     lanelet::ConstLineString2d centerline = lanelet::utils::to2D(lanelet.centerline());
--- a/carma_wm/src/CARMAWorldModel.cpp
+++ b/carma_wm/src/CARMAWorldModel.cpp
@@ -222,7 +222,7 @@ TrackPos CARMAWorldModel::routeTrackPos(const lanelet::BasicPoint2d& point) cons
   return tp;
 }
 
-std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double start, double end) const
+std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double start, double end, bool shortest_path_only = false) const
 {
   // Check if the route was loaded yet
   if (!route_)
@@ -239,6 +239,9 @@ std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double st
   auto lanelet_map = route_->laneletMap();
   for (lanelet::ConstLanelet lanelet : lanelet_map->laneletLayer)
   {
+    if (shortest_path_only && shortest_path_view_->laneletLayer.exists(lanelet.id())) {
+      continue; // Continue if we are only evaluating the shortest path and this lanelet is not part of it
+    }
     lanelet::ConstLineString2d centerline = lanelet::utils::to2D(lanelet.centerline());
 
     auto front = centerline.front();
--- a/carma_wm/src/CARMAWorldModel.cpp
+++ b/carma_wm/src/CARMAWorldModel.cpp
@@ -222,6 +222,23 @@ TrackPos CARMAWorldModel::routeTrackPos(const lanelet::BasicPoint2d& point) cons
   return tp;
 }
 
+class LaneletDowntrackPair
+{
+  public:
+    lanelet::ConstLanelet lanelet_;
+    double downtrack_ = 0;
+
+    LaneletDowntrackPair(lanelet::ConstLanelet lanelet, double downtrack) : lanelet_(lanelet), downtrack_(downtrack) {}
+    bool operator<(const LaneletDowntrackPair& pair) const
+    {
+      return this->downtrack_ < pair.downtrack_;
+    }
+    bool operator>(const LaneletDowntrackPair& pair) const
+    {
+      return this->downtrack_ > pair.downtrack_;
+    }
+};
+
 std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double start, double end, bool shortest_path_only) const
 {
   // Check if the route was loaded yet
@@ -234,7 +251,8 @@ std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double st
     throw std::invalid_argument("Start distance is greater than or equal to end distance");
   }
 
-  std::vector<lanelet::ConstLanelet> vec;
+  std::vector<lanelet::ConstLanelet> output;
+  std::priority_queue<LaneletDowntrackPair,  std::vector<LaneletDowntrackPair>, std::greater<LaneletDowntrackPair>> prioritized_lanelets;
 
   auto lanelet_map = route_->laneletMap();
   for (lanelet::ConstLanelet lanelet : lanelet_map->laneletLayer)
@@ -255,10 +273,18 @@ std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double st
       continue;
     }
     // Intersection has occurred so add lanelet to list
-    vec.push_back(lanelet);
+    LaneletDowntrackPair pair(lanelet, min.downtrack);
+    prioritized_lanelets.push(pair);
   }
 
-  return vec;
+  output.reserve(prioritized_lanelets.size());
+  while(!prioritized_lanelets.empty()) {
+    auto pair = prioritized_lanelets.top();
+    prioritized_lanelets.pop();
+    output.push_back(pair.lanelet_);
+  }
+
+  return output;
 }
 
 lanelet::LaneletMapConstPtr CARMAWorldModel::getMap() const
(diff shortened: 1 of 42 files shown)


carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -222,7 +222,24 @@ TrackPos CARMAWorldModel::routeTrackPos(const lanelet::BasicPoint2d& point) cons
   return tp;
 }
 
-std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double start, double end) const
+class LaneletDowntrackPair
```
kjrush, line 225: Is this downtrack relative to the contained lanelet or to the route overall? Also probably doesn't matter for now, but since this looks like an internal data structure might be worth some consideration about hiding this in a private namespace or something.

inlanecruising_plugin/include/inlanecruising_config.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
```
kjrush, line 1: Copyright notice needed

inlanecruising_plugin/include/inlanecruising_plugin.h
```diff
@@ -30,90 +30,52 @@
 #include <boost/geometry/index/rtree.hpp>
 #include <carma_wm/Geometry.h>
 #include <cav_srvs/PlanTrajectory.h>
+#include <tf2/LinearMath/Matrix3x3.h> // TODO it should be possible to use tf2_eigen instead which would be more efficient since lanelet2 uses eigen under the hood
```
kjrush, line 33: Any resolution on this TODO? Or is this a longer term migration?

inlanecruising_plugin/include/third_party_library/spline.h
```diff
@@ -0,0 +1,404 @@
+/*
```
kjrush, line 1: Just making a note that we still need to resolve this licensing issue. I'm not even sure if we're technically able to host them in the same repo, but we'll have to sort this out with a different library.

plan_delegator/src/plan_delegator.cpp
```diff
@@ -123,6 +127,7 @@ namespace plan_delegator
             auto time_diff_sec = time_diff.toSec();
             // this assumes the vehicle does not have significant lateral velocity
             plan_req.request.vehicle_state.longitudinal_vel = distance_diff / time_diff_sec;
+            // TODO develop way to set yaw value for future points
```
kjrush, line 130: Is this referring to the yaw at the end of a maneuver (as planned by an individual plugin)?

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -98,7 +98,7 @@ namespace route_following_plugin
         }
         double current_progress = wm_->routeTrackPos(current_loc).downtrack;
         double speed_progress = current_speed_;
-        double total_maneuver_length = current_progress + mvr_duration_ * RouteFollowingPlugin::TWENTY_FIVE_MPH_IN_MS;
+        double total_maneuver_length = current_progress + mvr_duration_ * RouteFollowingPlugin::FIFTEEN_MPH_IN_MS;
```
kjrush, line 101: Is it possible to at least make this a config value?

inlanecruising_plugin/src/calculation.cpp
```diff
@@ -0,0 +1,308 @@
...
+    if (first) {
+      output.push_back(p);
+      prev_pair = p;
+      first = false;
+      continue;
+    }
+
+    double delta_d = lanelet::geometry::distance2d(std::get<0>(p), std::get<0>(prev_pair));
+
+    current_time += delta_d / std::get<1>(prev_pair);
```
kjrush, line 66: Is it valid to assume we're traveling at constant prev_point speed as opposed to average of both point's speeds?

inlanecruising_plugin/src/calculation.cpp
```diff
@@ -0,0 +1,308 @@
...
+    if (first) {
+      if (starting_downtrack > max_starting_downtrack) {
+        starting_downtrack = max_starting_downtrack;
+      }
+      first = false;
+    }
+
+    ROS_WARN_STREAM("Used downtrack: " << starting_downtrack);
+
+    auto lanelets = wm->getLaneletsBetween(starting_downtrack, lane_following_maneuver.end_dist, true);
```
kjrush, line 102: Hmm, this is kinda complicated since we're working in distance domain here, and arbitrator really cares about total plan duration in seconds. I'm wondering if we need to just definitively pick one to use, since the transform between these frames is related to speed, which is what we're also changing/computing. Maybe it doesn't matter here, since we're at the tactical level, but it's kind of got my head spinning.

inlanecruising_plugin/src/calculation.cpp
```diff
@@ -0,0 +1,308 @@
...
+        continue; // Skip the first point if we have already added points from a previous maneuver to avoid duplicates
+      }
+      points_and_target_speeds.push_back(std::make_pair(p, lane_following_maneuver.end_speed));
+    }
+  }
+
+  return points_and_target_speeds;
+}
+
+std::vector<PointSpeedPair> downsample_points(const std::vector<PointSpeedPair>& points, int nth_point)
```
kjrush, line 130: At some point we really ought to refactor some of this functionality using templates, so that we have one generic "downsample" function in a utils package somewhere.

inlanecruising_plugin/src/calculation.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+  double r = 0.5 * (dist / std::sin(angle));
+
+  double max_curvature = 100000;
+  double curvature = std::min(1 / r, max_curvature);
+
+  return curvature;
+}
+
+std::vector<double> compute_orientation_from_fit(tk::spline curve, std::vector<lanelet::BasicPoint2d> sampling_points)
```
kjrush, line 268: I don't see any usage of the curve argument here, is that intended?

inlanecruising_plugin/src/calculation.cpp
```diff
@@ -0,0 +1,308 @@
...
+{
+  std::vector<double> orientations;
+  std::vector<double> cur_point{ 0.0, 0.0 };
+  std::vector<double> next_point{ 0.0, 0.0 };
+  double lookahead = 0.3;
```
kjrush, line 273: Unused variable?

inlanecruising_plugin/src/calculation.cpp
```diff
@@ -0,0 +1,308 @@
...
+    next_point[0] = sampling_points[i + 1].x();
+    next_point[1] = sampling_points[i + 1].y();
+    double res = calculate_yaw(cur_point, next_point);
+    orientations.push_back(res);
+  }
+  orientations.push_back(orientations.back());
+  return orientations;
+}
+
+std::vector<double> compute_curvature_from_fit(tk::spline curve,
```
kjrush, line 287: Also not seeing the curve argument used here either, maybe I'm missing something?

inlanecruising_plugin/src/calculation.cpp
```diff
@@ -0,0 +1,308 @@
...
+                                                                     std::vector<lanelet::BasicPoint2d> sampling_points)
+{
+  std::vector<double> curvatures;
+  std::vector<double> cur_point{ 0.0, 0.0 };
+  std::vector<double> next_point{ 0.0, 0.0 };
+  double lookahead = 0.3;
```
kjrush, line 293: Unused variable

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -18,253 +18,507 @@
...
+    double dist = sqrt(dx * dx + dy * dy);
+    current_dt += dist;
+    downtracks.push_back(current_dt);
+    prev_p = p;
+  }
+
+  return downtracks;
+}
 
+struct DiscreteCurve
```
kjrush, line 165: Not sure I love the name, as since this contains speeds as well it's a bit more than just a curve. But it's fine for now, that's a minor quibble.

kjrush (changes requested):

inlanecruising_plugin/include/third_party_library/spline.h
msmcconnell: Yep. It's being tracked here
https://github.com/usdot-fhwa-stol/carma-platform/issues/936

inlanecruising_plugin/src/calculation.cpp
msmcconnell: I agree I think we need to hash this out a bit more. In this case, I was grabbing the related lanelets and then doing the time constraint on a future step.

inlanecruising_plugin/src/calculation.cpp
msmcconnell: I can probably whip that up during my clean up work.

inlanecruising_plugin/src/calculation.cpp
msmcconnell: Left over from previous implementation. I'll take it out

inlanecruising_plugin/src/calculation.cpp
msmcconnell: Left over from previous implementation. I'll take it out

inlanecruising_plugin/src/calculation.cpp
msmcconnell: Left over from previous implementation. I'll take it out

inlanecruising_plugin/src/inlanecruising_plugin.cpp
msmcconnell: Easy enough to change if you have another name in mind

carma_wm/src/CARMAWorldModel.cpp
msmcconnell: The downtrack used by this method is route downtrack which is why is can be used with maneuvers. I'll wrap the class definition in an anonymous namespace {}
It does seem like a name that others might use on accident.

msmcconnell:
Closing in favor of https://github.com/usdot-fhwa-stol/carma-platform/pull/947

Issue #801, opened 2020-08-05
Trajectory Executor unit tests failing


msmcconnell:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
The trajectory executor has failing unit tests on develop
### Expected behavior
No failed unit tests on develop
### Software version this applies to

develop after 3.3.0
### Actual behavior
As seen here https://app.circleci.com/pipelines/github/usdot-fhwa-stol/carma-platform/2286/workflows/d0ab8818-abb6-4528-89b9-74373099df8d/jobs/2410

The following error is being thrown from the trajectory executor unit tests. CI still passed which is also concerning. 
```
[trajectory_executor.rosunit-trajectory_executor_test_1/test_message_no_crash][passed]
[trajectory_executor.rosunit-trajectory_executor_test_1/test_emit_traj][FAILURE]
/opt/carma/src/CARMAPlatform/trajectory_executor/src/test/trajectory_executor_test_1.cpp:42
Expected: (msg_count) > (0), actual: 0 vs 0
Failed to receive trajectory execution message from TrajectoryExecutor node.
--------------------------------------------------------------------------------

[trajectory_executor.rosunit-trajectory_executor_test_1/test_runover_shutdown][FAILURE]
/opt/carma/src/CARMAPlatform/trajectory_executor/src/test/trajectory_executor_test_1.cpp:56
Value of: recv_sys_alert
  Actual: false
Expected: true
Failed to receive system shutdown alert message from TrajectoryExecutor node.
--------------------------------------------------------------------------------


SUMMARY
 * RESULT: FAIL
 * TESTS: 3
 * ERRORS: 0
 * FAILURES: 2

rostest log file is in /home/carma/.ros/log/rostest-e5d17312fc19-24829.log
-- run_tests.py: verify result "/opt/carma/build/test_results/trajectory_executor/rostest-test_trajectory_executor_1.xml"
[100%] Built target _run_tests_trajectory_executor_rostest_test_trajectory_executor_1.test
Scanning dependencies of target _run_tests_trajectory_executor_rostest_test_trajectory_executor_3.test
-- run_tests.py: execute commands
[...]

PR #928, opened 2020-10-26
update trajectory executor unit tests

saina-ramyar:
# PR Details
## Description
Trajectory executor unit tests are updated so they can be passed on CI.

## Related Issue
#801 

## Types of changes

## Checklist:

### DIFF
--- a/trajectory_executor/src/test/trajectory_executor_test_1.cpp
+++ b/trajectory_executor/src/test/trajectory_executor_test_1.cpp
@@ -38,7 +38,7 @@ TEST_F(TrajectoryExecutorTestSuite, test_emit_traj) {
 
     traj_pub.publish(plan);
 
-    std::this_thread::sleep_for(std::chrono::milliseconds(200));
+    std::this_thread::sleep_for(std::chrono::milliseconds(500));
     ASSERT_GT(msg_count, 0) << "Failed to receive trajectory execution message from TrajectoryExecutor node.";
 } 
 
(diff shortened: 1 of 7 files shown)


trajectory_executor/src/test/trajectory_executor_test_1.cpp
```diff
@@ -53,7 +53,7 @@ TEST_F(TrajectoryExecutorTestSuite, test_runover_shutdown) {
     traj_pub.publish(plan);
 
     std::this_thread::sleep_for(std::chrono::milliseconds(1500));
-    ASSERT_TRUE(recv_sys_alert) << "Failed to receive system shutdown alert message from TrajectoryExecutor node.";
+    ASSERT_FALSE(recv_sys_alert) << "Failed to receive system shutdown alert message from TrajectoryExecutor node.";
```
kjrush, line 56: Don't we want this to be `ASSERT_TRUE()` still? I think the concept of this test is that we want to make sure that we do receive the shutdown alert when we run past the trajectory.

trajectory_executor/src/test/trajectory_executor_test_3.cpp
```diff
@@ -29,7 +29,7 @@ TEST_F(TrajectoryExecutorTestSuite, test_control_plugin_not_found) {
 
     std::this_thread::sleep_for(std::chrono::milliseconds(500));
 
-    ASSERT_TRUE(recv_sys_alert) << "Failed to receive system shutdown alert message from TrajectoryExecutor node.";
+    ASSERT_FALSE(recv_sys_alert) << "Failed to receive system shutdown alert message from TrajectoryExecutor node.";
```
kjrush, line 32: Same as prior comment

trajectory_executor/src/test/trajectory_executor_test_2.cpp
```diff
@@ -28,7 +28,7 @@ TEST_F(TrajectoryExecutorTestSuite, test_emit_multiple) {
 
     std::this_thread::sleep_for(std::chrono::milliseconds(2000));
 
-    ASSERT_LE(10, msg_count) << "Failed to receive whole trajectory from TrajectoryExecutor node.";
+    ASSERT_LE(msg_count, 10) << "Failed to receive whole trajectory from TrajectoryExecutor node.";
```
kjrush, line 31: I also think this was supposed to be the other way around as well, we want to check if the msg count is at least 10, so that we verify it's sending every point in the trajectory as it walks it's way through it.

Issue #908, opened 2020-10-14
CAR-3769: health_monitor publishes system alerts continuously 


msmcconnell:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

System alert publication from health_monitor is periodic (10Hz), but should be event driven.
The health_monitor publishes system alerts in a periodic fashion. This causes issues with the UI which is not expecting a continuous stream of error messages. 

Instead of publishing the same message at 10Hz the health_monitor should publish alerts only once on a latched publisher to guarantee all nodes receive the most recent alert. 

### Carma version where this issue was discovered

develop after 3.3.0
### Expected behavior
System alerts are latched and published once
### Actual behavior

Health monitor continues to publish the same system alert until a status change occurs. 
### Steps to reproduce the actual behavior

Monitor system_alert topic at any point during system operation. 

### Related work

This issue was created by splitting https://github.com/usdot-fhwa-stol/carma-platform/issues/865

TonysCousin:
I think it's a little more complicated than the description indicates.  If a sensor goes offline, then, indeed, we only want a single message published to indicate that.  But if it comes back to life and the system starts using it again, we want to track that state, so if it goes offline a second time, we should see a second message about that new event.

I'll acknowledge that, for now, the system probably is not sophisticated enough to handle a sensor failing then coming back online, even if health monitor reported it so.  Therefore, until such resiliency is built in, doing the latch as described should be okay.

msmcconnell:
Resolved by above PR.

PR #940, opened 2020-11-04
Fix/health monitor publishing

lewisid:
# PR Details
## Description

The new logic added to the health_monitor.cpp file would check two conditions before publishing system alerts:

1. Whether or not the drivers are active/have failed
2. Whether or not the system alert for the specified driver condition had been previously published

For the second condition, two boolean arrays were included in the health monitor header file with each element corresponding to a driver condition that prompts a system alert.

## Related Issue

https://github.com/usdot-fhwa-stol/carma-platform/issues/908#
## Motivation and Context

The health_monitor publishes system alerts in a periodic fashion. This causes issues with the UI which is not expecting a continuous stream of error messages.
## How Has This Been Tested?

Integration testing took place on the Ford Fusion at TFHRC. The changes were saved into a Docker image that was then used to build Carma. Throughout the testing procedures, the /system_alert topic was continuously monitored to determine if the changes yielded the expected outcome.
## Types of changes

## Checklist:


health_monitor/src/health_monitor.cpp
```diff
@@ -139,8 +139,121 @@ namespace health_monitor
...
+        }//End car
+
+
         return true;
     }
 
+
+    void HealthMonitor::setisPublishedFalse()
+    {
+        for(int i=0; i<9; i++)
```
msmcconnell, line 248: use size_t for index types in for loops. Using signed types generally results in a warning. 
The size of these arrays should be constants to avoid typos. Or make the arrays vectors and use .size().

msmcconnell: If you use a std::vector you could also initialize to false with the constructor which would be a lot cleaner than these loops

std::vector my_vec(NumElements, InitialValue)

health_monitor/src/health_monitor.cpp
```diff
@@ -139,8 +139,121 @@ namespace health_monitor
...
+        
+        /*Publish each system alert only once*/
+        auto crit_driver_status_truck = driver_manager_.are_critical_drivers_operational_truck(time_now);
+        auto crit_driver_status_car = driver_manager_.are_critical_drivers_operational_car(time_now);
+
+        if(truck_ == true)
+        {
+            if((crit_driver_status_truck == "s_1_l1_1_l2_1_g_1") && (is_published_truck[0] == false)) //"All essential drivers are ready"
+            {
+             	nh_.publishSystemAlert(driver_manager_.handleSpin(truck_,car_,time_now,start_up_timestamp_,start_duration,start_time_flag_.isZero()));
```
msmcconnell, line 151: Seems like you could call driver_manager_.handleSpin once before the conditionals and cache the return values. It would clean up the contents of these publish calls quite a bit.

msmcconnell (changes requested):
This needs unit tests

health_monitor/src/health_monitor.cpp
```diff
@@ -139,8 +150,138 @@ namespace health_monitor
...
+
+  		        is_published_truck[0] = true;
+            } 
+           else if((start_time_flag_.isZero()==true) || (time_now - start_up_timestamp_ <= start_duration) && (is_published_truck[1] == false)) //"System is starting up..."
+            {
+                nh_.publishSystemAlert(dm);
+
+  		        is_published_truck[1] = true;
+            } 
+            else if((crit_driver_status_truck == "s_1_l1_0_l2_1_g_1") || (crit_driver_status_truck=="s_1_l1_1_l2_0_g_1")&& (is_published_truck[2] =false)) //One LIDAR Failure
```
TonysCousin, line 174: In this if stmt and next, the third test is an assignment, not a comparison operator!  It will always return true.  Using a single = instead of a double == in if tests is a common C++ mistake (easy typo).  Therefore, it is advantageous to use the power of booleans - they don't need to be compared to anything!  if (a) { ... }   or   if (!a) { ... }.  This not only avoids the possibility of mistyping a ==, it also makes the code a lot more readable.

health_monitor/src/health_monitor.cpp
```diff
@@ -139,8 +150,138 @@ namespace health_monitor
...
+        for(int j = 0; j < 7; j++)
+        {
+            is_published_car[j] = false;
+        }
+    }*/
+
+
+    std::vector<bool> HealthMonitor::getPubStatusCar()
+    {
+        ROS_WARN_STREAM("THISISATEST");
```
TonysCousin, line 275: Be sure to remove this test code before finalizing this PR.

health_monitor/src/health_monitor.cpp
```diff
@@ -139,8 +153,223 @@ namespace health_monitor
...
+    }
+
+    void HealthMonitor::setCarTrue()
+    {
+        car_ = true;
+    }
+
+    void HealthMonitor::setTruckTrue()
+    {
+        truck_ = true;
```
msmcconnell, line 371: Probably worth setting car_ = false whenever truck_ = true and vice versa to prevent a possible mismatch

msmcconnell: Also would want to log in that event since it likely indicates an error if one was set true when the other was already true. 
Throwing an exception would also be acceptable.

health_monitor/test/test_health_monitor.cpp
```diff
@@ -0,0 +1,699 @@
...
+        HealthMonitor node;
+        bool publish_status_c, publish_status_t;
+        auto pub_stat_c = node.getPubStatusCar();
+        for(int i = 0; i< pub_stat_c.size(); i++)
+        {
+            if (pub_stat_c.at(i) == false)
+                publish_status_c = false;
+            else
+                publish_status_c = true;
+        }
```
msmcconnell, line 24: This loop is invalid since you do not assert that each element is False. Running it like this with the assert at the end will only assert that the final element in the list is false which invalidates the test.

health_monitor/test/test_health_monitor.cpp
```diff
@@ -0,0 +1,699 @@
...
+        auto pub_stat_c = node.getPubStatusCar();
+
+        for(int i = 0; i< pub_stat_c.size(); i++)
+        {
+            if (pub_stat_c.at(i) == false)
+                publish_status_c = false;
+            else
+                publish_status_c = true;
+        }
+        ASSERT_EQ(publish_status_c, false); //Assert that none of the car's system alerts have been published
```
msmcconnell, line 91: Same note about loop as above

health_monitor/test/test_health_monitor.cpp
```diff
@@ -0,0 +1,699 @@
...
```
msmcconnell, line 145: Same note about loop as above

msmcconnell: Just apply this fix to all the loops like this in the tests

health_monitor/include/health_monitor.h
```diff
@@ -29,18 +29,41 @@ namespace health_monitor
...
             void run();
 
+            // spin callback function
+            bool spin_cb();
+
+            //Publisher Handling
+            std::vector<bool> getPubStatusCar();
+            std::vector<bool> getPubStatusTruck();
+
+            void spinObject(long time_now);
```
msmcconnell, line 51: Since this method does not seem to replace the spin_cb I think a more descriptive name would be more appropriate. What exactly is the method meant to do?

health_monitor/src/health_monitor.cpp
```diff
@@ -22,38 +22,52 @@
 namespace health_monitor
 {
 
+
+    HealthMonitor::HealthMonitor()
+    {
+        std::vector<bool>pub_truck(8, false);
+        std::vector<bool>pub_car(6, false);
```
msmcconnell, line 29: These are local variables since you have declared them here.
The correct syntax would be
```
 pub_truck = std::vector<bool>(8, false)
```

msmcconnell (changes requested):

health_monitor/src/health_monitor.cpp
```diff
@@ -25,10 +25,10 @@ namespace health_monitor
 
     HealthMonitor::HealthMonitor()
     {
-        std::vector<bool>pub_truck(8, false);
-        std::vector<bool>pub_car(6, false);
-        is_published_truck = pub_truck;
-        is_published_car = pub_car;
+        //std::vector<bool>pub_truck(8, false);
+        //std::vector<bool>pub_car(6, false);
```
msmcconnell, line 29: Remove commented out code

msmcconnell (changes requested):

health_monitor/src/health_monitor.cpp
```diff
@@ -22,38 +22,52 @@
...
     void HealthMonitor::initialize()
     {
         // init node handles
         ROS_INFO_STREAM("Initalizing health_monitor node...");
-        nh_ = ros::CARMANodeHandle();
-        pnh_ = ros::CARMANodeHandle("~");
+        pnh_.reset(new ros::CARMANodeHandle("~"));
+        nh_.reset(new ros::CARMANodeHandle());
+        //nh_ = std::shared_ptr<ros::CARMANodeHandle>();
+        //pnh_ = std::shared_ptr<ros::CARMANodeHandle>("~");
```
msmcconnell, line 44: Remove commented out code

health_monitor/src/health_monitor.cpp
```diff
@@ -138,9 +152,209 @@ namespace health_monitor
...
+
+  		        is_published_truck[1] = true;
+            } 
+            else if((crit_driver_status_truck == "s_1_l1_0_l2_1_g_1") || (crit_driver_status_truck=="s_1_l1_1_l2_0_g_1")&& (is_published_truck[2] =false)) //One LIDAR Failure
+            {
+             	nh_->publishSystemAlert(dm);
+
+  		        is_published_truck[2] = true;
+            } 
+            else if((crit_driver_status_truck =="s_1_l1_0_l2_1_g_0") || (crit_driver_status_truck =="s_1_l1_1_l2_0_g_0")&& (is_published_truck[3] =false)) //One LIDAR and GPS Failure
```
msmcconnell, line 182: bad = instead of == in this call as well

msmcconnell (changes requested):

health_monitor/include/health_monitor.h
```diff
@@ -84,8 +108,12 @@ namespace health_monitor
             std::string tactical_plugin_service_suffix_;
             ros::Time start_time_flag_; //Bool for start up time
 
-
-            // spin callback function
-            bool spin_cb();
+            //Unit Testing Functions
+            int getStatusTruck(std::string status, long time_now);
+            int getStatusCar(std::string status, long time_now);
```
msmcconnell, line 113: time_now parameters of both of these methods are unused and should be removed

msmcconnell:
Closing in favor of
https://github.com/usdot-fhwa-stol/carma-platform/pull/954
which is targeted at the release branch

Issue #937, opened 2020-10-29
lightbar is not automatically controlled by carma (first noticed on trucks)


MishkaMN:
### Types of Issue

- [X] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

It was brought to our attention that the lightbar does not correctly reflect the automation states of the carma on trucks.
Involved components are lightbar_driver and lightbar_manager.
Lightbar_manager was never tested on passenger cars, but only was on mock drivers, so potential error could be there.

This does not affect functional operation of the carma, but introduces a potential safety concern as it removes visual indication of the car's automation states.

### Carma version where this issue was discovered

carma 3.3.0+ develop
### Expected behavior
On all passenger cars and trucks:
Essential states:
When the car starts, or carma inactive, all lights should be off. 
When the carma is engaged, the green should on (solid).
When the carma is disengaged, but active, the green flash should on.

In other words, expected state transition:
![image](https://user-images.githubusercontent.com/20613282/97618242-2bf39e00-19f5-11eb-866c-9750d80b88c7.png)
Expected different patterns:
![image](https://user-images.githubusercontent.com/20613282/97618339-46c61280-19f5-11eb-809e-cb9a7767226a.png)

### Actual behavior

Lights are on, but it doesn't reflect the various states of the automation.

### Steps to reproduce the actual behavior

Pull every develop branches and run carma start all.

TonysCousin:
In particular, this issue was brought up by Brian Reed at ATEF, who is concerned about its implications for safety on that track.

MishkaMN:
The issue is resolved by above PRs.
Any further issue is most likely related to lightbar wiring mismatch or reprogrammability

PR #941, opened 2020-11-04
state change fix

MishkaMN:
# PR Details
## Description

When CARMA is disengaged while running for some reason, yet is still able to be re-engaged, the lightbar should flash green.
That scenario corresponds to GUIDANCE::INACTIVE state. This PR fixes that.

Also this PR fixed lightbar_manager sonar scan, and fixed some unit tests that had been failing, and adds capability to detect whether if ssc is publishing a topic for left turn or right turn and changes the orange lights according to it. This pattern overwrites any other pattern, but continues it back after the turn is ended.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/937

## Motivation and Context
Lightbar was not automatically controlled by CARMA. This is part of the investigation for that.

## How Has This Been Tested?
New unit tests pass and integration tested on Blue Lexus.

## Types of changes

## Checklist:

- [X] I have added any new packages to the sonar-scanner.properties file | Not Needed

### DIFF
--- a/lightbar_manager/src/lightbar_manager/lightbar_manager_worker.cpp
+++ b/lightbar_manager/src/lightbar_manager/lightbar_manager_worker.cpp
@@ -75,8 +75,34 @@ namespace lightbar_manager
         return;
     }
 
+    std::vector<lightbar_manager::LightBarIndicator> LightBarManagerWorker::handleTurnSignal(const automotive_platform_msgs::TurnSignalCommandPtr& msg_ptr)
+    {
+        std::vector<lightbar_manager::LightBarIndicator> turn_signal;
+        if (msg_ptr->turn_signal == current_turn_signal_)
+        {
+            return {};
+        }
+        if (msg_ptr->turn_signal == automotive_platform_msgs::TurnSignalCommand::LEFT) //NONE -> LEFT
+        {
+            turn_signal.push_back(lightbar_manager::LightBarIndicator::YELLOW_ARROW_LEFT);  
+        }
+        else if (msg_ptr->turn_signal == automotive_platform_msgs::TurnSignalCommand::RIGHT) //NONE -> RIGHT
+        {
+            turn_signal.push_back(lightbar_manager::LightBarIndicator::YELLOW_ARROW_RIGHT);
+        }
+        else if (msg_ptr->turn_signal == automotive_platform_msgs::TurnSignalCommand::NONE) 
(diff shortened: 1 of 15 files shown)


MishkaMN:
Drivers ready state also shows flashing green light now per discussion with John

lightbar_manager/src/lightbar_manager/lightbar_manager_worker.cpp
```diff
@@ -75,8 +75,34 @@ namespace lightbar_manager
...
+            turn_signal.push_back(lightbar_manager::LightBarIndicator::YELLOW_ARROW_LEFT);  
+        }
+        else if (msg_ptr->turn_signal == automotive_platform_msgs::TurnSignalCommand::RIGHT) //NONE -> RIGHT
+        {
+            turn_signal.push_back(lightbar_manager::LightBarIndicator::YELLOW_ARROW_RIGHT);
+        }
+        else if (msg_ptr->turn_signal == automotive_platform_msgs::TurnSignalCommand::NONE) 
+        {
+            // check previous signal
+            if (current_turn_signal_ == automotive_platform_msgs::TurnSignalCommand::RIGHT) // RIGHT -> NONE
```
msmcconnell, line 96: The comments day Right-> None and Left->None but it looks like the code is doing Right->Right and Left->Left?

MishkaMN: This is because that logic hits when the light is instructed to turn off. So if the prev state was left, it should still return left so that the next logic knows which light to turn off

msmcconnell:
Sonar cloud is not detecting it but the ROS test for the turn signal callback is passing on CI
```
terminate called without an active exception
[Testcase: testlightbar_manager_test] ... ok

[ROSTEST]-----------------------------------------------------------------------

[lightbar_manager.rosunit-lightbar_manager_test/testSetIndicator][passed]
[lightbar_manager.rosunit-lightbar_manager_test/testTurnOffAll][passed]
[lightbar_manager.rosunit-lightbar_manager_test/testTurnSignalCallback][passed]

SUMMARY
 * RESULT: SUCCESS
 * TESTS: 3
 * ERRORS: 0
 * FAILURES: 0
```
Since this test and the others are passing. I do not believe the sonar cloud coverage report is accurate. I think we can merge this

PR #942, opened 2020-11-05
Feature/pure pursuit

itabrl:
# PR Details
## Description

This branch include changes required to use pure pursuit as controller 

## Motivation and Context

to use this for ACE Demo
## How Has This Been Tested?

White Truck 
## Types of changes

## Checklist:

### DIFF
--- a/arbitrator/config/arbitrator_params.yaml
+++ b/arbitrator/config/arbitrator_params.yaml
@@ -21,9 +21,9 @@ beam_width: 3
 # Bool: Use fixed priority cost function over using the cost system for 
 # evaluating maneuver plans
 # Unit: N/a
-use_fixed_costs: false
+use_fixed_costs: true
 
 # Map: The priorities/costs associated with each plugin during the planning 
 # process, values will be normalized at runtime
 # Unit: N/a
-plugin_priorities: {AutowarePlugin: 10.0, GlidepathPlugin: 5.0}
+plugin_priorities: {AutowarePlugin: 10.0, GlidepathPlugin: 5.0, RouteFollowingPlugin: 4.0}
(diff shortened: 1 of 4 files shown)


TonysCousin (commented):
I have no problem with these changes, but want Saina to do the approval, just to be sure.  I do have a concern about the topic remapping.  It seems that we are moving to a topic that is called mpc_follower/trajectory, instead of the very nicely generic name of trajectory_plan.  It seems to me it would be better to force mpc to use the generic name, rather than force the rest of the system to masquarade as mpc.

In any case, if this works for now,  I'm okay with it.  It is intended to be a short-term work-around for the ACE demo.  But it's worth maybe writing an issue to straighten out the topic names around trajectory data to avoid unnecessarily specific or presumptuous names.

TonysCousin (commented):
All looks good to me.

saina-ramyar:
> I have no problem with these changes, but want Saina to do the approval, just to be sure. I do have a concern about the topic remapping. It seems that we are moving to a topic that is called mpc_follower/trajectory, instead of the very nicely generic name of trajectory_plan. It seems to me it would be better to force mpc to use the generic name, rather than force the rest of the system to masquarade as mpc.
> 
> In any case, if this works for now, I'm okay with it. It is intended to be a short-term work-around for the ACE demo. But it's worth maybe writing an issue to straighten out the topic names around trajectory data to avoid unnecessarily specific or presumptuous names.

@TonysCousin I believe the trajectory topic's name includes the controller's name intentionally, because in a general case with both controllers working we do not want their trajectories to get mixed up. Right now, the default controller in trajectory executor is MPC, and we are using pure pursuit so the easiest change is to remap the topic.

saina-ramyar (approved):

Issue #975, opened 2020-11-19
Avoiding dirty rosbags


MishkaMN:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [X] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Currently CARMA doesn't cleanly save rosbag when it is forced to close. Forcing it to close with ctrl + c is common practice among developers, so rosbag gets corrupted every time. `carma stop all` while carma is running, successfuly saves the rosbag.
We need a better way to save the rosbag as well as to fix the corrupted ones. 
Solution approach should also be documented on confluence page.

### Carma version where this issue was discovered

3.3.0+ develop

### Expected behavior
/opt/carma/logs rosbags are re-playable and there is no bag.active files.

### Actual behavior

there are many bag.active files left.

### Steps to reproduce the actual behavior

ctrl + c double times will leave corrupted files
ctrl + c once is okay.

### Related work

Jira issue CAR-3975

MishkaMN:
Above issue was satisfied by this entry on confluence page https://usdot-carma.atlassian.net/wiki/spaces/CRMPLT/pages/1080819713/Dirty+rosbag+fix

PR #976, opened 2020-11-19
change carma script

MishkaMN:
# PR Details
## Related Issue

https://github.com/usdot-fhwa-stol/carma-platform/issues/975
## Motivation and Context
There are many `.bag.active` files left after carma is done.
While carma is running, `ctrl + c` double times will corrupt the files (doing once or `carma stop all` is okay).

However, the common practice among developers is they `ctrl + c` twice and then do `carma stop all`. This way we won't be able to have functional rosbag.

This solution adds another `record` option to carma stopping. So that after double `ctrl + c`, we can do following to save successfully.
`carma stop all record` or `carma stop record` will successfully.

The reason we added to the `carma stop` script is because if developer forcibly stopped carma, `carma stop` still is called anyway to avoid any unexpected error for the next run.
The reason we added a new option instead of baking into `carma stop` is that if the bag file is big, it might take time to reindex, and developer don't need the file that often. 

General rosbag reindexing help guide should also be added to the confluence page.

## How Has This Been Tested?

locally tested
## Types of changes

## Checklist:


engineering_tools/carma
```diff
@@ -204,7 +204,26 @@ carma-config__install() {
     echo "Cleaning up temporary container..."
     docker rm carma-config-tmp
 }
+carma__record() {
```
msmcconnell, line 207: I think the name record here is a bit misleading. It implies this command actually causes a recording action to be taken.
I think 
carma fix bag might be more appropriate or something similar.

engineering_tools/carma
```diff
@@ -204,7 +204,26 @@ carma-config__install() {
...
+    last_modified_file="$(find "/opt/carma/logs" -type f \( -name \*.bag.active -o -name \*.bag \) -printf "%T@ %p\0" | awk 'BEGIN {RS="\0";} {if (NR==1){minmtime=$1; $1=""; lastmodified=$0;} else if (minmtime<$1){minmtime=$1; $1=""; lastmodified=$0;}} END{print substr(lastmodified,2)}')"
```
msmcconnell, line 208: I think some commends explaining this complex one liner are needed, since it is quite hard to read

engineering_tools/carma
```diff
@@ -204,7 +204,26 @@ carma-config__install() {
...
+
+        if test -f "/opt/carma/logs/${rosbag_name}.bag.orig.active"; then
+            rm "/opt/carma/logs/${rosbag_name}.bag.orig.active"
+        fi
+
+        mv ${last_modified_file} "/opt/carma/logs/${rosbag_name}.bag"
+
+        printf "Created a rosbag named %s\n" "/opt/carma/logs/${rosbag_name}.bag"
+
+    fi
```
msmcconnell, line 225: What happens if no bag is found. There needs to be some sort of notice to the user

msmcconnell (changes requested):

msmcconnell (approved):

PR #983, opened 2020-12-03
Feature/ndt update

MishkaMN:
# PR Details
## Description

This adds timeout functionality for ndt and gnss when selecting between them.
If one is timed out, it tries to switch to the other. if both fails, it prints ERROR msg.

## Motivation and Context
Sometimes gnss fails to publish msg and we lose localization.

## How Has This Been Tested?
has not tested yet

## Types of changes

## Checklist:

### DIFF
--- a/gnss_ndt_selector/config/parameters.yaml
+++ b/gnss_ndt_selector/config/parameters.yaml
@@ -9,11 +9,11 @@ unreliable_message_upper_limit: 3
 
 # Integer: Maximum allowed time for failed gnss before using ndt_pose
 # Units: ms
-gnss_pose_timeout: 2000
+gnss_pose_timeout: 1500
 
 # Integer: Maximum allowed time for failed ndt before using gnss_pose
 # Units: ms
-ndt_pose_timeout: 2000
+ndt_pose_timeout: 1500
 
 # localization mode for the platform
 # 0 - ndt only
--- a/gnss_ndt_selector/config/parameters.yaml
+++ b/gnss_ndt_selector/config/parameters.yaml
@@ -7,8 +7,18 @@ score_upper_limit: 60.0
 # num of message thredhold to determin if ndt node is reliable or not
 unreliable_message_upper_limit: 3
 
+# Integer: Maximum allowed time for failed gnss before using ndt_pose
+# Units: ms
+gnss_pose_timeout: 2000
+
+# Integer: Maximum allowed time for failed ndt before using gnss_pose
+# Units: ms
+ndt_pose_timeout: 2000
+
 # localization mode for the platform
 # 0 - ndt only
 # 1 - GNSS only
 # 2 - auto select between ndt and GNSS
 localization_mode: 2
+
+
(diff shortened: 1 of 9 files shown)


gnss_ndt_selector/config/parameters.yaml
```diff
@@ -7,8 +7,18 @@ score_upper_limit: 60.0
 # num of message thredhold to determin if ndt node is reliable or not
 unreliable_message_upper_limit: 3
 
+# Integer: Maximum allowed time for failed gnss before using ndt_pose
+# Units: ms
+gnss_pose_timeout: 2000
+
+# Integer: Maximum allowed time for failed ndt before using gnss_pose
+# Units: ms
+ndt_pose_timeout: 2000
```
msmcconnell, line 16: Lets make this 1500. 
2s is a long time and I think even in poor localization areas we were able to keep the computation time below 1s

gnss_ndt_selector/src/localizer_node.cpp
```diff
@@ -71,6 +79,53 @@ namespace localizer
 		}
 	}
 
+	bool Localizer::spinCallback()
+	{
+		// once initialized check if sensors are timeout
+		if (gnss_initialized_)
+		{
+			// check if timeout
+			if (gnss_operational_ && (ros::Time::now() - gnss_last_received_).toSec() > double(gnss_time_out_ / 1000))
```
msmcconnell, line 88: double(my_int) is not a valid cast
It should be 
(double) my_int

gnss_ndt_selector/src/localizer_node.cpp
```diff
@@ -71,6 +79,53 @@ namespace localizer
...
+			// try to switch to ndt if gnss not operational
+			if (!gnss_operational_ && ndt_operational_)
+			{
+				ROS_WARN_STREAM("Since GNSS has timed out, switching to NDT");
+				localization_mode_ = LocalizerMode::NDT;
+			}
+			else if (!gnss_operational_ && !ndt_operational_) //if both not operational although initialized
+			{
+				ROS_ERROR_STREAM("Both NDT and GNSS have timed out! Please take manual control!");
+				return false;
```
msmcconnell, line 102: I don't think its nessasary to kill the node (which is what returning false will do). If operational state is restored that's not a bad thing.

gnss_ndt_selector/src/localizer_node.cpp
```diff
@@ -71,6 +79,53 @@ namespace localizer
...
+			// try to switch to gnss if ndt not operational
+			if (!ndt_operational_ && gnss_operational_)
+			{
+				ROS_WARN_STREAM("Since NDT has timed out, switching to GNSS");
+				localization_mode_ = LocalizerMode::GNSS;
+			}
+			else if (!gnss_operational_ && !ndt_operational_) //if both not operational although initialized
+			{
+				ROS_ERROR_STREAM("Both NDT and GNSS have timed out! Please take manual control!");
+				return false;
```
msmcconnell, line 123: Same note

msmcconnell (changes requested):

msmcconnell (approved):

Issue #988, opened 2020-12-03
Adaptive lookahead distance for in-lane cruising


saina-ramyar:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
Currently, the in-lane cruising has a constant lookahead distance for adjusting the speed of trajectory. However, depending on the vehicle's speed, the lookahead needs to be different. If the vehicle's velocity is high and it is approaching a low-speed area such as a tight curve, it must be aware well in advance.
Therefore, the lookahead distance should be adaptive with speed, so it can be large when the vehicle is moving fast, and small if the vehicle is moving slowly.

### Expected behavior'
Lookahead distance should be changed based n the vehicle's velocity.

### Actual behavior

The lookahead distance is a constant number of points ahead.

### Steps to reproduce the actual behavior

1. Do this
1. Then do this...

msmcconnell:
Resolved by above PR

PR #989, opened 2020-12-03
Feature/curve entry lookahead

saina-ramyar:
# PR Details
## Description

In this feature, an adaptive lookahead distance based on the vehicle's velocity is implemented that allows the in-lane cruising plugin to know about upcoming speed increases or decreases (such as curves and turns) in advance, to generate trajectories with smooth speeds.

## Related Issue
#988 

## Motivation and Context
This feature helps in-lane cruising plugin to generate smooth trajectories.

## How Has This Been Tested?
Unit tested and integration tested on Blue Lexus

## Types of changes

## Checklist:

### DIFF
--- a/inlanecruising_plugin/include/inlanecruising_plugin/inlanecruising_plugin.h
+++ b/inlanecruising_plugin/include/inlanecruising_plugin/inlanecruising_plugin.h
@@ -208,8 +208,25 @@ public:
    */ 
   Eigen::Isometry2d curvePointInMapTF(const Eigen::Isometry2d& curve_in_map, const lanelet::BasicPoint2d& p, double yaw) const;
 
+  /**
+   * \brief Returns the speeds of points closest to the lookahead distance.
+   * 
+   * \param points The points in the map frame that the trajectory will follow. Units m
+   * \param speeds Speeds assigned to points that trajectory will follow. Unit m/s
+   * \param lookahead The lookahead distance to obtain future points' speed. Unit m
+   * 
+   * \return A vector of speed values shifted by the lookahead distance.
+   */ 
+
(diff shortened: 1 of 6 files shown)


inlanecruising_plugin/include/inlanecruising_plugin/inlanecruising_plugin.h
```diff
@@ -208,6 +208,10 @@ class InLaneCruisingPlugin
    */ 
   Eigen::Isometry2d curvePointInMapTF(const Eigen::Isometry2d& curve_in_map, const lanelet::BasicPoint2d& p, double yaw) const;
 
+  std::vector<double> get_lookahead_speed(const std::vector<lanelet::BasicPoint2d>& points, const std::vector<double>& speeds, const double& lookahead);
```
kjrush, line 211: Would be nice to have doc comments for both of these methods, just for consistency.

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -392,6 +397,62 @@ std::vector<cav_msgs::TrajectoryPlanPoint> InLaneCruisingPlugin::compose_traject
...
+  std::vector<double> out;
+  out.reserve(speeds.size());
+
+  for (int i = 0; i < points.size(); i++)
+  {
+    int idx = i;
+    double min_dist = std::numeric_limits<double>::max();
+    for (int j=i+1; j < points.size(); j++){
+      double dist = lanelet::geometry::distance2d(points[i],points[j]);
+      if (abs(lookahead - dist) <= min_dist){
```
kjrush, line 445: Seems sensible to me. Could produce weirdness if we have some kind of ultra tight u-turn or something, where there's a point at a higher index that's actually closer to being lookahead dist from our current point, but that feels like an extremely esoteric case and such a trajectory should never exist so it doesn't really give me cause for concern.

PR #990, opened 2020-12-07
Feature/display route

aswath1:
# PR Details
## Description

Now that we are no longer using waypoints in the planning process, the route cannot be visualized on rviz via waypoints. An alternate visualization method should be employed. This story is to design and implement a simple way to visualize the loaded route in rviz without blocking the view of the vehicle or lanelets when in top down view on rviz. The easiest method would be to add a publisher to the route node which publishes rviz markers. Effort should be made to separate this logic from the core route planning logic so that could stays easy to read.
## How Has This Been Tested?
Unit Test
Local Integration tested

## Types of changes

## Checklist:


aswath1:
Related Enhancement 
https://github.com/usdot-fhwa-stol/carma-platform/issues/991

MishkaMN (approved):

PR #1007, opened 2020-12-10
Feature/cross track error

lewisid:
# PR Details
## Description

New logic has been added to the route node in order to detect and handle crosstrack error. Once the vehicle's position has been determined, a function will determine whether or not the vehicle has deviated from the route but remains in the lanelet bounds, departed from the route and is outside the route lanelet bounds, or if the vehicle has not left the route at all and no crosstrack error has occurred. This process evaluates the lanelet crosstrack and downtrack distances and decides whether or not the vehicle's position exceeds these values. Should the crosstrack error check return true, a message will be published to the /guidance/route_event topic indicating that the vehicle has departed from the route and the driver should manually take control.

## Related Issue

https://github.com/usdot-fhwa-stol/carma-platform/issues/1024
## Motivation and Context

As a tester or vehicle operator, I would like to know that Carma Platform has a safety mechanism for handling situations when it deviates from the planned route enough to put the vehicle in danger, such as running off the side of the road. 

The design needs to be more sophisticated than simply measuring CTE relative to the "route centerline", as that line may move laterally as the road geometry changes, even though the vehicles trajectory keeps it going in a straight line.
## How Has This Been Tested?

Unit testing was performed through gtest/gmock. Integration testing took place at TFHRC by running a Docker image in the White Ford Fusion.

## Types of changes

## Checklist:


lewisid:
12/10/2020
Tentative PR: Integration Testing has not been completed at this time.

msmcconnell:
Copying some thoughts from slack so other reviewers can see them. 

* WMListener is responsible only for providing an interface to access route and map data. It should not modify the state of the system. Any logic to broadcast a route departed event should be placed inside the route node which is the node responsible for monitoring route state.
* In the event we leave the route a ROUTE_DEPARTED event is the only thing that should be broadcast. Issuing a system shutdown is an overstep of authority. It is up to the guidance system to determine if a ROUTE_DEPARTED event should result in manual control it is not the responsibility of the route node to decide how to respond to route events.

carma_wm/include/carma_wm/WMListener.h
```diff
@@ -22,6 +22,8 @@
 #include <ros/callback_queue.h>
 #include <carma_wm/WorldModel.h>
 #include <carma_utils/CARMAUtils.h>
+#include <cav_msgs/RouteEvent.h>
```
MishkaMN, line 25: Let's remove these

carma_wm/src/WMListenerWorker.h
```diff
@@ -92,13 +101,14 @@ class WMListenerWorker
...
 
-
 private:
   std::shared_ptr<CARMAWorldModel> world_model_;
   std::function<void()> map_callback_;
   std::function<void()> route_callback_;
   void newRegemUpdateHelper(lanelet::Lanelet parent_llt, lanelet::RegulatoryElement* regem) const;
   double config_speed_limit_;
+  std::shared_ptr<ros::CARMANodeHandle> nh_;
+  lanelet::BasicPoint2d position;
```
MishkaMN, line 111: let's remove unused changes

route/include/route_generator_worker.h
```diff
@@ -153,6 +155,12 @@ namespace route {
          */
         visualization_msgs::MarkerArray compose_route_marker_msg(const lanelet::Optional<lanelet::routing::Route>& route);
 
+
+
+        void routeEventCallback(cav_msgs::RouteEvent status);
```
MishkaMN, line 160: seems like we are not using this

MishkaMN (changes requested):

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +442,65 @@ namespace route {
...
+    lanelet::BasicPoint2d position;
+
+    position.x()= msg->pose.position.x;
+    position.y()= msg->pose.position.y;  
+
+    if (boost::geometry::distance(position, current.polygon2d()) > cross_track_dist)
+        {
+            cte_count++;
+
+            if(cte_count > 4)
```
MishkaMN, line 466: let's make this parameter

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +442,65 @@ namespace route {
...
+    auto following_llts = world_model_->getMapRoutingGraph()->following(current);
+    lanelet::ConstLanelets filtered_llts;//TODO: Filter lanelets from following lanelets that are not a part ot the route
+    int out_count = 0;
+    for(auto i:following_llts)
+    {
+       
+        if (boost::geometry::within(position, i.polygon2d()) == false) 
+           out_count++;
+
+        if (boost::geometry::distance(position, current.polygon2d()) > 2.0)
```
MishkaMN, line 484: let's make this a parameter

MishkaMN: also let's use downtrack distance. See https://github.com/usdot-fhwa-stol/carma-platform/blob/fff790f974b53de0d0b2303dd7b63dfa88facc20/route/src/route_generator_worker.cpp#L348

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +442,65 @@ namespace route {
...
+        out_of_llt_bounds = true;
+      }
+
+    auto following_llts = world_model_->getMapRoutingGraph()->following(current);
+    lanelet::ConstLanelets filtered_llts;//TODO: Filter lanelets from following lanelets that are not a part ot the route
+    int out_count = 0;
+    for(auto i:following_llts)
+    {
+       
+        if (boost::geometry::within(position, i.polygon2d()) == false)
```
MishkaMN, line 481: I think this would all return false because the current position is in current lanelet, and guaranteed not to be inside following lanelets. I think we just need to find the lanelets on the route by comparing ids.

route/src/route_generator_worker.cpp
MishkaMN: And only find downtrack distance to those that are on the route. Once found acceptable, we have to check the crosstrack distance again to that lanelet, the following lanelet of current one. If it is within threshold we return true. If nothing returned true in the end, we return false.

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +442,65 @@ namespace route {
...
+            out_of_llt_bounds = true;
+
+    }
+    if(out_count == following_llts.size()) //If our current position is not within the bounds of any of the following lanelets, then return true
+        out_of_llt_bounds = true;
+
+
+    
+    
+    if(out_of_llt_bounds == true)
```
MishkaMN, line 494: `return out_of_llt_bounds` would achieve the same result

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +442,65 @@ namespace route {
...
+    {
+       
+        if (boost::geometry::within(position, i.polygon2d()) == false) 
+           out_count++;
+
+        if (boost::geometry::distance(position, current.polygon2d()) > 2.0)
+            out_of_llt_bounds = true;
+
+    }
+    if(out_count == following_llts.size()) //If our current position is not within the bounds of any of the following lanelets, then return true
```
MishkaMN, line 488: as mentioned above we don't need to check if it is within following lanelets, but we may need to create a counter like we did for crosstrack checking.

MishkaMN (changes requested):

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +442,77 @@ namespace route {
...
+    auto following_llts = world_model_->getMapRoutingGraph()->following(current);
+    lanelet::ConstLanelets filtered_llts;//TODO: Filter lanelets from following lanelets that are not a part of the route
+ 
+
+    for(auto i:following_llts)
+    {
+       for(auto j: route)
+       {
+           if (i.id() == j.id())
+            filtered_llts.push_back(i);
```
MishkaMN, line 484: i think we need to pass to this route object to be abel to use it. Anyways, on the second thought, if we are using route.get().shortestPath() I think we are guaranteed to have only 1 following lanalet. So we can just end the loop if we found the next one.

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +442,77 @@ namespace route {
...
+    {
+        if (boost::geometry::within(position, i.polygon2d()) == false) 
+           out_count++;
+           
+        if (boost::geometry::distance(position, current.polygon2d()) > ll_downtrack_distance_)
+            out_of_llt_bounds = true;
+    }
+
+    if(out_count == filtered_llts.size()) //If our current position is not within the bounds of any of the following lanelets, then return true
+        out_of_llt_bounds = true;
```
MishkaMN, line 498: Adding to the above comment, if we only have 1 next lanelet, we just need to check the downtrack distance to that. If the condition is met, update the counter and mark false until certain number is reached like you are doing above. 
Please use downtrack distance:
https://github.com/usdot-fhwa-stol/carma-platform/blob/582abd5a6a50878bb33266bd2a7bbab568dbd233/route/src/route_generator_worker.cpp#L345

MishkaMN (changes requested):

route/include/route_generator_worker.h
```diff
@@ -153,6 +155,14 @@ namespace route {
          */
         visualization_msgs::MarkerArray compose_route_marker_msg(const lanelet::Optional<lanelet::routing::Route>& route);
 
+
+        bool crossTrackErrorCheck(const geometry_msgs::PoseStampedConstPtr& msg, lanelet::ConstLanelet current_llt);
```
msmcconnell, line 159: Function comments are needed.

route/src/route_generator_worker.cpp
```diff
@@ -373,8 +373,12 @@ namespace route {
             // check if we left the seleted route by cross track error
             if(std::fabs(current_crosstrack_distance_) > cross_track_max_)
```
msmcconnell, line 374: Shouldn't the added functionality replace this behavior? The whole point is to not have a fixed cross track value that would need to change between roads.

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +442,77 @@ namespace route {
         return true;
     }
 
+bool RouteGeneratorWorker::crossTrackErrorCheck(const geometry_msgs::PoseStampedConstPtr& msg, lanelet::ConstLanelet current)
```
msmcconnell, line 445: Still looking at this, but I'm a bit skeptical of the logic here.

msmcconnell (changes requested):

route/src/route_generator_worker.cpp
msmcconnell: I believe the original approach we discussed was to have the alert trip if the vehicle was off the route. The route object contains all the lane lets on a given road that are in the same direction as the shortest path as described here 
https://github.com/usdot-fhwa-stol/carma-platform/tree/develop/carma_wm#route-definition

So I would think we would simply need to check if the vehicle was inside any of the lanelets on the route (not just shortest path) and  if not check what the cross-track distance to the nearest lanelet was to see if action needed to be taken. This logic seems to compare against the shortest path rather than the whole route which means it would trigger when passing another vehicle.

msmcconnell (changes requested):

route/config/route_config_params.yaml
```diff
@@ -4,4 +4,10 @@ max_crosstrack_error: 25
 
 # Double: downtrack_range
 # Units: meter
-destination_downtrack_range: 0
\ No newline at end of file
+destination_downtrack_range: 0
+
+# Int: cross-track deviation counter
```
kjrush, line 9: Is this measured in timesteps?

route/config/route_config_params.yaml
```diff
@@ -4,4 +4,10 @@ max_crosstrack_error: 25
...
+cte_count: 0
+
+# Int: lanelets outside of the route counter
```
kjrush, line 12: Same here

route/include/route_generator_worker.h
```diff
@@ -153,6 +155,21 @@ namespace route {
...
+        * 
+        *  \param msg Msg that contains the vehicle's current position
+        *  \param current_llt The lanelet that the vehicle is currently in
+        *  \param llt_track The crosstrack and downtrack distance of the current lanelet
+        * */
+        bool crosstrack_error_check(const geometry_msgs::PoseStampedConstPtr& msg, lanelet::ConstLanelet current_llt, carma_wm::TrackPos llt_track);
+
+        void setCTEcounter(int cte_counter);
+
+        void set_out_counter(int out_counter);
```
kjrush, line 170: Formatting consistencies between setCTEcounter and set_out_counter, also need doc comments.

route/config/route_config_params.yaml
kjrush: Also would be good for both of these to describe what happens when the count reaches these thresholds, so that someone modifying these would know what is changing.

route/src/route_generator_worker.cpp
```diff
@@ -369,13 +369,13 @@ namespace route {
...
             }
-
             // check if we left the seleted route by cross track error
-            if(std::fabs(current_crosstrack_distance_) > cross_track_max_)
-            {
-                this->rs_worker_.on_route_event(RouteStateWorker::RouteEvent::ROUTE_GEN_FAILED);
-                publish_route_event(cav_msgs::RouteEvent::ROUTE_DEPARTED);
-            }
+            bool departed = crosstrack_error_check(msg, current_lanelet, lanelet_track);
+            if (departed == true)
```
kjrush, line 374: Could just be `if (departed)`

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,91 @@ namespace route {
...
+    ROS_ERROR_STREAM("WMListener received a route before a map was available. Dropping route message.");
+    return true;
+  }
+
+   
+
+  auto route = lanelet::ConstLanelets();
+  for(auto id : route_msg_.route_path_lanelet_ids)
+  {
+    auto ll = world_model_->getMap()->laneletLayer.get(id);
```
kjrush, line 454: Is this just adding all the lanelets in the map to the the route check here? Or is this variable not actually describing the route? Might be a defect if the former, a misleading name if the latter.

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,91 @@ namespace route {
...
+    route.push_back(ll);
+  }
+
+  bool out_of_llt_bounds = false;
+
+    lanelet::BasicPoint2d position;
+
+    position.x()= msg->pose.position.x;
```
kjrush, line 462: Does this work? I don't know if assigning to `position.x()` actually updates the object's value. I'd think you just want `position.x =...`

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,91 @@ namespace route {
...
+    position.y()= msg->pose.position.y;  
+
+    if (boost::geometry::distance(position, current.polygon2d()) > cross_track_dist) //Evaluate lanelet crosstrack distance from vehicle
```
kjrush, line 465: Might not be a bad idea to parameterize `cross_track_dist` here

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,91 @@ namespace route {
...
+
+         if (boost::geometry::distance(position, current.polygon2d()) > llt_track.downtrack) //Evaluate lanelet downtrack distance
+        {
+            cte_count++;
+
+            if(cte_count > 4)
+                return true;
+        }
+
+  if(boost::geometry::within(position, current.polygon2d()) == false) //Determine whether or not the vehicle is in the lanelet polygon
```
kjrush, line 481: Stylistically, checking for "true" or "false" of a boolean inside an if is kind of redundant, could be simplified to ` if(!boost::geometry::within(position, current.polygon2d()))`

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,91 @@ namespace route {
...
+      {
+        out_of_llt_bounds = true;
+      }
+
+    auto following_llts = world_model_->getMapRoutingGraph()->following(current); //Get all subsequent lanelets from the map
+    lanelet::ConstLanelets filtered_llts;
+    for(auto i:following_llts)
+    {
+       for(auto j: route)
```
kjrush, line 490: If this is using the route variable I was looking at in a prior comment, this is all lanelets on the map to my understanding, so this whole for-loop wouldn't eliminate any lanelets?

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,91 @@ namespace route {
...
+    for(auto i: filtered_llts)
+    {
+        if (boost::geometry::within(position, i.polygon2d()) == false) 
+           out_count++;
+           
+        if (boost::geometry::distance(position, current.polygon2d()) > llt_track.downtrack)
+            out_of_llt_bounds = true;
+    }
+
+    if(out_count == filtered_llts.size()) //If our current position is not within the bounds of any of the following lanelets, then return true
```
kjrush, line 507: I think I understand what's being done here, but it may be algorithmically simpler (and faster) to iterate over the list of lanelets in the route, if we find one that we're in, we stop and return "No CTE violation", else if we hit the end of the list we return a CTE violation. The current way we're stepping through every lanelet every timestep it seems.

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,91 @@ namespace route {
...
+        out_of_llt_bounds = true;
+    
+    return out_of_llt_bounds;
+}
+
+    void RouteGeneratorWorker::set_out_counter(int out_counter)
```
kjrush, line 513: Are either of these setters used? Do we need corresponding getters?

kjrush (changes requested):

route/config/route_config_params.yaml
lewisid: I included the explanation in the route_generator_worker.cpp file where they are implemented, but I'll included here as well.

route/src/route_generator_worker.cpp
lewisid: Yes, this section is meant to get all of the lanelets in the map. These lanelets will later be filtered by id based on whether or not they are part of the route (lines 496 -505).

route/src/route_generator_worker.cpp
lewisid: This loop was initially made under the assumption that there may be lanelets in the map that were not a part of the shortest path route. This should be unnecessary, since the operation is meant to analyze all of the lanelets and not just the shortest path. I'll remove it if that's the case.

route/src/route_generator_worker.cpp
lewisid: I see your point. I'll modify it to match that logic.

route/src/route_generator_worker.cpp
lewisid: These setters are used in route.cpp to initialize the counters (which are declared as ros parameters). I might need to add getters for the test cases.

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,106 @@ namespace route {
...
+    
+    if (boost::geometry::distance(position, current.polygon2d()) > cross_track_dist) //Evaluate lanelet crosstrack distance from vehicle
+        {
+            cte_count++;
+
+            if(cte_count > 4) //If the distance exceeds the crosstrack distance a certain number of times, report that the route has been departed
+                return true;
+        }
+
+         if (boost::geometry::distance(position, current.polygon2d()) > llt_track.downtrack) //Evaluate lanelet downtrack distance
```
MishkaMN, line 480: Let me follow up on this through slack

MishkaMN (changes requested):

route/test/test_route_generator.cpp
```diff
@@ -47,7 +47,7 @@ Using this file:
 */
 
 
-TEST(RouteGeneratorTest, testRouteVisualizerCenterLineParser)
+TEST(RouteGeneratorTest, DISABLED_testRouteVisualizerCenterLineParser)
```
kjrush, line 50: Is this `DISABLED` prefix doing something, or just a note for us? If the tests aren't needed any more, I'd say we should delete or replace them, if they are needed but are disabled because they're broken, we need to fix them.

lewisid: That prefix is in reference to [this issue](https://github.com/usdot-fhwa-stol/carma-platform/issues/1013) that I noticed when I started unit testing. I'm not sure if it has been addressed yet.

route/src/route_generator_worker.cpp
kjrush: Has this been resolved? Just seems like a weird construct to me, setting a value this way. If this has been confirmed to work, I'd be curious to read the documentation for this method as I've just not seen this syntax before.

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,98 @@ namespace route {
...
+                return true;
+        }
+
+  ROS_DEBUG_STREAM("LLt Polygon Dimensions1: " << current.polygon2d().front().x()<< ", "<< current.polygon2d().front().y());
+  ROS_DEBUG_STREAM("LLt Polygon Dimensions2: " << current.polygon2d().back().x()<< ", "<< current.polygon2d().back().y());
+
+    if(!boost::geometry::within(position, current.polygon2d())) //Determine whether or not the vehicle is in the lanelet polygon
+    {
+        out_of_llt_bounds = true;
+    }
```
MishkaMN, line 484: I don't think we need this logic. We want to determine it is "out of bounds" (more than acceptable error) only after checking 2 logics. 1 being the crosstrack, 2nd being the downtrack, each has counters. But if we set it true here, and return it on line 511, it is invalidating the counter as it will return true at the very first counter. Let me know if I missing anything though.

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,98 @@ namespace route {
         return true;
     }
 
+bool RouteGeneratorWorker::crosstrack_error_check(const geometry_msgs::PoseStampedConstPtr& msg, lanelet::ConstLanelet current, carma_wm::TrackPos llt_track)
+{
```
MishkaMN, line 442: At the very beginning, if the point is within the lanelet, we will be optimizing it as we don't need to check run rest of the function. So including `boost::geometry::within(position, current.polygon2d())` if logic here might be a good idea

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,98 @@ namespace route {
...
+    ROS_DEBUG_STREAM("Distance1: "<< boost::geometry::distance(position, current.polygon2d())<<" Crosstrack: "<< cross_track_dist );
+    ROS_DEBUG_STREAM("Distance2: "<< boost::geometry::distance(position, current.polygon2d())<<" Downtrack: "<< llt_track.downtrack);
+
+    
+    if (boost::geometry::distance(position, current.polygon2d()) > cross_track_dist) //Evaluate lanelet crosstrack distance from vehicle
+        {
+            cte_count_++;
+
+            if(cte_count_ > 4) //If the distance exceeds the crosstrack distance a certain number of times, report that the route has been departed
+                return true;
```
MishkaMN, line 467: 1. We should have a counter that counts up - 1 variable. And another variable cte_count_ (consider renaming to cte_max_count_) which is the maximum allowed count to be reached. So we are actually setting that 4 with cte_count_. This applies to the other counter as well
2.  Also just realized we should only return true only if both counters are above 4. Because what if this "crosstrack" logic returns true, but the "downtrack" logic would have determined the car is still within bounds, and we returned true too early.

MishkaMN (changes requested):

route/src/route.cpp
```diff
@@ -41,8 +41,12 @@ namespace route {
         rg_worker_.setWorldModelPtr(wm_);
         // load params and pass to route generator worker
         double ct_error, dt_range;
+        int cte_count, out_count;
         pnh_->getParam("max_crosstrack_error", ct_error);
         pnh_->getParam("destination_downtrack_range", dt_range);
+        pnh_->getParam("cte_count", cte_count);
+        pnh_->getParam("out_count", out_count);
```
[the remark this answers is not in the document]
MishkaMN: @lewisid missing here still

route/src/route_generator_worker.cpp
MishkaMN: @lewisid pls add the protection here.

route/src/route_generator_worker.cpp
MishkaMN: @lewisid here

route/src/route_generator_worker.cpp
MishkaMN: @lewisid here

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,98 @@ namespace route {
...
+           {
+               return false; // iterate over the list of lanelets in the route. If the vehicle is inside of one, return that there is "No CTE violation"
+        
+           }
+        else if (!boost::geometry::within(position, i.polygon2d()) && boost::geometry::distance(position, current.polygon2d()) > cross_track_dist)
+        {
+            cte_count_++;
+
+            if(cte_count_ > 4)
+                out_of_following_llts = true;
```
MishkaMN: @lewisid here.

MishkaMN (changes requested):

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,95 @@ namespace route {
...
+           }
+        else if (!boost::geometry::within(position, i.polygon2d()) && boost::geometry::distance(position, current.polygon2d()) > cross_track_dist)
+        {
+            cte_count_++;
+
+            if(cte_count_ > cte_count_max_)
+                out_of_following_llts = true;
+        }
+        
+    }
```
MishkaMN, line 499: I don't think we need this logic anymore, now that we are not using the downtrack

MishkaMN (changes requested):

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,90 @@ namespace route {
...
+
+    ROS_DEBUG_STREAM("LLt Polygon Dimensions1: " << current.polygon2d().front().x()<< ", "<< current.polygon2d().front().y());
+    ROS_DEBUG_STREAM("LLt Polygon Dimensions2: " << current.polygon2d().back().x()<< ", "<< current.polygon2d().back().y());
+
+    if(!boost::geometry::within(position, current.polygon2d())) //Determine whether or not the vehicle is in the lanelet polygon
+    {
+        cte_count_++;
+        if(cte_count_ > cte_count_max_)
+            return true;
+    }
```
MishkaMN, line 465: We don't need this logic here since boost::geometry::distance(position, current.polygon2d()) logic later implicitly handles it

MishkaMN (changes requested):

route/include/route_generator_worker.h
```diff
@@ -204,6 +242,15 @@ namespace route {
 
         // private helper function to add a new route event into event queue
         void publish_route_event(uint8_t event_type);        
+
+        double cross_track_dist;
+
+        // counter to record how many times vehicle's position exceeds crosstrack distance
+        int cte_count_;
```
MishkaMN, line 249: intialize it like  `int cte_count_ = 0;`

route/include/route_generator_worker.h
```diff
@@ -153,6 +159,38 @@ namespace route {
...
+        *  \param llt_track The crosstrack and downtrack distance of the current lanelet
+        * */
+        bool crosstrack_error_check(const geometry_msgs::PoseStampedConstPtr& msg, lanelet::ConstLanelet current_llt);
+
+        /**
+         * \brief set the crosstrack error counter
+         * 
+         *  \param cte_counter initial value of the counter (set to zero as specified in the route_config_params.yaml file)
+        */
+        void set_CTE_counter(int cte_counter);
```
MishkaMN, line 177: Let's remove this as it is better to initialize it to 0 in the headerfile. Otherwise, it might be unnecessarily confusing.

route/src/route.cpp
```diff
@@ -41,9 +41,14 @@ namespace route {
         rg_worker_.setWorldModelPtr(wm_);
         // load params and pass to route generator worker
         double ct_error, dt_range;
+        int cte_count_max;
         pnh_->getParam("max_crosstrack_error", ct_error);
         pnh_->getParam("destination_downtrack_range", dt_range);
+        pnh_->getParam("cte_count_max", cte_count_max);
+        rg_worker_.set_CTE_counter(0);
```
MishkaMN, line 48: let's remove this as mentioned above.

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,51 @@ namespace route {
...
+    {
+  
+       lanelet::BasicPoint2d position;
+
+        position.x()= msg->pose.position.x;
+        position.y()= msg->pose.position.y;
+
+        if(boost::geometry::within(position, current.polygon2d())) //If vehicle is inside current_lanelet, there is no crosstrack error
+        {
+            return false;
```
MishkaMN, line 451: now before this, we could reset our counter to 0. `cte_count_ = 0` because we could be carrying counts generated from noise in the lanelets long ago.

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +438,51 @@ namespace route {
...
+        ROS_DEBUG_STREAM("LLt Polygon Dimensions2: " << current.polygon2d().back().x()<< ", "<< current.polygon2d().back().y());
+        ROS_DEBUG_STREAM("Distance1: "<< boost::geometry::distance(position, current.polygon2d())<<" Crosstrack: "<< cross_track_dist );
+    
+        if (boost::geometry::distance(position, current.polygon2d()) > cross_track_dist) //Evaluate lanelet crosstrack distance from vehicle
+            {
+                cte_count_++;
+
+                if(cte_count_ > cte_count_max_) //If the distance exceeds the crosstrack distance a certain number of times, report that the route has been departed
+                    return true;
+            }
```
MishkaMN, line 464: let's add `else logic` here that resets the count with a similar reasoning in the comment above.
`else
{
cte_count_ = 0;
}
`

MishkaMN (changes requested):

route/src/route_generator_worker.cpp
```diff
@@ -460,16 +470,48 @@ namespace route {
...
+        }
+
+        for (auto i: filtered)
+        {
+            if(i.first < min)
+                {
+                    min = i.first;
+                    min_llt = i.second;
+                }
+        }
```
MishkaMN, line 512: if filtered is empty, we are returning empty lanelet. So if filtered is empty, we should loop through route_llts and record each distance to the point. And return the one with minimum distance. I think current_position (name not verbatim) should be a member variable.

route/src/route_generator_worker.cpp
```diff
@@ -460,16 +470,48 @@ namespace route {
...
+        std::vector<std::pair<double, lanelet::ConstLanelet::ConstType>>  filtered;
+        double min = 1000.0;
+        lanelet::ConstLanelet min_llt;
+        for(auto i : list_of_pair)
+        {
+            for(auto j: route_llts)
+            {
+                if(i.second.id() == j.id())
+                    filtered.push_back(i);
+            }
```
MishkaMN, line 501: if we have route_llts, as a set, we would have O(N) operation, while this is O(N^2). We can save route_llt_ids as a set if it is hard to search through lanelets. And later we can query the lanelet itself back from laneletLayer with the ids.

route/src/route_generator_worker.cpp
```diff
@@ -340,8 +349,9 @@ namespace route {
                 ROS_WARN_STREAM("Routing has finished but carma_wm has not receive it!");
                 return;
             }
-            auto via_lanelet_vector = lanelet::geometry::findNearest(world_model_->getMap()->laneletLayer, current_loc, 1);
-            auto current_lanelet = lanelet::ConstLanelet(via_lanelet_vector[0].second.constData());
+            auto via_lanelet_vector = lanelet::geometry::findNearest(world_model_->getMap()->laneletLayer, current_loc, 10);
+            //auto current_lanelet = lanelet::ConstLanelet(via_lanelet_vector[0].second.constData());
```
MishkaMN, line 353: remove commented out code

route/include/route_generator_worker.h
```diff
@@ -190,6 +183,13 @@ namespace route {
         */
         void set_CTE_dist(double cte_dist);
 
+        /**
+         * \brief Get the closest lanelet from the list of llts
```
MishkaMN, line 187: "Get the closest lanelet on the route from the list of <distance, lanelet> pairs relative to the vehicle's current point. If the input list does not contain lanelets on the route, still closest lanelet from the route will be returned"

MishkaMN (changes requested):

route/config/route_config_params.yaml
```diff
@@ -4,4 +4,7 @@ max_crosstrack_error: 25
```
MishkaMN, line 4: please see line 3. This probably now has to change now that it is not measuring form the center of the road but from bounding lines. Please confirm with Mike or Kyle

MishkaMN (changes requested):

route/config/route_config_params.yaml
```diff
@@ -4,4 +4,7 @@ max_crosstrack_error: 25
 
 # Double: downtrack_range
 # Units: meter
-destination_downtrack_range: 0
\ No newline at end of file
+destination_downtrack_range: 0
+
+# Int: maximum cross track error counter
```
kjrush, line 9: I think something like "Maximum number of consecutive timesteps outside of the route allowable before triggering LEFT_ROUTE" or something similar would probably be preferable for this still. As is it leaves room for interpreting what a crosstrack error is and how/when they are counted.

route/config/route_config_params.yaml
kjrush: Agreed, this comment should be updated to reflect that, and the value should probably be trimmed way down to something reasonable like 2-4m, I'd think.

route/src/route_generator_worker.cpp
```diff
@@ -438,5 +448,83 @@ namespace route {
...
+            }
+ 
+    }
+
+
+
+    lanelet::ConstLanelet RouteGeneratorWorker::get_closest_lanelet_from_vector_llts (std::vector<std::pair<double, lanelet::ConstLanelet::ConstType>> list_of_pair)
+    {
+        std::vector<std::pair<double, lanelet::ConstLanelet::ConstType>>  filtered;
+        double min = 1000.0;
```
kjrush, line 493: I'd recommend using `std::limits<double>::infinity()` if this is supposed to represent an arbitrarily large value.

route/src/route_generator_worker.cpp
kjrush: Are you talking about a set in terms of some specific detail of the CPP version of a set? Is it backed by a hash-table of some kind? AFAIK a set on it's own would still require iteration in this internal loop to check all elements of the set to see if the ID in question belongs to the set.

Alternatively, is it plausible to restructure this computation in such a way that we only ever get the distance to route lanelets anyway? As I understand it, currently we're using a dijkstra-style method that gets the distance from a single point to all lanelets in our map-graph. Then we check all of those (distance, lanelet) pairs against a list of route_lanelets to determine which are relevant, then find the min among those. Is it possible/more efficient to just do a:

```
min_dist = inf
min_lanelet = null
FOR lanelet IN route:
  dist = get_dist(point, lanelet)
  if (dist < min_dist):
    min_lanelet = lanelet
    min_dist = dist
return min_lanelet
```

style solution? Would be O(N) in terms of our top-level lanelet search. If the underlying distance calculation is still djikstra's then I'm not sure we realize a net gain though, in fact it's probably worse if we run that calculation multiple times.

route/src/route_generator_worker.cpp
kjrush: Filtered would only be empty if route_llt is empty though, right? Or maybe I'm missing a detail that might mean we're not getting all the lanelets?

kjrush (changes requested):

route/src/route_generator_worker.cpp
kjrush: After looking at the documentation, I follow now on the set bit. The common CPP implementation of set lookup is O(log N) though, so overall it'd be O(N log N), I think we could use a hash-set or similar though to get the inner lookup to O(1) though. Probably overkill, I'm not sure how big this data set ends up being at run-time anyway to see these asymptotic performance differences matter.

route/src/route_generator_worker.cpp
MishkaMN: Filtered might be empty if the closest lanelet pairs to the point in the input, for some reason, does not contain any route lanelets. Chances are low.

route/src/route_generator_worker.cpp
MishkaMN: That makes a lot of sense. I was fixated on finding closest lanelets first and then filtering route from there. We could have just iterated through route lanelets and got the closest lanelet as current_lanelet there.

MishkaMN (approved):

kjrush (approved):

msmcconnell (approved):

kjrush:
Hmm, the line coverage is a point of concern here, I think. Looking at the report on Sonar, it's more than just the interfacing code that's uncovered too, there's good chunks of logic in the `route_generator_worker.cpp`. Can we add some more tests, particularly for `set_active_route_cb()`. It looks like there's a lot of logic in there that's not covered at the moment. It might make sense to refactor that method into several smaller ones, and then individually test those, especially if that would help decouple that logic from the ROS communications aspects.

PR #1008, opened 2020-12-10
Feature/pure pursuit next point

MishkaMN:
# PR Details
## Description

**carma-platform/pure_pursuit_wrapper:** 
 - (Same as Mike's branch) Uses the same set of functions and equations to derive the speeds back from traj points and time. So there should be no accuracy problem there.
 - (Same as Mike's branch) It was dropping the first point before, which we don't now.
 - Fixed plan_delegator where it was not setting the initial_velocity when it got the service callback from inlanecruising, which was not making the car move.

**autoware.ai/pure_pursuit:**
- getNextWaypointNumber: should give us the next waypoint that's in front of us:
   - it gets a vector from closest next potential waypoint to its next point in the waypoints.
   - it also gets a vector from current_pose to the next potential waypoint's position and get a vector from it. 
   - it compares the two vectors, and if the angle between them is within 90degrees, it picks that waypoint, otherwise it continues to check the following waypoints which increase in distance.
- everytime new waypoints are set, the pure_pursuit uses the above function to get the next-waypoint (which was hardcoded as first point before) and uses that point's speed.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1015

## Motivation and Context
See above issue.

## How Has This Been Tested?
unit tested and integration tested on black pacifica.

## Types of changes

## Checklist:


msmcconnell (approved):
LGTM

Issue #731, opened 2020-06-22
Tactical plugin for platooning


itabrl:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [`x`] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
Platooning Tactical plugin based on detail design to work as a service and provide trajectory for strategic plugin.

### Related work

Link to related tickets or prior related work here.

msmcconnell:
Resolved by above PRs

PR #1014, opened 2020-12-17
update platooning tactical plugin

saina-ramyar:
# PR Details
## Description
The platooning tactical plugin is updated to be compatible with the new routing mechanism of CARMA that uses vector map instead of recorded waypoints.

## Related Issue
#731 

## Types of changes

## Checklist:


platooning_tactical_plugin/include/platooning_tactical_plugin/platooning_tactical_plugin.h
```diff
@@ -1,104 +1,239 @@
...
+#include "third_party_library/spline.h"
+
+namespace platooning_tactical_plugin
+{
+using PublishPluginDiscoveryCB = std::function<void(const cav_msgs::Plugin&)>;
+
+/**
+ * \brief Convenience class for pairing 2d points with speeds
+ */ 
+struct PointSpeedPair
```
kjrush, line 42: Doesn't need to change, but wanted to comment that I think Mike has been using a cool technique for this kind of convenience struct: `using std::pair<TYPE_A, TYPE_B> as ...;`.

platooning_tactical_plugin/launch/platooning_tactical_plugin.launch
```diff
@@ -1,6 +1,5 @@
-<?xml version="1.0"?>
```
kjrush, line 1: I think we need this, no?

platooning_tactical_plugin/package.xml
```diff
@@ -1,79 +1,39 @@
-<?xml version="1.0"?>
```
kjrush, line 1: Same here

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
```diff
@@ -0,0 +1,617 @@
+#include "platooning_tactical_plugin/platooning_tactical_plugin.h"
```
kjrush, line 1: Copyright notice

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
```diff
@@ -0,0 +1,617 @@
...
+    for (auto p : route_geometry)
+    {
+      if (first && points_and_target_speeds.size() != 0)
+      {
+        first = false;
+        continue;  // Skip the first point if we have already added points from a previous maneuver to avoid duplicates
+      }
+      PointSpeedPair pair;
+      pair.point = p;
+      pair.speed = lane_following_maneuver.end_speed;
```
kjrush, line 360: This will be the baseline speed the control plugin modifies to maintain the gap, right? Just making sure I understand the data flow.

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
```diff
@@ -0,0 +1,617 @@
...
+  cav_msgs::TrajectoryPlan trajectory;
+  trajectory.header.frame_id = "map";
+  trajectory.header.stamp = ros::Time::now();
+  trajectory.trajectory_id = boost::uuids::to_string(boost::uuids::random_generator()());
+
+  trajectory.trajectory_points = compose_trajectory_from_centerline(downsampled_points, req.vehicle_state); // Compute the trajectory
+  trajectory.initial_longitudinal_velocity = std::max(req.vehicle_state.longitudinal_vel, config_.minimum_speed);
+
+  resp.trajectory_plan = trajectory;
+  resp.related_maneuvers.push_back(cav_msgs::Maneuver::LANE_FOLLOWING);
```
kjrush, line 75:
```
# A set of array indexes which map to the maneuvers in the provided maneuver_plan
# These indicate which maneuvers the returned trajectory is addressing
uint8[] related_maneuvers
```


Per PlanTrajectory.srv this should be the index (in the maneuver plan) of the maneuver this trajectory corresponds to, most likely 0 in this case.

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
```diff
@@ -0,0 +1,617 @@
...
+    ROS_DEBUG("SubCurve");
+
+    std::vector<double> speed_limits;
+    std::vector<lanelet::BasicPoint2d> curve_points;
+    splitPointSpeedPairs(discreet_curve.points, &curve_points, &speed_limits);
+
+    std::unique_ptr<smoothing::SplineI> fit_curve = compute_fit(curve_points); // Compute splines based on curve points
+
+    if (!fit_curve)
+    {  // TODO how better to handle this case
```
kjrush, line 465: Is this todo from the original ILC-plugin, or an additional thing we should be tracking with respect to platooning plugin?

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
```diff
@@ -0,0 +1,617 @@
...
+  // Apply accel limits
+  final_actual_speeds = trajectory_utils::apply_accel_limits_by_distance(downtracks, final_actual_speeds,
+                                                                         config_.max_accel, config_.max_accel);
+  log::printDoublesPerLineWithPrefix("postAccel[i]: ", final_actual_speeds);
+
+  
+  final_actual_speeds = smoothing::moving_average_filter(final_actual_speeds, config_.moving_average_window_size);
+  log::printDoublesPerLineWithPrefix("post_average[i]: ", final_actual_speeds);
+
+  for (auto& s : final_actual_speeds)  // Limit minimum speed. TODO how to handle stopping?
```
kjrush, line 594: Hmm, handling a stop normally would be covered by the Stop&Wait plugin, but in the case of platooning especially that may not be desired. Good catch, something for us to think about!

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
```diff
@@ -0,0 +1,617 @@
...
+  std::vector<double> final_actual_speeds;
+  std::vector<lanelet::BasicPoint2d> all_sampling_points;
+
+  for (const auto& discreet_curve : sub_curves)
+  {
+    ROS_DEBUG("SubCurve");
+
+    std::vector<double> speed_limits;
+    std::vector<lanelet::BasicPoint2d> curve_points;
+    splitPointSpeedPairs(discreet_curve.points, &curve_points, &speed_limits);
```
kjrush, line 460: So if I follow correctly, this is how the platoon speed is being applied? As a speed limit on the generated curvature-based speeds? That makes some sense. I think we might also want to still apply the regulatory speed limit if possible too though.

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
```diff
@@ -0,0 +1,617 @@
...
+
+  ROS_DEBUG_STREAM("PlanTrajectory");
+
+  cav_msgs::TrajectoryPlan trajectory;
+  trajectory.header.frame_id = "map";
+  trajectory.header.stamp = ros::Time::now();
+  trajectory.trajectory_id = boost::uuids::to_string(boost::uuids::random_generator()());
+
+  trajectory.trajectory_points = compose_trajectory_from_centerline(downsampled_points, req.vehicle_state); // Compute the trajectory
+  trajectory.initial_longitudinal_velocity = std::max(req.vehicle_state.longitudinal_vel, config_.minimum_speed);
```
kjrush, line 72: I believe this is also the actual vehicle speed at the start time of the trajectory, so likely just current vehicle speed if that data is available to this plugin.

kjrush (changes requested):
Overall looks good to me!

I think this does start to point towards us needing/wanting to refactor some of that basic waypoint generation logic into a carma_utils package in the very near future, so we can have one home for it and a centralized location to tweak/improve it. 

I think the only real thing I'd like to see here is the inclusion of the regulatory speed limit check as well as some minor quibbles regarding the data going into some of the messages.

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
kjrush: Should probably write up an issue to document that this edge case isn't covered.

platooning_tactical_plugin/include/platooning_tactical_plugin/log/log.h
```diff
@@ -0,0 +1,61 @@
+#include <sstream>
```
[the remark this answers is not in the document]
saina-ramyar: I agree. It would be helpful to move most of the logic to a utils location, as they are common between most of the tactical plugins. However, I suggest we do this switch once the enhancements on ILC is complete. As for the platooning tactical, since the ILC logic is functional, I used that in this plugin so we can better troubleshoot the strategic and control platooning plugins.

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
saina-ramyar: I know it says index, but I think it is actually the maneuver type, from Maneuver.msg

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
saina-ramyar: Yes, that is correct.

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
saina-ramyar: it is from ILC. I dont think that would be an issue in platooning, since platooning does not work well on highly curved roads, and probably not activated on that type of environment.

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
saina-ramyar: No, that is the actual regulatory speed limit, applied on top of the platoon speed communicated through the maneuver.

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
saina-ramyar: fixed

platooning_tactical_plugin/package.xml
saina-ramyar: fixed

platooning_tactical_plugin/launch/platooning_tactical_plugin.launch
saina-ramyar: fixed

platooning_tactical_plugin/config/parameters.yaml
```diff
@@ -0,0 +1,12 @@
+# The length of the trajectory in time domain, in seconds
```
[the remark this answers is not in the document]
saina-ramyar: addressed

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
saina-ramyar: added

platooning_tactical_plugin/src/platooning_tactical_plugin.cpp
kjrush: I think if it was maneuver type it'd be a single value instead of an array. This field isn't used at the moment, but I believe the intent was such that the plan delegator could know which maneuvers were now accounted for as a result of this trajectory plan.

kjrush:
Aside from us figuring out that one `related_maneuvers` field, and the CircleCI issue (looks like it's not seeing the test coverage again) I think this looks good to me

msmcconnell:
So we just removed the GPL licensed library in Misheel's PR which switched to a B-Spline Eigen library. If this code is merged it will reintroduce the GPL licensed code back into develop. Just wanted to point that out. @kjrush I'll let you decide if you want to address that now or later. I don't think we can do the release though until it is addressed.

kjrush:
Looks good to me overall. On the spline library point, let's fix that if possible @saina-ramyar. That code should be in develop now, and since we used an abstraction layer on top of the spline library itself it shouldn't be a big change I'd think to use the new implementation. There was one change to a method API I believe, it's worth giving it a look and we can re-evaluate if it's a big lift.

kjrush (approved):

PR #1022, opened 2020-12-24
Feature/add stopandwait plugin

adev4a:
# PR Details
Add Tactical Stop And Wait Maneuver Plugin
## Description
This PR adds the tactical plugin for a stop and wait maneuver.

A new  stop_and_wait_plugin package is added that handles maneuver plans of type Stop and Wait.
## How Has This Been Tested?
Unit Tested and Integration Tested

## Types of changes

## Checklist:


stop_and_wait_plugin/include/stop_and_wait_plugin.h
```diff
@@ -0,0 +1,181 @@
...
+        //Total time required to complete the maneuver
+        double maneuver_time_;
+
+        //Parameters loaded from config file initialized for unit tests
+        //The minimum duration of a trajectory length in seconds
+        double minimal_trajectory_duration_ = 6.0;
+        //The maximum acceptable jerk 
+        double max_jerk_limit_ = 3.0;
+        //The minimum acceptable jerk, after which constant speed is assumed
+        double min_jerk_ = 0.01;
```
adev4a, line 158: min_jerk_ used while testing is 0.001

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -0,0 +1,386 @@
...
+            traj_point.controller_plugin_name = "default";
+            traj_point.planner_plugin_name =plugin_discovery_msg_.name;
+            traj.push_back(traj_point);
+            traj_prev = traj_point;
+        }
+        
+        return traj;
+    }
+
+    void StopandWait::speed_to_time(const std::vector<double>& downtrack, const std::vector<double>& speeds, std::vector<double>* times, double jerk)
```
kjrush, line 311: Looking at this now, I'd just recommend making times a reference to avoid using raw pointers. Should accomplish the same results. And I'd make sure the doc comments mention that times will be modified as a result. Either that, or if it's imperative that times be an empty vector prior to this function, i'd say just make it the return value of this function instead of an input.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -0,0 +1,386 @@
...
+        }
+        if (downtrack.size() == 0)
+        {
+            throw std::invalid_argument("Input vectors are empty");
+        }
+
+        times->reserve(downtrack.size());  
+
+        //Uses equation 
+        //d_t = sqrt(2(d*v)/j)
```
kjrush, line 325: is the `d` term in `d*v` supposed to be displacement? If so, the implementation below lacks this term.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -0,0 +1,386 @@
...
+        double prev_speed = speeds[0];
+        double prev_time = 0.0;
+        double prev_pos = downtrack[0];
+        times->push_back(prev_time);
+        for(int i=1; i <downtrack.size();i++)
+        {
+            double cur_speed = speeds[i];
+            double delta_v = std::abs(cur_speed - prev_speed);
+            double dt = pow((2*delta_v/jerk),0.5);
```
kjrush, line 334: Doesn't matter a huge deal, but if I recall correctly the `sqrt()` function is substantially faster (and guaranteed to round correctly I believe) than `pow(..., 0.5)` for what it's worth.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -0,0 +1,386 @@
...
+            if(jerk < min_jerk_)    //If jerk is almost 0, treat as constant velocity
```
kjrush, line 335: If jerk is ~0, wouldn't that be constant acceleration not necessarily constant velocity?

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -0,0 +1,386 @@
...
+            }
+            else
+            {
+                double jerk_req = (2*current_speed_)/pow(maneuver_time_,2);
+                double start_speed = current_speed_;    //Get static value of current speed at start of planning
+                if(jerk_req > max_jerk_limit_)
+                {
+                    //unsafe to stop at the required jerk - reset to max_jerk and go beyond the maneuver end_dist
+                    jerk_ = max_jerk_limit_;  
+                    maneuver_time_ = pow(2*start_speed, 0.5);
```
kjrush: See other comment about sqrt()

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -0,0 +1,386 @@
...
+                    double travel_dist_new = start_speed * maneuver_time_ - (0.167 * jerk_ * pow(maneuver_time_,3));
+                    ending_downtrack = travel_dist_new + starting_downtrack;
+                }
+                else jerk_ = jerk_req;
+
+                //get all the lanelets in between starting and ending downtrack on shortest path
+                auto lanelets = wm_->getLaneletsBetween(starting_downtrack, ending_downtrack, true);
```
kjrush, line 198: if ending downtrack has been extended beyond the value from the maneuver (because to stop at the originally planned ending downtrack would violate the jerk limit), and that stop is at the end of the route how does this behave? Downtrack is only defined for lanelets on the route. Something to ponder a bit as an edge case.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -0,0 +1,386 @@
...
+                    {
+                        first = false;
+
+                        continue; //Skip the first point to avoid duplicates from previous maneuver
+                    }
+                    PointSpeedPair pair;
+                    pair.point = p;
+                    pair.speed = start_speed - (0.5 * jerk_ * pow(curr_time,2));
+
+                    if(p == route_geometry.back()) pair.speed = 0.0;    //force speed to 0 at last point
```
kjrush, line 230: Can't say I'm generally a fan of omitting {} for if statements like this, it makes it substantially less clear if line 232 is supposed to be part of that if-clause, or not. Doesn't appear to be a defect here, but implicit curly braces have always been a personal pet-peeve of mine ever since the large Apple SSL bug from 2014.

kjrush (changes requested):
Generally seems good to me, I have a couple comments I've noted about either curiosities, edge cases to check, or things I'd like to see adjusted. It's a bit challenging for me to precisely follow what's going on with the speed planning, since it's done largely as a function of downtrack instead of time (which is then derived afterwards) which makes it harder (in my opinion) to think about how speed changes relate to the jerk constraint. 

It might be helpful to take a step back and think about the higher-level sequence of steps in the trajectory callback here and reorganize things a bit more (that might mean refactoring a bit to group related bits into functions) so that the overall steps that are being taken can be understood easier.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
kjrush: Generally, I'd favor return values almost all the time unless it's expected that the object being returned will be very large in size (and even then, I'd likely want to see performance profiling data indicating that we'd actually have a problem before optimizing like that). I just think that's easier to follow most of the time as it better highlights where data transformations are happening and how.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
adev4a: This should be d_v, change in speed

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
adev4a: It would still ignore the ending downtrack from the maneuver, in order to make sure the jerk limit is met.
Since there is no distinction between stop at the end of a route vs elsewhere, this check would have to happen in the strategic plugin

kjrush: I would probably recommend throwing some kind of exception, ROS_ERROR logging it, or something along those lines then, since we won't be able to access the route-lanelets for the part of the stop maneuver which extends beyond the end of the route.

kjrush: Following up on this, did we come to a decision on throwing an exception here or no?

kjrush:
Followup on that comment aside, lets see if we can get Sonar coverage working on this branch as well so we know if we're hitting our target there.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
adev4a: Yes we did, It is throwing an exceptions before this getLaneletsBetween call

kjrush (approved):

kjrush:
👍 , that one security hotspot flagged by Sonar seems like a non-issue but I don't have permission to mark it as reviewed.

PR #1023, opened 2020-12-24
add filter to incoming message plan

adev4a:
# PR Details
This PR fixes Issue 1012.
## Description

A filter was added to the incoming trajectory plan request so inlanecruising only works on LANE FOLLOWING maneuver types.
## Types of changes

## Checklist:


inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -66,8 +66,16 @@ bool InLaneCruisingPlugin::plan_trajectory_cb(cav_srvs::PlanTrajectoryRequest& r
...
+  //only work on lane_following maneuver plans
+  std::vector<cav_msgs::Maneuver> maneuver_plan;
+  for(int i=0;i<req.maneuver_plan.maneuvers.size();i++)
+  {
+    if(req.maneuver_plan.maneuvers[i].type == cav_msgs::Maneuver::LANE_FOLLOWING)
+    {
+      maneuver_plan.push_back(req.maneuver_plan.maneuvers[i]);
+    }
+  }
+  auto points_and_target_speeds = maneuvers_to_points(maneuver_plan, current_downtrack, wm_); // Convert maneuvers to points
```
msmcconnell, line 78: If this received a plan of three maneuvers at the given downtrack distances. [ LANE_FOLLOW (0-10), LANE_CHANGE(10-20), LANE_FOLLOW(20-30) ]

Then it would select [LANE_FOLLOW (0-10), LANE_FOLLOW(20-30)]
This means inlanecruising would then try to connect the two lane following maneuvers right? That seems pretty problematic.

msmcconnell (changes requested):

inlanecruising_plugin/src/inlanecruising_plugin.cpp
msmcconnell: If throwing an exception is the wrong response (as the issue described) then what would be the right response? It seems to me that InlaneCruising should not ever receive these maneuvers in the first place and that throwing an exception would be the correct response. It seems the Strategic plugin which send inlane cruising an unsupported maneuver is at fault not In-LaneCruising.

kjrush (approved):

Issue #936, opened 2020-10-29
Review usage of GPL licensed code in the codebase for spline library


msmcconnell:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [x] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

We would like to use the spline library https://github.com/ttk592/spline
This library is licensed under GPL v2.0. Which in my non-legal opinion is incompatible with Apache 2.0 based off https://www.apache.org/licenses/GPL-compatibility.html

@snallamothu Has said we can continue to prototype with this library while investigation of use and alternatives proceeds.

Some alternative libraries with usages issues include:
https://www.boost.org/doc/libs/1_65_0/libs/math/doc/html/math_toolkit/interpolate/barycentric.html : Requires boost 1.65 which is not available on current ROS Kinetic and Ubuntu 16.04 installations.

https://github.com/msteinbeck/tinyspline : Requires cmake 3.12 we are using cmake 3.5.1 which comes with Ubuntu 16.04 and ROS Kinetic.

https://github.com/ejmahler/SplineLibrary : Building uses qmake so it will take some time to determine how to build. Also the TODO comments in the readme do not make me overly confident in the accuracy of the library. 

https://github.com/mfcats/SplineLib : Requires C++17 while the highest we currently support is C++14. Uses the LGPL license which I think is compatible with our use case so long as we do not modify the library, but further investigation is needed. 

https://github.com/NCAR/bspline : A c++ wrapper for FORTRAN that uses scons to build. Since scons is a tool we are unfamiliar with it will take some learning time. 

https://github.com/roboticslibrary/rl : RoboticsLibrary seems like is might work, but it is a very large library so setup will take some time. 

### Carma version where this issue was discovered

integration/routing branch
### Expected behavior
All licenses are meeting legal requirements
### Actual behavior

See above

msmcconnell:
Use of the referenced GPL spline library has been replaced with https://eigen.tuxfamily.org/dox/unsupported/structEigen_1_1SplineFitting.html for vanden-plas release which is the release directly following the opening of this issue.

PR #1026, opened 2021-01-10
Feature/spline

MishkaMN:
# PR Details
## Description

Previously, develop was using cubic spline library that was under incompatible license with Apache2 and it was not capable of fitting every curves like S. This PR replaces that with Eigen::Spline which satisfies above criteria.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/936 License issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1027 Spline issue

## Motivation and Context
See above.

## How Has This Been Tested?
Existing unit tests pass.
Included unit test to cover straight line which is an edge case.
Included a unit test (normally disabled) that uses whole TFHRC map to generate a route centerline and recreate it using the library. Plotting those two graphs also resulted in same.

## Types of changes

## Checklist:


msmcconnell (changes requested):
Leaving some initial comments to get things rolling on this.

inlanecruising_plugin/include/inlanecruising_plugin/smoothing/SplineI.h
```diff
@@ -41,15 +41,15 @@ class SplineI
...
+   * \brief Get the Eigen::VectorXf for the given t parameter
    * 
-   * \param x The value to solve the spline for
+   * \param t The value to solve the spline for
    * 
-   * \return The y value that matches x
+   * \return The Eigen::VectorXf vector with x, y that matches t parameter
    */ 
-  virtual double operator()(double x) const = 0;
+  virtual Eigen::VectorXf operator()(double t) const = 0;
```
kjrush, line 52: If the intent is to use this class as an abstraction layer over top of the spline implementation, does it make sense to use an Eigen-specific type here? I'm kind of split personally, Eigen is the underlying implementation we're abstracting over, but also just a support library we'd be using regardless.

inlanecruising_plugin/include/inlanecruising_plugin/smoothing/BSpline.h
```diff
@@ -18,25 +18,27 @@
 
 #include <vector>
 #include <carma_wm/Geometry.h>
-#include <inlanecruising_plugin/third_party_library/spline.h>
 #include <inlanecruising_plugin/smoothing/SplineI.h>
+#include <unsupported/Eigen/Splines>
```
kjrush, line 22: I'm curious to know more what "unsupported" here means, do we have a feel for what the implications of using this are? Does it run the risk of being removed or broken in later versions?

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -228,108 +229,99 @@ std::vector<cav_msgs::TrajectoryPlanPoint> InLaneCruisingPlugin::compose_traject
...
+  splitPointSpeedPairs(time_bound_points, &curve_points, &speed_limits);
+  
+  std::unique_ptr<smoothing::SplineI> fit_curve = compute_fit(curve_points); // Compute splines based on curve points
+  if (!fit_curve)
+  {  // TODO how better to handle this case
+    ROS_ERROR_STREAM("Could not fit!");
+    for (size_t i = 0; i < curve_points.size() - 1; i++)
+    {
+      all_sampling_points.push_back(curve_points[i]);
+      final_yaw_values.push_back(final_yaw_values.back());
```
kjrush, line 248: I think this yields undefined behavior per the C++ standard if the vector is empty, you're effectively referencing an uninitialized element.

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -228,108 +229,99 @@ std::vector<cav_msgs::TrajectoryPlanPoint> InLaneCruisingPlugin::compose_traject
...
+      final_actual_speeds.push_back(final_actual_speeds.back());
     }
```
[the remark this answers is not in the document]
kjrush: Hmm, I'm kind of split. If it'd be possible to proceed (but with lower accuracy) without the spline fit succeeding in hopes that maybe next timestep (for whatever reason) the fit succeeds, then I think we'd want to continue rather than crashing with an exception. Is that plausible?

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -228,108 +229,99 @@ std::vector<cav_msgs::TrajectoryPlanPoint> InLaneCruisingPlugin::compose_traject
...
-      for (size_t i = current_speed_index; i < curve_points.size(); i++)
-      {
-        if (curve_points[i].x() >= current_dist)
-        {
-          current_speed_index = i;
-          break;
-        }
-      }
+  double next_speed_parameter_threshold = parameter_length / total_point_size;
+  double scaled_parameter = 0.0;
```
kjrush, line 277: I think I understand what this variable is, but "parameter" is a bit vague. Not really sure what a better name would be though, but maybe a comment can help make it more obvious if we can't think of a more direct name?

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -228,108 +229,99 @@ std::vector<cav_msgs::TrajectoryPlanPoint> InLaneCruisingPlugin::compose_traject
...
-
-    curvatures = smoothing::moving_average_filter(curvatures, config_.moving_average_window_size);
+  log::printDebugPerLine(sampling_points, &log::basicPointToStream);
 
-    log::printDoublesPerLineWithPrefix("curvatures[i]: ", curvatures);
+  std::vector<double> yaw_values = carma_wm::geometry::compute_tangent_orientations(sampling_points);
 
-    std::vector<double> ideal_speeds =
-        trajectory_utils::constrained_speeds_for_curvatures(curvatures, config_.lateral_accel_limit);
+  std::vector<double> curvatures = carma_wm::geometry::local_circular_arc_curvatures(
```
kjrush, line 298: This is the method that the other branch is pending changes on, just to make sure I follow?

kjrush (changes requested):

inlanecruising_plugin/src/inlanecruising_plugin.cpp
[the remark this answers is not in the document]
msmcconnell: I don't like the idea of this failing silently. What happens if the next timestep it can't recover? How long should we allow it to operate like that? An automated vehicle that has had a core failure in its algorithm should return control to the user or hand control over to some other more robust safe stop component. We could allow some number of failed steps but I don't think it could be more than a handful and it would need to be tracking that in the code.

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -228,108 +229,99 @@ std::vector<cav_msgs::TrajectoryPlanPoint> InLaneCruisingPlugin::compose_traject
...
-    distributed_speed_limits.reserve(1 + discreet_curve.points.size() * 2);
+  std::vector<double> distributed_speed_limits;
+  distributed_speed_limits.reserve(1 + curve_points.size() * 2);
 
-    double max_x = curve_points.back().x();
-    double current_dist = 0;
-    double step_size = config_.curve_resample_step_size;
-    int current_speed_index = 0;
+  // compute subcurves to get correct time steps
+  std::vector<DiscreteCurve> sub_curves = compute_sub_curves(time_bound_points);
```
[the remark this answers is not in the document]
MishkaMN: I removed this as well as curvePointInMapTF and DiscreteCurve struct that is related to it

inlanecruising_plugin/include/inlanecruising_plugin/smoothing/SplineI.h
MishkaMN: I'll make it std::pair

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -313,17 +256,14 @@ std::vector<cav_msgs::TrajectoryPlanPoint> InLaneCruisingPlugin::compose_traject
 
   log::printDoublesPerLineWithPrefix("yaw_values[i]: ", yaw_values);
 
-  for (int i = 0; i < yaw_values.size() - 1; i++)
-  {  // Drop last point
-    final_yaw_values.push_back(yaw_values[i]);
-    all_sampling_points.push_back(sampling_points[i]);
-  }
-
+  // Drop last point
```
msmcconnell, line 259: Is it still necessary to drop the last point when subcurves are no longer a thing? I think this logic was originally meant to support the concatenation of the sub-curves.

msmcconnell (changes requested):

inlanecruising_plugin/src/inlanecruising_plugin.cpp
MishkaMN: ah okay. I don't think it is necessary then

msmcconnell (approved):

kjrush (approved):

PR #1031, opened 2021-01-12
Feature/rosbag exclude

pcguglielmino:
# PR Details
## Description

Added a new carma_record package which allows for carma config parameters to control the recording of certain topics. This also encapsulates the rosbag record functionality and will run it with the desired exclusions.

## Motivation and Context

This allows carma to be run in specific environments and avoid collecting sensitive data

## How Has This Been Tested?

Because this node relies on the ros interface, this has been tested by hand, making sure that it does not record any undesired topics and responds to changes in the rosparam sever

## Types of changes

## Checklist:


.sonarqube/sonar-scanner.properties
```diff
@@ -52,8 +52,9 @@ sonar.modules= bsm_generator, \
   mobilitypath_publisher, \
   platoon_control_plugin, \
   rosbag_mock_drivers, \
-  lightbar_manager,
-  inlanecruising_plugin
+  lightbar_manager, \
+  inlanecruising_plugin,
```
msmcconnell, line 56: add a \ after inlanecruising_plugin,

carma_record/CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
+cmake_minimum_required(VERSION 3.0.2)
+project(carma_record)
+
+## Compile as C++11, supported in ROS Kinetic and newer
+add_compile_options(-std=c++11)
```
msmcconnell, line 5: Make this c++14 since that is what we are moving to. It should cause any issues

carma_record/CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
+
+# Mark cpp header files for installation
+install(DIRECTORY include/${PROJECT_NAME}/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+  FILES_MATCHING PATTERN "*.h"
+  PATTERN ".svn" EXCLUDE
```
msmcconnell, line 80: No need for the FILES_MATCHING or PATTERN lines.

carma_record/launch/carma_record.launch
```diff
@@ -0,0 +1,19 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+  License for the specific language governing permissions and limitations under
+  the License.
+-->
+
+<launch>
+  <!-- The carma_record_node will load in the parameters from the rosparam server, and export a exclude_regex param -->
+  <node name="carma_record_load_regex" pkg="carma_record" type="carma_record_node" output="screen"/>
```
msmcconnell, line 16: don't set output="screen" unless using it for testing. This will prevent logging to a file which is preferable when trying to debug in the future.

carma_record/package.xml
```diff
@@ -0,0 +1,87 @@
...
+
+  <!-- One maintainer tag required, multiple allowed, one person per tag -->
+  <!-- Example:  -->
+  <!-- <maintainer email="jane.doe@example.com">Jane Doe</maintainer> -->
+  <maintainer email="carma@todo.todo">carma</maintainer>
+
+
+  <!-- One license tag required, multiple allowed, one license per tag -->
+  <!-- Commonly used license strings: -->
+  <!--   BSD, MIT, Boost Software License, GPLv2, GPLv3, LGPLv2.1, LGPLv3 -->
```
msmcconnell, line 32: You can remove these extra template comments in this file.

carma_record/package.xml
```diff
@@ -0,0 +1,87 @@
...
+  <exec_depend>roscpp</exec_depend>
+
+  <build_depend>std_msgs</build_depend>
+  <build_export_depend>std_msgs</build_export_depend>
+  <exec_depend>std_msgs</exec_depend>
+
+
+  <build_depend>carma_utils</build_depend>
+  <build_export_depend>carma_utils</build_export_depend>
+  <exec_depend>carma_utils</exec_depend>
```
msmcconnell, line 80: You can reduce all three of these tags to 
```<depend></depend>```

File: carma_record/src/CarmaRecordNode.cpp [hunk 7]
```diff
@@ -0,0 +1,94 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+#include <carma_record/CarmaRecordNode.h>
+
+namespace carma_record
+{
+
+CarmaRecordNode::CarmaRecordNode(){};
```
msmcconnell, line 22: This constructor can be reduced to CarmaRecordNode()=default; in the header file and this line can be removed.

carma_record/scripts/carma_record.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+while true
+    do
+        # Get the regex of the excluded topics
+        excluded_topics="$(rosparam get exclude_regex)"
+        if [ ! -z "${excluded_topics}" ]
+        then 
+            # Run the rosbag record command
+            rosbag record -o /opt/carma/logs/ --lz4 -a -x '/rosout(.*)|(.*)/received_messages|(.*)/sent_messages|(.*)/scan|$($excluded_topics)'
```
msmcconnell, line 10: Since there is a default option, I think it would make more sense to include this exclusions in that default set rather than hard coded.

carma_record/src/CarmaRecordNode.cpp
```diff
@@ -0,0 +1,94 @@
...
+  // exclude_regex is the final list of topics to be excluded
+  std::string exclude_regex = ""; // Add default exclusion to this string
+
+  // if a set of topics needs to be excluded, it will be iterated through and added to the exclude regex
+  if(exclude_default){
+    std::vector<std::string> excluded_default_topics;
+    cnh_.getParam("/excluded_default_topics", excluded_default_topics);
+
+    for (param_itr = excluded_default_topics.begin(); param_itr < excluded_default_topics.end(); param_itr++){
+      exclude_regex += ((*param_itr) + "|");
```
msmcconnell, line 51: Is using the iterator a necessity here or would c++ ranged loops work?

Could this be reduced to?
```
    for (auto exclusion_str : excluded_default_topics){
      exclude_regex += exclusion_str + "|";
      }
```

It would make this look much cleaner if this code be.

carma_record/src/CarmaRecordNode.cpp
```diff
@@ -0,0 +1,94 @@
...
+    for (param_itr = excluded_can_topics.begin(); param_itr < excluded_can_topics.end(); param_itr++){
+      exclude_regex += ((*param_itr) + "|");
+    }
+  }
+
+  // remove the final "|" from the regex
+  exclude_regex.pop_back();
+ 
+  // set the exclude_regex as a param in the param server
+  // cnh_.setParam("exclude_regex", exclude_regex);
```
msmcconnell, line 86: remove unused line

carma_record/src/CarmaRecordNode.cpp
```diff
@@ -0,0 +1,94 @@
...
+  ros::param::set("exclude_regex", exclude_regex);
+
+  // Spin
+  cnh_.spin();
```
msmcconnell, line 90: Seems like we could just exist rather than spin. Since there is no callback functionality in this node, there is no need to eat up CPU cycles.

carma_record/src/main.cpp
```diff
@@ -0,0 +1,29 @@
...
+// Main execution
+int main(int argc, char** argv)
+{
+  // Initialize node
+  ros::init(argc, argv, "carma_record");
+  carma_record::CarmaRecordNode node;
+
+  // Start execution
+  node.run();
+  return 0;
```
msmcconnell, line 28: I think you would want to forward the return value of node.run() here otherwise it is dropped and that method could return void. 
return node.run();

carma_record/scripts/carma_record.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+while true
+    do
+        # Get the regex of the excluded topics
+        excluded_topics="$(rosparam get exclude_regex)"
+        if [ ! -z "${excluded_topics}" ]
```
msmcconnell, line 7: Since this will not record unless at least one exclusion is supplied we need to make sure that is documented somewhere. Maybe where the actual exclusion list parameter is set as well as the top of this file.

msmcconnell (changes requested):

carma_record/src/CarmaRecordNode.cpp
```diff
@@ -47,47 +42,52 @@ int CarmaRecordNode::run()
...
   // remove the final "|" from the regex
   exclude_regex.pop_back();
  
   // set the exclude_regex as a param in the param server
-  // cnh_.setParam("exclude_regex", exclude_regex);
-  ros::param::set("exclude_regex", exclude_regex);
+  cnh_.setParam("exclude_regex", exclude_regex);
+
+  // if no topics are being excluded, set the no_exclusions parameter so the record script will still run
+  if(!(exclude_default && exclude_lidar && exclude_camera && exclude_can)){
```
msmcconnell, line 84: Shouldn't this be
```
!(exclude_default || exclude_lidar || exclude_camera || exclude_can)
```

carma_record/scripts/carma_record.sh
```diff
@@ -4,10 +4,11 @@ while true
     do
         # Get the regex of the excluded topics
         excluded_topics="$(rosparam get exclude_regex)"
```
msmcconnell, line 6: If excluded_topics is unset I would think excluded topics would store something like
```
ERROR: Parameter [/excluded_topics] is not set
```
That will resolve as not empty in the following check. 
Seems like you might want an additional check for "ERROR: Parameter"

msmcconnell (changes requested):

carma_record/scripts/carma_record.sh
pcguglielmino: From my testing if a param is not set then it will store an empty string, so I don't think an error check is necessary

msmcconnell: Seems you are correct. Looks like they must be routing the log to stderr instead of stdout so it doesn't get stored.
image

msmcconnell (approved):

PR #1033, opened 2021-01-13
Feature/route end behavior

adev4a:
# PR Details
This branch adds the end of route behavior.
## Description
At the end of the route, the vehicle should come to a reasonable stop and disengage automation.  
The end of route uses the stop and wait tactical plugin to bring the vehicle to complete stop at the end of the route.

Route following calculated the stopping distance based on the current speed, when the vehicle has only stopping distance left on the known route, the stop and wait maneuver plan is created to bring it to a halt. 
The guidance state is changed to inactive and then drivers ready if system alert is still drivers ready.
## How Has This Been Tested?

Unit tested and integration tested

## Types of changes

## Checklist:


guidance/src/guidance/guidance_state_machine.cpp
```diff
@@ -60,6 +60,8 @@ namespace guidance
                 if(signal == Signal::ACTIVATED)
                 {
                     current_guidance_state_ = State::ACTIVE;
+                }else if(signal == Signal::INITIALIZED){
```
kjrush, line 63: Where is this INITIALIZED coming from? Normally that's from health monitor I believe, is that the same thing that's happening at end of route? If not, maybe it makes more sense to use another signal

route/src/route_generator_worker.cpp
```diff
@@ -377,14 +377,19 @@ namespace route {
                 publish_route_event(cav_msgs::RouteEvent::ROUTE_DEPARTED);
             }
             // check if we reached our destination be remaining down track distance
-            if(current_downtrack_distance_ > world_model_->getRoute()->length2d() - down_track_target_range_)
+            if(current_downtrack_distance_ > world_model_->getRoute()->length2d() - down_track_target_range_ && current_speed_ == 0 || current_downtrack_distance_ > world_model_->getRoute()->length2d())
```
kjrush, line 380: Comparing current_speed as a double to 0 is potentially complicated, floating point precision is a tricky thing. I'd recommend the more robust `current_speed < epsilon` where epsilon is a small (0.001) constant.

kjrush: I'm also not 100% sure this works as intended without additional parenthesis around the first "... & ..." clause. It short circuit the evaluation of the whole expression of either of those terms is not-true, and then it wouldn't even check the or-clause. I'm not 100% sure, but parens would make it explicit.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -162,12 +162,22 @@ namespace stop_and_wait_plugin
...
-            maneuver_time_ = ros::Duration(stop_and_wait_maneuver.end_time - stop_and_wait_maneuver.start_time).toSec();
+            double ending_downtrack = stop_and_wait_maneuver.end_dist; 
+            double start_speed = current_speed_;    //Get static value of current speed at start of planning
+            //maneuver_time_ = ros::Duration(stop_and_wait_maneuver.end_time - stop_and_wait_maneuver.start_time).toSec();
             
+            maneuver_time_ = (3*(ending_downtrack - starting_downtrack))/(2*start_speed);
+
             double delta_time, curr_time;
-            if(current_speed_ == 0 && starting_downtrack >= ending_downtrack)  //If at end_dist return zero speed trajectory
+            if(start_speed == 0 )  //If at end_dist return zero speed trajectory
```
kjrush, line 172: Same comment as earlier with regards to doubles, generally speaking comparing doubles using `==` is a no-no given the quirks of how they're stored. You can have cases where if `x` is a double `(x * 2)/2) == x` is false and other sorts of weird issues which end up being very challenging to debug when they come up.

kjrush (changes requested):

guidance/src/guidance/guidance_state_machine.cpp
adev4a: This initialized comes from Line 122,  at the end of the route the system_alert message is checked, if that is "DRIVERS READY", this INITIALIZED signal is created

kjrush (approved):

PR #1056, opened 2021-02-04
Feature/mobility conversion

MishkaMN:
# PR Details
## Description
To support the cooperative lane change design, we want to use MobilityPath data to get accurate predictions of vehicle motion for external object prediction. We updated our motion_computation node to do a direct conversion of MobilityPath data to ExternalObject. This operational mode will be controlled by 
a flag with 3 states:

MOBILITY_PATH_ONLY: MobilityPath used as only source of external object data
SENSORS_ONLY: Sensors used as only source of external object data (mobility paths dropped)
PATH_AND_SENSORS: Both MobilityPath and sensors used without fusion but synchronized so the output message contains both.

--------------
Implementation:
MOBILITY_PATH_ONLY & SENSORS_ONLY publishes incoming respective directly when it comes (mobility is synchronized with sensors still even if I are not using it)
PATH_AND_SENSORS: When this mode is selected, published ExternalObjects are synchronized with the timestamp of the sensor_data, and it is only published when new sensor data is received.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1067

## Motivation and Context
See above

## How Has This Been Tested?
Unit and local integration tested

## Types of changes

## Checklist:

### DIFF
--- a/motion_computation/include/motion_computation_worker.h
+++ b/motion_computation/include/motion_computation_worker.h
@@ -20,9 +20,21 @@
 #include <ros/ros.h>
 #include <cav_msgs/ExternalObject.h>
 #include <cav_msgs/ExternalObjectList.h>
+#include <cav_msgs/MobilityPath.h>
+#include <std_msgs/String.h>
 #include <functional>
+#include <lanelet2_extension/projection/local_frame_projector.h>
+#include <tf2/LinearMath/Transform.h>
 
 namespace object{
+
+//! @brief Enum describing the possible operational modes of the MotionComputation
+enum MotionComputationMode
+{
+  MOBILITY_PATH_ONLY = 0,   // MobilityPath used as only source of external object data
+  SENSORS_ONLY = 1,         // Sensors used as only source of external object data (mobility paths dropped)
+  PATH_AND_SENSORS = 2,     // Both MobilityPath and sensors used without fusion but synchronized so the output message contains both
+};
 
 class MotionComputationWorker
 {
@@ -45,26 +57,49 @@ class MotionComputationWorker
 
   // Setters for the prediction parameters
   void setPredictionTimeStep(double time_step);
+  void setMobilityPathPredictionTimeStep(double time_step);
   void setPredictionPeriod(double period);
   void setXAccelerationNoise(double noise);
   void setYAccelerationNoise(double noise);
   void setProcessNoiseMax(double noise_max);
   void setConfidenceDropRate(double drop_rate);
- 
+  void setExternalObjectPredictionMode(int external_object_prediction_mode);
+
+  //callbacks
+  void mobilityPathCallback(const cav_msgs::MobilityPath& msg);
+  void geoReferenceCallback(const std_msgs::String& georef);
+  
+  // TODO
+  cav_msgs::PredictedState composePredictedState(const lanelet::BasicPoint3d& curr_pt, const lanelet::BasicPoint3d& prev_pt, const ros::Time& prev_time_stamp);
+
+  // msg converter
+  cav_msgs::ExternalObject mobilityPathToExternalObject(const cav_msgs::MobilityPath& msg);
+  
  private:
 
-  // local copy of external object publihsers
+  // local copy of external object publishers
 
   PublishObjectCallback obj_pub_;
 
   // Prediction parameters
   double prediction_time_step_ = 0.1;
+  double mobility_path_prediction_time_step_ = 0.1;
   double prediction_period_ = 2.0;
   double cv_x_accel_noise_ = 9.0;
   double cv_y_accel_noise_ = 9.0;
   double prediction_process_noise_max_ = 1000.0;
   double prediction_confidence_drop_rate_ = 0.9;
   
+  // External object conversion mode
+  MotionComputationMode external_object_prediction_mode_ = MOBILITY_PATH_ONLY;
+
+  // Queue for mobility path msgs to synchronize them with sensor msgs 
+  cav_msgs::ExternalObjectList mobility_path_list_;
+
+  // Projector variables to convert from ECEF to Map frame
+  std::string georeference_ = "";
+  std::shared_ptr<lanelet::projection::LocalFrameProjector> local_projector;
+
 };
 
 }//object
--- a/motion_computation/include/motion_computation_worker.h
+++ b/motion_computation/include/motion_computation_worker.h
@@ -25,9 +25,9 @@
 #include <functional>
 #include <motion_predict/motion_predict.h>
 #include <motion_predict/predict_ctrv.h>
-#include <lanelet2_extension/projection/local_frame_projector.h>
 #include <carma_utils/CARMAUtils.h>
 #include <tf2/LinearMath/Transform.h>
+#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 
 namespace object{
 
@@ -45,6 +45,7 @@ class MotionComputationWorker
  public:
 
   using PublishObjectCallback = std::function<void(const cav_msgs::ExternalObjectList&)>;
+  using LookUpTransform = std::function<void()>;
 
   /*!
    * \brief Constructor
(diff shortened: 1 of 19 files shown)


motion_computation/include/motion_computation_worker.h
```diff
@@ -45,28 +60,55 @@ class MotionComputationWorker
...
   void setConfidenceDropRate(double drop_rate);
- 
- private:
+  void setExternalObjectPredictionMode(int external_object_prediction_mode);
+  
+  //callbacks
+  void mobilityPathCallback(const cav_msgs::MobilityPath& msg);
+  void geoReferenceCallback(const std_msgs::String& georef);
+  
+  // TODO
```
msmcconnell, line 75: todo comment?

motion_computation/include/motion_computation_worker.h
```diff
@@ -45,28 +60,55 @@ class MotionComputationWorker
...
+  cav_msgs::PredictedState composePredictedState(const lanelet::BasicPoint3d& curr_pt, const lanelet::BasicPoint3d& prev_pt, const ros::Time& prev_time_stamp);
+
+  // msg converter
+  cav_msgs::ExternalObject mobilityPathToExternalObject(const cav_msgs::MobilityPath& msg);
 
-  // local copy of external object publihsers
+  // TODO synchronizes and interpolates
```
msmcconnell, line 81: todo?

motion_computation/src/motion_computation_worker.cpp
```diff
@@ -90,20 +80,39 @@ void MotionComputationWorker::predictionLogic(cav_msgs::ExternalObjectListPtr ob
...
-  obj_pub_(list);
+  // Determine mode
+  switch(external_object_prediction_mode_)
+  {
+    case SENSORS_ONLY:
+      output_list = sensor_list;
+    break;
+    case PATH_AND_SENSORS:
+      output_list = synchronizeAndAppend(sensor_list, mobility_path_list_);
+    break;
```
msmcconnell, line 94: Where is mobility path only mode?

motion_computation/src/motion_computation_worker.cpp
```diff
@@ -129,4 +138,218 @@ void MotionComputationWorker::setConfidenceDropRate(double drop_rate)
...
+    return output;
+  }
+  output.dynamic_obj = true;
+
+  // get planned trajectory points
+  auto prev_pt_msg = msg.trajectory.offsets[0]; // setup first point to be processed later
+  cav_msgs::PredictedState prev_state;
+  lanelet::BasicPoint3d prev_pt_ecef {ecef_x + (double)prev_pt_msg.offset_x /100.0, ecef_y + (double)prev_pt_msg.offset_y /100.0, ecef_z + (double)prev_pt_msg.offset_z /100.0};
+
+  auto prev_pt_map = local_projector_->projectECEF(prev_pt_ecef, -1);
```
msmcconnell, line 207: Adding lanelet2_extension and lanelet2 as a dep seems excessive just to get this functionality. 
The "earth" frame is already ECEF, so if you just lookup and cache the earth->map transform once you can apply it to all the points without depending on lanelet2.  If you think the current approach is cleaner though I'm open to an explanation of that.

msmcconnell (changes requested):

motion_computation/src/motion_computation_worker.cpp
```diff
@@ -129,4 +138,218 @@ void MotionComputationWorker::setConfidenceDropRate(double drop_rate)
...
+      curr_state.predicted_position.position.y = curr_pt_map.y();
+      curr_state.predicted_position.position.z = curr_pt_map.z();
+      output.predictions.push_back(curr_state);
+    }
+    
+    prev_state = curr_state;
+    prev_pt_map = curr_pt_map;
+  }
+
+  return output;
```
msmcconnell, line 242: It looks to me like we are reaching this point without ever setting the current state for the external object. Only the positions seem to be set. Is that being done somewhere I am not seeing?

motion_computation/src/motion_computation_worker.cpp
```diff
@@ -129,4 +138,218 @@ void MotionComputationWorker::setConfidenceDropRate(double drop_rate)
...
+
+  tf2::Quaternion vehicle_orientation;
+  vehicle_orientation.setRPY(0, 0, yaw);
+  output_state.predicted_position.orientation.x = vehicle_orientation.getX();
+  output_state.predicted_position.orientation.y = vehicle_orientation.getY();
+  output_state.predicted_position.orientation.z = vehicle_orientation.getZ();
+  output_state.predicted_position.orientation.w = vehicle_orientation.getW();
+
+  // Set velocity
+  output_state.predicted_velocity.linear.x = vehicle_vector.norm() / mobility_path_prediction_time_step_;
```
msmcconnell, line 266: It might also worth setting the angular.z velocity as the rate of change of angle between this and the last point.

roadway_objects/src/RoadwayObjectsNode.cpp
```diff
@@ -25,6 +25,8 @@ RoadwayObjectsNode::RoadwayObjectsNode()
   external_objects_sub_ =
       nh_.subscribe("external_objects", 10, &RoadwayObjectsWorker::externalObjectsCallback, &object_worker_);
   roadway_obs_pub_ = nh_.advertise<cav_msgs::RoadwayObstacleList>("roadway_objects", 10);
+  georef_sub_ = nh_.advertise<cav_msgs::RoadwayObstacleList>("roadway_objects", 10);
```
msmcconnell, line 28: What is this change for?

msmcconnell (changes requested):

motion_computation/src/motion_computation_worker.cpp
MishkaMN: if (i==1) is handling the current state of the object. Pose includes position and orientation. Velocity is set there as well. Let me know if I am missing anything

roadway_objects/src/RoadwayObjectsNode.cpp
MishkaMN: sorry this was error

motion_computation/src/motion_computation_worker.cpp
MishkaMN: I was not sure if I could put it in here because I needed next prediction's orientation to calculate the angular speed. Next prediction's orientation only gets calculated on the next iteration, so I added extra function to loop over all the predictions and fill in the angular velocity. Lmk if that makes sense

motion_computation/src/motion_computation_worker.cpp
```diff
@@ -129,4 +138,218 @@ void MotionComputationWorker::setConfidenceDropRate(double drop_rate)
...
+  }
+  local_projector_ = std::make_shared<lanelet::projection::LocalFrameProjector>(lanelet::projection::LocalFrameProjector(georeference_.c_str()));
+}
+
+void MotionComputationWorker::mobilityPathCallback(const cav_msgs::MobilityPath& msg)
+{
+  mobility_path_list_.objects.push_back(mobilityPathToExternalObject(msg));
+  if (external_object_prediction_mode_ == MOBILITY_PATH_ONLY)
+  {
+    obj_pub_(mobility_path_list_);
```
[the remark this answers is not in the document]
MishkaMN: I added it to the other switch statement

motion_computation/src/motion_computation_node.cpp
```diff
@@ -24,34 +24,45 @@ namespace object{
...
     motion_worker_.setYAccelerationNoise(ay);
     motion_worker_.setProcessNoiseMax(process_noise_max);
     motion_worker_.setConfidenceDropRate(drop_rate);
+    motion_worker_.setExternalObjectPredictionMode(external_object_prediction_mode);
+
 
     // Setup pub/sub
     motion_comp_sub_=nh_.subscribe("external_objects",1,&MotionComputationWorker::predictionLogic,&motion_worker_);
     carma_obj_pub_=nh_.advertise<cav_msgs::ExternalObjectList>("external_object_predictions", 2);
+    mobility_path_sub_=nh_.subscribe("incoming_mobility_path",1,&MotionComputationWorker::mobilityPathCallback,&motion_worker_);
```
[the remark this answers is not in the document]
MishkaMN: changed to 20

motion_computation/src/motion_computation_node.cpp
```diff
@@ -61,4 +69,18 @@ namespace object{
...
+    tf2::Transform map_in_earth;
+    try
+    {
+      tf2::convert(tf_buffer_.lookupTransform("earth", "map", ros::Time(0)).transform, map_in_earth); //save to local copy of transform
+    }
+    catch (const tf2::TransformException &ex)
+    {
+      ROS_ERROR_STREAM("Could not lookup transform with exception: " << ex.what());
+    }
+    return map_in_earth;
```
msmcconnell, line 83: There is a possiblity the transform would not be available before. This is called. I think there is a tf2::waitForTransform method you could use to handle this

motion_computation/src/motion_computation_worker.cpp
msmcconnell: Looks like you have the pose and velocity set. What about the object type, precense vector, and id?

motion_computation/src/motion_computation_worker.cpp
msmcconnell: I'm good with that.

msmcconnell (changes requested):

motion_computation/src/motion_computation_node.cpp
MishkaMN: looks like tf2 doesn't have it, and it was replaced by canTransform with timeout duration. Let me know if that is okay

motion_computation/src/motion_computation_worker.cpp
MishkaMN: Added those as well

motion_computation/src/motion_computation_node.cpp
```diff
@@ -13,52 +13,78 @@
...
+  tf2::Transform MotionComputationNode::lookupECEFtoMapTransform()
+  {
+    tf2::Transform map_in_earth;
+    try
+    {
+      std::string errstr;
+      if (tf_buffer_.canTransform("earth", "map", ros::Time(0), ros::Duration(20), &errstr)) //20 second timeout
+        tf2::convert(tf_buffer_.lookupTransform("earth", "map", ros::Time(0)).transform, map_in_earth); //save to local copy of transform
+      else
+        ROS_ERROR_STREAM("Look up transform timed out: " << errstr);
```
msmcconnell, line 81: We need to be throwing an exception here. We cannot allow the vehicle to operate without object detection.

motion_computation/src/motion_computation_node.cpp
```diff
@@ -13,52 +13,78 @@
...
+    }
+    catch (const tf2::TransformException &ex)
+    {
+      ROS_ERROR_STREAM("Could not lookup transform with exception: " << ex.what());
```
msmcconnell, line 85: Same as above

msmcconnell (changes requested):

motion_computation/src/motion_computation_node.cpp
msmcconnell: More specifically a system alert needs to occur. I believe this could be most cleanly achieved with a call to CARMANodeHandle::handleException since this initialize function is probably not wrapped in the default handler.

msmcconnell (changes requested):

msmcconnell (approved):
Looks good if CI passes

Issue #1061, opened 2021-02-05
Route end point can only be at the end of a lanelet


msmcconnell:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

The current route description does not provide a mechanism for allowing the route to end in the middle of a lanelet. This is problamatic on many of the maps where the lanelets are very large. Ideally, there would be some kind of end point or ending downtrack distance which can be checked against for planning the stopping maneuver as well as the route completed event. 
### Carma version where this issue was discovered

release/vanden-plas
### Expected behavior
Route can be completed in the middle of a lanelet
### Actual behavior

Route can only end at the end of a lanelet. 
### Steps to reproduce the actual behavior

Run the vanden-plas testing route.

msmcconnell:
Resolved by above PRs

PR #1078, opened 2021-02-08
Add stop point for route

MishkaMN:
# PR Details
## Description

The current route description does not provide a mechanism for allowing the route to end in the middle of a lanelet. This is problamatic on many of the maps where the lanelets are very large. Ideally, there would be some kind of end point or ending downtrack distance which can be checked against for planning the stopping maneuver as well as the route completed event.

## Related Issue

https://github.com/usdot-fhwa-stol/carma-platform/issues/1061
## Motivation and Context
See above.

## How Has This Been Tested?
Unit tested.

## Types of changes

## Checklist:


carma_wm/include/carma_wm/WorldModel.h
```diff
@@ -140,6 +140,13 @@ class WorldModel
    */
   virtual LaneletRouteConstPtr getRoute() const = 0;
 
+    /*! \brief Get trackpos of the end of route point relative to the last lanelet in the route
+   *
+   * \return Trackpos 
+   */
+  virtual TrackPos getRouteEndTrackPos() const = 0;
```
msmcconnell, line 147: I think it would make more sense to use the route track position rather than the lanelet track position. This is because the maneuvers are planned relative to the route track position, so it will take less logic for them to compare against that. 
See carma_wm::geometry::routeTrackPos()

msmcconnell (changes requested):

carma_wm/include/carma_wm/WorldModel.h
MishkaMN: using route track position now

carma_wm/include/carma_wm/WorldModel.h
```diff
@@ -140,6 +140,16 @@ class WorldModel
...
 
+    /*! \brief Get trackpos of the end of route point relative to the route
+   *
+   * \return Trackpos 
+   */
+  virtual TrackPos getRouteEndTrackPos() const = 0;
+
+  /*! \brief Set endpoint of the route
+   */
+  virtual void setRouteEndPoint(const lanelet::BasicPoint3d& end_point) = 0;
```
msmcconnell, line 151: I don't think this method needs to be part of this interface. I try to keep the WorldModel.h interface read only whenever possible. Since this is only ever called on CARMAWorldModel in the WMListener it doesn't need to be part of the public interface here.

msmcconnell (changes requested):

carma_wm/include/carma_wm/WorldModel.h
MishkaMN: ah okay makes sense

msmcconnell:
@MishkaMN This needs to be pointing at the release/vanden-plas branch

carma/launch/environment.launch
```diff
@@ -187,7 +187,7 @@
   <arg name="num_prediction" value="10"/>
   <arg name="sensor_height_calibration_params_file" value="$(arg vehicle_calibration_dir)/naive_motion_predict/calibration.yaml"/>
   <arg name="filter_out_close_object_threshold" value="1.5"/>
-  <arg name="input_topic" value="detection/objects"/>
+  <arg name="input_topic" value="/detection/fusion_tools/objects"/>
```
MishkaMN, line 190: I am not sure why it is saying it is changed. I confirmed this change is already in vanden-plas

MishkaMN:
Closed this PR as another PR is created here https://github.com/usdot-fhwa-stol/carma-platform/pull/1123
in order to use fixes from another branch to test this

Issue #1059, opened 2021-02-05
Silver and Blue lexus cannot make left hand turn at East Intersection at TFHRC


msmcconnell:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Niether the silver or blue lexus can safely make the left hand turn at the easy intersection. The vehicles approach the turn at far too high of a speed. 
Observaton of the map, route, and trajectory position all show them to be reasonable. This implies the issue lies in the speed generation approach. 
### Carma version where this issue was discovered

release/vanden-plas
### Expected behavior
Vehicle can make all turns
### Actual behavior

See description
### Steps to reproduce the actual behavior

Drive vanden-plas test plan route through east intersection at TFHRC

msmcconnell:
Resolved by above PR

lewisid:
During testing, the Blue Lexus failed to make the left-hand turn at the second stoplight/east intersection at TFHRC it would drive too fast when approaching the entrance and make a very wide turn (almost ending up on the sidewalk).

msmcconnell:
Per @lewisid comment, I believe this issue has reemerged as of vanden-plas commit 6f046fae17ed69ae5979d94c6f48d91fcd6c6cea though the cause could come from one of the other repos so the date on the commit would need to be compared.

msmcconnell:
After deploying the most recent vanden-plas build with default parameters this was not an issue. Will reclose this issue.

PR #1088, opened 2021-02-11
Hotfix/correct initial velocity

msmcconnell:
# PR Details
## Description
This PR applies some fixes to the plan_delegator to prevent duplicate maneuvers from being planned by inlane-cruising and correctly setting the entry speed for the trajectory.
It also adds a python script which parses the inlane-cruising log file to generate graphs which can be scrubbed by timestep to easily view trajectory generation over time. 
Other Changes
* The moving average filter in inlane-cruising is updated so that it does not shift minimum points by using a center based average instead of a following point average. 
* Pure pursuit wrapper is updated to apply a vehicle response lag in seconds to each recieved plan. 

## Related Issue
Resolves https://github.com/usdot-fhwa-stol/carma-platform/issues/1059

## Motivation and Context
Vehicle can drive routes at Turner

## How Has This Been Tested?
In the silver lexus

## Types of changes

## Checklist:

### DIFF
--- a/plan_delegator/src/plan_delegator.cpp
+++ b/plan_delegator/src/plan_delegator.cpp
@@ -30,6 +30,7 @@ namespace plan_delegator
         pnh_.param<std::string>("planning_topic_suffix", planning_topic_suffix_, "/plan_trajectory");
         pnh_.param<double>("spin_rate", spin_rate_, 10.0);
         pnh_.param<double>("trajectory_duration_threshold", max_trajectory_duration_, 6.0);
+        pnh_.param<double>("min_speed", min_crawl_speed_, min_crawl_speed_);
 
         traj_pub_ = nh_.advertise<cav_msgs::TrajectoryPlan>("plan_trajectory", 5);
         plan_sub_ = nh_.subscribe("final_maneuver_plan", 5, &PlanDelegator::maneuverPlanCallback, this);
@@ -96,7 +97,7 @@ namespace plan_delegator
 
     bool PlanDelegator::isManeuverExpired(const cav_msgs::Maneuver& maneuver, ros::Time current_time) const
     {
-        return GET_MANEUVER_PROPERTY(maneuver, end_time) <= current_time; // TODO maneuver expiration should maybe be based off of distance not time?
+        return GET_MANEUVER_PROPERTY(maneuver, end_time) <= current_time; // TODO maneuver expiration should maybe be based off of distance not time? https://github.com/usdot-fhwa-stol/carma-platform/issues/1107
     }
 
     cav_srvs::PlanTrajectory PlanDelegator::composePlanTrajectoryRequest(const cav_msgs::TrajectoryPlan& latest_trajectory_plan) const
@@ -159,7 +160,7 @@ namespace plan_delegator
             auto maneuver_planner = GET_MANEUVER_PROPERTY(maneuver, parameters.planning_tactical_plugin);
 
             //////////
-            // TODO REMOVE THE FOLLOWING IF STATEMENT AFTER VANDEN-PLAS release
+            // TODO REMOVE THE FOLLOWING IF STATEMENT AFTER VANDEN-PLAS release https://github.com/usdot-fhwa-stol/carma-platform/issues/1106
             /////////
             if (maneuver_planner.compare("InLaneCruisingPlugin") == 0) {
                 if (already_planned_inlane_cruising) {
@@ -198,7 +199,7 @@ namespace plan_delegator
                 break;
             }
         }
-        latest_trajectory_plan.initial_longitudinal_velocity = std::max(latest_twist_.twist.linear.x, 2.2352); // TODO make config paramete or evaluate if this max call needed? Would 0 cause an issue?
+        latest_trajectory_plan.initial_longitudinal_velocity = std::max(latest_twist_.twist.linear.x, min_crawl_speed_); 
         return latest_trajectory_plan;
     }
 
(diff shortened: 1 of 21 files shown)


route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -96,26 +96,30 @@ namespace route_following_plugin
 
         double target_speed=findSpeedLimit(current_lanelet);   //get Speed Limit
 
-        //double total_maneuver_length = current_progress + mvr_duration_ * target_speed; // TODO STOPPING_ISSUE we can do a min of this line and the line below. 
-        double total_maneuver_length = wm_->routeTrackPos(shortest_path.back().centerline2d().back()).downtrack; // TODO STOPPING_ISSUE Maneuver extends to the end of the route???
+        double total_maneuver_length = current_progress + mvr_duration_ * target_speed; // TODO STOPPING_ISSUE we can do a min of this line and the line below.
```
msmcconnell, line 99: This is a bit simplistic since the speed limit could change before the 15s maneuver duration has expired. But we can leave that change till later since vanden-plas only has one strategic plugin.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -96,26 +96,30 @@ namespace route_following_plugin
...
         stopping_dist = target_speed*time_req_to_stop - (0.167 * jerk_ * pow(time_req_to_stop,3));
-        if(total_maneuver_length - current_progress <= stopping_dist){
+        
+        if(route_length - current_progress <= stopping_dist){
             approaching_route_end = true;
         }
         while(current_progress < total_maneuver_length && !approaching_route_end)
         {
             ROS_DEBUG_STREAM("Lanlet: " << shortest_path[last_lanelet_index].id());
             auto p = shortest_path[last_lanelet_index].centerline2d().back();
```
msmcconnell, line 114: p is not used. We can leave for now, but remove before merging to vanden-plas

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -96,26 +96,30 @@ namespace route_following_plugin
...
+            end_dist = std::min(end_dist, total_maneuver_length);
+
             double dist_diff = end_dist - current_progress;
-            time_req_to_stop = sqrt(2*target_speed/jerk_);
-            stopping_dist = target_speed*time_req_to_stop - (0.167 * jerk_ * pow(time_req_to_stop,3));
             
-            if(total_maneuver_length - end_dist <= stopping_dist){
-                end_dist -= stopping_dist; // TODO STOPPING_ISSUE this will cause you to stop early if the final lanelet is shorter than stopping distance
+            if(route_length - end_dist <= stopping_dist){
+                end_dist -= stopping_dist; // TODO STOPPING_ISSUE this will cause you to stop early if the final lanelet is shorter than stopping distance
```
msmcconnell, line 121: This still allows for stopping before the end if end_distance does not equal route length but the delta is less than stopping distance. 
I think better logic would be
end_dist = route_length - stopping_distance

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -115,6 +115,10 @@ namespace stop_and_wait_plugin
             }
         }
 
+        if(current_downtrack < maneuver_plan[0].stop_and_wait.start_dist){
```
msmcconnell, line 118: Is there any possibility that meneuver_plan will not have at least one element? If so this can fail silently with a segfault. If we can guarentee this will not happen this is fine. Otherwise we should throw an exception here.

msmcconnell (commented):
@adev4a Here are my comments on your commit. Only the last two need changes.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
adev4a: Since the plan delegator will only request stop&wait for a plan, if there is an associated maneuver, there shouldn't be a case where this is zero

msmcconnell: Ok sounds good.

plan_delegator/src/plan_delegator.cpp
```diff
@@ -183,6 +198,7 @@ namespace plan_delegator
                 break;
             }
         }
+        latest_trajectory_plan.initial_longitudinal_velocity = std::max(latest_twist_.twist.linear.x, 2.2352); // TODO make config paramete or evaluate if this max call needed? Would 0 cause an issue?
```
kjrush, line 201: I think per our discussion 0 would cause some issues with starting up from a standstill. I'd be in favor of making this a config param though, probably a global one. We should just define a minimum "crawl speed" and anything below that should only occur on the way to 0.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -115,6 +115,10 @@ namespace stop_and_wait_plugin
             }
         }
 
+        if(current_downtrack < maneuver_plan[0].stop_and_wait_maneuver.start_dist){
```
kjrush, line 118: Hmm, this works as a bandaid but means that our plan shrinks pretty badly as we approach the junction between the maneuver before stop&wait and the stop&wait maneuver itself, right? Ideally we'd want them concatenated with the stop&wait maneuver starting at the end of whatever the prior maneuver is.

kjrush (changes requested):

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
msmcconnell: @adev4a Any thoughts on that ^

Issue #1060, opened 2021-02-05
Vehicle comes to a stop but does not produce ROUTE COMPLETED event


msmcconnell:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

While driving the silver lexus, the vehicle was able to come to a smooth stop at the end of the route, but did not recieve a ROUTE COMPLETED event. After manually driving a few feet forward the ROUTE COMPLETED event was recieved. This seems to imply that the vehicle needs to be providing some buffer for that event that is triggered by the 0 speed. Sometimes the notification did appear in the blue lexus so there might be a threshold somewhere that needs to be tweaked. 
### Carma version where this issue was discovered

release/vanden-plas
### Expected behavior
When the vehicle comes to a full stop at the end of the route a ROUTE COMPLETED event appears. 
### Actual behavior

See description
### Steps to reproduce the actual behavior

Drive vanden-plas test route to the end and observe stopping behavior.

msmcconnell:
Resolved by above PR

PR #1102, opened 2021-02-18
Fix/pure pursuit stopandwait

adev4a:
# PR Details
This pr fixes issue found in issue 1060 https://github.com/usdot-fhwa-stol/carma-platform/issues/1060
where the vehicle doesn't publish a route completed event after stopping at the end of the route.
## Description

Constant jerk calculation for converting time to speed was added in the pure pursuit wrapper and the time step for stop and wait
trajectories was changed to 0.1s. 
This PR requires corresponding change in carma_utils on branch feature/const_jerk_update.
## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1060

## Types of changes

## Checklist:


msmcconnell (changes requested):
I thought we had discussed breaking this custom logic into a different control plugin such that trajectory_executor would forward the trajectory to the relevant plugin.

route/config/route_config_params.yaml
```diff
@@ -4,7 +4,7 @@ max_crosstrack_error: 2.0
 
 # Double: downtrack_range
 # Units: meter
-destination_downtrack_range: 0
+destination_downtrack_range: 10
```
msmcconnell, line 7: What does this parameter do and why was it changed?

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -79,6 +81,9 @@ namespace stop_and_wait_plugin
         ros::CARMANodeHandle::setSpinCallback([this]() -> bool
         {
             plugin_discovery_pub_.publish(plugin_discovery_msg_);
+            std_msgs::Float64 jerk_msg;
+            jerk_msg.data = jerk_;
+            jerk_pub_.publish(jerk_msg);
```
msmcconnell, line 86: Is this parameter constant?
This whole approach to setting the jerk value via a subscription seems both pretty computationally expensive assuming you are publishing a constant value each timestep, and overly complex. Why not just have both components load the same jerk ros parameter once at startup?

msmcconnell (changes requested):

route/config/route_config_params.yaml
adev4a: This parameter tracks how far from the end of the route can we stop to get the route_completed message.
It was changed because we are stopping slightly before the end point of route

pure_pursuit_wrapper/src/pure_pursuit_wrapper.cpp
```diff
@@ -45,8 +49,17 @@ void PurePursuitWrapper::trajectoryPlanHandler(const cav_msgs::TrajectoryPlan::C
   std::vector<double> downtracks;
   trajectory_utils::conversions::trajectory_to_downtrack_time(tp->trajectory_points, &downtracks, &times);
 
+//Tag all the trajectory points needing constant jerk calc
+  std::vector<bool> isStopandWait;
+  isStopandWait.resize(tp->trajectory_points.size(),0);
+  for(int i=0;i<tp->trajectory_points.size();i++){
+    if(tp->trajectory_points[i].planner_plugin_name == "StopandWaitPlugin"){
```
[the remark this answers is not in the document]
adev4a: yes it is "StopandWaitPlugin", would changing it to "StopAndWaitPlugin" be more consistent?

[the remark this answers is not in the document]
msmcconnell: It's fine as is, I just wanted to double check since it wasn't quite camel case.

route/config/route_config_params.yaml
msmcconnell: Ok. Please update the description then, so the use of the parameter is clearer.

pure_pursuit_wrapper/include/pure_pursuit_wrapper/pure_pursuit_wrapper.hpp
```diff
@@ -36,14 +37,22 @@ class PurePursuitWrapper {
...
+         * \param jerk float64 message with jerk value from stop_and_wait_plugin
+         */
+        void updatejerk(std_msgs::Float64 jerk);
 
         bool onSpin();
 
     private:
     WaypointPub waypoint_pub_;
     PluginDiscoveryPub plugin_discovery_pub_;
+    ros::Subscriber jerk_sub_;  //Jerk value from stop&wait plugin
```
kjrush, line 51: This could be used by any plugin, so let's make sure this solution works for those cases as well.

pure_pursuit_wrapper/include/pure_pursuit_wrapper/pure_pursuit_wrapper.hpp
```diff
@@ -36,14 +37,22 @@ class PurePursuitWrapper {
...
     cav_msgs::Plugin plugin_discovery_msg_;
 
+    double stop_and_wait_jerk_;
```
kjrush, line 54: Same as above, let's keep this generic

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
kjrush: I think the problem is that while the jerk is constant for a given trajectory, it may not be constant across all trajectories (other than as a max limit).

I think the best solution may be to expand the trajectory message to include metadata fields like the maneuver plan does. That way, we can make it part of the API contract for this control plugin that we can specify the jerk via those metadata fields. Shouldn't require too much work to add that, but I'm not sure if that data needs only go on the trajectory object itself or if it needs to go on each point. Putting it on each point gives us finer control, but bloats our message pretty substantially.

msmcconnell:
Need to review this for this PR
https://github.com/usdot-fhwa-stol/carma-platform/pull/1088#discussion_r580559505

pure_pursuit_wrapper/include/pure_pursuit_wrapper/pure_pursuit_wrapper.hpp
adev4a: This PR is specific to handling jerk calculation for stop&wait. This pr includes a separate wrapper for plugins needing constant jerk https://github.com/usdot-fhwa-stol/carma-platform/compare/feature/add_const_jerk_wrapper

pure_pursuit_jerk_wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,217 @@
...
+#   FILES
+#   Action1.action
+#   Action2.action
+# )
+
+## Generate added messages and services with any dependencies listed here
+# generate_messages(
+#   DEPENDENCIES
+#   cav_msgs#   std_msgs
+# )
```
msmcconnell, line 91: Remove these large irrelevant comment blocks

pure_pursuit_jerk_wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,217 @@
+cmake_minimum_required(VERSION 3.0.2)
+project(pure_pursuit_jerk_wrapper)
+
+## Compile as C++11, supported in ROS Kinetic and newer
```
msmcconnell, line 4: c++14

pure_pursuit_jerk_wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,217 @@
...
+  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
+
+# Mark executables for installation
+install(DIRECTORY include/${PROJECT_NAME}/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+  FILES_MATCHING PATTERN "*.h"
+  PATTERN ".svn" EXCLUDE
```
msmcconnell, line 195: you can remove the FILES_MATCHING and PATTERN lines.  They are not needed

pure_pursuit_jerk_wrapper/src/pure_pursuit_jerk_wrapper_node.cpp
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (C) 2018-2020 LEIDOS.
```
msmcconnell, line 2: 2021

msmcconnell: Please double check all the copyrights

pure_pursuit_wrapper/src/pure_pursuit_wrapper.cpp
```diff
@@ -69,7 +70,7 @@ void PurePursuitWrapper::trajectoryPlanHandler(const cav_msgs::TrajectoryPlan::C
     wp.twist.twist.linear.x = lag_speeds[i];
     ROS_DEBUG_STREAM("Setting waypoint idx: " << i <<", x: " << tp->trajectory_points[i].x << 
                             ", y: " << tp->trajectory_points[i].y <<
-                            ", speed: " << lag_speeds[i]* 2.23694 << "mph");
+                            ", speed: " << speeds[i]* 2.23694 << "mph");
```
msmcconnell, line 73: This should stay as lag_speeds

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -103,7 +108,7 @@ namespace stop_and_wait_plugin
     
     bool StopandWait::plan_trajectory_cb(cav_srvs::PlanTrajectoryRequest& req, cav_srvs::PlanTrajectoryResponse& resp)
     {
-        lanelet::BasicPoint2d veh_pos(req.vehicle_state.X_pos_global,req.vehicle_state.Y_pos_global);
+        lanelet::BasicPoint2d veh_pos(req.vehicle_state.X_pos_global,req.vehicle_state.Y_pos_global); // TODO STOPPING_ISSUE this is why we plan to the end of the route
```
msmcconnell, line 111: The comment here can be removed right?

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -199,8 +201,9 @@ namespace stop_and_wait_plugin
             }
             else
             {
-                double jerk_req = (2*start_speed)/pow(maneuver_time_,2);
-                //double start_speed = current_speed_;    //Get static value of current speed at start of planning
+                //double jerk_req = (2*start_speed)/pow(maneuver_time_,2);
+                double jerk_req = 2.0; //Setting constant for testing:vanden-plas
```
msmcconnell, line 205: Is this hardcoded Jerk expected?

msmcconnell: If so shouldn't it be a parameter?

pure_pursuit_jerk_wrapper/launch/pure_pursuit_jerk_wrapper.launch
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<launch>
+    <remap from="final_waypoints" to="carma_final_waypoints"/>
```
msmcconnell, line 3: Looks like this wrapper is using the same pure_pursuit instance as the other wrapper. This launch file should probably launch a seperate instance of the pure_pursuit node with a different input topic so that there is less risk of a race condition causing overlapping commands. But if @kjrush wants to wait I imagine its a bit of a low risk at the moment.

msmcconnell (changes requested):
Generally its looking good.

pure_pursuit_jerk_wrapper/src/pure_pursuit_jerk_wrapper_node.cpp
adev4a: Should this be 2020-2021?

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
adev4a: This shouldn't be hardcoded, but currently making it dynamic is causing the vehicle to prompt a route_completed message without completely bringing it to a stop.

pure_pursuit_jerk_wrapper/src/pure_pursuit_jerk_wrapper_node.cpp
msmcconnell: Just 2021 since that is the year the file was created

pure_pursuit_jerk_wrapper/src/pure_pursuit_jerk_wrapper.cpp
```diff
@@ -71,7 +71,7 @@ void PurePursuitJerkWrapper::trajectoryPlanHandler(const cav_msgs::TrajectoryPla
     wp.twist.twist.linear.x = speeds[i];
     ROS_DEBUG_STREAM("Setting waypoint idx: " << i <<", x: " << tp->trajectory_points[i].x << 
                             ", y: " << tp->trajectory_points[i].y <<
-                            ", speed: " << speeds[i]* 2.23694 << "mph" << "  Tactical plugin:"<< tp->trajectory_points[i].planner_plugin_name);
+                            ", speed: " << lag_speeds[i]* 2.23694 << "mph" << "  Tactical plugin:"<< tp->trajectory_points[i].planner_plugin_name);
```
msmcconnell, line 74: I think there might have been some git weirness here. speeds[i] is correct for this file, but lag_speeds[i] is correct for the regular wrapper.

msmcconnell (changes requested):

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
msmcconnell: Ok we will leave it for later per your other investigation

trajectory_executor/src/trajectory_executor/trajectory_executor.cpp
```diff
@@ -56,6 +56,11 @@ namespace trajectory_executor
         _private_nh->param<std::string>("default_control_plugin_topic", default_control_plugin_topic_, "NULL");
 
         out[default_control_plugin_] = default_control_plugin_topic_;
+
+        //Hardcoding jerk control wrapper
+        std::string default_jerk_control_plugin = "Pure Pursuit Jerk";
+        std::string default_jerk_control_plugin_topic_ = "/guidance/pure_pursuit/plan_jerk_trajectory";
+        out[default_jerk_control_plugin]=default_jerk_control_plugin_topic_
```
msmcconnell, line 63: Didn't see this before. I seem to recall you and Kyle discussing this on slack. So I want confirmation from @kjrush this is ok to have here.

msmcconnell (approved):

PR #1109, opened 2021-02-24
Implementation of Route Invalidated event

aswath1:
# PR Details
## Description
When a geofence is received, it is possible for it to make the current route invalid for example if a lane is closed that was part of the route. To support this, the route package state machine includes a Route Invalidated event which is meant to force a reroute operation.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1148

## Types of changes

## Checklist:


carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -233,6 +233,7 @@ void WMBroadcaster::addPassingControlLineFromMsg(std::shared_ptr<Geofence> gf_pt
 void WMBroadcaster::addRegionAccessRule(std::shared_ptr<Geofence> gf_ptr, const cav_msgs::TrafficControlMessageV01& msg_v01, const std::vector<lanelet::Lanelet>& affected_llts) const
 {
    gf_ptr->regulatory_element_ = std::make_shared<lanelet::RegionAccessRule>(lanelet::RegionAccessRule::buildData(lanelet::utils::getId(),affected_llts,{},participantsChecker(msg_v01)));
+   route_flag_=true;
```
msmcconnell, line 236: I think it would be safer for this flag to be part of the Geofence object gf_ptr refers to. The addGeofence callback is triggered based on a geofence's  schedule, so its possible you could encounter a race condition where this is set for another geofence that is not yet triggered, but get applied for a different one that is. 

If it is part of the Geofence object this would not be an issue

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -571,9 +572,18 @@ void WMBroadcaster::addGeofence(std::shared_ptr<Geofence> gf_ptr)
   autoware_lanelet2_msgs::MapBin gf_msg;
   auto send_data = std::make_shared<carma_wm::TrafficControl>(carma_wm::TrafficControl(gf_ptr->id_, gf_ptr->update_list_, gf_ptr->remove_list_));
   carma_wm::toBinMsg(send_data, &gf_msg);
-  map_update_pub_(gf_msg);
   
-
+  if(route_flag_==true)
```
msmcconnell, line 576: Lets name this something other than route_flag that does not convey meaning. invalidates_route or something similar would work.

msmcconnell, line 576: rather than having an if/else statement could could just assign the flag directing to the message field since both are boolean with a 1-1 mapping.

route/include/route_generator_worker.h
```diff
@@ -56,6 +56,10 @@ namespace route {
 
     public:
 
+        std::function<bool()> localFunction;
```
msmcconnell, line 59: localFunction as a variable name does not convey meaning of this field.

route/include/route_generator_worker.h
```diff
@@ -56,6 +56,10 @@ namespace route {
...
+
+        void setReroutingChecker(std::function<bool()> inputFunction);
```
msmcconnell, line 61: Needs function comment.

route/src/route_generator_worker.cpp
```diff
@@ -418,6 +424,20 @@ namespace route {
...
+        if(localFunction()==true)
+        {
+           this->rs_worker_.on_route_event(RouteStateWorker::RouteEvent::ROUTE_INVALIDATION);
+           publish_route_event(cav_msgs::RouteEvent::ROUTE_INVALIDATION);
+           
+           auto route=routing(destination_points_in_map_.front(),
+                                std::vector<lanelet::BasicPoint2d>(destination_points_in_map_.begin() + 1, destination_points_in_map_.end() - 1),
+                                destination_points_in_map_.back(),
+                                world_model_->getMap(), world_model_->getMapRoutingGraph());
+
```
msmcconnell, line 436: There needs to be a check here for if routing succeeded or not. If it failed we need to log something and call the route_aborted callback. If it succeeds the state needs to transition from routing to following.

route/src/route_generator_worker.cpp
```diff
@@ -418,6 +424,20 @@ namespace route {
     
     bool RouteGeneratorWorker::spin_callback()
     {
+        if(localFunction()==true)
+        {
+           this->rs_worker_.on_route_event(RouteStateWorker::RouteEvent::ROUTE_INVALIDATION);
+           publish_route_event(cav_msgs::RouteEvent::ROUTE_INVALIDATION);
+           
+           auto route=routing(destination_points_in_map_.front(),
```
msmcconnell, line 432: On a basic level this is fine, but I think there is an edge case we need to consider. 
The route file is made up of a list of destination points. If the vehicle, has already passed one of these points it will cause the new reroute to loop back on itself. I would suggest adding some logic to this node which tracks which points have been passed and then removes them from the destination_points_in_map_ list. 

I think the way to approach this would be to compute the corresponding route downtrack distances for each point after the original route it generated, then you could compare the output vehicle downtrack from this node against those points on each spin loop to remove points as needed.

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -137,6 +152,12 @@ void WMListenerWorker::roadwayObjectListCallback(const cav_msgs::RoadwayObstacle
 
 void WMListenerWorker::routeCallback(const cav_msgs::RouteConstPtr& route_msg)
 {
+
+  if(set_flag_==true)
```
msmcconnell, line 156: Flag name should be something more meaningful

route/src/route.cpp
```diff
@@ -53,6 +56,7 @@ namespace route {
         pnh_->getParam("route_file_path", route_file_location);
         rg_worker_.set_route_file_path(route_file_location);
         rg_worker_.set_publishers(route_event_pub_, route_state_pub_, route_pub_,route_marker_pub_);
+        rg_worker_.set_publishers(wml_);
```
msmcconnell, line 59: I highly doubt this line works the way you think it does. The wml_ object is not a ros::publisher. I'm pretty amazed this even compiles.

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -52,8 +52,21 @@ void WMListenerWorker::mapCallback(const autoware_lanelet2_msgs::MapBinConstPtr&
...
+}
+
 void WMListenerWorker::mapUpdateCallback(const autoware_lanelet2_msgs::MapBinConstPtr& geofence_msg) const
 {
+
+  if(geofence_msg.invalidates_route==true)
+  {
+  local_geofence_msg_=geofence_msg;
+  set_flag_=true;
+  }
```
msmcconnell, line 68: This is supposed to not apply the map until the route is received. You need to return in that if call so that the rest of this method does not run.

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -52,8 +52,21 @@ void WMListenerWorker::mapCallback(const autoware_lanelet2_msgs::MapBinConstPtr&
     map_callback_();
   }
 }
+
+bool WMListenerWorker::checkIfReRoutingNeeded()
```
msmcconnell, line 56: This method works as a way for the route node to know if it needs to update the route, but there is still no mechanism to allow the route node specifically to update the map while other nodes do not. You need something the route node can set in the wm listener to handle that.

msmcconnell (changes requested):

route/src/route.cpp
aswath1: It shouldn't be there. Came from somewhere.

carma_wm_ctrl/src/WMBroadcaster.cpp
aswath1: Done

carma_wm_ctrl/src/WMBroadcaster.cpp
aswath1: Done

aswath1: Done

carma_wm/src/WMListenerWorker.cpp
aswath1: Done

route/include/route_generator_worker.h
aswath1: Done

route/src/route_generator_worker.cpp
aswath1: Done.

aswath1:
https://github.com/usdot-fhwa-stol/carma-platform/issues/1148

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -52,8 +51,32 @@ void WMListenerWorker::mapCallback(const autoware_lanelet2_msgs::MapBinConstPtr&
...
 {
+
+  if(geofence_msg->invalidates_route==true)
+  {  
+    rerouting_flag_=true;
+    local_geofence_msg_ = boost::make_shared<autoware_lanelet2_msgs::MapBin>(*geofence_msg);
+
+    if(route_node_flag_!=true)
+    {
+     ROS_INFO_STREAM("Route is not invalidated/available yet");
```
msmcconnell, line 75: This alert should just say Route is not available yet. "Not invalidated" is a bit confusing since it was invalidated.

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -232,13 +232,23 @@ void WMBroadcaster::addPassingControlLineFromMsg(std::shared_ptr<Geofence> gf_pt
 
 void WMBroadcaster::addRegionAccessRule(std::shared_ptr<Geofence> gf_ptr, const cav_msgs::TrafficControlMessageV01& msg_v01, const std::vector<lanelet::Lanelet>& affected_llts) const
 {
-   gf_ptr->regulatory_element_ = std::make_shared<lanelet::RegionAccessRule>(lanelet::RegionAccessRule::buildData(lanelet::utils::getId(),affected_llts,{},participantsChecker(msg_v01)));
+  auto regulatory_element = std::make_shared<lanelet::RegionAccessRule>(lanelet::RegionAccessRule::buildData(lanelet::utils::getId(),affected_llts,{},participantsChecker(msg_v01)));
+
+  if(!regulatory_element->accessable("car"))
```
msmcconnell, line 237: I believe the participant type should be "participant:vehicle" to match the lanelet2 standards.

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -267,6 +277,7 @@ ros::V_string WMBroadcaster::participantsChecker(const cav_msgs::TrafficControlM
             participant.vehicle_class == j2735_msgs::TrafficControlVehClass::PASSENGER_CAR)
     {
       participants.push_back(lanelet::Participants::VehicleCar);
+      ROS_WARN_STREAM("HEYYYYYYYY");
```
msmcconnell, line 280: log cleanup

msmcconnell (changes requested):

docker/checkout.bash
```diff
@@ -38,12 +38,12 @@ done
...
 
 if [[ "$BRANCH" = "develop" ]]; then
-      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch $BRANCH
+      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch feature/routeEventUpdate
       git clone --depth=1 https://github.com/usdot-fhwa-stol/novatel_gps_driver.git --branch $BRANCH
       git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-utils.git --branch $BRANCH
       git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-messenger.git --branch $BRANCH
 else
-      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch develop
+      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch feature/routeEventUpdate
```
msmcconnell, line 46: Before merging this needs to be reverted to develop

docker/checkout.bash
```diff
@@ -38,12 +38,12 @@ done
 cd ${dir}/src
 
 if [[ "$BRANCH" = "develop" ]]; then
-      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch $BRANCH
+      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch feature/routeEventUpdate
```
msmcconnell, line 41: Before merging this needs to be reverted to $BRANCH

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -232,13 +232,25 @@ void WMBroadcaster::addPassingControlLineFromMsg(std::shared_ptr<Geofence> gf_pt
 
 void WMBroadcaster::addRegionAccessRule(std::shared_ptr<Geofence> gf_ptr, const cav_msgs::TrafficControlMessageV01& msg_v01, const std::vector<lanelet::Lanelet>& affected_llts) const
 {
-   gf_ptr->regulatory_element_ = std::make_shared<lanelet::RegionAccessRule>(lanelet::RegionAccessRule::buildData(lanelet::utils::getId(),affected_llts,{},participantsChecker(msg_v01)));
+  auto regulatory_element = std::make_shared<lanelet::RegionAccessRule>(lanelet::RegionAccessRule::buildData(lanelet::utils::getId(),affected_llts,{},participantsChecker(msg_v01)));
+
+  if(regulatory_element->accessable("vehicle:car")) //TODO: region_access rule by default blocks everything except participant
```
msmcconnell, line 237: This says participant "vehicle:car" but what I suggested was "participant:vehicle". 
Just to be safe you can just use lanelet::Participants::Vehicle

msmcconnell (changes requested):

msmcconnell:
Please address this sonar smell. Seems like a potential bug
https://sonarcloud.io/project/issues?id=usdot-fhwa-stol_CARMAPlatform&issues=AXgzv7KhHf9EeM931bb4&open=AXgzv7KhHf9EeM931bb4&pullRequest=1109

route/src/route_generator_worker.cpp
```diff
@@ -339,17 +345,17 @@ namespace route {
     {
         if(this->rs_worker_.get_route_state() == RouteStateWorker::RouteState::FOLLOWING) {
             // convert from pose stamp into lanelet basic 2D point
-            lanelet::BasicPoint2d current_loc(msg->pose.position.x, msg->pose.position.y);
+            current_loc_(msg->pose.position.x, msg->pose.position.y);
```
msmcconnell, line 348: This is a bug. You cannot call a constructor on an existing object. This should be current_loc_ = lanelet::BasicPoint2d(msg->pose.position.x, msg->pose.position.y)

msmcconnell (changes requested):

msmcconnell (approved):

Issue #1114, opened 2021-02-25
In-Lane cruising does not always select the shortest path lanelet when there is an adjacent lanelet going in the same direction


msmcconnell:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

The picture below shows an issue which occurred at summit point during vande-plas verification testing
![image](https://user-images.githubusercontent.com/10292079/109208426-473ea200-7778-11eb-8f44-f313d075d289.png)
Investigation showed that inlane-cruising was trying to plan using the adjacent lane's centerline. 
### Carma version where this issue was discovered

release/vanden-plas
### Expected behavior
In-lanecruising only uses the shortest path centerline to plan
### Actual behavior

See image and description
### Steps to reproduce the actual behavior

Run the vandenplas build on the two lane summit point map and drive a route through an intersection.

msmcconnell:
Resolved by above PR

PR #1115, opened 2021-02-25
Fix #114

### LINKED ISSUE #114: Upgrade TrafficSignalPlugin to use external queue prediction data
Contract TO 13 produced a report describing how roadway sensors could be used to feed an infrastructure-based intersection queue discharge prediction algorithm.  This prediction could then be transmitted via a new type of J2735 message to approaching vehicles.  It should give them a better understanding of expected queue discharge, since they may not be able to detect how many cars are in the queue by themselves, and thus their onboard prediction would be limited.

If this enhancement is planned for implementation there must be an accompanying infrastructure component developed simultaneously to read the roadway sensors, implement the queue prediction algorithm, and transmit that prediction data to the vehicles.  This ticket is only for the Carma Platform onboard plugin enhancement.

msmcconnell:
# PR Details
## Description
Issue #1114 was caused because the shortest_path_view variable in carma world model was created using the lanelet::createConstMap() call provided with only the lanelets on the shortest path. However, because there was a regulatory element shared by lanelets on the shortest path and other lanelets these non-shortest path lanelets were also added to the map. 
This PR updates the call to use lanelet2's SubMap interface which only adds the elements you provide and not their children's children. 

## Related Issue
#1114 

## Motivation and Context
Proper route following behavior

## How Has This Been Tested?
Unit tests pass in carma_wm

## Types of changes

## Checklist:

### DIFF
--- a/carma_wm/include/carma_wm/CARMAWorldModel.h
+++ b/carma_wm/include/carma_wm/CARMAWorldModel.h
@@ -173,7 +173,7 @@ private:
   std::vector<lanelet::LineString3d> shortest_path_centerlines_;  // List of disjoint centerlines seperated by lane
                                                                   // changes along the shortest path
   IndexedDistanceMap shortest_path_distance_map_;
-  lanelet::LaneletSubmapUPtr shortest_path_filtered_centerline_view_;  // Lanelet map view of shortest path center lines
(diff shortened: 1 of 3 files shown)


msmcconnell:
Don't merge yet

aswath1 (approved):

PR #1119, opened 2021-03-01
update yield to work with clc

saina-ramyar:
# PR Details
## Description
The yield plugin is updated to participate in cooperative behavior by receiving the mobility request messages, adjust the trajectory based on the incoming requested trajectory, and publishing mobility response.

## Motivation and Context
To expand the functionality of the Yield plugin to take part in cooperative behavior.

## How Has This Been Tested?
unit test and ros test.

## Types of changes

## Checklist:


yield_plugin/src/yield_plugin.cpp
```diff
@@ -37,72 +37,344 @@ using oss = std::ostringstream;
...
+    req_trajectory_points_ = req_trajectory;
+    req_target_speed_ = req_speed;
+    req_target_plan_time_ = req_planning_time;
+  }
+
+
+  void YieldPlugin::bsm_cb(const cav_msgs::BSMConstPtr& msg)
+  {
+    cav_msgs::BSMCoreData bsm_core_ = msg->core_data;
+    host_bsm_id_ = bsmIDtoString(bsm_core_);
```
kjrush, line 226: Not sure I understand the naming, is this actually the host vehicle's BSM ID? Or would this be the ID of the vehicle sending us the request? if the latter, we should probably rename this for clarity.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -37,72 +37,344 @@ using oss = std::ostringstream;
...
 
-    trajectory = update_traj_for_object(original_trajectory, req.vehicle_state.longitudinal_vel); // Compute the trajectory
-    trajectory.initial_longitudinal_velocity = original_trajectory.initial_longitudinal_velocity;//copy the original trajectory's desired speed for now. 
+    // seperating cooperative yield with regular object detection for better performance.
+    if (config_.enable_cooperative_behavior && received_cooperative_request_)
+    {
+      tf2_listener_.reset(new tf2_ros::TransformListener(tf2_buffer_));
+      yield_trajectory = update_traj_for_cooperative_behavior(original_trajectory, req.vehicle_state.longitudinal_vel);
+      // reset the flag
+      received_cooperative_request_ = false;
```
kjrush, line 244: If we reset this flag every timestep, are we in danger of having some strange behavior? I'm concerned if they're a little late next timestep with sending us the message, will that cause us to not yield or to kind of toggle in and out of yield behavior on every other timestep?

yield_plugin/include/yield_plugin/yield_plugin.h
```diff
@@ -86,48 +96,145 @@ class YieldPlugin
...
+   * \brief callback for mobility request
+   * \param msg mobility request message 
+   */
+  void mobilityrequest_cb(const cav_msgs::MobilityRequestConstPtr& msg);
+
+  /**
+   * \brief callback for bsm message
+   * \param msg mobility bsm message 
+   */
+  void bsm_cb(const cav_msgs::BSMConstPtr& msg);
```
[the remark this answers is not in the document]
saina-ramyar: I checked with @msmcconnell and he said the host bsm id is not available in wm, and it changes frequently so we need to subscribe to the msg to get the correct value.

yield_plugin/include/yield_plugin/yield_plugin_node.h
```diff
@@ -46,6 +48,8 @@ class YieldPluginNode
     auto wm_ = wml.getWorldModel();
 
     ros::Publisher discovery_pub = nh.advertise<cav_msgs::Plugin>("plugin_discovery", 1);
+    ros::Publisher mob_resp_pub = nh.advertise<cav_msgs::MobilityResponse>("outgoing_mobility_response", 1);
```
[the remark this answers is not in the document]
saina-ramyar: yes,it was always in .h, similar to ilc plugin. Let me know if you want me to change it

yield_plugin/src/yield_plugin.cpp
```diff
@@ -37,72 +37,344 @@ using oss = std::ostringstream;
...
     plugin_discovery_publisher_(plugin_discovery_msg_);
     return true;
   }
 
+  void YieldPlugin::set_lanechange_status_publisher(const ros::Publisher& publisher)
+  {
+    lanechange_status_pub_ = publisher;
+  }
+
+  std::vector<lanelet::BasicPoint2d> YieldPlugin::detect_trajectories_intersection(std::vector<lanelet::BasicPoint2d> trajectory1, std::vector<lanelet::BasicPoint2d> trajectory2) const
```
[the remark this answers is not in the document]
saina-ramyar: 1- It only finds the intersection point and does not consider vehicle size/vicinity. I updated this function to detect close points as well. 
2- It does not consider time, it just finds the first overlapping position. The function `update_traj_for_cooperative_behavior` performs speed adjustments before reaching the intersection point.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -37,72 +37,344 @@ using oss = std::ostringstream;
...
+    cav_msgs::MobilityRequest incoming_request = *msg;
+    cav_msgs::LaneChangeStatus lc_status_msg;
+    if (incoming_request.strategy == "carma/cooperative-lane-change")
+    {
+      if (incoming_request.plan_type.type == cav_msgs::PlanType::CHANGE_LANE_LEFT || incoming_request.plan_type.type == cav_msgs::PlanType::CHANGE_LANE_RIGHT)
+      {
+        ROS_DEBUG_STREAM("Cooperative Lane Change Request Received");
+        lc_status_msg.status = cav_msgs::LaneChangeStatus::REQUEST_RECEIVED;
+        lc_status_msg.description = "Cooperative Lane Change Request Received";
+        if (incoming_request.header.recipient_id == config_.vehicle_id)
```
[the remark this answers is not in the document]
saina-ramyar: just an extra check to make sure it is receiving and processing a request that is meant for host vehicle.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -37,72 +37,344 @@ using oss = std::ostringstream;
...
+        double req_traj_speed = pt.get<double>("speed");
+        int start_lanelet_id = pt.get<int>("start_lanelet");
+        int end_lanelet_id = pt.get<int>("end_lanelet");
+
+
+        std::vector<lanelet::BasicPoint2d> req_traj_plan = {};
+        try
+        {
+          geometry_msgs::TransformStamped tf = tf2_buffer_.lookupTransform("map", "earth", ros::Time(0));
+          req_traj_plan = convert_eceftrajectory_to_mappoints(incoming_trajectory, tf);
```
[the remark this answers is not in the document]
saina-ramyar: moved to the constructor.

yield_plugin/src/yield_plugin.cpp
saina-ramyar: if the request is not received, the yield plugin performs its regular object avoidance logic.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -37,72 +37,381 @@ using oss = std::ostringstream;
...
+      boost::geometry::append(traj2, tpp2);
+    }
+
+    std::vector<lanelet::BasicPoint2d> intersection_points;
+    boost::geometry::intersection(traj1, traj2, intersection_points);
+
+    return intersection_points;
+  }
+
+  std::vector<lanelet::BasicPoint2d> YieldPlugin::detect_trajectories_intersection2(std::vector<lanelet::BasicPoint2d> self_trajectory, std::vector<lanelet::BasicPoint2d> incoming_trajectory) const
```
kjrush, line 100: Do we still use both variations of these functions? If not, I think we should just replace whichever one we don't use instead of having a "...2(...)" version. I'm not seeing this function get referenced, so I'm unsure what it's purpose is.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -37,72 +37,381 @@ using oss = std::ostringstream;
...
+  {
+    std::vector<lanelet::BasicPoint2d> intersection_points;
+    boost::geometry::model::linestring<lanelet::BasicPoint2d> self_traj;
+    for (auto tpp:self_trajectory)
+    {
+      boost::geometry::append(self_traj, tpp);
+    }
+    // distance to consider trajectories colliding (chosen based on lane width and vehicle size)
+    double acceptable_distance = 2.0;
```
kjrush, line 109: We should definitely parameterize this as a ROS param. The bounding box (or I suppose circle here) size will likely be an important tuning parameter for this scenario. I would probably double this as a default value as well (or more) since it needs to account for sufficient space for both vehicles and since our vehicles are longer than they are wide.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -37,72 +37,381 @@ using oss = std::ostringstream;
...
+    for (size_t i=0; i<incoming_trajectory.size(); i++)
+    {
+      double res = boost::geometry::distance(incoming_trajectory[i], self_traj);
```
kjrush, line 112: So this computing the distance between that point and the closest point on our current path? That might work, but I'm still not 100% sure it of it's soundness without considering the time-aspect of the conflict/collision warning. 

If the proposed lane change (from the mobility request) comes from a vehicle behind us (but in an adjacent lane) such that the last point of their trajectory aligns basically with the first few points of our current trajectory (but 6s after we'd be there), are we going to flag that as a conflict and slow down (thus causing an actual lane change conflict)?

yield_plugin/src/yield_plugin.cpp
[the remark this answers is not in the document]
kjrush: Sounds good, I believe we had also defined an empty-string value here as being the "broadcast recipient", so it'd be good to check for that as well as the merging vehicle may choose to broadcast this message rather than uni-cast it. Unless the CLC plugin as developed by @adev4a is guaranteed to target a specific vehicle with its mobility request, in which case no change is needed. We just need to make sure these are aligned.

yield_plugin/src/yield_plugin.cpp
kjrush: That makes sense, but in the case where we're slowing down to avoid a future collision or to make space for the merging vehicle (so object avoidance would not yet be triggered) running alternating trajectories of regular behavior and yield could produce some strange results. 

Another possible consideration here is network latency and bandwidth, a full negotiation cycle (mobility request -> mobility ack, TX and RX for both) can take several (potentially 3-5 timesteps) to complete even in nominal conditions. If we're relying on completing a handshake on lane change plan every timestep we may be setting ourselves up for issues, especially as we scale out the number of participants in the network.

yield_plugin/src/yield_plugin.cpp
saina-ramyar: this function is computing the distance between the point and the linestring, which would be the closest point in the line. And yes, I believe it would detect the conflict in the lane change scenario.
About time, since we are checking the times a few times, so it should be okay.
Also, I had the same concern when developing this, and then I realized the time is more important if it was car following where trajectories just overlap, not in the case of an incoming lane change where a point can be found as the first point of intersection. Here, it is more about speed adjustment reaching that point.

kjrush (approved):

PR #1131, opened 2021-03-03
adding reference to veh calibration

adev4a:
# PR Details
## Types of changes

## Checklist:


pure_pursuit_wrapper/launch/pure_pursuit_wrapper.launch
```diff
@@ -10,6 +10,7 @@
             <arg name="is_linear_interpolation" value="True"/>
             <arg name="publishes_for_steering_robot" value="True"/>
         </include>
+        <rosparam command="load" file="$(arg vehicle_calibration_dir)/pure_pursuit/calibration.yaml"/>
```
msmcconnell, line 13: There is a good chance these parameters will not be used by pure_pursuit because they are loaded after it is launched. 
There is also a publisher in guidance.launch that needs to be removed so this is not overwritten. Misheel can point you to it.

msmcconnell (changes requested):

pure_pursuit_wrapper/launch/pure_pursuit_wrapper.launch
```diff
@@ -9,6 +9,7 @@
         <include file="$(find pure_pursuit)/launch/pure_pursuit.launch">
             <arg name="is_linear_interpolation" value="True"/>
             <arg name="publishes_for_steering_robot" value="True"/>
+            <rosparam command="load" file="$(arg vehicle_calibration_dir)/pure_pursuit/calibration.yaml"/>
```
msmcconnell, line 12: Does this run?
rosparam is not part of the supported xml set for the include tag
http://wiki.ros.org/roslaunch/XML/include

carma/launch/guidance.launch
```diff
@@ -91,7 +91,7 @@
   <!-- Control Stack -->
   <!-- Waypoint Follower Config-->
   <node pkg="rostopic" type="rostopic" name="config_waypoint_follower_rostopic"
-        args="pub -l config/waypoint_follower autoware_config_msgs/ConfigWaypointFollower '{ header: auto, param_flag: 0, velocity: 5.0, lookahead_distance: 4.0, lookahead_ratio: 2.0, minimum_lookahead_distance: 6.0, displacement_threshold: 0.0, relative_angle_threshold: 0.0 }' "
+        args="pub -l config/waypoint_follower autoware_config_msgs/ConfigWaypointFollower '{ header: auto, param_flag: 0, velocity: 5.0, displacement_threshold: 0.0, relative_angle_threshold: 0.0 }' "
```
msmcconnell, line 94: If we take those three parameters out, is this line at all still necessary? Could we remove it?

msmcconnell (changes requested):

pure_pursuit_wrapper/launch/pure_pursuit_wrapper.launch
adev4a: Yeah, that makes sense. It's supported under the group tag. I'll move it back outside

carma/launch/guidance.launch
adev4a: Pure pursuit is still reading the param flag and velocity values from here. The rest are not being used. Do we want to move those two to the calibration folder as well?

pure_pursuit_wrapper/launch/pure_pursuit_wrapper.launch
msmcconnell: If you put in above the desired scope it will not be loaded into the namespace you want. Since all the these parameters are private they need to end up under the namespace of the pure_pursuit node.

carma/launch/guidance.launch
msmcconnell: You could move them or if these are not at all likely to change and they are the defaults just let them remain as default and remove the pub.

pure_pursuit_wrapper/launch/pure_pursuit_wrapper.launch
```diff
@@ -3,13 +3,14 @@
...
     <!-- Pure Pursuit Node -->
     <group>
-        <remap from="final_waypoints" to="carma_final_waypoints"/>
-        <remap from="/guidance/trajectory_plan" to="/guidance/mpc_follower/trajectory"/>
-        <remap from="vehicle_info/wheel_base" to="/vehicle_wheel_base"/>
-        <include file="$(find pure_pursuit)/launch/pure_pursuit.launch">
+        <node pkg="pure_pursuit" type="pure_pursuit" name="pure_pursuit" output="log">
+            <remap from="final_waypoints" to="carma_final_waypoints"/>
+            <remap from="/guidance/trajectory_plan" to="/guidance/mpc_follower/trajectory"/>
+            <remap from="vehicle_info/wheel_base" to="/vehicle_wheel_base"/>
```
msmcconnell, line 9: Remove this remapping or your wheel_base param might not load correctly

msmcconnell (changes requested):

msmcconnell (approved):

PR #1133, opened 2021-03-04
feature/gps_with_ndt_init

MishkaMN:
# PR Details
## Description

This PR is create new GPS_ONLY_WITH_NDT_INITIALIZATION mdoe in localization_manager. it would operate like the GPS only mode with the exception that it requires NDT reach the OPERATIONAL state for 5 sequential timesteps before switching to GPS. This allows the GPS offset with the map to be computed which gives the resulting GPS only pose far more accuracy.

## Related Issue
N/A as of now

## Motivation and Context
See above. 

## How Has This Been Tested?
Not tested yet

## Types of changes

## Checklist:

- [ ] I have updated the documentation accordingly. TODO

### DIFF
--- a/localization_manager/config/parameters.yaml
+++ b/localization_manager/config/parameters.yaml
@@ -24,9 +24,15 @@ auto_initialization_timeout: 30000
 # Units: ms
 gnss_only_operation_timeout: 20000
 
+# Integer: Maximum allowed number of sequential timesteps to let lidar initialize before switching to GPS only mode
+# NOTE: Only used in GNSS only with NDT initialization mode
+# Units: Int
+sequential_timesteps_until_gps_operation: 5
+
 # Integer Enum: Localization mode for the platform
 # 0 - NDT only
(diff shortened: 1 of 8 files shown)


localization_manager/config/parameters.yaml
```diff
@@ -24,9 +24,15 @@ auto_initialization_timeout: 30000
 # Units: ms
 gnss_only_operation_timeout: 20000
 
+# Integer: Maximum allowed number of sequential timesteps to let lidar initialize before switching to GPS only mode
+# NOTE: Only used in GNSS only with NDT initialization mode
+# Units: Int
+sequential_timesteps_until_gps_operation: 100
```
msmcconnell, line 30: 100 is probably a bit high. I would suggest a lower value like 5 or 10. The system rarely stays operational for that long.

msmcconnell (changes requested):

msmcconnell (approved):

Issue #1089, opened 2021-02-11
Add Yield Tactical Plugin


saina-ramyar:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

A Yield tactical plugin is developed for modifying trajectories to avoid surrounding objects.

### Expected behavior
The vehicle follows moving objects at a safe distance and decelerates to a stop for stationary obstacles without collision.
### Steps to reproduce the actual behavior

N/A

msmcconnell:
Resolved by above PRs

PR #1136, opened 2021-03-05
add communication with yield to lanechange

saina-ramyar:
# PR Details
## Description
Communication with Yield plugin is added to the un-obstructed lane change plugin, so the lane change trajectory can be modified for object avoidance purposes.

## Related Issue
#1089 

## Motivation and Context
Originally, the lane change trajectory is generated assuming there are no surrounding vehicles. With this chage, the lane change trajectory is modified to avoid and yield to objects or vehicles in its path.

## How Has This Been Tested?
unit tested and ros tested

## Types of changes

## Checklist:

### DIFF
--- a/unobstructed_lanechange/src/unobstructed_lanechange.cpp
+++ b/unobstructed_lanechange/src/unobstructed_lanechange.cpp
@@ -59,6 +59,7 @@ namespace unobstructed_lanechange
         pnh_->param<double>("curvature_calc_lookahead_count", curvature_calc_lookahead_count_, 1);
         pnh_->param<int>("downsample_ratio", downsample_ratio_, 8);
         pnh_->param<bool>("enable_object_avoidance_lc", enable_object_avoidance_lc_, false);
+        pnh_->param<double>("acceptable_time_difference_", acceptable_time_difference_, 1.0);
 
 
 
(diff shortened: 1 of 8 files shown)


unobstructed_lanechange/src/unobstructed_lanechange.cpp
```diff
@@ -611,4 +648,24 @@ namespace unobstructed_lanechange
         
     }
 
+    bool UnobstructedLaneChangePlugin::validate_yield_plan(const cav_msgs::TrajectoryPlan& yield_plan)
+    {
+        if (yield_plan.trajectory_points.size()>= 2)
+        {
+            if (yield_plan.trajectory_points[0].target_time > ros::Time::now())
```
kjrush, line 655: Is this guaranteed to be the case? I'd think yield plugins plan would start at `ros::Time::now()` when it generated the plan, which would be less than the `ros::Time::now()` value here.

unobstructed_lanechange/src/unobstructed_lanechange.cpp
```diff
@@ -59,6 +59,7 @@ namespace unobstructed_lanechange
         pnh_->param<double>("curvature_calc_lookahead_count", curvature_calc_lookahead_count_, 1);
         pnh_->param<int>("downsample_ratio", downsample_ratio_, 8);
         pnh_->param<bool>("enable_object_avoidance_lc", enable_object_avoidance_lc_, false);
+        pnh_->param<double>("acceptable_time_difference_", acceptable_time_difference_, 1.0);
```
kjrush, line 62: I think you need to have a `time_dur_ = ros::Duration(acceptable_time_difference_)` here as well, otherwise this just updates that double value member but the ros::Duration won't reflect that change.

kjrush (approved):

PR #1152, opened 2021-03-15
hotfix/fusion-calibration

MishkaMN:
Ford fusion tuning at Summit Point at 25mph


MishkaMN:
This is obsolete as further calibration has been done since this

PR #1156, opened 2021-03-17
Fix/route complete

adev4a:
# PR Details
This PR increases the stopping distance to allow for the vehicle to come to a complete stop. Since the SSC treats any speed at or below 0.4 as 0, a min crawl speed is also added in stop and wait plugin so that until the vehicle is within an acceptable distance from the end, the speed doesn't reach that low.
This implementation was integration tested.
## Types of changes

## Checklist:

### DIFF
--- a/route_following_plugin/src/route_following_plugin.cpp
+++ b/route_following_plugin/src/route_following_plugin.cpp
@@ -46,7 +46,7 @@ namespace route_following_plugin
         
         pnh_->param<double>("minimal_maneuver_duration", mvr_duration_, 16.0);
         pnh2_->param<double>("config_speed_limit",config_limit);
-        pnh_->param<double>("route_end_jerk", jerk_);
+        pnh_->param<double>("/guidance/route_end_jerk", jerk_, 1.0);
         wml_.reset(new carma_wm::WMListener());
         // set world model point form wm listener
         wm_ = wml_->getWorldModel();
--- a/route_following_plugin/src/route_following_plugin.cpp
+++ b/route_following_plugin/src/route_following_plugin.cpp
@@ -272,7 +272,7 @@ namespace route_following_plugin
             
         }
         else{
-            ROS_WARN("Failed to set the current speed limit. Valid traffic rules object could not be built.");
+            ROS_WARN(" Valid traffic rules object could not be built.");
         }
 
         if(config_limit > 0.0 && config_limit < hardcoded_max)
@@ -285,6 +285,7 @@ namespace route_following_plugin
             param_speed = hardcoded_max;
             ROS_DEBUG(" Using Hardcoded maximum");
         }
+        //If either value is 0, use the other valid limit
         if(traffic_speed <= 0.001 || param_speed <= 0.001){
             target_speed = std::max(traffic_speed, param_speed);
         }
--- a/route_following_plugin/src/route_following_plugin.cpp
+++ b/route_following_plugin/src/route_following_plugin.cpp
@@ -285,7 +285,12 @@ namespace route_following_plugin
             param_speed = hardcoded_max;
             ROS_DEBUG(" Using Hardcoded maximum");
         }
-        target_speed = std::max(traffic_speed, param_speed);
+        if(traffic_speed <= 0.001 || param_speed <= 0.001){
+            target_speed = std::max(traffic_speed, param_speed);
+        }
+        else{
+            target_speed = std::min(traffic_speed,param_speed);
+        }
         
         return target_speed;
     }
--- a/route_following_plugin/src/route_following_plugin.cpp
+++ b/route_following_plugin/src/route_following_plugin.cpp
@@ -263,7 +263,7 @@ namespace route_following_plugin
     double RouteFollowingPlugin::findSpeedLimit(const lanelet::ConstLanelet& llt)
     {
         lanelet::Optional<carma_wm::TrafficRulesConstPtr> traffic_rules = wm_->getTrafficRules();
-        double target_speed, traffic_speed =0.0, param_speed =0.0;
+        double target_speed = 0.0, traffic_speed =0.0, param_speed =0.0;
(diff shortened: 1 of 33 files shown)


stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -217,7 +226,18 @@ namespace stop_and_wait_plugin
                         throw std::invalid_argument("Ending distance is beyond known route"); 
                     }
                 }
-                else jerk_ = jerk_req;
+                else {
+                    //find distance to the end
+                    lanelet::BasicPoint2d curr_pose (pose_msg_.pose.position.x,pose_msg_.pose.position.y);
+                    double current_downtrack = wm_->routeTrackPos(curr_pose).downtrack;
+                    //stay approximately at crawl speed until within destination downtrack range (defined in route)
+                    if(start_speed <= min_crawl_speed_ && current_downtrack < ending_downtrack - 10.0){
```
msmcconnell, line 234: 10.0 should probably be a parameter or loaded directly from the route end distance parameter?

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -46,7 +46,7 @@ namespace route_following_plugin
         
         pnh_->param<double>("minimal_maneuver_duration", mvr_duration_, 16.0);
         pnh2_->param<double>("config_speed_limit",config_limit);
-        pnh_->param<double>("route_end_jerk", jerk_);
+        pnh_->param<double>("/guidance/route_end_jerk", jerk_, 1.0);
```
kjrush, line 49: This parameter loading pattern still seems a bit weird to me. Is there a reason why only this param is getting mapped into the guidance namespace? Or why we have 2 different private node handles (`pnh_` and `pnh2_`) doing this loading? 

Looking at it further, I'd wager that all of route_following_plugins params are being misloaded, they're being put into this guidance namespace instead of under the route_following_plugin's own namespace like they should be. Compare how route_following's launch file looks:
```
    <rosparam command="load" file="$(find route_following_plugin)/config/parameters.yaml"/>
    <node name="route_following_plugin" pkg="route_following_plugin" type="route_following_plugin" output="log"/>
```
to how trajectory_executor is:
```
<node pkg="trajectory_executor" type="trajectory_executor_node" name="trajectory_executor_node">
        <rosparam command="load" file="$(find trajectory_executor)/config/trajectory_executor.yaml" />
    </node>
```

With the param loading outside vs inside the node tag, I think that's probably putting all of route_following's params in the /guidance/ namespace. Let's plan to fix this on develop, but if there are any other of these params that might be giving us issues in the release let's see what we can do to fix it here with minimal breakage risk (by changing the param like you have here)

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -263,29 +263,36 @@ namespace route_following_plugin
     double RouteFollowingPlugin::findSpeedLimit(const lanelet::ConstLanelet& llt)
     {
         lanelet::Optional<carma_wm::TrafficRulesConstPtr> traffic_rules = wm_->getTrafficRules();
-        double target_speed;
+        double target_speed, traffic_speed =0.0, param_speed =0.0;
```
kjrush, line 266: Is there a reason to not initialize target_speed to 0 as well?

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -263,29 +263,36 @@ namespace route_following_plugin
...
-            }
-            
+        if(config_limit > 0.0 && config_limit < hardcoded_max)
+        {
+            param_speed = config_limit;
+            ROS_DEBUG("Using Configurable value");
+        }
+        else 
+        {
+            param_speed = hardcoded_max;
```
kjrush, line 285: I assume this is a known or intended consequence, but it does strike me as a little bit odd that if someone were to tweak the config limit here to be 0.0 that they'd end up with a limit of ~80mph if I'm understanding. That could be pretty surprising behavior for someone who thought they essentially turned off a feature by limiting it to 0.0 mph.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -263,29 +263,36 @@ namespace route_following_plugin
...
+            ROS_DEBUG(" Using Hardcoded maximum");
         }
+        //If either value is 0, use the other valid limit
+        if(traffic_speed <= 0.001 || param_speed <= 0.001){
```
kjrush, line 289: Let's use an epsilon constant here instead of literals

kjrush (changes requested):

route_following_plugin/src/route_following_plugin.cpp
adev4a: No reason here, I can add it.

route_following_plugin/src/route_following_plugin.cpp
adev4a: Yes, this is what i think is happening too. With the other params, its not giving an issue because the default values match the values in the config file.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
adev4a: Yes, this can be loaded from the param in route

route_following_plugin/src/route_following_plugin.cpp
kjrush: Okay, if the other params are fine for now we can leave it and plan to fix it on develop. Let's make an issue to track this.

adev4a: https://github.com/usdot-fhwa-stol/carma-platform/issues/1157

kjrush (approved):

kjrush (approved):

Issue #1173, opened 2021-03-22
Sudden steering event near start of stop and wait trajectory on summit point straight away


msmcconnell:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Near the start of the summit point stop and wait trajectory at the end of the straight route, the vehicles experience a sudden steering event when running on 3.5.2. This issue forced validation testing to continue on 3.5.3. 
### Carma version where this issue was discovered

3.5.2
### Expected behavior
Vehicle does not abruptly steer on a straight road. 
### Actual behavior

See description
### Steps to reproduce the actual behavior

See description. Steer is most obvious in the black pacifica.

msmcconnell:
Has not reappeared after hotfix. Closing

Issue #1169, opened 2021-03-22
Inlane cruising issues std::bad alloc at tfhrc


adev4a:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

While testing at tfhrc, it was noticed that an exception from inlane cruising "std::bad alloc" occurs at the intersection into the main parking lot.
The exception was consistently seen while trying to engage from this location. Testing was conducted and reproducible on white pacifica.

JonSmet:
Closed by PR #1189.

PR #1174, opened 2021-03-22
fix/steering_issue

MishkaMN:
# PR Details
## Description
At TFHRC, we initially noticed that there was `bad_alloc` error from stop_and_wait_plugin near the route end. This was because vehicle_state that was passed into stop_and_wait_plugin was not current_state of the vehicle (we were using position as if it was vehicle's position), which made lanelets_between function to fail as we are trying to get a vector starting from an index that is way past the end() index. . 
Therefore, we used vehicle's realtime pose. However, at Summit Point, this created additional random steering event supposedly when inlanecruising stops and stop_and_wait_plugin takes over, most likely due to that inconsistency in which pose to use in planning.
This PR is to revert that change and potentially fix the random steering event

The issue below is related to inlanecruising's bad_alloc, and this story is for stop_and_wait but they could be related
## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1173

## Motivation and Context
See above.

## How Has This Been Tested?
Not tested yet

## Types of changes

## Checklist:

### DIFF
--- a/route_following_plugin/src/route_following_plugin.cpp
+++ b/route_following_plugin/src/route_following_plugin.cpp
@@ -46,7 +46,7 @@ namespace route_following_plugin
         
         pnh_->param<double>("minimal_maneuver_duration", mvr_duration_, 16.0);
         pnh2_->param<double>("config_speed_limit",config_limit);
-        pnh_->param<double>("/guidance/route_end_jerk", jerk_, 1.0);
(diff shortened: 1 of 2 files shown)


msmcconnell:
Also related to https://github.com/usdot-fhwa-stol/carma-platform/issues/1169

MishkaMN:
Going to open a new PR to point to develop

PR #1176, opened 2021-03-24
Feature/cooperative lanechange

adev4a:
# PR Details
## Description
This PR adds the Cooperative Lanechange tactical plugin.

Cooperative Lanechange package is added, the working algorithm is taken from the design document. 
Code is tested with rostests.
## Types of changes

## Checklist:


cooperative_lanechange/include/cooperative_lanechange.h
```diff
@@ -0,0 +1,356 @@
...
+            private:
+
+            // node handles
+            std::shared_ptr<ros::CARMANodeHandle> nh_, pnh_;
+
+            ros::Publisher cooperative_lanechange_plugin_discovery_pub_;
+
+            // ros service servers
+            ros::ServiceServer trajectory_srv_;
+            ros::ServiceServer maneuver_srv_;
```
kjrush, line 256: Is this doing both maneuver and trajectory planning?

cooperative_lanechange/include/cooperative_lanechange.h
```diff
@@ -0,0 +1,356 @@
...
+
+            // ROS params
+            //Vehicle params
+            std::string sender_id_ = DEFAULT_STRING_;
+            cav_msgs::BSMCoreData bsm_core_;
+
+            //Plugin specific params
+            double desired_time_gap_ = 3.0;
+            double trajectory_time_length_ = 6;
+            std::string control_plugin_name_ = "mpc_follower";
```
kjrush, line 282: This should probably be pure_pursuit under the current control system

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -0,0 +1,867 @@
...
+#include <cav_msgs/ManeuverPlan.h>
+
+namespace cooperative_lanechange
+{
+    void CooperativeLaneChangePlugin::initialize()
+    {
+        nh_.reset(new ros::CARMANodeHandle());
+        pnh_.reset(new ros::CARMANodeHandle("~"));
+        
+        trajectory_srv_ = nh_->advertiseService("plugins/CooperativeLaneChangePlugin/plan_trajectory", &CooperativeLaneChangePlugin::plan_trajectory_cb, this);
```
kjrush, line 48: Either we need to initialize the maneuver service as well, or we should delete that private field. I'm not seeing it referenced so was it's inclusion a copy/paste error?

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -0,0 +1,867 @@
...
+        double veh2_speed = rwol[0].object.velocity.twist.linear.x;
+
+        //get current_gap
+        double current_gap = find_current_gap(veh2_lanelet_id,veh2_downtrack);
+        //get desired gap - desired time gap (default 3s)* relative velocity
+        double relative_velocity = current_speed_ - veh2_speed;
+        double desired_gap = desired_time_gap_ * relative_velocity;      
+         
+        if(current_gap < desired_gap){
+            return true;
```
kjrush, line 222: If we have enough gap we exit the trajectory plan callback without any other planning being done?  Is the expectation then that this plugin should *only* be invoked if there is a definite conflict between the host vehicle and a known other vehicle in the target lane?

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -0,0 +1,867 @@
...
+        int current_lanelet_id = current_lanelets[0].second.id();
+        if(current_lanelet_id == target_lanelet_id && current_downtrack >= target_downtrack - destination_range_){
+            cav_msgs::LaneChangeStatus lc_status_msg;
+            lc_status_msg.status = cav_msgs::LaneChangeStatus::PLANNING_SUCCESS;
+            //No description as per UI documentation
+            lanechange_status_pub_.publish(lc_status_msg);
+        }
+
+        std::vector<cav_msgs::RoadwayObstacle> rwol = wm_->getRoadwayObjects();
+        //Assuming only one object in list
```
kjrush, line 210: I'm not sure this is a valid assumption, we might have multiple objects, such as a vehicle in front of us, or a BSM from a vehicle behind us, being tracked. It might be worth thinking about how to identify that vehicle in the other lane.

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -0,0 +1,867 @@
...
+
+        return true;
+    }
+    
+    cav_msgs::MobilityRequest CooperativeLaneChangePlugin::create_mobility_request(std::vector<cav_msgs::TrajectoryPlanPoint>& trajectory_plan, cav_msgs::Maneuver maneuver){
+
+        cav_msgs::MobilityRequest request_msg;
+        cav_msgs::MobilityHeader header;
+        header.sender_id = sender_id_;
+        header.recipient_id = DEFAULT_STRING_;      //To do- find way to query recipient id
```
kjrush, line 277: We don't necessarily need to implement it this way, but in CARMA 2 this mobility request was treated as a broadcast most of the time, not a uni-cast. We used it mostly to say "Heres the path I would like to do, but is currently not possible, can anyone in the surrounding area do something (read: slow down, etc.) to allow me to complete this path. Any vehicle that didn't feel it was relevant would simply not respond, any vehicle that deemed it to be a fundamentally unsafe action would send a MobilityResponse NACK, and any vehicle that was willing and able to help would send a MobilityResponse ACK. If we got at least 1 ACK, and non NACKs we would proceed with the plan, otherwise we'd abort that operation or try a different proposed plan.

Just a thought.

If you think it makes sense to treat this as a uni-cast negotiation, then we'll need to query the world model of the vehicle we identified as being in conflict with us and get it's BSM data that way.

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -0,0 +1,867 @@
...
+        location.ecef_y = pose_msg_.pose.position.y;
+        location.ecef_z = pose_msg_.pose.position.z;
+        //Using trajectory first point time as location timestamp;
+        location.timestamp = trajectory_plan.front().target_time.toSec();
+        request_msg.location = location;
+        //Strategy params
+        //Encode JSON with Boost Property Tree
+        using boost::property_tree::ptree;
+        ptree pt;
+        pt.put("speed",maneuver.lane_change_maneuver.start_speed);
```
kjrush, line 299: This needs to be considered carefully. I was talking with @saina-ramyar about exactly how we're determining the parameters for the jerk-minimizing-trajectory used to cause the yielding vehicle to slow down. You should check with her to make sure we're keeping this two aligned in terms of how we're coordinating this gap creation.

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -0,0 +1,867 @@
...
+            //Throw exception
+            ROS_WARN("%s", ex.what());
+        }
+        request_msg.trajectory = trajectory;
+        request_msg.expiration = trajectory_plan.back().target_time.toSec();
+        
+        return request_msg;
+    }
+
+    cav_msgs::Trajectory CooperativeLaneChangePlugin::TrajectoryPlantoTrajectory(const std::vector<cav_msgs::TrajectoryPlanPoint>& traj_points, const geometry_msgs::TransformStamped& tf) const{
```
kjrush, line 324: methods should be lower case, snake_case for consistency of naming patterns.

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -0,0 +1,867 @@
...
+        cav_msgs::Trajectory traj;
+        cav_msgs::LocationECEF ecef_location = TrajectoryPointtoECEF(traj_points[0], tf);
+
+        if (traj_points.size()<2){
+            ROS_WARN("Received Trajectory Plan is too small");
+            traj.offsets = {};
+        }
+        else{
+            for (size_t i=1; i<traj_points.size(); i++){
```
kjrush, line 333: I want to double check that we're not missing an implicit requirement here: 
```
# This message uses one start location in ECEF frame and a list of offset
# points relative to X, Y and Z of start point to represent a trajectory.
# The time interval between each pair of points should be 0.1 seconds.
```
Are we guaranteeing that time interval between each of these points?

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -0,0 +1,867 @@
...
+                
+            }
+        }
+        
+        
+        traj.location = ecef_location;
+        return traj;
+    }
+
+    cav_msgs::LocationECEF CooperativeLaneChangePlugin::TrajectoryPointtoECEF(const cav_msgs::TrajectoryPlanPoint& traj_point, const geometry_msgs::TransformStamped& tf) const{
```
kjrush, line 350: Same thing RE: method naming consistency

kjrush (changes requested):
One other question, are we using the latest waypoint generation behavior here with the tweaks from vanden-plas? It looks like it RE: lookahead, etc. but if it's not too much work to double check how closely these implementations compare and update as needed that might be valuable. Or check with @saina-ramyar how the overall refactor is looking, and if that'd cover this case anyway it may be unnecessary.

cooperative_lanechange/src/cooperative_lanechange.cpp
adev4a: The planning is being done regardless of there being enough gap. It's only when the gap is large enough that the plan is passed to the response. 
There doesn't have to be a definite conflict, but this plugin is different from unobstructed lanechange it doesn't always return a trajectory plan when its called. It will look for that conflict first.
I changed the default for "is_lanechange_accepted" to true, so if there's no other vehicle found, it should work like unobstructed lc

carma/launch/guidance.launch
```diff
@@ -157,6 +157,12 @@
     <include file="$(find unobstructed_lanechange)/launch/unobstructed_lanechange.launch" />
   </group>
 
+  <!-- Cooperative lanechange -->
+  <group>
+    <remap from = "final waypoints" to="base_waypoints"/>
```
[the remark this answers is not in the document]
adev4a: Looks like this final_waypoints isn't a parameter at all. This seems to be a common mapping for all the tactical plugins

cooperative_lanechange/src/cooperative_lanechange.cpp
adev4a: Would this mean that the interval should be atleast 0.1s? or exactly 0.1s?

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -0,0 +1,897 @@
...
+            
+        }
+        else{
+            ROS_WARN_STREAM("Did not find a connected and automated vehicle roadway object");
+        }
+
+        //plan lanechange without filling in response
+        std::vector<cav_msgs::TrajectoryPlanPoint> planned_trajectory_points = plan_lanechange(req);
+        
+        ros::Time request_sent_time;
```
kjrush, line 245: Might need to initialize this somehow, not sure what the default value for ros::Time is, but I can't imagine it's relevant if we're going to use this data element later.

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -0,0 +1,897 @@
...
+            }
+            resp.maneuver_status.push_back(cav_srvs::PlanTrajectory::Response::MANEUVER_IN_PROGRESS);
+
+        }
+        else{
+            if(!negotiate){
+                request_sent_time = ros::Time::now();
+            }
+            ros::Time planning_end_time = ros::Time::now();
+            ros::Duration passed_time = planning_end_time - request_sent_time;
```
kjrush, line 284: `request_sent_time` is allocated on the stack in line 245 right? So here we're just measuring the time it takes us to navigate from line 245 to line 285? Is the intent here to track this time across multiple iterations of this callback?

kjrush: Does it make more sense to track the request state and timing as a class member variable to ensure that the data persists across multiple calls to this callback?

cooperative_lanechange/src/cooperative_lanechange.cpp
kjrush: Should be exactly 0.1s, within reasonable assumptions about constant acceleration

PR #1202, opened 2021-04-12
fix output order from getLaneletsBetween

adev4a:
# PR Details
This pr adds a fix in getLaneletsBetween (carma_wm) which sorts the order of the output from the function, according to the shortest path.
## Description
This fix is required to correctly generate lane change paths.

## Types of changes

## Checklist:

### DIFF
--- a/carma_wm/src/CARMAWorldModel.cpp
+++ b/carma_wm/src/CARMAWorldModel.cpp
@@ -280,8 +280,19 @@ std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double st
     prioritized_lanelets.pop();
     output.push_back(pair.lanelet_);
   }
+  
+  //Sort lanelets according to shortest path
+  std::vector<lanelet::ConstLanelet> sorted_output;
+  for(auto llt : route_->shortestPath()){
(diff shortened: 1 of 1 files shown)


MishkaMN (approved):
looks good to me

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -280,8 +280,19 @@ std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double st
...
+  //Sort lanelets according to shortest path
+  std::vector<lanelet::ConstLanelet> sorted_output;
+  for(auto llt : route_->shortestPath()){
+    for(int i=0; i < output.size();i++){
+      if(llt.id() == output[i].id()){
+        sorted_output.push_back(llt);
+        break;
+      }
+    }
+  }
```
msmcconnell, line 293: This loop should only run if shortest_path_only is set to true. There is no way to order the non-shortest route lanelets in the alternative case.

msmcconnell (changes requested):

msmcconnell (approved):
Looks good. Waiting on CI

PR #1209, opened 2021-04-19
Feature/yield with adjustable gap

saina-ramyar:
# PR Details
## Description
Yield plugin updated to receive adjustable inter-vehicle gap from the map and modify the trajectory accordingly.

## Related Issue
Issue #1089 

## Motivation and Context
In scenarios where a geofence message is received, carma wm updates the map for the digital gap or speed limit, etc. 
Now, Yield plugin is updated to utilize the updated information.

## How Has This Been Tested?
Unit tested.

## Types of changes

## Checklist:


yield_plugin/src/yield_plugin.cpp
```diff
@@ -417,8 +418,20 @@ namespace yield_plugin
...
-      
-      double goal_pos = x_lead - std::max(goal_velocity * gap_time, config_.x_gap); 
+      // determine the safety inter-vehicle gap based on speed
+      double safety_gap = std::max(goal_velocity * gap_time, config_.x_gap);
+      if (config_.enable_adjustable_gap)
+      {
+        // check if a digital_gap is available
+        double digital_gap = check_traj_for_digital_min_gap(original_tp);
+        // if a digital gap is available, it is replaced as safety gap 
+        if (digital_gap < std::numeric_limits<double>::max())
```
kjrush, line 428: Does it make sense to also compare to the current configured gap and use the larger of the two? Or do we want CARMA Cloud to be able to tell vehicles to shrink their headway as well? There might be benefits both ways.

yield_plugin/include/yield_plugin/yield_plugin.h
```diff
@@ -207,6 +208,14 @@ class YieldPlugin
    */
   void lookupECEFtoMapTransform();
 
+  /**
+   * \brief checks trajectory for minimum gap associated with it
+   * \param original_tp original trajectory plan
+   * \return minumum required
+   */
+  double check_traj_for_digital_min_gap(const cav_msgs::TrajectoryPlan& original_tp);
```
kjrush, line 216: I'd recommend adding a note to this comment that it returns `std::numeric_limits<double>::max()` if no such limit can be found, just so that anyone who might call this is aware later.

yield_plugin/src/yield_plugin.cpp
msmcconnell: At the moment the regulatory element is strictly a minimum gap. We would need to update the implementation to support a max gap which is not currently even part of the traffic control message spec.

saina-ramyar: @msmcconnell it already is updated to minimum gap in the new commit.

msmcconnell: Yeah what you have is fine Saina, I was mainly referring to Kyle's question on if carma-cloud would provide max headways (forced gap shrinkage). The answer is it will not.

kjrush (approved):

Issue #1211, opened 2021-04-21
Spin rate of localization_manager can introduce latency in inputs to EKF


msmcconnell:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

![image](https://user-images.githubusercontent.com/10292079/115568357-35661080-a28a-11eb-9515-af72a401443e.png)
The graph above shows a comparison of the timestamp of localization poses from various components with their receipt time by the rosbag. The split in the middle of blue plot is from an NDT to GPS transition. What is notable about the plot is that the latency of the outputs from localization_manager fluctuate rapidly at an amplitude nearly exactly period defined by the spin rate of that node which is 10Hz (0.1s per loop). 

I believe what is happening is that slight phase shifts between the localization_manager and its inputs are causing it to sometimes miss messages for one loop and not process them till the next loop.

For localization, the EKF which spins at 50Hz is likely preventing this specific problem instance from impacting the rest of the system heavily. However, I think this is likely an issue with other components in the system operating at matching frequencies with each other. It would be a good idea to either up the operational spin rate of many components and sperate it from their output frequencies or make them event driven by updating CARMANodeHandle to support ros::spin() type functionality. 

Currently marked as Sev2 since the issue in localization is essentially being resolved by EKF, however, feel this could be having a notable impact on other components that don't have this protection. 
### Carma version where this issue was discovered

develop after 3.5.3
### Expected behavior
Latency is always well below operational frequency of critical components like localization
### Actual behavior

See description
### Steps to reproduce the actual behavior

Run carma manually localize and plot the resulting data from the rosbag

msmcconnell:
This was resolved by the above PRs

PR #1216, opened 2021-04-22
Feature/update spin behavior

msmcconnell:
# PR Details
## Description
This PR refactors most nodes in carma-platform to use the new default carma node handle spin behavior defined here https://github.com/usdot-fhwa-stol/carma-utils/pull/64 . This change is meant to reduce latency in the system by preventing matching pub/sub subscriptions between nodes wherever possible. Instead nodes are updated to be event driven and use ros::Timer's to perform periodic publication in most cases. I have made an effort to leave spin callbacks in place for nodes that are not part of the control stack or its inputs and do not provide rapid data updates to downstream components. 

The previous localization latency can be seen in the linked issue. Under this new system for the same test I see the following:
image

As can be seen in the image above, the latency of the localization manager now reflects the ndt_node latency and the latency in the GPS pose has been effectively removed. 

Automated run in the Black Pacifica produced the following trajectory latency graph for the trajectory sent to pure pursuit
image

The latency is consistently between 0.2 and 0.15s with 
Mean: 0.17112644930039683
Median: 0.16291487216949463

After this change (on fusion someone else needed the pacifica)
image

After the change the result is a bit more noisy but the mean and median are noticeably improved. 
Mean: 0.09036078631320846
Median: 0.08326888084411621

## Related Issue
usdot-fhwa-stol/carma-platform#1211

## Motivation and Context
Improved vehicle control by reducing planning and feedback communications latencies.

## How Has This Been Tested?
In the black pacifica under manual drive and the ford fusion under automated control. 

## Types of changes

## Checklist:


route/src/route.cpp
```diff
@@ -63,8 +63,12 @@ namespace route {
     void Route::run()
     {
         initialize();
+
         // spin with spin_callback function from RouteGeneratorWorker
-        ros::CARMANodeHandle::setSpinCallback(std::bind(&RouteGeneratorWorker::spin_callback, &rg_worker_));
+        ros::Timer discovery_pub_timer_ = pnh_->createTimer(
```
kjrush, line 68: `discovery_pub_timer` is probably not the best name for this, since there's no plugin discovery going on here. `spin_timer`?

route_following_plugin/include/route_following_plugin.h
```diff
@@ -125,6 +125,7 @@ namespace route_following_plugin
         ros::Publisher plugin_discovery_pub_;
         ros::Subscriber pose_sub_;
         ros::Subscriber twist_sub_;
+        ros::Timer discovery_pub_timer_;
```
kjrush, line 128: See above, I assume this was copy pasted to save time

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -50,11 +50,10 @@ namespace route_following_plugin
         wml_.reset(new carma_wm::WMListener());
         // set world model point form wm listener
         wm_ = wml_->getWorldModel();
-        ros::CARMANodeHandle::setSpinCallback([this]() -> bool 
-        {
-           plugin_discovery_pub_.publish(plugin_discovery_msg_);
-            return true;
-        });
+
+        discovery_pub_timer_ = pnh_->createTimer(
```
kjrush, line 54: Likewise

route_following_plugin/include/route_following_plugin.h
msmcconnell: This is a discovery pub timer though. If you look in the cpp file you can see it only publishes the plugin discovery message.

route_following_plugin/src/route_following_plugin.cpp
msmcconnell: See above

platooning_control/include/platoon_control.hpp
```diff
@@ -86,6 +86,8 @@ namespace platoon_control
         	
         	ros::Publisher plugin_discovery_pub_;
 
+					ros::Timer discovery_pub_timer_;
```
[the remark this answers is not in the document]
msmcconnell: Done

route/src/route.cpp
msmcconnell: Done

PR #1226, opened 2021-04-28
Test/unobs lc yield

MishkaMN:
# PR Details
## Description
This PR contains only the necessary components for cooperative lanechange functionality that only relates to yield plugin. 
 

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1227

## Motivation and Context
See above.

## How Has This Been Tested?
Integration tested at Summit Point Fusion

## Types of changes

## Checklist:


inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -701,7 +704,7 @@ bool InLaneCruisingPlugin::validate_yield_plan(const cav_msgs::TrajectoryPlan& y
   {
     ROS_DEBUG_STREAM("Invalid Yield Trajectory"); 
   }
-  return false;
+  return true;
```
msmcconnell, line 707: The line right above this states this is an invalid plan and yet we are returning true???

yield_plugin/src/yield_plugin.cpp
```diff
@@ -170,22 +192,26 @@ namespace yield_plugin
...
+        ROS_DEBUG_STREAM("req_traj_speed" << req_traj_speed);
+        ROS_DEBUG_STREAM("start_lanelet_id" << start_lanelet_id);
+        ROS_DEBUG_STREAM("end_lanelet_id" << end_lanelet_id);
 
         std::vector<lanelet::BasicPoint2d> req_traj_plan = {};
 
         req_traj_plan = convert_eceftrajectory_to_mappoints(incoming_trajectory, tf_);
 
-        double req_expiration_sec = incoming_request.expiration/1000;
+        double req_expiration_sec = (double)incoming_request.expiration; //ms to second
```
msmcconnell, line 208: What part of this line is ms/seconds???

MishkaMN: Ah that is bad comment thanks

inlanecruising_plugin/src/inlanecruising_plugin.cpp
kjrush: Agreed, not sure I understand the intent of this change, it appears that in the original code this was a fallthrough case to return false, which makes sense.

mobilitypath_publisher/src/mobilitypath_publisher.cpp
```diff
@@ -131,10 +131,6 @@ namespace mobilitypath_publisher
                 traj.offsets.push_back(offset);
             }
         }
-        //according to msg standard, ecef_location is in meters
```
kjrush, line 134: Is the message standard wrong or was this comment incorrect before?

mobilitypath_visualizer/src/mobilitypath_visualizer.cpp
```diff
@@ -134,11 +134,6 @@ namespace mobilitypath_visualizer {
 
         auto curr_location_msg = msg; //variable to update on each iteration as offsets are measured since last traj point
         
-        //location message standard is specified in meters, while the logic here expects cm to work with offsets
```
kjrush, line 137: Same as above, one or the other should be updated if this is no longer needed.

yield_plugin/config/parameters.yaml
```diff
@@ -47,11 +47,11 @@ enable_cooperative_behavior: true
 always_accept_mobility_request: true
 # Acceptable number of timesteps to use the latest known mobility request before switching to yield
 # Value type: Desired
-acceptable_passed_timesteps: 3
+acceptable_passed_timesteps: 10
```
kjrush, line 50: A whole second seems like a lot, but if it's been working in the field I suppose it's best not to play around with it.

yield_plugin/include/yield_plugin/yield_plugin_node.h
```diff
@@ -77,7 +77,7 @@ class YieldPluginNode
     
     worker.set_lanechange_status_publisher(lc_status_pub);
 
-    ros::ServiceServer trajectory_srv_ = nh.advertiseService("plugins/Yieldlugin/plan_trajectory",
+    ros::ServiceServer trajectory_srv_ = nh.advertiseService("plugins/YieldPlugin/plan_trajectory",
```
kjrush, line 80: Ah, good catch. This is the kind of thing that hopefully refactoring this boilerplate into a base class can solve for us, that these service creations/etc can be automatically defined for us.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -248,11 +277,13 @@ namespace yield_plugin
     {
       if (timesteps_since_last_req_ < config_.acceptable_passed_timesteps)
       {
+        ROS_DEBUG_STREAM("Yield for CLC");
         yield_trajectory = update_traj_for_cooperative_behavior(original_trajectory, req.vehicle_state.longitudinal_vel);
-        timesteps_since_last_req_++;
+        // timesteps_since_last_req_++;
```
kjrush, line 282: So are we never expiring the yield plan if we're not incrementing this?

kjrush: The log statement here should probably also note that we haven't received an updated negotiation this timestep, so it's obvious in post-analysis.

mobilitypath_publisher/src/mobilitypath_publisher.cpp
MishkaMN: the comment and this /100 logic was incorrect before. So I updated it to use cm, so we don't have this code. Rest of the packages also use this metrics

mobilitypath_visualizer/src/mobilitypath_visualizer.cpp
MishkaMN: Same as above

yield_plugin/config/parameters.yaml
saina-ramyar: back to 3, it is working well with both values.

yield_plugin/src/yield_plugin.cpp
saina-ramyar: removed to test its impact. It is back now.

kjrush (approved):

PR #1220, opened 2021-04-27
Refactor plan_delegator and tactical plugins for plan_trajectory_cb() maneuver sequencing

JonSmet:
# PR Details
## Description

This PR refactors inlanecruising_plugin's plan_trajectory_cb() to enable it to plan a trajectory across sequential LANE_FOLLOWING maneuvers. Other tactical plugins' (unobstructed_lanechange, cooperative_lanechange, stop_and_wait_plugin) plan_trajectory_cb() have been updated to only plan a trajectory for one maneuver of the appropriate type at a time. Plan Delegator's planTrajectory() has been updated to track the current index of the maneuver plan throughout the trajectory planning process.

Additionally, Plan Delegator's planTrajectory() has been updated to assign its published trajectory's initial longitudinal velocity with the initial longitudinal velocity used by the tactical plugin that planned the first used (not expired) maneuver in the maneuver plan.

Related PR in carma-msgs: [Update PlanTrajectory.srv request to communicate maneuver index to be planned](https://github.com/usdot-fhwa-stol/carma-platform/issues/1036)

## Related Issue

[Issue 1106: Refactor plan_delegator and InLane-cruising to properly sequence maneuver/trajectory planning](https://github.com/usdot-fhwa-stol/carma-platform/issues/1106)
[Isseu 1036: Inlane cruising plugin cannot create trajectory plan for distributed maneuvers](https://github.com/usdot-fhwa-stol/carma-platform/issues/1036)

## Motivation and Context

Prior to these updates, if a Strategic Plugin published a maneuver plan with non-sequential maneuvers of the same type, (i.e. LANE_FOLLOWING, LANE_CHANGE, LANE_FOLLOWING), the appropriate tactical plugin for the non-sequential maneuver type would plan a trajectory across all of the maneuvers of that type, regardless of whether they were sequential or not. With these changes, inlanecruising_plugin can plan a trajectory across all sequential LANE_FOLLOWING maneuvers, and other tactical plugins can only plan a trajectory for a single maneuver of the appropriate type at a time.

## How Has This Been Tested?

Unit tested and integration tested on both Silver Lexus and Ford Fusion. 

## Types of changes

- [x] Defect fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that cause existing functionality to change)

## Checklist:

- [] I have added any new packages to the sonar-scanner.properties file
- [ ] My change requires a change to the documentation.
[...]

PR #1230, opened 2021-04-29
Test/unobs lc test

adev4a:
# PR Details
This PR adds changes to cooperative lane change after integration testing as well as adds the lane change logic to route following.
## Types of changes

## Checklist:


cooperative_lanechange/include/cooperative_lanechange.h
```diff
@@ -251,9 +251,21 @@ namespace cooperative_lanechange
...
+                          /**
+             * \brief Given the curvature fit, computes the curvature at the given step along the curve
+             * 
+             * \param step_along_the_curve Value in double from 0.0 (curvature start) to 1.0 (curvature end) representing where to calculate the curvature
+             * 
+             * \param fit_curve curvature fit
+             * 
+             * \return Curvature (k = 1/r, 1/meter)
+             */ 
+            double compute_curvature_at(const cooperative_lanechange::smoothing::SplineI& fit_curve, double step_along_the_curve) const;
```
kjrush, line 267: I'll write up a user story to finish the basic autonomy refactor for this plugin.

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -134,22 +137,40 @@ namespace cooperative_lanechange
...
-        double downtrack_1 = wm_->routeTrackPos(p).downtrack;
-        double remaining_downtrack = downtrack_1 - veh2_downtrack + ego_current_downtrack;
+        double veh2_lanelet_start_downtrack = wm_->routeTrackPos(shortest_path2.front().centerline2d().front()).downtrack;
+        ROS_DEBUG_STREAM("veh2_lanelet_start_downtrack:" << veh2_lanelet_start_downtrack);
+
+        double ego_current_downtrack = wm_->routeTrackPos(ego_pos).downtrack;      
+        ROS_DEBUG_STREAM("ego_current_downtrack:" << ego_current_downtrack);
+        
+        
+        double remaining_downtrack = ego_current_downtrack - veh2_downtrack - veh2_lanelet_start_downtrack;
```
kjrush, line 171: This whole block here is kind of confusing, both in it's purpose and it's naming. `remaining downtrack` implies we're using some of it up? Are both of these downtracks in lanelet-reference frame or just veh2? Do we need to be in the same lanelet as veh2 for this to hold? Some comments and clearer variable names could help out here.

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -225,25 +246,37 @@ namespace cooperative_lanechange
...
+            ROS_DEBUG_STREAM("Relative velocity:"<<relative_velocity);
             double desired_gap = desired_time_gap_ * relative_velocity;      
-            
-            if(current_gap > desired_gap){
-                negotiate = false;  //No need for negotiation
+            ROS_DEBUG_STREAM("Desired gap:"<<desired_gap);
+            if(desired_gap< 5.0){
+                desired_gap = 5.0;
             }
+            // if(current_gap > desired_gap){
```
kjrush, line 263: If we're going to leave this as-is let's write up a user story to patch both this and the above desired gap calculation.

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -260,6 +293,10 @@ namespace cooperative_lanechange
 
         //if ack mobility response, send lanechange response
         if(!negotiate || is_lanechange_accepted_){
+            ROS_DEBUG_STREAM("negotiate" << negotiate);
+            ROS_DEBUG_STREAM("negotis_lanechange_accepted_iate" << is_lanechange_accepted_);
```
kjrush, line 297: Looks like typos in the log message, let's clean some of those up a bit before release.

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -310,83 +347,109 @@ namespace cooperative_lanechange
...
+        double end_speed_floor = std::floor(maneuver.lane_change_maneuver.end_speed);
+        int end_speed_fractional = (maneuver.lane_change_maneuver.end_speed - end_speed_floor) * 10;
+
+        ROS_DEBUG_STREAM("end_speed_floor" << end_speed_floor);
+        ROS_DEBUG_STREAM("end_speed_fractional" << end_speed_fractional);
+        ROS_DEBUG_STREAM("start_lanelet_id" << maneuver.lane_change_maneuver.starting_lane_id);
+        ROS_DEBUG_STREAM("end_lanelet_id" << maneuver.lane_change_maneuver.ending_lane_id);
+
+        pt.put("s",(int)end_speed_floor); 
+        pt.put("f",end_speed_fractional);
```
kjrush, line 370: Did we shorten these just to save space in the message? Looks a lot harder to interpret what these are now, it would be nice to revert them to meaningful names if possible.

kjrush: Or maybe - if not - thats an indicator that we need to increase the max payload size for our mobility messages again.

plan_delegator/src/plan_delegator.cpp
```diff
@@ -206,7 +206,8 @@ namespace plan_delegator
         // TODO UPDATE THE FOLLOWING IF STATEMENT AFTER VANDEN-PLAS release https://github.com/usdot-fhwa-stol/carma-platform/issues/1106
         /////////
         if (latest_trajectory_plan.trajectory_points.size() > 0) {
-            if(latest_trajectory_plan.trajectory_points[0].planner_plugin_name == "InLaneCruisingPlugin"){
+            if(latest_trajectory_plan.trajectory_points[0].planner_plugin_name == "InLaneCruisingPlugin" ||
```
kjrush, line 209: Is this needed in case we're starting up and our first maneuver is a CLC maneuver?

route_following_plugin/CMakeLists.txt
```diff
@@ -91,7 +91,7 @@ install(DIRECTORY launch config
 #############
 ## Testing ##
 #############
-catkin_add_gmock(${PROJECT_NAME}-test
-  test/test_route_following_plugin.cpp
-  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
-target_link_libraries(${PROJECT_NAME}-test route_following_plugin_lib ${catkin_LIBRARIES})
+##catkin_add_gmock(${PROJECT_NAME}-test
```
kjrush, line 94: Are we just turning off tests for route following? Were these failing? If so, let's either fix them or make issues/stories to fix them, but removing the tests seems like a problem.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -46,7 +46,8 @@ namespace route_following_plugin
         
         pnh_->param<double>("minimal_maneuver_duration", mvr_duration_, 16.0);
         pnh2_->param<double>("config_speed_limit",config_limit);
-        pnh_->param<double>("/guidance/route_end_jerk", jerk_, 0.05);
+        pnh_->param<double>("buffer_time_lanechange",buffer_lanechange_time_);
+        pnh_->param<double>("/guidance/route_end_jerk", jerk_, 1.0);
```
kjrush, line 50: Not strictly a defect, but I'm not sure a private node handle should be referencing parameters starting with `/`, this makes it really hard to remap this parameter in a launch file if we need to. I can make a note of this to sweep up in another story or something.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -133,39 +142,59 @@ namespace route_following_plugin
...
+            auto following_lanelets = wm_->getRoute()->followingRelations(shortest_path[last_lanelet_index]);
 
-            resp.new_plan.maneuvers.push_back(
+            //if not already on last lanelet in path, check relation with next lanelet- follow lane change procedure if req, else lane follow
+            if(last_lanelet_index!= (shortest_path.size()-1) && identifyLaneChange(following_lanelets, shortest_path[last_lanelet_index + 1].id()))
+            {
+                //calculate constant start distance for lane change
+                double longl_travel_dist = (target_speed*(LANE_CHANGE_TIME_MAX + buffer_lanechange_time_)); //*config_param for time
+                double lane_change_start_dist;               
+                //lane change start distance needs to be constant
```
kjrush, line 154: I don't understand this comment, we update it every timestep just a few lines down? Looks like we could also just directly initialize it to current_progress on 153 instead of splitting definition and initialization across 2 lines.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -133,39 +142,59 @@ namespace route_following_plugin
...
+            {
+                resp.new_plan.maneuvers.push_back(
                 composeManeuverMessage(current_progress, end_dist,  
                                     speed_progress, target_speed, 
                                     shortest_path[last_lanelet_index].id(), time_progress));
+                
+            }
             current_progress += dist_diff;
-            time_progress = resp.new_plan.maneuvers.back().lane_following_maneuver.end_time;
+            //time_progress = resp.new_plan.maneuvers.back().lane_following_maneuver.end_time;
```
kjrush, line 181: So this time value at the end of each maneuver is now no longer updated? Are we able to plan maneuvers after a lane change?

yield_plugin/src/yield_plugin.cpp
```diff
@@ -100,26 +100,49 @@ namespace yield_plugin
   {
     cav_msgs::TrajectoryPlan trajectory_plan;
     std::vector<lanelet::BasicPoint2d> map_points;
-    lanelet::BasicPoint2d first_point;
 
-    first_point.x() = ecef_trajectory.location.ecef_x * tf.transform.translation.x;
-    first_point.y() = ecef_trajectory.location.ecef_y * tf.transform.translation.y;
+    tf2::Stamped<tf2::Transform> transform;
+    tf2::fromMsg(tf, transform);
```
kjrush, line 105: Does this include the changes from #1226 ? These look very similar to that branch. If this includes both changesets let's just merge this one after addressing the review comments from that other PR. If we're mixing and matching like that though, let's also note in the PR body which other branches/PRs it relates to and constraints on merge order/conflict resolution/etc.

kjrush (commented):
And it won't let me leave a comment, but let's delete that .carma.env.swp file, I'm not sure how it got added but it looks like it's junk binary data. Be careful when adding files to git using something like `git add *` or `git add .`, I'd recommend `git add -u .` or `git add -u *` which only stages changes to files that are already tracked in git and wont add any new files to the repo.

cooperative_lanechange/src/cooperative_lanechange.cpp
adev4a: In this part we are creating a temporary route from the veh2 to veh1 and using the shortest path from that, to calculate the downtrack distance between the two vehicles.
This downtrack distance is used as the gap between the two.
I'll add comments and rename the variables

cooperative_lanechange/src/cooperative_lanechange.cpp
adev4a: Yes that makes sense. An issue we were seeing while calculating the gap was that sometimes the map has been modified in a way that there is no path from veh2 to veh1. 
Currently we have had to modify the map in a way that lanechange is possible in both direction, so that a path exists.
Would it make sense to use the (x,y) map coordinates to calculate the gap? It should always be larger than the downtrack distance, i think

cooperative_lanechange/src/cooperative_lanechange.cpp
adev4a: The latter, the messages dont encode with the proper names in the string.
@MishkaMN

plan_delegator/src/plan_delegator.cpp
adev4a: Yes that's correct, we are not able to move from 0 without this change (if clc is the first maneuver)

route_following_plugin/src/route_following_plugin.cpp
adev4a: This comment got left behind from when lane following was being called before lane change. I'll remove it.

cooperative_lanechange/src/cooperative_lanechange.cpp
kjrush: Good to know, I'll write up a story for that then and we can consider it for later.

cooperative_lanechange/src/cooperative_lanechange.cpp
kjrush: Hmm, not really sure about using cartesian coordinates, there's some tradeoffs there too. I think ideally we'd fix what the underlying issue is with the lanelet graph, this kind of a gap should be something we're able to compute. Is the issue particular to our map at summit or is it more related to how we use lanelet2 in general?

adev4a: That makes sense, the issue was particular to the map at summit and even tfhrc

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -216,6 +220,9 @@ namespace cooperative_lanechange
         //get subject vehicle info
         lanelet::BasicPoint2d veh_pos(req.vehicle_state.X_pos_global, req.vehicle_state.Y_pos_global);
         double current_downtrack = wm_->routeTrackPos(veh_pos).downtrack;
+        if(current_downtrack < maneuver_plan[0].lane_change_maneuver.start_dist - 5.0){
```
kjrush, line 221: Should this be using that `starting_downtrack_range` parameter instead of a fixed 5.0?

kjrush (approved):

plan_delegator/src/plan_delegator.cpp
JonSmet: This TODO block is cleaned up/fixed by my plan delegator refactor PR (https://github.com/usdot-fhwa-stol/carma-platform/pull/1220), which is shown above in lines 194-198. It was integration tested for inlane-cruising, CLC, and stop and wait. So these lines (220-232) can be removed.

PR #1259, opened 2021-05-12
Fix/inlane double lane

MishkaMN:
# PR Details
## Description

This fixes logic in inlanecruising_plugin where it includes both of previous adjacent lanelets used for lanechanging in its curvature calculation logic. It only includes the lanelets in the straight part of the road (doesn't include adjacent lanelets that would have occured in the future as well).
Other changes include:
- ... TODO

This branch's changes in route_following_plugin should be ignored. However, until other PR in https://github.com/usdot-fhwa-stol/carma-platform/pull/1272 is merged, we are keeping the changes just for completeness.

Note:
Currently it is pointing to temporary branch named fix/zephyr-revert in order to better check difference as release/zephyr branch is corrupted.

## Motivation and Context

See above.

## How Has This Been Tested?

## Types of changes

## Checklist:

### DIFF
--- a/inlanecruising_plugin/src/inlanecruising_plugin.cpp
+++ b/inlanecruising_plugin/src/inlanecruising_plugin.cpp
@@ -611,9 +611,9 @@ std::vector<PointSpeedPair> InLaneCruisingPlugin::maneuvers_to_points(const std:
       while (curr_idx + 1 < lanelets.size() && 
               std::find(following_lanelets.begin(),following_lanelets.end(), lanelets[curr_idx + 1]) != following_lanelets.end())
       {
-        ROS_DEBUG_STREAM("Added following lanelet Id: " << lanelets[curr_idx].id());
-        straight_lanelets.push_back(lanelets[curr_idx + 1]);
         curr_idx++;
+        ROS_DEBUG_STREAM("Added following lanelet Id: " << lanelets[curr_idx].id());
+        straight_lanelets.push_back(lanelets[curr_idx]);
         following_lanelets = wm->getMapRoutingGraph()->following(lanelets[curr_idx]);
       }
     }
--- a/inlanecruising_plugin/src/inlanecruising_plugin.cpp
+++ b/inlanecruising_plugin/src/inlanecruising_plugin.cpp
@@ -411,20 +411,22 @@ std::vector<cav_msgs::TrajectoryPlanPoint> InLaneCruisingPlugin::compose_traject
   {
     ROS_WARN_STREAM("No trajectory points could be generated");
     return {};
-  }
+  };
 
   // Add current vehicle point to front of the trajectory
 
   nearest_pt_index = get_nearest_point_index(all_sampling_points, state);
   ROS_DEBUG_STREAM("Curvature right now: " << better_curvature[nearest_pt_index] << ", at x: " << state.X_pos_global << ", y: " << state.Y_pos_global);
 
+  int buffer_pt_index = get_nearest_point_index(all_sampling_points, ending_state_before_buffer);
+//drop buffer points here
   std::vector<lanelet::BasicPoint2d> future_basic_points(all_sampling_points.begin() + nearest_pt_index + 1,
-                                            all_sampling_points.end());  // Points in front of current vehicle position
+                                            all_sampling_points.begin()+ buffer_pt_index);  // Points in front of current vehicle position
 
   std::vector<double> future_speeds(constrained_speed_limits.begin() + nearest_pt_index + 1,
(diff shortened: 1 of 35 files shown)


msmcconnell (changes requested):

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -572,22 +572,65 @@ std::vector<PointSpeedPair> InLaneCruisingPlugin::maneuvers_to_points(const std:
...
+    }
+
     ROS_DEBUG_STREAM("Maneuver");
 
     lanelet::BasicLineString2d downsampled_centerline;
     downsampled_centerline.reserve(200);
 
-    for (auto l : lanelets)
+    // getLaneletsBetween returns shortest path that is ordered
+    // exclude lanechanges and plan for only the straight part
```
JonSmet, line 587: Can we add a comment to explain how/why a lane change may exist within inlanecruising plugin's maneuvers_to_points() function?

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -616,8 +659,40 @@ std::vector<PointSpeedPair> InLaneCruisingPlugin::maneuvers_to_points(const std:
...
+        continue;
+      }
+
+      double delta_d = lanelet::geometry::distance2d(prev_point, current_point);
+      dist_accumulator += delta_d;
+
+      double ending_downtrack = maneuvers.back().lane_following_maneuver.end_dist;
+      if(ending_downtrack + config_.buffer_ending_downtrack < wm_->getRouteEndTrackPos().downtrack){
+        ending_downtrack = ending_downtrack + config_.buffer_ending_downtrack;
+      }
```
msmcconnell, line 681: This ending downtrack change should be before this for loop as it returns a constant value. No need to have it in the loop.

JonSmet (changes requested):

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -616,8 +659,40 @@ std::vector<PointSpeedPair> InLaneCruisingPlugin::maneuvers_to_points(const std:
...
+      prev_point = current_point;
+      i++;
+    }
+    if (max_i == 0) {
+      max_i = points_and_target_speeds.size() - 1;
+    }
 
-  return points_and_target_speeds;
+    std::vector<PointSpeedPair> constrained_points(points_and_target_speeds.begin(), points_and_target_speeds.begin() + max_i);
```
msmcconnell, line 694: There needs to be a way to remove the extra points added just for supporting the spline fit after computing the curvature. This is what we discussed and you implemented in the car on 5/14 I believe.

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -572,22 +579,65 @@ std::vector<PointSpeedPair> InLaneCruisingPlugin::maneuvers_to_points(const std:
...
+      straight_lanelets.push_back(lanelets[curr_idx]);
+
+      // add all lanelets on the straight road until next lanechange
+      while (curr_idx + 1 < lanelets.size() && 
+              std::find(following_lanelets.begin(),following_lanelets.end(), lanelets[curr_idx + 1]) != following_lanelets.end())
+      {
+        curr_idx++;
+        ROS_DEBUG_STREAM("Added following lanelet Id: " << lanelets[curr_idx].id());
+        straight_lanelets.push_back(lanelets[curr_idx]);
+        following_lanelets = wm->getMapRoutingGraph()->following(lanelets[curr_idx]);
```
msmcconnell, line 626: Using the RoutingGraph does not mean you will only get lanelets on the shortest path. I think you would also need to check if the returned following lanelets were in the shortest path. No?

msmcconnell (changes requested):

inlanecruising_plugin/src/inlanecruising_plugin.cpp
MishkaMN: Thanks! I added extra comments to clarify

inlanecruising_plugin/src/inlanecruising_plugin.cpp
MishkaMN: we are dropping the extra points around the line which has comment `//drop buffer points here`

MishkaMN:
closed this PR as new one is updated here https://github.com/usdot-fhwa-stol/carma-platform/pull/1307

PR #1272, opened 2021-05-20
refactor route following

adev4a:
# PR Details
This PR refactors the route following plugin to be more simplistic is forming maneuver plans.
## Description
In this implementation the maneuver planning is done as a call back when the route is set. Following that the maneuvers are added to the response when a request is sent. 
The time, starting distance and starting speed values are updated each time the plan is added to the response.

## Motivation and Context

This change is needed to allow lane change maneuvers to be formed correctly in between lane following maneuvers.
## How Has This Been Tested?

The code was unit tested.

## Types of changes

## Checklist:

### DIFF
--- a/route_following_plugin/CMakeLists.txt
+++ b/route_following_plugin/CMakeLists.txt
@@ -91,7 +91,7 @@ install(DIRECTORY launch config
 #############
 ## Testing ##
 #############
-# catkin_add_gmock(${PROJECT_NAME}-test
-# test/test_route_following_plugin.cpp
-#  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
-# target_link_libraries(${PROJECT_NAME}-test route_following_plugin_lib ${catkin_LIBRARIES})
+catkin_add_gmock(${PROJECT_NAME}-test
+test/test_route_following_plugin.cpp
+ WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
+target_link_libraries(${PROJECT_NAME}-test route_following_plugin_lib ${catkin_LIBRARIES})
--- a/route_following_plugin/CMakeLists.txt
+++ b/route_following_plugin/CMakeLists.txt
@@ -91,7 +91,7 @@ install(DIRECTORY launch config
 #############
 ## Testing ##
 #############
-catkin_add_gmock(${PROJECT_NAME}-test
-test/test_route_following_plugin.cpp
- WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
-target_link_libraries(${PROJECT_NAME}-test route_following_plugin_lib ${catkin_LIBRARIES})
+# catkin_add_gmock(${PROJECT_NAME}-test
+# test/test_route_following_plugin.cpp
+#  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
+# target_link_libraries(${PROJECT_NAME}-test route_following_plugin_lib ${catkin_LIBRARIES})
--- a/route_following_plugin/include/route_following_plugin.h
+++ b/route_following_plugin/include/route_following_plugin.h
@@ -26,7 +26,7 @@
 #include <cav_srvs/PlanManeuvers.h>
 
 /**
- * \brief Macro definition to enable easier access to fields shared across the maneuver typees
+ * \brief Macro definition to enable easier access to fields shared across the maneuver types
  * \param mvr The maneuver object to invoke the accessors on
  * \param property The name of the field to access on the specific maneuver types. Must be shared by all extant maneuver types
  * \return Expands to an expression (in the form of chained ternary operators) that evalutes to the desired field
@@ -36,16 +36,15 @@
             ((mvr).type == cav_msgs::Maneuver::INTERSECTION_TRANSIT_RIGHT_TURN ? (mvr).intersection_transit_right_turn_maneuver.property :\
                 ((mvr).type == cav_msgs::Maneuver::INTERSECTION_TRANSIT_STRAIGHT ? (mvr).intersection_transit_straight_maneuver.property :\
                     ((mvr).type == cav_msgs::Maneuver::LANE_CHANGE ? (mvr).lane_change_maneuver.property :\
+                        (mvr).type == cav_msgs::Maneuver::LANE_FOLLOWING ? (mvr).lane_following_maneuver.property :\
                         throw new std::invalid_argument("GET_MANEUVER_PROPERTY (property) called on maneuver with invalid type id"))))))
-
-
+                        
 namespace route_following_plugin
 {
 
     class RouteFollowingPlugin
     {
-
-    public:
+        public:
 
         // constant speed limit for test
         // TODO Once world_model and vector map is ready, it should be removed
@@ -56,6 +55,7 @@ namespace route_following_plugin
          * \brief Default constructor for RouteFollowingPlugin class
          */
         RouteFollowingPlugin();
+
 
         /**
          * \brief General entry point to begin the operation of this class
@@ -80,18 +80,8 @@ namespace route_following_plugin
          * \param current_time Start time of the current maneuver passed as reference
          * \return A lane keeping maneuver message which is ready to be published
          */
-        cav_msgs::Maneuver composeManeuverMessage(double current_dist, double end_dist, double current_speed, double target_speed, int lane_id, ros::Time& current_time);
-        /**
-         * \brief Compose a stop and wait maneuver message based on input params
-         * \param current_dist Start downtrack distance of the current maneuver
-         * \param end_dist End downtrack distance of the current maneuver
-         * \param current_speed Start speed of the current maneuver
-         * \param start_lane_id Starting Lanelet ID of the current maneuver
-         * \param target_speed Target speed of the current maneuver
-         * \param current_time Start time of the current maneuver passed as reference
-         * \return A lane keeping maneuver message which is ready to be published
-         */
-        cav_msgs::Maneuver composeStopandWaitManeuverMessage(double current_dist, double end_dist, double current_speed, int start_lane_id, int target_lane_id, ros::Time& current_time, double end_time);
+        cav_msgs::Maneuver composeManeuverMessage(double current_dist, double end_dist, double current_speed, double target_speed, int lane_id, ros::Time current_time);
+
         /**
          * \brief Compose a lane change maneuver message based on input params
          * \param current_dist Start downtrack distance of the current maneuver
@@ -103,16 +93,8 @@ namespace route_following_plugin
          * \param current_time Start time of the current maneuver passed as reference
          * \return A lane keeping maneuver message which is ready to be published
          */
-        cav_msgs::Maneuver composeLaneChangeManeuverMessage(double current_dist, double end_dist, double current_speed, double target_speed, int starting_lane_id, int ending_lane_id, ros::Time& current_time);
-        /**
-         * \brief Given a prior maneuver plan, update current speed, current_progress and current lanelet id for requested plan
-         * \param maneuver A maneuver message from the end of the prior plan, used to update current request
-         * \param speed current speed value passed as reference, updated in the function
-         * \param current_progress the current_progress passed as reference, updated in the function
-         * \param lane_id The lanelet id of the current lanelet passed as reference, updated in the function
-         * \return Whether we need a lanechange to reach to the next lanelet in the shortest path
-         */
-        void updateCurrentStatus(cav_msgs::Maneuver maneuver, double& speed, double& current_progress, int& lane_id);
+        cav_msgs::Maneuver composeLaneChangeManeuverMessage(double current_dist, double end_dist, double current_speed, double target_speed, int starting_lane_id, int ending_lane_id, ros::Time current_time);
+        
         /**
          * \brief Given a LaneletRelations and ID of the next lanelet in the shortest path
          * \param relations LaneletRelations relative to the previous lanelet
@@ -121,6 +103,8 @@ namespace route_following_plugin
          */
         bool identifyLaneChange(lanelet::routing::LaneletRelations relations, int target_id);
 
+        void set_maneuver_start_dist(cav_msgs::Maneuver& maneuver, double start_dist);
+
         /**
          * \brief Service callback for arbitrator maneuver planning
          * \param req Plan maneuver request
@@ -128,12 +112,21 @@ namespace route_following_plugin
          * \return If service call successed
          */
         bool plan_maneuver_cb(cav_srvs::PlanManeuversRequest &req, cav_srvs::PlanManeuversResponse &resp);
+
+        double get_final_speed(cav_msgs::Maneuver maneuver);
+
+        double get_end_dist(cav_msgs::Maneuver maneuver);
+
         /**
          * \brief Given a Lanelet, find it's associated Speed Limit from the vector map
          * \param llt Constant Lanelet object
          * \return value of speed limit as a double, returns default value of 25 mph
          */
         double findSpeedLimit(const lanelet::ConstLanelet& llt);
+
+        void route_cb();
+
+        std::function<void()> route_set; 
 
         //Internal Variables used in unit tests
         // Current vehicle forward speed
@@ -152,7 +145,7 @@ namespace route_following_plugin
         //Tactical plugin being used for planning lane change
         std::string lane_change_plugin_ = "CooperativeLaneChangePlugin";
 
-    private:
+        private:
 
         // CARMA ROS node handles
         std::shared_ptr<ros::CARMANodeHandle> nh_, pnh_;
@@ -166,25 +159,20 @@ namespace route_following_plugin
         ros::Timer discovery_pub_timer_;
 
         // ROS service servers
-        ros::ServiceServer plan_maneuver_srv_;        
+        ros::ServiceServer plan_maneuver_srv_;  
+
+        //Queue of maneuver plans
+        std::vector<cav_msgs::Maneuver> maneuvers;
 
         // Minimal duration of maneuver, loaded from config file
         double mvr_duration_;
-        //Jerk used to come to stop at end of route
-        double jerk_ = 0.05;
-        //extra time allowed for lane changing, in order to make transition smooth
-        double buffer_lanechange_time_ = 1.0;
 
-        //lane change constant
-        static constexpr double LATERAL_ACCELERATION_LIMIT_IN_MS=2.00;
-        static const int MAX_LANE_WIDTH=3.70;
-        static constexpr double LANE_CHANGE_TIME_MAX=sqrt(2*MAX_LANE_WIDTH/LATERAL_ACCELERATION_LIMIT_IN_MS);
-
-        //Small constant to compare double with approx zero
-        const double epislon_ = 0.001;
 
         // Plugin discovery message
         cav_msgs::Plugin plugin_discovery_msg_;
+
+        //Small constant to compare double with approx zero
+        const double epislon_ = 0.001;
 
         /**
          * \brief Initialize ROS publishers, subscribers, service servers and service clients
@@ -202,9 +190,6 @@ namespace route_following_plugin
          * \param msg Latest twist message
          */
         void twist_cd(const geometry_msgs::TwistStampedConstPtr& msg);
- 
-    
-    
-    };
 
+    };
 }
--- a/route_following_plugin/include/route_following_plugin.h
+++ b/route_following_plugin/include/route_following_plugin.h
@@ -66,7 +66,7 @@ namespace route_following_plugin
          * \param path A list of lanelet with different lanelet IDs
          * \return Index of the target lanelet in the list
          */
-        int findLaneletIndexFromPath(int target_id,const lanelet::routing::LaneletPath& path);
+        int findLaneletIndexFromPath(int target_id,const lanelet::routing::LaneletPath& path) const;
 
         /**
          * \brief Compose a lane keeping maneuver message based on input params
@@ -78,7 +78,7 @@ namespace route_following_plugin
          * \param start_time Start time of the current maneuver passed as reference
          * \return A lane keeping maneuver message which is ready to be published
          */
-        cav_msgs::Maneuver composeLaneFollowingManeuverMessage(double start_dist, double end_dist, double start_speed, double target_speed, lanelet::Id lane_id, ros::Time start_time);
+        cav_msgs::Maneuver composeLaneFollowingManeuverMessage(double start_dist, double end_dist, double start_speed, double target_speed, lanelet::Id lane_id, ros::Time start_time) const;
 
         /**
          * \brief Compose a lane change maneuver message based on input params
@@ -91,7 +91,7 @@ namespace route_following_plugin
          * \param start_time Start time of the current maneuver passed as reference
          * \return A lane keeping maneuver message which is ready to be published
          */
-        cav_msgs::Maneuver composeLaneChangeManeuverMessage(double start_dist, double end_dist, double start_speed, double target_speed, lanelet::Id starting_lane_id,lanelet::Id ending_lane_id, ros::Time start_time);
+        cav_msgs::Maneuver composeLaneChangeManeuverMessage(double start_dist, double end_dist, double start_speed, double target_speed, lanelet::Id starting_lane_id,lanelet::Id ending_lane_id, ros::Time start_time) const;
         
         /**
          * \brief Given a LaneletRelations and ID of the next lanelet in the shortest path
@@ -106,20 +106,20 @@ namespace route_following_plugin
          * \param maneuver A maneuver (non-specific to type) to be performed
          * \param start_dist the starting distance that the maneuver need to be updated to
          */
-        void setManeuverStartDist(cav_msgs::Maneuver& maneuver, double start_dist);
+        void setManeuverStartDist(cav_msgs::Maneuver& maneuver, double start_dist) const;
 
         /**
          * \brief Given an array of maneuvers update the starting time for each
          * \param maneuvers An array of maneuvers (non-specific to type) 
          * \param start_time The starting time for the first maneuver in the sequence, each consequent maneuver is pushed ahead by same amount
          */
-        void updateTimeProgress(std::vector<cav_msgs::Maneuver>& maneuvers, ros::Time start_time);
+        void updateTimeProgress(std::vector<cav_msgs::Maneuver>& maneuvers, ros::Time start_time) const;
         /**
          * \brief Given an maneuver update the starting speed
          * \param maneuver maneuver to update the starting speed for
          * \param start_time The starting speed for the maneuver passed as argument
          */
-        void updateStartingSpeed(cav_msgs::Maneuver& maneuver, double start_speed);
+        void updateStartingSpeed(cav_msgs::Maneuver& maneuver, double start_speed) const;
         /**
          * \brief Service callback for arbitrator maneuver planning
          * \param req Plan maneuver request
--- a/route_following_plugin/include/route_following_plugin.h
+++ b/route_following_plugin/include/route_following_plugin.h
@@ -99,7 +99,7 @@ namespace route_following_plugin
          * \param target_id ID of the next lanelet in the shortest path
          * \return Whether we need a lanechange to reach to the next lanelet in the shortest path
          */
-        bool identifyLaneChange(lanelet::routing::LaneletRelations relations, int target_id);
+        bool isLaneChangeNeeded(lanelet::routing::LaneletRelations relations, int target_id);
         
         /**
          * \brief Set the start distance of a maneuver based on the progress along the route
@@ -124,9 +124,9 @@ namespace route_following_plugin
         bool plan_maneuver_cb(cav_srvs::PlanManeuversRequest &req, cav_srvs::PlanManeuversResponse &resp);
 
         /**
-         * \brief Given a Lanelet, find it's associated Speed Limit from the vector map
+         * \brief Given a Lanelet, find it's associated Speed Limit
          * \param llt Constant Lanelet object
-         * \return value of speed limit as a double, returns default value of 25 mph
+         * \return value of speed limit in mps
          */
         double findSpeedLimit(const lanelet::ConstLanelet& llt);
         /**
@@ -170,7 +170,7 @@ namespace route_following_plugin
 
 
         // Minimal duration of maneuver, loaded from config file
-        double mvr_duration_;
+        double min_mvr_duration_;
 
 
         // Plugin discovery message
@@ -192,7 +192,7 @@ namespace route_following_plugin
          * \brief Callback for the twist subscriber, which will store latest twist locally
          * \param msg Latest twist message
          */
-        void twist_cd(const geometry_msgs::TwistStampedConstPtr& msg);
+        void twist_cb(const geometry_msgs::TwistStampedConstPtr& msg);
 
     };
 }
--- a/route_following_plugin/include/route_following_plugin.h
+++ b/route_following_plugin/include/route_following_plugin.h
@@ -97,7 +97,7 @@ namespace route_following_plugin
          * \param target_id ID of the next lanelet in the shortest path
          * \return Whether we need a lanechange to reach to the next lanelet in the shortest path.
          */
-        bool isLaneChangeNeeded(lanelet::routing::LaneletRelations relations, int target_id);
+        bool isLaneChangeNeeded(lanelet::routing::LaneletRelations relations, lanelet::Id target_id) const;
         
         /**
          * \brief Set the start distance of a maneuver based on the progress along the route
--- a/route_following_plugin/include/route_following_plugin.h
+++ b/route_following_plugin/include/route_following_plugin.h
@@ -176,7 +176,7 @@ namespace route_following_plugin
 
 
         // Minimal duration of maneuver, loaded from config file
-        double min_mvr_duration_;
+        double min_plan_duration_;
 
 
         // Plugin discovery message
--- a/route_following_plugin/include/route_following_plugin.h
+++ b/route_following_plugin/include/route_following_plugin.h
@@ -186,6 +186,9 @@ namespace route_following_plugin
         std::string planning_strategic_plugin_ = "RouteFollowingPlugin";
         std::string lanefollow_planning_tactical_plugin_ = "InLaneCruisingPlugin"; 
 
+        //Small constant to compare doubles against
+        double epsilon_ = 0.0001;
+
         /**
          * \brief Callback for the pose subscriber, which will store latest pose locally
          * \param msg Latest pose message
--- a/route_following_plugin/include/route_following_plugin.h
+++ b/route_following_plugin/include/route_following_plugin.h
@@ -114,7 +114,12 @@ namespace route_following_plugin
          * \param start_time The starting time for the first maneuver in the sequence, each consequent maneuver is pushed ahead by same amount
          */
         void updateTimeProgress(std::vector<cav_msgs::Maneuver>& maneuvers, ros::Time start_time);
-
+        /**
+         * \brief Given an maneuver update the starting speed
+         * \param maneuver maneuver to update the starting speed for
+         * \param start_time The starting speed for the maneuver passed as argument
+         */
+        void updateStartingSpeed(cav_msgs::Maneuver& maneuver, double start_speed);
         /**
          * \brief Service callback for arbitrator maneuver planning
          * \param req Plan maneuver request
@@ -133,7 +138,7 @@ namespace route_following_plugin
          * \brief Calculate maneuver plan for remaining route. This callback is triggered when a new route has been received and processed by the world model
          *
          */
-        std::vector<cav_msgs::Maneuver> route_cb(lanelet::routing::LaneletPath route_shortest_path, lanelet::BasicPoint2d& current_loc);
+        std::vector<cav_msgs::Maneuver> route_cb(lanelet::routing::LaneletPath route_shortest_path);
 
         //Internal Variables used in unit tests
         // Current vehicle forward speed
@@ -141,6 +146,7 @@ namespace route_following_plugin
 
         // Current vehicle pose in map
         geometry_msgs::PoseStamped pose_msg_;
+        lanelet::BasicPoint2d current_loc_;
 
         // wm listener pointer and pointer to the actual wm object
         std::shared_ptr<carma_wm::WMListener> wml_;
--- a/route_following_plugin/include/route_following_plugin.h
+++ b/route_following_plugin/include/route_following_plugin.h
@@ -73,10 +73,9 @@ namespace route_following_plugin
          * \param start_speed Start speed of the current maneuver
          * \param target_speed Target speed pf the current maneuver, usually it is the lanelet speed limit
          * \param lane_id Lanelet ID of the current maneuver
-         * \param start_time Start time of the current maneuver passed as reference
          * \return A lane keeping maneuver message which is ready to be published
          */
-        cav_msgs::Maneuver composeLaneFollowingManeuverMessage(double start_dist, double end_dist, double start_speed, double target_speed, lanelet::Id lane_id, ros::Time start_time) const;
+        cav_msgs::Maneuver composeLaneFollowingManeuverMessage(double start_dist, double end_dist, double start_speed, double target_speed, lanelet::Id lane_id) const;
 
         /**
          * \brief Compose a lane change maneuver message based on input params
@@ -86,10 +85,9 @@ namespace route_following_plugin
          * \param start_lane_id Starting Lanelet ID of the current maneuver
          * \param ending_lane_id Ending Lanelet ID of the current maneuver
          * \param target_speed Target speed of the current maneuver
-         * \param start_time Start time of the current maneuver passed as reference
          * \return A lane keeping maneuver message which is ready to be published
          */
-        cav_msgs::Maneuver composeLaneChangeManeuverMessage(double start_dist, double end_dist, double start_speed, double target_speed, lanelet::Id starting_lane_id,lanelet::Id ending_lane_id, ros::Time start_time) const;
+        cav_msgs::Maneuver composeLaneChangeManeuverMessage(double start_dist, double end_dist, double start_speed, double target_speed, lanelet::Id starting_lane_id,lanelet::Id ending_lane_id) const;
         
         /**
          * \brief Given a LaneletRelations and ID of the next lanelet in the shortest path
@@ -124,7 +122,7 @@ namespace route_following_plugin
          * \param resp Plan maneuver response with a list of maneuver plan
          * \return If service call successed
          */
-        bool plan_maneuver_cb(cav_srvs::PlanManeuversRequest &req, cav_srvs::PlanManeuversResponse &resp);
+        bool planManeuverCb(cav_srvs::PlanManeuversRequest &req, cav_srvs::PlanManeuversResponse &resp);
 
         /**
          * \brief Given a Lanelet, find it's associated Speed Limit
@@ -136,7 +134,7 @@ namespace route_following_plugin
          * \brief Calculate maneuver plan for remaining route. This callback is triggered when a new route has been received and processed by the world model
          * \param route_shortest_path A list of lanelets along the shortest path of the route using which the maneuver plan is calculated.
          */
-        std::vector<cav_msgs::Maneuver> route_cb(const lanelet::routing::LaneletPath& route_shortest_path);
+        std::vector<cav_msgs::Maneuver> routeCb(const lanelet::routing::LaneletPath& route_shortest_path);
 
         //Internal Variables used in unit tests
         // Current vehicle forward speed
@@ -201,6 +199,12 @@ namespace route_following_plugin
          * \param msg Latest twist message
          */
         void twist_cb(const geometry_msgs::TwistStampedConstPtr& msg);
+
+        /**
+         * \brief returns duration as ros::Duration required to complete maneuver given its start dist, end dist, start speed and end speed
+         * \param maneuver The maneuver message to calculate duration for
+         */
+        ros::Duration maneuverDuration(cav_msgs::Maneuver &maneuver) const;
 
     };
 }
--- a/route_following_plugin/include/route_following_plugin.h
+++ b/route_following_plugin/include/route_following_plugin.h
@@ -24,6 +24,7 @@
 #include <carma_wm/WMListener.h>
 #include <carma_wm/WorldModel.h>
 #include <cav_srvs/PlanManeuvers.h>
+#include <gtest/gtest_prod.h>
 
 /**
  * \brief Macro definition to enable easier access to fields shared across the maneuver types
@@ -136,24 +137,6 @@ namespace route_following_plugin
          */
         std::vector<cav_msgs::Maneuver> routeCb(const lanelet::routing::LaneletPath& route_shortest_path);
 
-        //Internal Variables used in unit tests
-        // Current vehicle forward speed
-        double current_speed_;
-
-        // Current vehicle pose in map
-        geometry_msgs::PoseStampedConstPtr pose_msg_;
-        lanelet::BasicPoint2d current_loc_;
-
-        // wm listener pointer and pointer to the actual wm object
-        std::shared_ptr<carma_wm::WMListener> wml_;
-        carma_wm::WorldModelConstPtr wm_;
-
-        //Queue of maneuver plans
-        std::vector<cav_msgs::Maneuver> latest_maneuver_plan_;
-
-        //Tactical plugin being used for planning lane change
-        std::string lane_change_plugin_ = "CooperativeLaneChangePlugin";
-
         private:
 
         // CARMA ROS node handles
@@ -178,10 +161,29 @@ namespace route_following_plugin
         // Plugin discovery message
         cav_msgs::Plugin plugin_discovery_msg_;
 
+        // Current vehicle forward speed
+        double current_speed_;
+
+        // Current vehicle pose in map
+        geometry_msgs::PoseStampedConstPtr pose_msg_;
+        lanelet::BasicPoint2d current_loc_;
+
+        // wm listener pointer and pointer to the actual wm object
+        std::shared_ptr<carma_wm::WMListener> wml_;
+        carma_wm::WorldModelConstPtr wm_;
+
+        //Queue of maneuver plans
+        std::vector<cav_msgs::Maneuver> latest_maneuver_plan_;
+
+        //Tactical plugin being used for planning lane change
+        std::string lane_change_plugin_ = "CooperativeLaneChangePlugin";
+
         std::string planning_strategic_plugin_ = "RouteFollowingPlugin";
         std::string lanefollow_planning_tactical_plugin_ = "InLaneCruisingPlugin"; 
         //lane change tactical plugin declared as a public member
 
+        //constant small value to compare doubles approaching zero value
+        const double epsilon_ = 0.00001;
 
         /**
          * \brief Initialize ROS publishers, subscribers, service servers and service clients
@@ -205,6 +207,14 @@ namespace route_following_plugin
          * \param maneuver The maneuver message to calculate duration for
          */
         ros::Duration maneuverDuration(cav_msgs::Maneuver &maneuver) const;
+
+        //Unit Tests
+        FRIEND_TEST(RouteFollowingPluginTest, testFindLaneletIndexFromPath);
+        FRIEND_TEST(RouteFollowingPluginTest, testComposeManeuverMessage);
+        FRIEND_TEST(RouteFollowingPluginTest, testIdentifyLaneChange);
+        FRIEND_TEST(RouteFollowingPlugin, TestAssociateSpeedLimit);
+        FRIEND_TEST(RouteFollowingPlugin, TestAssociateSpeedLimitusingosm);
+        FRIEND_TEST(RouteFollowingPlugin, TestHelperfunctions);
 
     };
 }
--- a/route_following_plugin/include/route_following_plugin.h
+++ b/route_following_plugin/include/route_following_plugin.h
@@ -60,6 +60,13 @@ namespace route_following_plugin
         void run();
 
         /**
+         * \brief Initialize ROS publishers, subscribers, service servers and service clients
+         */
+        void initialize();
+
+        private:
+
+        /**
          * \brief Given a LaneletPath object, find index of the lanelet which has target_id as its lanelet ID
          * \param target_id The laenlet ID this function is looking for
          * \param path A list of lanelet with different lanelet IDs
@@ -137,8 +144,6 @@ namespace route_following_plugin
          */
         std::vector<cav_msgs::Maneuver> routeCb(const lanelet::routing::LaneletPath& route_shortest_path);
 
-        private:
-
         // CARMA ROS node handles
         std::shared_ptr<ros::CARMANodeHandle> nh_, pnh_;
 
@@ -180,15 +185,6 @@ namespace route_following_plugin
 
         std::string planning_strategic_plugin_ = "RouteFollowingPlugin";
         std::string lanefollow_planning_tactical_plugin_ = "InLaneCruisingPlugin"; 
-        //lane change tactical plugin declared as a public member
-
-        //constant small value to compare doubles approaching zero value
-        const double epsilon_ = 0.00001;
-
-        /**
-         * \brief Initialize ROS publishers, subscribers, service servers and service clients
-         */
-        void initialize();
 
         /**
          * \brief Callback for the pose subscriber, which will store latest pose locally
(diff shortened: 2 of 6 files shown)


route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -43,35 +42,47 @@ namespace route_following_plugin
...
         wml_.reset(new carma_wm::WMListener());
         // set world model point form wm listener
         wm_ = wml_->getWorldModel();
-
+        //set a route callback to update route and calculate maneuver
+        wml_->setRouteCallback(route_set);
+        if(route_set)    
+        {
+            route_cb();
+        }
```
msmcconnell, line 54: I know what you are trying to do here, but this isn't the correct syntax. 
setRouteCallback() takes in a function which will be triggered when a new route has been received and processed by the world model. The way this is currently written route_set will be called every time a route is received which is probably an exception since its not a populated function object. What you are really looking for would be something like
 ```
 // Using std::bind to generate an std::function object from a function pointer
  wml_->setRouteCallback(std::bind(&RouteFollowingPlugin::route_cb, this));
 
 // OR
 
 // Using C++ lambda syntax
 wml_->setRouteCallback([this]() { route_cb(); });
 ```

msmcconnell (changes requested):

route_following_plugin/include/route_following_plugin.h
```diff
@@ -57,6 +56,7 @@ namespace route_following_plugin
          */
```
msmcconnell, line 56: TWENTY_FIVE_MPH_IN_MS and FIFTEEN_MPH_IN_MS seem like they are no longer needed

route_following_plugin/include/route_following_plugin.h
```diff
@@ -80,18 +80,8 @@ namespace route_following_plugin
...
-         * \param current_dist Start downtrack distance of the current maneuver
-         * \param end_dist End downtrack distance of the current maneuver
-         * \param current_speed Start speed of the current maneuver
-         * \param start_lane_id Starting Lanelet ID of the current maneuver
-         * \param target_speed Target speed of the current maneuver
-         * \param current_time Start time of the current maneuver passed as reference
-         * \return A lane keeping maneuver message which is ready to be published
-         */
-        cav_msgs::Maneuver composeStopandWaitManeuverMessage(double current_dist, double end_dist, double current_speed, int start_lane_id, int target_lane_id, ros::Time& current_time, double end_time);
+        cav_msgs::Maneuver composeManeuverMessage(double current_dist, double end_dist, double current_speed, double target_speed, int lane_id, ros::Time current_time);
```
msmcconnell, line 83: If this method is just for lane following maneuvers as described by the comments it should be named as such.

msmcconnell: Having current_ in the parameter names is also a bit unclear as its not really the current state of the vehicle so much as the desired start state of the maneuver.

route_following_plugin/include/route_following_plugin.h
```diff
@@ -121,20 +103,31 @@ namespace route_following_plugin
          */
         bool identifyLaneChange(lanelet::routing::LaneletRelations relations, int target_id);
 
+        void set_maneuver_start_dist(cav_msgs::Maneuver& maneuver, double start_dist);
```
msmcconnell, line 106: comments needed. Same for get_final_speed and get_end_distance.

route_following_plugin/include/route_following_plugin.h
```diff
@@ -121,20 +103,31 @@ namespace route_following_plugin
...
+        double get_end_dist(cav_msgs::Maneuver maneuver);
+
         /**
          * \brief Given a Lanelet, find it's associated Speed Limit from the vector map
          * \param llt Constant Lanelet object
          * \return value of speed limit as a double, returns default value of 25 mph
          */
         double findSpeedLimit(const lanelet::ConstLanelet& llt);
 
+        void route_cb();
```
msmcconnell, line 127: comments needed. Also it seems like there is a mixing of camelCase and snake_case method names in this file which is a purely stylistic issue but it improves readability if everything in the same file follows similar formatting.

route_following_plugin/include/route_following_plugin.h
```diff
@@ -121,20 +103,31 @@ namespace route_following_plugin
...
+
+        std::function<void()> route_set;
```
msmcconnell, line 129: Per my first comment this variable is not needed.

route_following_plugin/include/route_following_plugin.h
```diff
@@ -152,7 +145,7 @@ namespace route_following_plugin
         //Tactical plugin being used for planning lane change
```
msmcconnell, line 145: The config_limit variable above is really not needed at all. The usage of the config_limit and hardcoded max speed limit is handled inside carma_wm not need to cache it separately.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -43,35 +42,47 @@ namespace route_following_plugin
...
+    {
+        current_speed_ = msg->twist.linear.x;
+    }
+
+    void RouteFollowingPlugin::route_cb(){
+        //This function calculates the maneuver every time a route is set
+        //Go through route - identify lane changes and fill in the spaces with lane following
         lanelet::BasicPoint2d current_loc(pose_msg_.pose.position.x, pose_msg_.pose.position.y);
-        auto current_lanelets = lanelet::geometry::findNearest(wm_->getMap()->laneletLayer, current_loc, 10);       
+        auto current_lanelets = lanelet::geometry::findNearest(wm_->getMap()->laneletLayer, current_loc, 10);
```
msmcconnell, line 80: nearest_lanelets would be a better name for this vector since only a few of them may actually be a lanelet we are on top of.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -92,139 +103,105 @@ namespace route_following_plugin
         if(last_lanelet_index == -1)
```
msmcconnell, line 103: A comment describing what the above loop is meant to do would be good to include

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -326,25 +281,6 @@ namespace route_following_plugin
...
-            lane_id =  stoi(maneuver.lane_following_maneuver.lane_id);
-        }
-        else{
-            speed = GET_MANEUVER_PROPERTY(maneuver,end_speed);
-            current_progress =GET_MANEUVER_PROPERTY(maneuver,end_dist);
-            lane_id = stoi(GET_MANEUVER_PROPERTY(maneuver,ending_lane_id));
-        }
-
-    }
-
```
msmcconnell, line 347: The findSpeedLimit method below can be substantially simplified. 
1. Get the traffic_rules optional
2. Check if it is valid. If not throw an exception as something has gone very very wrong
3. If the traffic rules object does exist all you need to do is call speedLimit(llt); Internally the carma_wm handles the config_limit and hard coded max speed limit constraints. You do not need to check them yourself

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -92,139 +103,105 @@ namespace route_following_plugin
...
         }
-        double current_progress = wm_->routeTrackPos(current_loc).downtrack;
-        double speed_progress = current_speed_;
-        ros::Time time_progress = ros::Time::now();
-        double target_speed=findSpeedLimit(current_lanelet);   //get Speed Limit
 
-        double total_maneuver_length = current_progress + mvr_duration_ * target_speed;
-        double route_length=  wm_->getRouteEndTrackPos().downtrack; 
-        total_maneuver_length = std::min(total_maneuver_length, route_length);
+        double route_length = wm_->getRouteEndTrackPos().downtrack;;
```
msmcconnell, line 108: minor thing but you have duplicate semicolons here

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -43,35 +42,47 @@ namespace route_following_plugin
...
     {
         initialize();
         ros::CARMANodeHandle::spin();
     }
-    bool RouteFollowingPlugin::plan_maneuver_cb(cav_srvs::PlanManeuversRequest &req, cav_srvs::PlanManeuversResponse &resp)
-    {        
+
+    void RouteFollowingPlugin::pose_cb(const geometry_msgs::PoseStampedConstPtr& msg)
+    {
+        pose_msg_ = geometry_msgs::PoseStamped(*msg.get());
```
msmcconnell, line 69: Probably out of scope for this PR, but I find it problematic that the strategic plugins subscribe to pose and current speed. Similar to the trajectory planners the arbitrator should provide a consistent state to all strategic plugins such that when the arbitrator compares two plans they both start from the same downtrack distance. Under the current system there can be mismatch between starting states. Could potentially cause issues with switching over to something like platooning.  
@kjrush can prioritize such a change.

route_following_plugin/CMakeLists.txt
```diff
@@ -91,7 +91,7 @@ install(DIRECTORY launch config
...
 ## Testing ##
 #############
-catkin_add_gmock(${PROJECT_NAME}-test
-test/test_route_following_plugin.cpp
- WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
-target_link_libraries(${PROJECT_NAME}-test route_following_plugin_lib ${catkin_LIBRARIES})
+# catkin_add_gmock(${PROJECT_NAME}-test
+# test/test_route_following_plugin.cpp
+#  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
+# target_link_libraries(${PROJECT_NAME}-test route_following_plugin_lib ${catkin_LIBRARIES})
```
msmcconnell, line 97: Don't forget to reenable testing and run the tests before we merge

msmcconnell (changes requested):
Some initial comments. I haven't gone through the main logic changes yet

route_following_plugin/include/route_following_plugin.h
```diff
@@ -121,20 +103,31 @@ namespace route_following_plugin
...
+
         /**
          * \brief Service callback for arbitrator maneuver planning
          * \param req Plan maneuver request
          * \param resp Plan maneuver response with a list of maneuver plan
          * \return If service call successed
          */
         bool plan_maneuver_cb(cav_srvs::PlanManeuversRequest &req, cav_srvs::PlanManeuversResponse &resp);
+
+        double get_final_speed(cav_msgs::Maneuver maneuver);
```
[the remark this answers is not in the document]
kjrush: Let's also make sure to doc-comment these new members before we close this out.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -92,139 +103,105 @@ namespace route_following_plugin
...
-            ++last_lanelet_index;
+        //update start distance of first maneuver and time for all maneuvers
+           set_maneuver_start_dist(maneuvers.front(), current_progress);
 
+        //Return the set of maneuvers which intersect those 15s
+        int i=0;
+        while(current_progress < total_maneuver_length && i < maneuvers.size()){
+           double maneuver_end_dist = GET_MANEUVER_PROPERTY(maneuvers[i],end_dist);
+           current_progress += maneuver_end_dist;
+           //Update time progress for maneuvers
```
kjrush, line 159: We're updating the distances, but is this time value being updated elsewhere?

route_following_plugin/src/route_following_plugin.cpp
kjrush: That's interesting, I actually hadn't considered that possibility, that a new state update could occur in between two plugins generating plans, but that kind of a race condition does make sense. I'd hope the time difference would not be too large (given that the planning itself would hopefully not take long) but I don't have data to confirm or deny that, and if the difference isn't too large the state delta shouldn't be large either. But I can definitely see how that might impact us in edge cases (first plan is at very end of lanelet 1, then we get an update, second plan is entirely within lanelet 2). And if it the time difference is large that's a whole problem unto itself.

I'll have to think on that, but that's a good note. The fix here would just be to push the same kind of state info into PlanManeuver.srv like we have for trajectories (as you mentioned) and then remove those subscribers from the strategic plugin. I do like that flow.

route_following_plugin/CMakeLists.txt
kjrush: If anything I'd argue we'd want to leave testing on all the time no? The circle CI warning that tests failed (or failed to build) is another reminder to not merge things until all the issues are fixed, and locally we should be able to do some kind of skip tests command line option right?

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -326,25 +281,6 @@ namespace route_following_plugin
...
-            speed = GET_MANEUVER_PROPERTY(maneuver,end_speed);
-            current_progress =GET_MANEUVER_PROPERTY(maneuver,end_dist);
-            lane_id = stoi(GET_MANEUVER_PROPERTY(maneuver,ending_lane_id));
-        }
-
-    }
-
     double RouteFollowingPlugin::findSpeedLimit(const lanelet::ConstLanelet& llt)
     {
         lanelet::Optional<carma_wm::TrafficRulesConstPtr> traffic_rules = wm_->getTrafficRules();
```
TonysCousin, line 286: Several comments on findSpeedLimit() since Github won't let me write in there...
On next line (287), prefer to declare one variable per line. There's no need to save space like this, and it makes the declarations harder to see.  Also, spaces after the = signs (also on line 288 & 292)!
Line 302 & 307 DEBUG statements should indicate what kind of values, and even what method is writing the message. Reading the log file isn't going to be very meaningful with these vague phrases.
Readability - it's good to have a blank line prior to introducing a new logic block, esp when it begins with a comment (as most logic blocks should), like at line 309.
If block at line 310 testing for 0 speeds - since either of these conditions is possible, it is also possible that both will happen simultaneously, yet we don't handle that condition.  I'd like to see an embedded if test here that says if they're both 0 then write a warning message.
Space after comma on line 314.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -264,7 +242,7 @@ namespace route_following_plugin
         return maneuver_msg;
```
TonysCousin, line 242: On line 233 above there is a literal constant 0.00001.  Its value seems arbitrary, as is that of EPSILON.  Can we use the same EPSILON in all of these places?  There is another literal a little further down also.

TonysCousin (changes requested):
Anish, I'm not done yet, but that's all for today.  Have fun with these, and feel free to ask questions.  I hope this gives you a feel for the kinds of things we need people to be thinking about.

route_following_plugin/CMakeLists.txt
msmcconnell: Generally agree, but Anish put up this PR before updating unit tests, so its possible they were not compiling as this point.

route_following_plugin/src/route_following_plugin.cpp
msmcconnell: Sounds good

route_following_plugin/include/route_following_plugin.h
```diff
@@ -166,26 +159,21 @@ namespace route_following_plugin
         ros::Timer discovery_pub_timer_;
 
         // ROS service servers
-        ros::ServiceServer plan_maneuver_srv_;        
+        ros::ServiceServer plan_maneuver_srv_;  
+
+        //Queue of maneuver plans
```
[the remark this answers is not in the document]
msmcconnell: I think there might be a risk of using a strict queue since you need to walk the queue at points before a maneuver is completed, but I haven't looked at that logic yet.

route_following_plugin/src/route_following_plugin.cpp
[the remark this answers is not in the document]
msmcconnell: I'm actually a bit unclear how this compiles. The input is a boost:shared_ptr which can be dereferenced with either .get() or *. The use of both here would imply a double dereference which I think would be null.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -291,29 +269,6 @@ namespace route_following_plugin
...
-            end_time = mvr_duration_;
-        } 
-        maneuver_msg.stop_and_wait_maneuver.end_time = current_time + ros::Duration(end_time);
-        current_time = maneuver_msg.stop_and_wait_maneuver.end_time;
-        ROS_INFO_STREAM("Creating stop and wait start dist:"<<current_dist<<" end dist:"<<end_dist);
-        ROS_INFO_STREAM("Duration:"<<maneuver_msg.stop_and_wait_maneuver.end_time.toSec() - maneuver_msg.stop_and_wait_maneuver.start_time.toSec());
-        return maneuver_msg;
-    }
     bool RouteFollowingPlugin::identifyLaneChange(lanelet::routing::LaneletRelations relations, int target_id)
     {
```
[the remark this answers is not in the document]
msmcconnell: Just a caution here on John's comment. There are multiple valid ways to extract lane changes from the lanelet2 route and routinggraph objects. If you use the left/right relations calls for detecting lane changes its important to distinguish between the left relation which is routable and the left_adjacent relation which is not routable. Additionally, left and right relations are route methods not shortest path methods, so there is a possibility they could return lanelets not on the shortest path.

route_following_plugin/include/route_following_plugin.h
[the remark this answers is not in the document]
adev4a: These functions were made public so they can be individually unit tested

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -43,35 +42,47 @@ namespace route_following_plugin
...
         if(current_lanelets.size() == 0)
         {
             ROS_WARN_STREAM("Cannot find any lanelet in map!");
-            return true;
         }
-        
+
         auto shortest_path = wm_->getRoute()->shortestPath();
 
         lanelet::ConstLanelet current_lanelet;
```
TonysCousin, line 88: General advice (e.g. for next line (89)).  It is a common mistake to name a variable as "last_...", when what is really meant is "previous_...".  The word "last" means the final one!  That is seldom the intended meaning in code, especially where loops are concerned.  But occasionally it is, so casual use of this term can occasionally be really confusing.  I believe in this case the intended meaning is prev_lanelet_index.

route_following_plugin/src/route_following_plugin.cpp
TonysCousin: Agree with Mike.  Also a comment quickly explaining what the 10 argument means (number of points? search radius?), since I spent several minutes looking for the proper lanelet header file that declares this function (there are at least three LaneletMap.h files!!!).  And since it is number of neighbors, why is 10 the desired number?

msmcconnell: @TonysCousin findNearest is a k-nearest-neighbors method so 10 means the number of nearest lanelets to return. They are returned in distance order

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -92,139 +103,105 @@ namespace route_following_plugin
...
+            default :
+                ROS_DEBUG_STREAM("Invalid maneuver type");
+        }
+
     }
-    void RouteFollowingPlugin::twist_cd(const geometry_msgs::TwistStampedConstPtr& msg)
-    {
-        current_speed_ = msg->twist.linear.x;
+    
+    double RouteFollowingPlugin::get_final_speed(cav_msgs::Maneuver maneuver){
```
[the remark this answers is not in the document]
adev4a: This one is a member function because not all maneuvers have target_speed as a parameter, for stop and wait the end speed is always 0.0, its not a part of the message parameters

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -43,188 +42,187 @@ namespace route_following_plugin
...
-        pnh2_->param<std::string>("guidance/lane_change_plugin", lane_change_plugin_);
+        pnh_->param<std::string>("lane_change_plugin", lane_change_plugin_);
+
         wml_.reset(new carma_wm::WMListener());
         // set world model point form wm listener
         wm_ = wml_->getWorldModel();
-
+        //set a route callback to update route and calculate maneuver
+        lanelet::BasicPoint2d current_loc(pose_msg_.pose.position.x, pose_msg_.pose.position.y);
+        wml_->setRouteCallback([this, &current_loc]() { this->latest_maneuver_plan_ = route_cb(wm_->getRoute()->shortestPath(), current_loc);});
```
msmcconnell, line 54: Don't try to capture a local variable which will go out of scope like this. The current_location variable, the way this is written, will never change even if rerouting occurs. Instead you want to extract it from "this" inside the lambda similar to what you did for latest_maneuver_plan_

route_following_plugin/include/route_following_plugin.h
```diff
@@ -121,20 +103,31 @@ namespace route_following_plugin
...
          * \param llt Constant Lanelet object
          * \return value of speed limit as a double, returns default value of 25 mph
          */
         double findSpeedLimit(const lanelet::ConstLanelet& llt);
 
+        void route_cb();
+
+        std::function<void()> route_set; 
+
         //Internal Variables used in unit tests
```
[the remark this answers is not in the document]
adev4a: These variables are being used in the unit tests. They need to be assigned values because some of the functions use their values directly

route_following_plugin/include/route_following_plugin.h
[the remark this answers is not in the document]
adev4a: The stop and wait maneuver tactical plugin requires some refactoring and hasn't been included in this implementation

route_following_plugin/include/route_following_plugin.h
```diff
@@ -36,16 +36,15 @@
...
 {
 
     class RouteFollowingPlugin
     {
-
-    public:
+        public:
 
         // constant speed limit for test
         // TODO Once world_model and vector map is ready, it should be removed
```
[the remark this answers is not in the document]
TonysCousin: Good, but there's still a comment describing a constant speed limit that no longer exists :)

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,210 @@ namespace route_following_plugin
...
-                    current_lanelet = shortest_path[last_lanelet_index];
+                    current_lanelet_index = potential_index;
                     break;
                 }
             }
         }
-        if(last_lanelet_index == -1)
+        if(current_lanelet_index == -1)
         {
             ROS_ERROR_STREAM("Current position is not on the shortest path! Returning an empty maneuver");
```
TonysCousin, line 112: I think the design needs to be reconsidered for situations where vehicle is off the shortest path.  In the meantime, the quick solution is to change this error statement to a warning and remove the second sentence (since there is no return statement here), and just let the method continue building a maneuvers list as it does.  This will allow steering control to bring things back to the shortest path eventually.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,210 @@ namespace route_following_plugin
...
 
-            if(current_progress >= total_maneuver_length || last_lanelet_index == shortest_path.size() - 1)
-            {
-                break;
-            }
+    bool RouteFollowingPlugin::plan_maneuver_cb(cav_srvs::PlanManeuversRequest &req, cav_srvs::PlanManeuversResponse &resp)
+    { 
+        if(latest_maneuver_plan_.empty()){
+            ROS_ERROR_STREAM("A maneuver plan has not been generated");
+            return true;
```
TonysCousin, line 149: I believe this should return false to indicate unable to make a plan.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,210 @@ namespace route_following_plugin
...
     }
-    bool RouteFollowingPlugin::plan_maneuver_cb(cav_srvs::PlanManeuversRequest &req, cav_srvs::PlanManeuversResponse &resp)
-    {        
-        lanelet::BasicPoint2d current_loc(pose_msg_.pose.position.x, pose_msg_.pose.position.y);
-        auto current_lanelets = lanelet::geometry::findNearest(wm_->getMap()->laneletLayer, current_loc, 10);       
-        if(current_lanelets.size() == 0)
+
+    void RouteFollowingPlugin::pose_cb(const geometry_msgs::PoseStampedConstPtr& msg)
+    {
+        pose_msg_ = geometry_msgs::PoseStamped(*msg.get());
```
TonysCousin, line 76: It would be cleaner to have this callback populate a member variable called current_loc_, rather than to have multiple other methods creating their local variables current_loc and populating them repeatedly.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,210 @@ namespace route_following_plugin
...
+        if(latest_maneuver_plan_.empty()){
+            ROS_ERROR_STREAM("A maneuver plan has not been generated");
+            return true;
+        }
 
-            ++last_lanelet_index;
+        lanelet::BasicPoint2d current_loc(pose_msg_.pose.position.x, pose_msg_.pose.position.y);
+        
+        double current_downtrack = wm_->routeTrackPos(current_loc).downtrack;
+        double speed_progress = current_speed_;
```
TonysCousin, line 155: Neither of these variables is used in this method.  Remove this line.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,210 @@ namespace route_following_plugin
...
-        pose_msg_ = geometry_msgs::PoseStamped(*msg.get());
+
+    void RouteFollowingPlugin::updateTimeProgress(std::vector<cav_msgs::Maneuver>& maneuvers, ros::Time start_time){
+        ros::Time time_progress = start_time;
+        ros::Time prev_time = time_progress;
+
+        for(auto maneuver : maneuvers){
+            ros::Duration maneuver_duration =GET_MANEUVER_PROPERTY(maneuver,end_time) - GET_MANEUVER_PROPERTY(maneuver, start_time);
+            time_progress += maneuver_duration;
+            switch(maneuver.type){
```
TonysCousin, line 189: If the class structure was refactored to put these time params in a base class, then this whole switch statement could go away and you could just say maneuver.start_time = ..., for example.  Would simplify other areas of the code as well.  Something for a future enhancement.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,210 @@ namespace route_following_plugin
...
+        }
 
-            ++last_lanelet_index;
+        lanelet::BasicPoint2d current_loc(pose_msg_.pose.position.x, pose_msg_.pose.position.y);
+        
+        double current_downtrack = wm_->routeTrackPos(current_loc).downtrack;
+        double speed_progress = current_speed_;
+        
+        //update start distance of first maneuver and time for all maneuvers
+        setManeuverStartDist(latest_maneuver_plan_.front(), current_downtrack);
```
TonysCousin, line 158: I think both of these calls are in error (the logic in the methods, actually).  They assume vehicle is at the beginning of the route.  It may be actually several maneuvers downtrack at this point.

TonysCousin (changes requested):
Done for today.  I'll do a re-review after you have addressed everything currently open.

route_following_plugin/include/route_following_plugin.h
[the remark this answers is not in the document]
TonysCousin: route_cb now has a partial header description, but it needs to include description of params & return value.

route_following_plugin/include/route_following_plugin.h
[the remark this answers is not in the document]
TonysCousin: Hmm, okay.  Can't unit tests be constructed to see private (or at least protected) variables?  If they inherit the class under test they should have visibility into the protected members, which would be one good reason to use protected instead of private, as a rule.  If fundamental restructuring of how unit tests are being done is what I'm asking, then never mind.  Not worth it at this time.

route_following_plugin/src/route_following_plugin.cpp
TonysCousin: Good explanation, Mike, but it's nice to have it written in the code so I don't have to go dig for 10 minutes next year to refresh my memory.

route_following_plugin/src/route_following_plugin.cpp
TonysCousin: speed_progress is now used (in call to updateStartingSpeed(), but there's no need for it.  Just use current_speed_ in that call, and this local variable can be eliminated.

route_following_plugin/src/route_following_plugin.cpp
TonysCousin: I still don't get what is being attempted in these two methods.  Remember that the current_downtrack and current time is generally well into the route, probably several maneuvers have been completed already.  These methods are assuming we are at beginning of route.  I think the first step we need in this method is to find which maneuver we are currently in. These methods are changing the master route list!  They should not do that.  It should be passed in as a const.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -138,44 +136,48 @@ namespace route_following_plugin
...
+        double current_downtrack = wm_->routeTrackPos(current_loc_).downtrack;
         double speed_progress = current_speed_;
         
         //update start distance of first maneuver and time for all maneuvers
         setManeuverStartDist(latest_maneuver_plan_.front(), current_downtrack);
         //Update time progress for maneuvers
         updateTimeProgress(latest_maneuver_plan_, ros::Time::now());
-        //update starting speed
         
         //Return the set of maneuvers which intersect those 15s
```
TonysCousin, line 158: I don't like this comment.  First, 15s should be a parameter, not hard-coded (or hard-commented).  Second, it assumes a previous conversation about some 15s time frame, which is not in the above context, so I don't know what it's talking about.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -138,44 +136,48 @@ namespace route_following_plugin
...
         double planned_time = 0.0;
         while(planned_time < min_mvr_duration_ && i < latest_maneuver_plan_.size()){
+            //Ignore plans for distance already covered
+            if(GET_MANEUVER_PROPERTY(latest_maneuver_plan_[i], end_dist) < current_downtrack){
+                ++i;
+                continue;
+            }
             double maneuver_start_time = GET_MANEUVER_PROPERTY(latest_maneuver_plan_[i], start_time).toSec();
             double maneuver_end_time = GET_MANEUVER_PROPERTY(latest_maneuver_plan_[i],end_time).toSec();
             planned_time += maneuver_end_time - maneuver_start_time;
```
TonysCousin, line 169: I believe this computation of planned_time is erroneous if we are already in the middle of this lanelet/maneuver.  It should only add the time remaining to complete the lanelet if time_progress > maneuver_start_time.

route_following_plugin/include/route_following_plugin.h
```diff
@@ -166,26 +159,21 @@ namespace route_following_plugin
         ros::Timer discovery_pub_timer_;
 
         // ROS service servers
-        ros::ServiceServer plan_maneuver_srv_;        
+        ros::ServiceServer plan_maneuver_srv_;  
+
+        //Queue of maneuver plans
+        std::vector<cav_msgs::Maneuver> maneuvers;
 
         // Minimal duration of maneuver, loaded from config file
```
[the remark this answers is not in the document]
TonysCousin: Yes, min_mvr_duration_ is definitely a misleading name.  It is being used to represent the minimum plan duration (which may include several maneuvers).  Please rename.

TonysCousin (changes requested):
@adev4a it is definitely looking better!  Thanks for your hard work here (and on a Sunday, too).  But we are not finished yet.  There are quite a few unaddressed comments (I just added some more, as well).  Please take another pass and ensure every comment is resolved either with a code change or a reply by you.  I have marked resolved a bunch of my early comments. 
 @kjrush @msmcconnell please check all your conversations and mark them resolved as soon as appropriate.  Since the volume of comments & edits is getting pretty big, it's kinda hard to see everything, so easy to overlook.  Please ask questions if uncertain.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,210 @@ namespace route_following_plugin
...
-                                    shortest_path[last_lanelet_index].id(), time_progress));
-                
-            }
-            current_progress += dist_diff;
-            //time_progress = resp.new_plan.maneuvers.back().lane_following_maneuver.end_time;
-            speed_progress = target_speed;
-            
+            prev_mvr_end_speed = target_speed_in_lanelet;
+        }
+        //add lane follow as last maneuver
```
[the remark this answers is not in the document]
adev4a: The last maneuver doesn't have to be a lane follow, this implementation required assigning lane changing maneuvers and filling in the rest of the plan with lane following, that is what this is attempting to do. I've changed the loop plan till the second to last lanelet

route_following_plugin/include/route_following_plugin.h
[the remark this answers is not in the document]
kjrush: I'd recommend using the friend test pattern like this: https://stackoverflow.com/questions/2396370/how-to-make-google-test-classes-friends-with-my-classes That should let the unit tests access the private/protected members of the classes without making them public in a broader sense. Either that or it might indicate that these functions might be worth factoring out into a separate class, from the perspective of the responsibilities of the route_following plugin I'm not sure they make sense to be part of the external API.

route_following_plugin/include/route_following_plugin.h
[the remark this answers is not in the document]
kjrush: I believe using GTest's Friend Test feature they can be, it makes the unit test framework a friend class of the class under test letting it access private members. I linked a stack overflow response about it on another comment in this PR.

route_following_plugin/src/route_following_plugin.cpp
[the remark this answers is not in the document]
kjrush: Any thoughts on this? This one I'm not too strong on, since it's not a functional change. But generally parameterizing functions over direct access to class variables would be my preference for testability where possible.

route_following_plugin/include/route_following_plugin.h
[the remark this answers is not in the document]
kjrush: That's a fair consideration, if we need random access or simply to non-destructively walk it then a std::queue likely wouldn't match our use case.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,238 @@ namespace route_following_plugin
...
-            end_dist = std::min(end_dist, total_maneuver_length);
-            ROS_DEBUG_STREAM("end_dist: " << end_dist);
-            double dist_diff = end_dist - current_progress;
-            ROS_DEBUG_STREAM("dist_diff: " << dist_diff);
-            if(route_length - end_dist <= stopping_dist){
-                end_dist = route_length - stopping_dist; 
-                dist_diff = end_dist - current_progress;
-                approaching_route_end = true;
+            if(isLaneChangeNeeded(following_lanelets, route_shortest_path[i + 1].id())){
+                maneuvers.push_back(composeLaneChangeManeuverMessage(start_dist, end_dist, prev_mvr_end_speed , target_speed_in_lanelet, route_shortest_path[i].id(), route_shortest_path[i+1].id(), ros::Time::now()));
```
kjrush, line 125: I'm not sure that ros::Time::now() is accurate, considering this could be the second or third or so-on maneuver in the maneuver plan, for which the start time would need to be ~= to the end time of the maneuver before it.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,238 @@ namespace route_following_plugin
...
             }
-            if(end_dist < current_progress){
-                break;
+            else{
+                maneuvers.push_back(composeLaneFollowingManeuverMessage(start_dist, end_dist, prev_mvr_end_speed, target_speed_in_lanelet, route_shortest_path[i].id(),ros::Time::now()));
```
kjrush, line 128: Same as above comment re: time

route_following_plugin/src/route_following_plugin.cpp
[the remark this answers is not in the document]
kjrush: Could the macro not be updated to account for that instead of a separate function? If it needs to be separate from the macro, a free function would still be preferable to a member function I'd think. While it's used by the RFP, it still doesn't really have any real relationship to it at a logical level that would necessitate this function belonging to the RFP.

route_following_plugin/src/route_following_plugin.cpp
adev4a: This starting time is updated in plan_maneuver_cb before the maneuvers are added to response

route_following_plugin/src/route_following_plugin.cpp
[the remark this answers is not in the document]
adev4a: This function is not being used anymore

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,238 @@ namespace route_following_plugin
...
-        ros::Time time_progress = ros::Time::now();
-        double target_speed=findSpeedLimit(current_lanelet);   //get Speed Limit
 
-        double total_maneuver_length = current_progress + mvr_duration_ * target_speed;
-        double route_length=  wm_->getRouteEndTrackPos().downtrack; 
-        total_maneuver_length = std::min(total_maneuver_length, route_length);
+        double route_length = wm_->getRouteEndTrackPos().downtrack;
+        double prev_mvr_end_speed = 0.0;
+        //Find lane changes in path - up to the second to last lanelet in path (till lane change is possible)
+        for(int i = 0; i < route_shortest_path.size() - 1; ++i){
```
msmcconnell, line 116: Use size_t when defining iterators for for loops. Sonar cloud will throw a warning over this.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -236,85 +284,61 @@ namespace route_following_plugin
...
-        current_time = maneuver_msg.stop_and_wait_maneuver.end_time;
-        ROS_INFO_STREAM("Creating stop and wait start dist:"<<current_dist<<" end dist:"<<end_dist);
-        ROS_INFO_STREAM("Duration:"<<maneuver_msg.stop_and_wait_maneuver.end_time.toSec() - maneuver_msg.stop_and_wait_maneuver.start_time.toSec());
-        return maneuver_msg;
-    }
-    bool RouteFollowingPlugin::identifyLaneChange(lanelet::routing::LaneletRelations relations, int target_id)
+    bool RouteFollowingPlugin::isLaneChangeNeeded(lanelet::routing::LaneletRelations relations, int target_id)
     {
         for(auto& relation : relations)
         {
```
msmcconnell, line 344: Is SUCCESSOR really the correct check for identifying lane changes? It's been a while since I looked at the possible routing graph relations but couldn't a lane change also be a successor?

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,238 @@ namespace route_following_plugin
...
-        double route_length=  wm_->getRouteEndTrackPos().downtrack; 
-        total_maneuver_length = std::min(total_maneuver_length, route_length);
+        double route_length = wm_->getRouteEndTrackPos().downtrack;
+        double prev_mvr_end_speed = 0.0;
+        //Find lane changes in path - up to the second to last lanelet in path (till lane change is possible)
+        for(int i = 0; i < route_shortest_path.size() - 1; ++i){
+            auto following_lanelets = wm_->getRoute()->followingRelations(route_shortest_path[i]);
+            double target_speed_in_lanelet = findSpeedLimit(route_shortest_path[i]);
+            double start_dist = wm_->routeTrackPos(route_shortest_path[i].centerline2d().front()).downtrack;
+            double end_dist = wm_->routeTrackPos(route_shortest_path[i].centerline2d().back()).downtrack;
```
msmcconnell, line 120: The start distance and start time should be assigned from the previous maneuver so that we avoid possibly small floating point deltas on lanelet downtrack calculations. There is actually no need to track these values as unique variables like you did with prev_maver_end_speed you can just extract them from maneuvers.back() on each loop.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -236,85 +284,61 @@ namespace route_following_plugin
...
     }
 
-    cav_msgs::Maneuver RouteFollowingPlugin::composeLaneChangeManeuverMessage(double current_dist, double end_dist, double current_speed, double target_speed, int starting_lane_id,int ending_lane_id, ros::Time& current_time){
+    cav_msgs::Maneuver RouteFollowingPlugin::composeLaneChangeManeuverMessage(double start_dist, double end_dist, double start_speed, double target_speed, int starting_lane_id,int ending_lane_id, ros::Time start_time){
         cav_msgs::Maneuver maneuver_msg;
         maneuver_msg.type = cav_msgs::Maneuver::LANE_CHANGE;
         maneuver_msg.lane_change_maneuver.parameters.neogition_type = cav_msgs::ManeuverParameters::NO_NEGOTIATION;
         maneuver_msg.lane_change_maneuver.parameters.presence_vector = cav_msgs::ManeuverParameters::HAS_TACTICAL_PLUGIN;
         maneuver_msg.lane_change_maneuver.parameters.planning_tactical_plugin = lane_change_plugin_;
         maneuver_msg.lane_change_maneuver.parameters.planning_strategic_plugin = "RouteFollowingPlugin";
```
msmcconnell, line 320: The name of the plugin should be a constant in the class which is reused to prevent future errors where we change it in one place but not others.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -236,85 +284,61 @@ namespace route_following_plugin
...
         } else {
-            maneuver_msg.lane_change_maneuver.end_time = current_time + ros::Duration((end_dist - current_dist) / (0.5 * cur_plus_target));
+            maneuver_msg.lane_change_maneuver.end_time = start_time + ros::Duration((end_dist - start_dist) / (0.5 * cur_plus_target));
         }
         maneuver_msg.lane_change_maneuver.starting_lane_id = std::to_string(starting_lane_id);
         maneuver_msg.lane_change_maneuver.ending_lane_id = std::to_string(ending_lane_id);
-        current_time = maneuver_msg.lane_change_maneuver.end_time;
-        ROS_INFO_STREAM("Creating lane change start dist:"<<current_dist<<" end dist:"<<end_dist<<" Starting llt:"<<starting_lane_id<<" Ending llt:"<<ending_lane_id);
+        
+        ROS_INFO_STREAM("Creating lane change start dist:"<<start_dist<<" end dist:"<<end_dist<<" Starting llt:"<<starting_lane_id<<" Ending llt:"<<ending_lane_id);
```
msmcconnell, line 336: please add spaces in your prints quoted text "" so that the output is actually readable like 
start dist: 5
instead of 
start dist:5

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -236,85 +284,61 @@ namespace route_following_plugin
...
+        maneuver_msg.lane_change_maneuver.start_speed = start_speed;
+        maneuver_msg.lane_change_maneuver.start_time = start_time;
         maneuver_msg.lane_change_maneuver.end_dist = end_dist;
         maneuver_msg.lane_change_maneuver.end_speed = target_speed;
         // because it is a rough plan, assume vehicle can always reach to the target speed in a lanelet
-        double cur_plus_target = current_speed + target_speed;
+        double cur_plus_target = start_speed + target_speed;
         if (cur_plus_target < 0.00001) {
-            maneuver_msg.lane_change_maneuver.end_time = current_time + ros::Duration(mvr_duration_);
+            maneuver_msg.lane_change_maneuver.end_time = start_time + ros::Duration(min_plan_duration_);
```
msmcconnell, line 329: why is the length of the entire maneuver plan used in this case. It seems like this is basically an error case and it would be better to log a warning and provide some small value like 1s instead of 16s for the bad maneuver. Same applies to the other method like this.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -236,85 +284,61 @@ namespace route_following_plugin
...
         maneuver_msg.lane_following_maneuver.lane_id = std::to_string(lane_id);
-        current_time = maneuver_msg.lane_following_maneuver.end_time;
-        ROS_INFO_STREAM("Creating lane follow start dist:"<<current_dist<<" end dist:"<<end_dist);
+        ROS_INFO_STREAM("Creating lane follow start dist:"<<start_dist<<" end dist:"<<end_dist);
         ROS_INFO_STREAM("Duration:"<<maneuver_msg.lane_following_maneuver.end_time.toSec() - maneuver_msg.lane_following_maneuver.start_time.toSec());
         return maneuver_msg;
     }
 
-    cav_msgs::Maneuver RouteFollowingPlugin::composeLaneChangeManeuverMessage(double current_dist, double end_dist, double current_speed, double target_speed, int starting_lane_id,int ending_lane_id, ros::Time& current_time){
+    cav_msgs::Maneuver RouteFollowingPlugin::composeLaneChangeManeuverMessage(double start_dist, double end_dist, double start_speed, double target_speed, int starting_lane_id,int ending_lane_id, ros::Time start_time){
```
msmcconnell, line 314: Recommend all lanelet Id use the lanelet::Id type instead of int to prevent any possibility of overflow as some of our lanelet ids end up being really large when the map is converted.

route_following_plugin/src/route_following_plugin.cpp
msmcconnell, line 125: is an extra i++ needed here? I think the way it currently is the next iteration will plan a lane follow overtop of our lane change.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,238 @@ namespace route_following_plugin
...
-            if(current_progress >= total_maneuver_length || last_lanelet_index == shortest_path.size() - 1)
-            {
-                break;
-            }
-
-            ++last_lanelet_index;
+            prev_mvr_end_speed = target_speed_in_lanelet;
+        }
+        //add lane follow as last maneuver
+        if(maneuvers.size() < route_shortest_path.size()){
```
msmcconnell, line 133: Add TODO comment about the need to add stop and wait back into this logic

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,238 @@ namespace route_following_plugin
...
+            double target_speed_in_lanelet = findSpeedLimit(route_shortest_path.back());
+            double start_dist = wm_->routeTrackPos(route_shortest_path.back().centerline2d().front()).downtrack;
+            double end_dist = wm_->routeTrackPos(route_shortest_path.back().centerline2d().back()).downtrack; 
+            maneuvers.push_back(composeLaneFollowingManeuverMessage(start_dist, end_dist, prev_mvr_end_speed, target_speed_in_lanelet, route_shortest_path.back().id(),ros::Time::now()));
```
msmcconnell, line 137: there us a ros::Time now field here as well which might need to change based on your discussion with Kyle

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,238 @@ namespace route_following_plugin
...
-                speed_progress,shortest_path[last_lanelet_index].id(),
-                shortest_path[last_lanelet_index].id(),end_time,time_req_to_stop)); 
+            resp.new_plan.maneuvers.push_back(latest_maneuver_plan_[i]);
+            ++i;
         }
+        
         if(resp.new_plan.maneuvers.size() == 0)
         {
             ROS_WARN_STREAM("Cannot plan maneuver because no route is found");
         }
```
msmcconnell, line 172: Should there be a return false here? @kjrush What is the desired behavior if planning is not successful? Return false or just return an empty list.

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,190 +41,238 @@ namespace route_following_plugin
...
+            case cav_msgs::Maneuver::INTERSECTION_TRANSIT_RIGHT_TURN :
+                maneuver.intersection_transit_right_turn_maneuver.start_time = prev_time;
+                maneuver.intersection_transit_right_turn_maneuver.end_time = time_progress;
+                break;
+            case cav_msgs::Maneuver::STOP_AND_WAIT :
+                maneuver.stop_and_wait_maneuver.start_time = prev_time;
+                maneuver.stop_and_wait_maneuver.start_time = time_progress;
+                break;
+            default :
+                ROS_DEBUG_STREAM("Invalid maneuver type, cannot update time progress for maneuver");
```
msmcconnell, line 218: This should be an exception as the output will be garbage after failing to set these fields. Same with the other two methods like this.

msmcconnell (changes requested):

route_following_plugin/src/route_following_plugin.cpp
adev4a: Yeah, we are still creating a lane change over 2 lanelets, so we need to push the maneuver twice and have an extra i++.
Making this change now

msmcconnell:
Also please address the code smells

route_following_plugin/src/route_following_plugin.cpp
adev4a: I'm a little confused about this one, the description for Successor says " A (the only) direct, reachable successor. Not merging and not diverging." which does sound like it would not pick up a lane change, so our !Successor check should be good. But it does mean that at this point in the route, Successor should be an invalid check since there is no 'un-diverging' lanelet directly reachable

route_following_plugin/src/route_following_plugin.cpp
adev4a: I'm guessing the intention here was that if the current and target speed are zero, we should not move, so we just send an entire plan with that 0 speed

msmcconnell: That seems like a plausible reason, but such a case really should be handled by stop and wait and would therefore be an error for a lane following or lane change maneuver.

route_following_plugin/src/route_following_plugin.cpp
msmcconnell: @adev4a 
I looked into the source code and I think what you have will work correctly. Though there is a constraint on the input that the target lanelet be accessible from the other lanelet. If you had a non-accessible lanelet you would get the wrong result. I think this is fine given how you are using the method but you should probably document this constraint in the method comments

route_following_plugin/include/route_following_plugin.h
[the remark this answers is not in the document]
TonysCousin: @adev4a thanks for changing this, but the config file needs to be updated accordingly, as well.  Need to try think of all the side effects of every change :)

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -43,188 +42,187 @@ namespace route_following_plugin
...
-        auto shortest_path = wm_->getRoute()->shortestPath();
 
-        lanelet::ConstLanelet current_lanelet;
-        int last_lanelet_index = -1;
-        for (auto llt : current_lanelets)
+        //auto shortest_path = wm_->getRoute()->shortestPath();
+
+        //Check if current position is on shortest path, if not on shortest path return empty maneuver  
+        int current_lanelet_index = -1;
+        for (auto llt : nearest_lanelets)
```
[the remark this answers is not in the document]
TonysCousin: That's a very good explanation, Mike, and one that ought to be captured.  It's definitely not obvious by looking at the code, unless you really know lanelet2 inside and out.  Therefore, this is the kind of info that is really useful in code comments!  Consider how much time we all have spent rehashing this topic.  5-6 lines of comments explaining what Mike says above would have saved all that expensive time.

Bottom line: I agree with Mike now, this whole code block should be removed.

route_following_plugin/src/route_following_plugin.cpp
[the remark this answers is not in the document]
TonysCousin: Agree with what it is attempting to do & why (for now).  I'm still concerned that it isn't clear to a reader (e.g. the next person who works on this node).  A comment explaining _why_ the final maneuver cannot be a lane change would be a big help.  Also I still believe that last if test is unnecessary.  Isn't maneuvers.size() always going to be 1 less than route_shortest_path.size() at this point?

route_following_plugin/src/route_following_plugin.cpp
adev4a: That makes sense to me. But also at this point the vehicle is already stopped, so there is no way for route following to switch to a different maneuver (since the current implementation compares against downtrack to advance maneuvers). Should this be an exception still? I'm thinking about cases where we try to engage in areas with complete zero speed limit

msmcconnell: Any lanelet with a speed limit of zero really should be considered a bug in the map but I think my point stands. These methods are specific to lane change and lane follow both of which would inherently require non-zero speed since those maneuver requires motion by definition. So providing either method with 0 speed is an error.

route_following_plugin/include/route_following_plugin.h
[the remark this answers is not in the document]
TonysCousin: In that case, please move these 5 variables to the private section.

route_following_plugin/src/route_following_plugin.cpp
TonysCousin: I think the fix still has a problem in it.

route_following_plugin/src/route_following_plugin.cpp
adev4a: The ros::Time now here is just a place holder, the time is updated without using this info when writing to response.
This happens in updateTimeProgress

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -41,191 +41,233 @@ namespace route_following_plugin
...
-        pose_msg_ = geometry_msgs::PoseStamped(*msg.get());
+
+    void RouteFollowingPlugin::updateTimeProgress(std::vector<cav_msgs::Maneuver>& maneuvers, ros::Time start_time) const {
+        ros::Time time_progress = start_time;
+        ros::Time prev_time = time_progress;
+
+        for(auto& maneuver : maneuvers){
+            double cur_plus_target = GET_MANEUVER_PROPERTY(maneuver,start_speed) + GET_MANEUVER_PROPERTY(maneuver, end_speed);
+            ros::Duration maneuver_duration;
+            maneuver_duration = ros::Duration((GET_MANEUVER_PROPERTY(maneuver,end_dist) - GET_MANEUVER_PROPERTY(maneuver,start_dist)) / (0.5 * cur_plus_target));
```
msmcconnell, line 183: Calling these GET_MANEUVER_PROPERTY macros twice on the same line is causing a pretty concerning error message on sonar cloud. You should store their output as a local variable 
https://sonarcloud.io/project/issues?id=usdot-fhwa-stol_CARMAPlatform&issues=AXmkZrrBdYWyuZksBLpj&open=AXmkZrrBdYWyuZksBLpj&pullRequest=1272

route_following_plugin/src/route_following_plugin.cpp
msmcconnell: @TonysCousin what are you referring too here?

route_following_plugin/include/route_following_plugin.h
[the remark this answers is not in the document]
TonysCousin: @adev4a I trust you'll either move these or leave a TODO in place, as we discussed.  Once you see this comment, I'm fine with resolving the conversation.

route_following_plugin/src/route_following_plugin.cpp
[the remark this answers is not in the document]
TonysCousin: @adev4a I'll trust you to resolve this as we discussed, then mark this conversation as resolved.

TonysCousin:
@adev4a I am heading out on travel, and my availability will be really sporadic over the next 3 weeks.  Please do not wait for me on anything.  Good job pushing to get this important component whipped into shape!

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -24,360 +24,361 @@
...
+
+            resp.new_plan.maneuvers.push_back(latest_maneuver_plan_[i]);
+            ++i;
         }
-        if(resp.new_plan.maneuvers.size() == 0)
+
+        if (resp.new_plan.maneuvers.size() == 0)
         {
             ROS_WARN_STREAM("Cannot plan maneuver because no route is found");
+            return true;
```
msmcconnell, line 183: Should this be false?

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -207,7 +207,7 @@ namespace route_following_plugin
 
         for (auto &maneuver : maneuvers)
         {
-            time_progress += maneuverDuration(maneuver);
+            time_progress += getManeuverDuration(maneuver, 0.001);
```
msmcconnell, line 218: Here the epsilon is 0.001 while on line 167 it is 0.0001 what is the reason for the discrepancy? Seems like it should be a constant

msmcconnell (changes requested):

msmcconnell (approved):

msmcconnell:
Also please address code smells

kjrush (approved):

PR #1280, opened 2021-05-26
fix_1

aswath1:
# PR Details
## Description
Route Visualization 

## Motivation and Context
 marker id's are bring set based on index, so when the new route it smaller than the previous route some points are not cleared

## Types of changes

## Checklist:


route/src/route_generator_worker.cpp
```diff
@@ -129,11 +129,9 @@ namespace route {
 
     bool RouteGeneratorWorker::set_active_route_cb(cav_srvs::SetActiveRouteRequest &req, cav_srvs::SetActiveRouteResponse &resp)
     {
-        ROS_INFO_STREAM("set_active_route_cb: Selected ID:" << req.routeID);
         // only allow activate a new route in route selection state
         if(this->rs_worker_.get_route_state() == RouteStateWorker::RouteState::SELECTION)
         {
-            ROS_DEBUG_STREAM("Valid state proceeding with selection");
```
msmcconnell, line 136: I don't see the reason to remove these two log lines. Lets leave them in

route/src/route_generator_worker.cpp
```diff
@@ -358,48 +355,42 @@ namespace route {
...
         }
 
-        route_marker_msg_.markers={};
+        route_marker_msg_.points={};
 
-        if (points.empty())
+        if (!points.empty())
         {
             ROS_WARN_STREAM("No central line points! Returning");
-            return route_marker_msg_;
```
msmcconnell, line 366: Why remove this return? It still seems valid since the route is empty

msmcconnell: Actually probably best to move this if statement to line 383 before the for loop so that a routing failure will clear the points from rviz

route/src/route_state_worker.cpp
```diff
@@ -64,7 +60,5 @@ namespace route {
             // should not reach here
             throw std::invalid_argument("Current state is illegal: " + std::to_string(state_));
         }
-
-        ROS_INFO_STREAM("Received Route Event: " << event << " transitioning from: " << old_state << " to: " << state_);
```
msmcconnell, line 68: No reason for this file to be changed. Please revert changes

msmcconnell (changes requested):

route/src/route_generator_worker.cpp
```diff
@@ -129,11 +129,10 @@ namespace route {
 
     bool RouteGeneratorWorker::set_active_route_cb(cav_srvs::SetActiveRouteRequest &req, cav_srvs::SetActiveRouteResponse &resp)
     {
-        ROS_INFO_STREAM("set_active_route_cb: Selected ID:" << req.routeID);
```
msmcconnell, line 132: You missed adding back this comment

msmcconnell (approved):
Looks good. Will try to test it on the track before mering

msmcconnell (approved):

PR #1281, opened 2021-05-27
Feature/freight arrival

JonSmet:
# PR Details
## Description
This PR updates PortDrayageWorker's compose_arrival_message() function, which is used to broadcast a Mobility Operation message from the host vehicle when it arrives at a destination in the port drayage use case. More specifically, several fields have been added to the broadcasted message's strategy_params string because these additional fields are required for V2XHub to conduct its Mobility Operation message processing.

In addition to the updates described above, the data types of several data members within PortDrayageWorker have been updated to match the expectations of V2XHub to support communications. For example, V2XHub expects/required cmv_id, cargo_id, action_id, and next_action to be integers.

## How Has This Been Tested?

Unit tested and integration tested at TFHRC with the White Pacifica and V2XHub.

## Types of changes

## Checklist:

### DIFF
--- a/port_drayage_plugin/config/params.yml
+++ b/port_drayage_plugin/config/params.yml
@@ -1,3 +1,3 @@
 stop_speed_epsilon: 1.0
-cmv_id: CARMA_TEST_CMV
-cargo_id: TEST_CARGO
+cmv_id: 123 // Test value; CARMA Streets expects cmv_id to be an integer
+cargo_id: 321 // Test value; CARMA Streets expects cargo_id to be an integer
--- a/port_drayage_plugin/src/port_drayage_plugin.cpp
+++ b/port_drayage_plugin/src/port_drayage_plugin.cpp
@@ -29,10 +29,10 @@ namespace port_drayage_plugin
 
         double speed_epsilon = _pnh->param("stop_speed_epsilon", 1.0);
         declaration = _pnh->param("declaration", 1.0);
-        std::string cmv_id;
-        _pnh->param<std::string>("cmv_id", cmv_id, "");
-        std::string cargo_id;
-        _pnh->param<std::string>("cargo_id", cargo_id, "");
+        int cmv_id;
+        _pnh->param<int>("cmv_id", cmv_id, 123);
+        int cargo_id;
+        _pnh->param<int>("cargo_id", cargo_id, 321);
 
         ros::Publisher outbound_mob_op = _nh->advertise<cav_msgs::MobilityOperation>("outgoing_mobility_operation", 5);
         _outbound_mobility_operations_publisher = std::make_shared<ros::Publisher>(outbound_mob_op);
--- a/port_drayage_plugin/src/port_drayage_worker.cpp
+++ b/port_drayage_plugin/src/port_drayage_worker.cpp
@@ -84,6 +84,20 @@ namespace port_drayage_plugin
         pt.put("cmv_id", _cmv_id);
         pt.put("cargo_id", _cargo_id);
         pt.put("operation", PORT_DRAYAGE_ARRIVAL_OPERATION_ID);
+
+        // Certain fields required by CARMA Streets are unnecessary for arrival messages and can be zero-filled
+        pt.put("cargo", 0); 
+        ptree location;
+        location.put("latitude", 0);
+        location.put("longitude", 0);
+        pt.put_child("location", location);
+        ptree destination;
+        destination.put("latitude", 0);
+        destination.put("longitude", 0);
+        pt.put_child("destination", destination);
+        pt.put("action_id", 0);
+        pt.put("next_action", 0);
+
         std::stringstream body_stream;
         boost::property_tree::json_parser::write_json(body_stream, pt);
         msg.strategy_params = body_stream.str();
@@ -99,7 +113,7 @@ namespace port_drayage_plugin
             ptree pt;
             std::istringstream strategy_params_ss(msg->strategy_params);
             boost::property_tree::json_parser::read_json(strategy_params_ss, pt);
-            std::string mobility_operation_cmv_id = pt.get<std::string>("cmv_id");
+            int mobility_operation_cmv_id = pt.get<int>("cmv_id");
 
             // Check if the received MobilityOperation message is intended for this vehicle's cmv_id   
             if(mobility_operation_cmv_id == _cmv_id) {
@@ -131,10 +145,10 @@ namespace port_drayage_plugin
         std::istringstream mobility_operation_strategy_params_ss(mobility_operation_strategy_params);
         boost::property_tree::json_parser::read_json(mobility_operation_strategy_params_ss, pt);
 
-        _latest_mobility_operation_msg.cargo_id = pt.get<std::string>("cargo_id");
+        _latest_mobility_operation_msg.cargo_id = pt.get<int>("cargo_id");
         _latest_mobility_operation_msg.has_cargo = pt.get<bool>("cargo");
-        _latest_mobility_operation_msg.current_action_id = pt.get<std::string>("action_id");
-        _latest_mobility_operation_msg.next_action_id = pt.get<std::string>("next_action");
+        _latest_mobility_operation_msg.current_action_id = pt.get<int>("action_id");
+        _latest_mobility_operation_msg.next_action_id = pt.get<int>("next_action");
         _latest_mobility_operation_msg.operation = pt.get<std::string>("operation");
 
         if(_latest_mobility_operation_msg.operation == "MOVING_TO_LOADING_AREA") {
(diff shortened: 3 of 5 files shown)


port_drayage_plugin/config/params.yml
```diff
@@ -1,3 +1,3 @@
 stop_speed_epsilon: 1.0
-cmv_id: CARMA_TEST_CMV
-cargo_id: TEST_CARGO
+cmv_id: 123 // Test value; CARMA Streets expects cmv_id to be an integer
```
kjrush, line 2: Does this work? I thought yaml used # as a comment start?

kjrush, line 2: This representation is also a little weird, if this is an ID I imagine it's more like the BSM ID where it's intended to be a raw byte-string. Not to say that we can't integerize it, but it may be more accurate to use an unsigned int to avoid any weirdness or misinterpretation. I'm not 100% sure what the interpretation is on CARMA streets side, so I could be off base though.

port_drayage_plugin/src/port_drayage_worker.cpp
```diff
@@ -84,6 +84,20 @@ namespace port_drayage_plugin
         pt.put("cmv_id", _cmv_id);
         pt.put("cargo_id", _cargo_id);
         pt.put("operation", PORT_DRAYAGE_ARRIVAL_OPERATION_ID);
+
+        // Certain fields required by CARMA Streets are unnecessary for arrival messages and can be zero-filled
```
kjrush, line 88: Would it be cleaner to just omit certain fields or have different message types? Unless we're using some sort of JSON schema, it should be fine to only encode the fields that the receiving end is unpacking. If we need to strictly define these messages, it may be better to add presence bitstrings for the optional fields, to distinguish between empty data and actual zeros. Or perhaps consider using several different messages differentiated by a type field in the message.

kjrush (commented):
Apologies for the delay, I didn't even notice this was on my docket for so long. What's here seems totally fine, mostly just wondering if there might be some benefits to some more thoughts on the message set for usage with CARMA Streets.

JonSmet:
Please note that the TODO comment in port_drayage_worker.cpp (also identified as a Code Smell by sonarcloud) is intended to be addressed in future stories when Cargo-Loading- and Cargo-Unloading-related features are implemented.

port_drayage_plugin/src/port_drayage_worker.cpp
```diff
@@ -108,6 +128,7 @@ namespace port_drayage_plugin
                 _previous_strategy_params = msg->strategy_params;
                 
                 // Process event based on the PortDrayageEvent associated with the received MobilityOperation message
+                // TODO: Update this object's '_has_cargo' flag and '_cargo_id' based on the received status message from a loading/unloading device
```
kjrush, line 131: This todo is being covered by a separate story, correct?

port_drayage_plugin/src/port_drayage_worker.cpp
```diff
@@ -82,8 +82,28 @@ namespace port_drayage_plugin
         using boost::property_tree::ptree;
         ptree pt;
         pt.put("cmv_id", _cmv_id);
-        pt.put("cargo_id", _cargo_id);
         pt.put("operation", PORT_DRAYAGE_ARRIVAL_OPERATION_ID);
+
+        // Add current vehicle location (latitude and longitude)
+        ptree location;
+        location.put("latitude", _current_gps_position.latitude * 10000000); // Convert degrees to 1/10 microdegrees
```
kjrush, line 89: Is there a particular reason this is being read straight from the GPS itself (based on the INSPVA message)? Fundamentally that should work fine, mostly. But I think the more idiomatic approach here would involve still subscribing to the final output of localization, but then transforming it into lat/lon via the proj library.

port_drayage_plugin/src/port_drayage_plugin.cpp
```diff
@@ -29,10 +29,13 @@ namespace port_drayage_plugin
...
         declaration = _pnh->param("declaration", 1.0);
-        std::string cmv_id;
-        _pnh->param<std::string>("cmv_id", cmv_id, "");
         std::string cargo_id;
-        _pnh->param<std::string>("cargo_id", cargo_id, "");
+        _pnh->param<std::string>("cargo_id", cargo_id, ""); 
+
+        // Read in 'cmv_id' parameter as a string, then convert to an unsigned long before initializing the PortDrayageWorker object
+        std::string cmv_id_string;
+        _pnh->param<std::string>("cmv_id", cmv_id_string, "123");
```
kjrush, line 37: When possible let's make these kind of default values very obviously wrong "UNDEFINED" or something similar, easier to track if they do show up in some data stream somewhere somehow.

port_drayage_plugin/src/port_drayage_worker.cpp
```diff
@@ -82,8 +82,28 @@ namespace port_drayage_plugin
...
+        location.put("longitude", _current_gps_position.longitude * 10000000); // Convert degrees to 1/10 microdegrees
```
kjrush, line 90: Is there a reason to encode degrees vs 1/10 microdegrees? Json has support for floats, so I think it'd be all the same under-the-hood, no? 1/10th microdegrees lines up nicely with J2735, but every application I've ever seen interact with that data from J2735 immediately converts back to degrees for usability. Is floating point precision the concern?

port_drayage_plugin/src/port_drayage_worker.cpp
```diff
@@ -99,7 +119,7 @@ namespace port_drayage_plugin
             ptree pt;
             std::istringstream strategy_params_ss(msg->strategy_params);
             boost::property_tree::json_parser::read_json(strategy_params_ss, pt);
-            std::string mobility_operation_cmv_id = pt.get<std::string>("cmv_id");
+            unsigned long mobility_operation_cmv_id = pt.get<unsigned long>("cmv_id");
```
kjrush, line 122: Nitpick, but for whatever it's worth, the sizeof(unsigned long) is implementation/compiler/architecture specific, so this may result in some undefined behavior if we ever attempt to run this on something with a smaller u-long than our 4-byte IDs. Maybe not worth anything more than a comment, but food for thought I suppose.

port_drayage_plugin/src/port_drayage_worker.cpp
JonSmet: Correct, this is being addressed in CF-9 (CMV Broadcasts Arrival at Loading Point).

port_drayage_plugin/src/port_drayage_worker.cpp
JonSmet: The only reason it was decided to encode latitude and longitude values as 1/10 microdegrees was to line up with the J2735 specification. I am fine with encoding as degrees, and will make sure the change is applied on the V2XHub side as well.

port_drayage_plugin/src/port_drayage_plugin.cpp
JonSmet: Great point. I've updated cargo_id to "UNDEFINED-CARGO-ID".  I was unable to change cmv_id to a more descriptive default value since it is converted to an unsigned long.

port_drayage_plugin/src/port_drayage_worker.cpp
JonSmet: I've fixed this to follow the approach used in other plugins.

port_drayage_plugin/src/port_drayage_worker.cpp
JonSmet: Added a comment to note this in the code.

port_drayage_plugin/src/port_drayage_worker.cpp
```diff
@@ -121,8 +121,8 @@ namespace port_drayage_plugin
 
         // Add current vehicle location (latitude and longitude)
         ptree location;
-        location.put("latitude", _current_gps_position.latitude * 10000000); // Convert degrees to 1/10 microdegrees
-        location.put("longitude", _current_gps_position.longitude * 10000000); // Convert degrees to 1/10 microdegrees
+        location.put("latitude", _current_gps_position.latitude);
```
kjrush, line 124: I can't add a comment on the line, but where this is initialized I think it might be wise (if extremely cautious) to either check, and make sure that this `_current_gps_position` struct has meaningful data (we've gotten a pose update after having a georeference) OR at least initializing it explicitly to `{0.0, 0.0}` where it's defined as a member variable or as a default value for the struct type.

JonSmet: Good catch, I've initialized the latitude and longitude values within the LatLonCoordinate struct to 0.0 to address this.

kjrush (approved):

PR #1283, opened 2021-05-27
Bugfix/timer tcr

dan-du-car:
# PR Details
## Description
After selecting a route, the wmb should broadcast constantly every 10 seconds.

## Motivation and Context
Zephyr testing

## How Has This Been Tested?
Local VM

## Types of changes

## Checklist:

### DIFF
--- a/carma_wm_ctrl/src/WMBroadcasterNode.cpp
+++ b/carma_wm_ctrl/src/WMBroadcasterNode.cpp
@@ -60,7 +60,6 @@ int WMBroadcasterNode::run()
   map_update_pub_ = cnh_.advertise<autoware_lanelet2_msgs::MapBin>("map_update", 200, [this](auto& pub){ wmb_.newUpdateSubscriber(pub); });
   //Route Message Publisher
   control_msg_pub_= cnh_.advertise<cav_msgs::TrafficControlRequest>("outgoing_geofence_request", 1, true);
-  TCR_timer = cnh_.createTimer(ros::Duration(10), &WMBroadcaster::TCR_timerCallBack,&wmb_ );
   //Check Active Geofence Publisher
   active_pub_ = cnh_.advertise<cav_msgs::CheckActiveGeofence>("active_geofence", 200, true);
   // Base Map Sub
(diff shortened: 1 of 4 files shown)


carma_wm_ctrl/src/WMBroadcasterNode.cpp
```diff
@@ -60,6 +60,7 @@ int WMBroadcasterNode::run()
   map_update_pub_ = cnh_.advertise<autoware_lanelet2_msgs::MapBin>("map_update", 200, [this](auto& pub){ wmb_.newUpdateSubscriber(pub); });
   //Route Message Publisher
   control_msg_pub_= cnh_.advertise<cav_msgs::TrafficControlRequest>("outgoing_geofence_request", 1, true);
+  TCR_timer = cnh_.createTimer(ros::Duration(10), &WMBroadcaster::TCR_timerCallBack,&wmb_ );
```
msmcconnell, line 63: This looks like the same timer as that done on line 86? Why would this code work and that would fail?

dan-du-car: The lambda callback seems executed for one time only on carma-platform start up and carma_wm_ctl run() is called.
This is the local testing debug info.
Time-1 is this PR change
Timer and 0 is the line 86 print log.
image

dan-du-car: I see posts where lambda callback is also used for ros::timer, but I do not find a good explanation why this lambda function is not executed every 10s in this case.

msmcconnell: @dan-du-car Looking at the lambda callback it looks like the address for wmb_ is being passed in as the oneshot argument for the timer. This is likely casting to true for that boolean argument preventing that callback from triggering more than once. My guess is you could replace &wmb_ on line 89 with false and it would work.

dan-du-car: Thanks!

msmcconnell (approved):

Issue #1289, opened 2021-06-03
Rosbag Matching


jsun6065:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Rosbags are difficult to match to their corresponding log directory because the log directories have unintuitive names that do not indicate which bag file they belong to. 

### Carma version where this issue was discovered

Latest version
### Actual behavior

Log directories are named as an ID which makes it difficult to match them to a rosbag
### Steps to reproduce the actual behavior

1. Gather a Rosbag
2. Try to match it to a large log file that has many log directories

TonysCousin:
Closed as implemented in PR #1290.

PR #1290, opened 2021-06-03
Rosbag matching

jsun6065:
# PR Details
## Description

I wrote a script that matches rosbag files to their log directory in a logfile using timestamps extracted from both. The script renames the log directory so that it is clear which bag it belongs to. The current threshold for indicating a match is if the log directory was created within a minute of the rosbag.
## Related Issue

https://github.com/usdot-fhwa-stol/carma-platform/issues/1289
## Motivation and Context

Rosbags are difficult to match to their log directory because log directories are named with an ID that does not clearly indicate which bag that it belongs to. This script will address that problem.
## How Has This Been Tested?
I tested this script with two rosbags and two logfiles 

## Types of changes

## Checklist:

- [ x I have read the **CONTRIBUTING** document.


tyagi86 (approved):

engineering_tools/rosbag_matching/README.txt
```diff
@@ -0,0 +1,13 @@
+Rosbag Matcher
+
+Matches a given rosbag to a log directory in a given logfile by examining timestamps in the directories and renaming it to match the bag if the difference in timestamps is a minute or less.
```
TonysCousin, line 3: This statement is a little confusing, in that it appears to be casually inconsistent between referring to files vs directories.  It is unfortunate that ROS puts all the log dirs into a directory named "logfiles"!  But it is worth explaining that (and using that exact name).  Probably good to state the assumption that the script is run from a directory that contains one or more .bag files and a directory named logfiles, under which are all the log directories (each of which contains a bunch of files from a single run).

engineering_tools/rosbag_matching/rosbag_match
```diff
@@ -0,0 +1,66 @@
...
+cd $logfile
+
+
+match="0"
+# go through every log directory in logfiles
+for dir in */; do
+	cd $dir
+	echo Log Directory:
+	echo $dir
+	while read -r line ; do
```
TonysCousin, line 30: It is possible that users may run this (by accident, or lack of familiarity, or by choice) in a dir that has other subdirs in it that do not contain logs at all.  Therefore, it would be good to check if the rosout.log file exists before opening it to read (before this while loop).  If none exists, just skip this dir (maybe silently, before printing its name).

TonysCousin (changes requested):

engineering_tools/rosbag_matching/rosbag_match
```diff
@@ -0,0 +1,66 @@
...
+			if [[ "$(echo "$difference<60" | bc -l)" -eq "1"  ]]; then
+				echo "Match!" ; echo
+				match="1"
+				match_dir=$dir
+				match_time=${time:0:13}
+				match_diff=$difference
+								
+				# rename the directory
+				cd ..
+				match_log=${rosbag%.bag}.log
```
TonysCousin, line 52: I don't like the ".log" extension for the name of a directory.  Text log files inside this dir are already named *.log.  Also, directory names typically do not have a . extension, so I think it would be very confusing.  I suggest a suffix of "_logs" instead.

TonysCousin (changes requested):

TonysCousin (approved):

Issue #1227, opened 2021-04-28
Integration test for cooperative lanechange at Summit Point


MishkaMN:
### Types of Issue

- [X] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

We need integration test for cooperative lanechange

### Expected behavior
A car should lanechange successfully while the other yields and slows down.

### Actual behavior

It is not integration tested yet.

MishkaMN:
This has been integration tested and resolved by above PRs

PR #1307, opened 2021-06-11
Fix/all-remaining-fixes-for-lanechange

MishkaMN:
# PR Details
## Description

Initial Problems:
We were getting numerous issues such as :
There were numerous exceptions like "empty vector" "bad_alloc" that result from not having enough points in curve fitting. We added extra buffer for only planning and cut it later now
Car is not lanechanging early enough so that it ends up overshooting in small maps like TFHRC (buffer points were not getting cut, so that CLC for example was planning longer.Sort of related to above)
Due to duplication of points during concatenating or some other places, speed_to_time time_to_speed functions would create garbage numbers which makes pure pursuit unresponsive and the car was slowing down to 0mph in the middle of the road
Inlanecruising were never called with cooperative lanechange before, so it didn't have notion of other lanelets that are used for lanechange. So now it filters only lane_follow portion. Route_folllowing in this case is giving correct start and end distance. But functions used in ILC can pick up extra lanelets, so we fixed that

Fixes made for above:
This fixes logic in inlanecruising_plugin where it includes both of previous adjacent lanelets used for lanechanging in its curvature calculation logic. It only includes the lanelets in the straight part of the road (doesn't include adjacent lanelets that would have occured in the future as well).
Other changes include:
- CLC uses plan index now
- CLC and ILC uses different nearest point index logic using downtrack (as cartesian distance would pick a point out of the selected segment)
- CLC and ILC added extra buffer points beyond every maneuver start and end dist to create a curve fit, and those points are removed later
- Plan delegator has protection from duplicate points so that there is no INF or NAN points in speeds that are planned
- Extra protection in PPwrapper that also checks same timestamps to protect from above which results in garbage data being sent to PP that slowly stops the vehicle

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1227

## Motivation and Context

See above.

## How Has This Been Tested?

## Types of changes

## Checklist:


MishkaMN:
branch is not ready to be reviewed yet

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -792,60 +818,81 @@ namespace cooperative_lanechange
...
+                break;
             }
-            i++;
         }
-        return best_index;
+        if (!set_index)
+        {
+            best_index = points.size() - 1;
+        }
+        ROS_DEBUG_STREAM("get_nearest_route_point_index>> Found best_idx: " << best_index<<", points[i].x(): " << points[best_index].x() << ", points[i].y(): " << points[best_index].y());
```
adev4a, line 838: If you initialize best_index as points.size() -1, you can avoid having this if(!set_index) loop

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -792,60 +818,81 @@ namespace cooperative_lanechange
...
-                best_index = i;
+
+            if(downtrack > ending_downtrack){
+                best_index = i - 1;
+                ROS_DEBUG_STREAM("get_nearest_point_index>> Found best_idx: " << best_index<<", points[i].x(): " << points[best_index].x() << ", points[i].y(): " << points[best_index].y() << ", downtrack: "<< downtrack);
+                set_index = true;
                 break;
             }
         }
+        if (!set_index)
```
adev4a, line 860: Same as above. best_index can be initialized with points.size() - 1, to remove the need for this check

cooperative_lanechange/src/cooperative_lanechange.cpp
adev4a: Just a thought, would it make sense to call this function, get_nearest_point_index as well?

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -292,21 +292,26 @@ std::vector<PointSpeedPair> InLaneCruisingPlugin::constrain_to_time_boundary(con
...
+      ROS_DEBUG_STREAM("get_nearest_point_index>> points[i].x(): " << points[i].x() << ", points[i].y(): " << points[i].y() << ", downtrack: "<< downtrack);
+
+      if(downtrack > ending_downtrack){
+          best_index = i - 1;
+          ROS_DEBUG_STREAM("get_nearest_point_index>> Found best_idx: " << best_index<<", points[i].x(): " << points[best_index].x() << ", points[i].y(): " << points[best_index].y() << ", downtrack: "<< downtrack);
+          set_index = true;
+          break;
+      }
+  }
+  if (!set_index)
```
adev4a, line 311: Same note as other get_nearest_point functions

cooperative_lanechange/src/cooperative_lanechange.cpp
MishkaMN: That's a good idea. fixed best_index and renamed the function

msmcconnell (changes requested):

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -292,21 +292,20 @@ std::vector<PointSpeedPair> InLaneCruisingPlugin::constrain_to_time_boundary(con
 int InLaneCruisingPlugin::get_nearest_point_index(const std::vector<lanelet::BasicPoint2d>& points,
                                                const cav_msgs::VehicleState& state) const
 {
-  lanelet::BasicPoint2d veh_point(state.X_pos_global, state.Y_pos_global);
-  double min_distance = std::numeric_limits<double>::max();
-  int i = 0;
+  lanelet::BasicPoint2d state_pos(state.X_pos_global, state.Y_pos_global);
```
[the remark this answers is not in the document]
kjrush: Would basic autonomy be a good place to host this function?

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -616,8 +667,50 @@ std::vector<PointSpeedPair> InLaneCruisingPlugin::maneuvers_to_points(const std:
...
   }
+    // Here we are limiting the trajectory length to the given length by maneuver end dist as opposed to the end of lanelets involved.
+    double starting_route_downtrack = wm_->routeTrackPos(points_and_target_speeds.front().point).downtrack; 
+    size_t i = 0;
+    size_t max_i = 0;
+    double dist_accumulator = starting_route_downtrack;
+    lanelet::BasicPoint2d prev_point;
+
+    double ending_downtrack = maneuvers.back().lane_following_maneuver.end_dist;
+    if(ending_downtrack + config_.buffer_ending_downtrack < wm_->getRouteEndTrackPos().downtrack){
```
kjrush, line 678: If ending_downtrack + buffer is > route end, does it make more sense to just set end downtrack to route end? Right now if our buffer is 20m and the ending downtrack is 19m, we still don't use any buffer at all.

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -616,8 +667,50 @@ std::vector<PointSpeedPair> InLaneCruisingPlugin::maneuvers_to_points(const std:
...
+      ending_downtrack = ending_downtrack + config_.buffer_ending_downtrack;
+    } 
+
+    for (auto point_speed_pair : points_and_target_speeds) {
```
kjrush, line 682: foreach with an index accumulator, is it more clear to refactor as a standard for loop?

pure_pursuit_wrapper/src/pure_pursuit_wrapper.cpp
```diff
@@ -95,4 +100,34 @@ std::vector<double> PurePursuitWrapper::apply_response_lag(const std::vector<dou
   output = trajectory_utils::shift_by_lookahead(speeds, (unsigned int) lookahead_count);
   return output;
 }
+
+std::vector<cav_msgs::TrajectoryPlanPoint> PurePursuitWrapper::remove_repeated_timestamps(const std::vector<cav_msgs::TrajectoryPlanPoint>& traj_points){
```
kjrush, line 104: This assumes that all repeated timestamps are sequential, e.g. it catches [1, 2, 3, 4, 4, 5, 6, 7] but it does not catch [1, 2, 3, 2, 3, 4], just want to double check that the latter is not an expected case or worth handling. Might be worth documenting explicitly, since the function name would lead me to potentially believe otherwise.

pure_pursuit_wrapper/include/pure_pursuit_wrapper/pure_pursuit_wrapper.hpp
```diff
@@ -52,6 +52,7 @@ class PurePursuitWrapper {
          */ 
         std::vector<double> apply_response_lag(const std::vector<double>& speeds, const std::vector<double> downtracks, double response_lag) const;
 
+        std::vector<cav_msgs::TrajectoryPlanPoint> remove_repeated_timestamps(const std::vector<cav_msgs::TrajectoryPlanPoint>& traj_points);
```
kjrush, line 55: Needs doc comment

cooperative_lanechange/config/parameters.yaml
```diff
@@ -58,7 +58,7 @@ curvature_moving_average_window_size : 9
 curvature_calc_lookahead_count : 1
 
 # Int: Amount to downsample input lanelet centerline data. Value corresponds to saving each nth point.
-downsample_ratio : 8
+downsample_ratio : 1
```
kjrush, line 61: Is a 1:1 point downsample ratio really intended?

inlanecruising_plugin/src/inlanecruising_plugin.cpp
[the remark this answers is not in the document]
adev4a: Basic Autonomy does host this function now. It needs to be updated with the implementation that comes out of this pr

cooperative_lanechange/include/cooperative_lanechange.h
```diff
@@ -142,13 +131,33 @@ namespace cooperative_lanechange
...
                                                const cav_msgs::VehicleState& state) const;
+            /**
+             * \brief Overload: Returns the nearest point to the provided vehicle pose in the provided list
+             * 
+             * \param points BasicLineString2d points
+             * \param ending_downtrack ending downtrack along the route to get index of
+             * 
+             * \return index of nearest point in points
+             */
+            int get_nearest_point_index(lanelet::BasicLineString2d& points, double ending_downtrack);
```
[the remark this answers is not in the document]
MishkaMN: discussed offline and kept it as it is

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -792,60 +818,63 @@ namespace cooperative_lanechange
...
-        return best_index;
+        ROS_DEBUG_STREAM("get_nearest_point_index>> Found best_idx: " << best_index<<", points[i].x(): " << points[best_index].x() << ", points[i].y(): " << points[best_index].y());
 
+        return best_index;
     }
 
-        int CooperativeLaneChangePlugin::get_ending_point_index(lanelet::BasicLineString2d& points, double ending_downtrack){
-        int best_index = points.size()-1;
-        for(int i=points.size()-1;i>=0;i--){
+    int CooperativeLaneChangePlugin::get_nearest_point_index(const std::vector<lanelet::BasicPoint2d>& points,
```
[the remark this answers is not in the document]
MishkaMN: refactored as such

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -792,60 +818,63 @@ namespace cooperative_lanechange
...
             }
         }
+        ROS_DEBUG_STREAM("get_nearest_point_index>> Found best_idx: " << best_index<<", points[i].x(): " << points[best_index].x() << ", points[i].y(): " << points[best_index].y());
+
         return best_index;
     }
-
-    int CooperativeLaneChangePlugin::getNearestPointIndex(const std::vector<PointSpeedPair>& points,
+    
+    int CooperativeLaneChangePlugin::get_nearest_point_index(const std::vector<PointSpeedPair>& points,
```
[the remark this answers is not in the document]
MishkaMN: refactored

inlanecruising_plugin/src/inlanecruising_plugin.cpp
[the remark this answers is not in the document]
MishkaMN: it is refactored out to carma_wm for the time being

cooperative_lanechange/config/parameters.yaml
MishkaMN: yeah because sometimes in a small map like TFHRC, we run out of points if we sample it down.

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -868,33 +898,34 @@ namespace cooperative_lanechange
...
         return lc_route;
     }
 
     lanelet::BasicLineString2d CooperativeLaneChangePlugin::create_route_geom( double starting_downtrack, int starting_lane_id, double ending_downtrack, const carma_wm::WorldModelConstPtr& wm)
     {
         //Create route geometry for lane change maneuver
-        //Starting downtrack maybe in the previous lanelet, if it is, have a lane follow till new lanelet starts
-        std::vector<lanelet::ConstLanelet> lanelets_in_path = wm->getLaneletsBetween(starting_downtrack, ending_downtrack, true);
+        //Starting downtrack maybe in the previous lanelet, if it is, have a lane follow until new lanelet starts
+        std::vector<lanelet::ConstLanelet> lanelets_in_path = wm->getLaneletsBetween(starting_downtrack + 0.1, ending_downtrack -0.1, true); // +/-0.1 meters: try our best to avoid picking up before and next lanelets
```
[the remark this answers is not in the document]
MishkaMN: included non-inclusive variant in carma_wm

inlanecruising_plugin/src/inlanecruising_plugin.cpp
MishkaMN: good point. I added another clause

inlanecruising_plugin/src/inlanecruising_plugin.cpp
MishkaMN: switched to normal for loop

pure_pursuit_wrapper/src/pure_pursuit_wrapper.cpp
MishkaMN: I dont think we ever saw timestamps that go back, so I have documented explicitly.

pure_pursuit_wrapper/include/pure_pursuit_wrapper/pure_pursuit_wrapper.hpp
MishkaMN: added doc comment

carma_wm/include/carma_wm/WorldModel.h
```diff
@@ -338,6 +351,42 @@ class WorldModel
...
+
+  /**
+   * \brief Overload: Returns the nearest point to the provided vehicle pose in the provided list
+   * 
+   * \param points The points and speed pairs to evaluate
+   * \param state The current vehicle state
+   * 
+   * \return index of nearest point in points
+   */
+  virtual int get_nearest_point_index(const std::vector<PointSpeedPair>& points,
```
msmcconnell, line 363: I think we need a different name for this function. It is not clear to a user why this function uses downtrack distance instead of cartesian distance. Additionally, the comments should be updated to provide more detail about the implementation, why they might be used, and high cost of these methods. 

It seems Anish has them in basic autonomy as well. Should we move these to that library?

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -232,7 +232,7 @@ class LaneletDowntrackPair
   }
 };
 
-std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double start, double end,
+std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double start, double end, bool bounds_inclusive,
```
msmcconnell, line 235: This argument should not be included before the existing default valued argument as that will introduce unknown behavior in other components already relying on that default argument. It would be better to add this as another default argument after the shortest_path so that the old API is not broken.

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -263,11 +263,23 @@ std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double st
     TrackPos min = routeTrackPos(front);
     TrackPos max = routeTrackPos(back);
 
-    if (std::max(min.downtrack, start) > std::min(max.downtrack, end))
-    {  // Check for 1d intersection
-      // No intersection so continue
-      continue;
+    if (!bounds_inclusive) // reduce bounds slightly to avoid including exact bounds
+    {
+      if (std::max(min.downtrack, start + 0.1) > std::min(max.downtrack, end - 0.1))
```
msmcconnell, line 268: Let's use smaller bounds like 0.00001 for the comparison as 0.1 is a common increment size developers use. 
Additionally, the bound size needs to be documented in this methods comments to let users know that the non-inclusive behavior is not equivalent to a != check.

msmcconnell (changes requested):

carma_wm/src/CARMAWorldModel.cpp
MishkaMN: Reduced the number and added the comment in headerfile

carma_wm/src/CARMAWorldModel.cpp
MishkaMN: Switched the order and reversed other components that were affected

carma_wm/include/carma_wm/WorldModel.h
MishkaMN: Moved these functions to basic_autonomy library

cooperative_lanechange/include/cooperative_lanechange.h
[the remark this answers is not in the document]
MishkaMN: Lanechanges worked fine with this current logicat Summit Point. But I am planning to test again today with 2d version in places I think it's safe to use

basic_autonomy/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+
+add_library(basic_autonomy 
+  src/helper_functions.cpp
+)
+
+add_dependencies(basic_autonomy ${catkin_EXPORTED_TARGETS})
+target_link_libraries(basic_autonomy ${catkin_LIBRARIES})
+
+#target_link_libraries(basic_autonomy ${catkin_LIBRARIES} ${Boost_LIBRARIES})
+#add_dependencies(basic_autonomy ${catkin_EXPORTED_TARGETS} ${${PROJECT_NAME}_EXPORTED_TARGETS})
```
msmcconnell, line 71: Commented lines here can be deleted

basic_autonomy/include/basic_autonomy/helper_functions.h
```diff
@@ -0,0 +1,95 @@
...
+     * \param state The current vehicle state
+     * 
+     * \return index of nearest point in points
+     */
+    int get_nearest_point_index(const std::vector<PointSpeedPair>& points,
+                                                 const cav_msgs::VehicleState& state);
+    /**
+     * \brief Returns the nearest point to the provided vehicle pose in the provided list by utilizing the downtrack measured along the route
+     * This function compares the downtrack, provided by routeTrackPos, of each points in the list to get the closest one to the given downtrack.
+     *
```
msmcconnell, line 47: Should have same NOTE: section as the other two downtrack methods

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -263,10 +263,21 @@ std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double st
...
```
msmcconnell, line 268: Looks like you might have clobbered you previous fix for the bounds size here and additional comment describing this assumption.

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -570,24 +557,68 @@ std::vector<PointSpeedPair> InLaneCruisingPlugin::maneuvers_to_points(const std:
 
     ROS_DEBUG_STREAM("Used downtrack: " << starting_downtrack);
 
-    auto lanelets = wm->getLaneletsBetween(starting_downtrack, lane_following_maneuver.end_dist, true);
+    auto lanelets = wm->getLaneletsBetween(starting_downtrack, lane_following_maneuver.end_dist, true, true);
```
msmcconnell, line 560: If this were to be made non-inclusive what is the impact on the complex logic below? Could it simplify anything?

msmcconnell (changes requested):

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -589,17 +613,20 @@ namespace cooperative_lanechange
         std::vector<double> times;
         trajectory_utils::conversions::speed_to_time(downtracks, final_actual_speeds, &times);
 
+        // Remove extra points
+        ROS_DEBUG_STREAM("Before removing extra buffer points, future_geom_points.size()" << future_geom_points.size());
+        int end_dist_pt_index = basic_autonomy::waypoint_generation::get_nearest_index_by_downtrack(future_geom_points, wm_, ending_state_before_buffer_);
+        future_geom_points.resize(end_dist_pt_index);
```
adev4a, line 619: Should the resizing here and in the next two lines be to index + 1?

cooperative_lanechange/src/cooperative_lanechange.cpp
```diff
@@ -589,17 +613,20 @@ namespace cooperative_lanechange
...
+        times.resize(end_dist_pt_index);
+        final_yaw_values.resize(end_dist_pt_index);
+        ROS_DEBUG_STREAM("After removing extra buffer points, future_geom_points.size()" << future_geom_points.size());
+
         // Build trajectory points
         // TODO When more plugins are implemented that might share trajectory planning the start time will need to be based
```
adev4a, line 625: Looks like this was fixed by PR#1088 https://github.com/usdot-fhwa-stol/carma-platform/pull/1088.
This comment can be removed.

basic_autonomy/CMakeLists.txt
MishkaMN: deleted

basic_autonomy/include/basic_autonomy/helper_functions.h
MishkaMN: added the note

inlanecruising_plugin/src/inlanecruising_plugin.cpp
MishkaMN: I think it's best to keep that logic, as the input `max_starting_downtrack` passed to its parent function has that back_distance buffer, which would inevitably pick up lanechanging lanelets.

cooperative_lanechange/src/cooperative_lanechange.cpp
MishkaMN: that's a good catch. Changed to + 1

cooperative_lanechange/src/cooperative_lanechange.cpp
MishkaMN: Deleted

carma_wm/src/CARMAWorldModel.cpp
MishkaMN: added back the smaller size and comments in the header

cooperative_lanechange/include/cooperative_lanechange.h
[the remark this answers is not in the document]
msmcconnell: Ok I'll leave it up to you whether to use the 2d version or downtrack version. 2d is preferable but only if it is error free now.

msmcconnell (approved):

PR #1193, opened 2021-04-08
add basic autonomy lib

saina-ramyar:
# PR Details
## Description
Basic autonomy library is added to include the waypoint generation logic which is shared between different tactical plugins

## Types of changes

- [ ] Defect fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that cause existing functionality to change)

## Checklist:

- [x] I have added any new packages to the sonar-scanner.properties file
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
[CARMA Contributing Guide](https://github.com/usdot-fhwa-stol/carma-platform/blob/develop/Contributing.md) 
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Issue #1170, opened 2021-03-22
Sync inlane-cruising fixes with other plugins using the same algorithm. 


msmcconnell:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

The inlane-cruising plugin on 3.5.0 contains a number of fixes that will likely be needed in other components such as lane change.
These need to be synced with the 3.5.0 release since they contain some copy and pasted code. The best solution here would be to refactor out the dependency or have those components call inlane-cruising somehow
### Carma version where this issue was discovered

develop after 3.5.0
### Expected behavior
Duplicated algorithms do not have duplicated bugs. 
### Steps to reproduce the actual behavior

1. Do this
1. Then do this...

PR #1313, opened 2021-06-16
Feature/basic autonomy lanechange

adev4a:
# PR Details

## Description
This PR adds the basic autonomy library for trajectory planning. 

Inlane cruising, cooperative_lanechange and unobstructed_lanechange are refactored to make use of this library.
Since the lane following and lane change algorithms are considerably different at the time this of this pr, the larger trajectory composing functions are separate for the 2 different types of maneuver.

## How Has This Been Tested?

These changes were unit tested.

## Types of changes

## Checklist:


msmcconnell:
This PR should point at develop not master

msmcconnell:
Linking this PR to this relevant additional issue 
https://github.com/usdot-fhwa-stol/carma-platform/issues/1170

msmcconnell:
Just cross posting my comment from Saina's PR which was the original PR for this library
https://github.com/usdot-fhwa-stol/carma-platform/pull/1193#issuecomment-850487495

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -0,0 +1,880 @@
...
+         std::vector<PointSpeedPair> create_geometry_profile(const std::vector<cav_msgs::Maneuver> &maneuvers, double max_starting_downtrack,const carma_wm::WorldModelConstPtr &wm,
+                                                                   cav_msgs::VehicleState &ending_state_before_buffer,const cav_msgs::VehicleState& state,
+                                                                   const GeneralTrajConfig &general_config, const DetailedTrajConfig &detailed_config){
+            std::vector<PointSpeedPair> points_and_target_speeds;
+            
+            bool first = true;
+            ROS_DEBUG_STREAM("VehDowntrack:"<<max_starting_downtrack);
+            for(const auto &maneuver : maneuvers)
+            {
+                if(std::find(lane_follow_plugins_list.begin(), lane_follow_plugins_list.end(), general_config.trajectory_type) !=lane_follow_plugins_list.end() && maneuver.type != cav_msgs::Maneuver::LANE_FOLLOWING)
```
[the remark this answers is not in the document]
kjrush: Hmm, also trajectory_type seems to be a bit of a confusing name here. If it needs to match a planning plugin's name, the variable name should reflect that somehow. I'm also not 100% sure about this pattern or if it's necessary, let's have a quick chat on this to think it through and so I can make sure I understand.

Also the exception below refers to "lane following tactical plugin" when in theory this error is a result of no plugins supporting the maneuver type right?

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -0,0 +1,880 @@
...
+                {
+                    throw std::invalid_argument("lane following tactical plugin doesn't support this maneuver type");
+                }
+                else if(std::find(lane_change_plugins_list.begin(), lane_change_plugins_list.end(), general_config.trajectory_type) !=lane_change_plugins_list.end() && maneuver.type != cav_msgs::Maneuver::LANE_CHANGE)
```
kjrush, line 37: Same comments as above.

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -0,0 +1,880 @@
...
+
+            }
+            return points_and_target_speeds;
+
+        }
+
+        std::vector<PointSpeedPair> create_lanefollow_geometry(const cav_msgs::Maneuver &maneuver, double starting_downtrack,
+                                                                   const carma_wm::WorldModelConstPtr &wm, cav_msgs::VehicleState &ending_state_before_buffer,
+                                                                    const GeneralTrajConfig &general_config, const DetailedTrajConfig &detailed_config)
+        {
```
kjrush, line 77: Might not hurt to defensively check here that the config's maneuver type matches lane follow, just to be sure.

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -0,0 +1,880 @@
...
+                //tpp.planner_plugin_name        //Planner plugin name is filled in the tactical plugin
+
+                traj.push_back(tpp);
+            }
+
+            return traj;
+        }
+
+
+        std::vector<PointSpeedPair> attach_back_points(const std::vector<PointSpeedPair> &points,
```
kjrush, line 484: Doesn't have to change, but "attach_back_points" is also kind of a weak name. What are "back_points" and why do I need them?

cooperative_lanechange/config/parameters.yaml
```diff
@@ -86,6 +86,15 @@ mid_fraction : 0.5
...
-ending_buffer_downtrack : 5.0
+##Unused parameters added for using basic autonomy library
+#Int: Amount to downsample input lanelet centerline data on turns. Value corresponds to saving each nth point.
+turn_downsample_ratio: 0
+
+#Double: Curve re-sampling step size in m
+curve_resample_step_size: 1.0
+
+#Double: Number of meters behind the first maneuver that need to be included in points for curvature calculation
+back_distance: 0.0
```
kjrush, line 97: Still not a huge fan of this name.

unobstructed_lanechange/config/parameters.yaml
```diff
@@ -67,6 +67,20 @@ enable_object_avoidance_lc : false
 #Units: seconds
 min_timestep : 0.1
 
+<<<<<<< HEAD
```
kjrush, line 70: Whoa, careful with these merge conflict resolutions. Definitely want to make sure we clean this up, looks like the unresolved conflict got committed and pushed. I'm guessing we want to take the HEAD branch here, but take a look at it please.

kjrush:
Also if possible, It'd be nice to make sure we're not duplicating the map data used for unit tests. Having a +55k addition that we might not strictly need is less than ideal. If it's possible to use the prior ones, i'd recommend that but let me know if I'm off base.

basic_autonomy/src/basic_autonomy.cpp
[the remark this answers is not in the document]
adev4a: This list is local to the basic autonomy class. Since the plugins are either following a lane follow planning approach or lane change, the list just defines which plugin will follow which one. 
This is needed because we're checking if the correct maneuver request was sent to the plugin. I think this could be done in the tactical plugin as well. In case you think that's cleaner.

basic_autonomy/src/basic_autonomy.cpp
adev4a: Would this be the same check as line 33 and line 37?

kjrush (approved):

PR #1316, opened 2021-06-21
Fix/near zero comparison

adev4a:
# PR Details
## Description
This PR fixes an issue with get_nearest_index_by_downtrack in basic autonomy where a query for a target 0.0 downtrack results in an exception. 
Adding this value results in an assumptions that the points being queried are at least 0.1mm apart, which seems reasonably small to allow.

## How Has This Been Tested?

These changes were unit tested and integration tested.
## Types of changes

## Checklist:

### DIFF
--- a/basic_autonomy/src/helper_functions.cpp
+++ b/basic_autonomy/src/helper_functions.cpp
@@ -65,9 +65,9 @@ namespace waypoint_generation
     int get_nearest_index_by_downtrack(const std::vector<lanelet::BasicPoint2d>& points, const carma_wm::WorldModelConstPtr& wm, double ending_downtrack)
     {
         int best_index = points.size() - 1;
-        for(int i=0;i < points.size();i++){
+        for(int i = 0;i < points.size(); i++){
             double downtrack = wm->routeTrackPos(points[i]).downtrack;
-            if(downtrack > ending_downtrack){
+            if(downtrack > ending_downtrack + epsilon_){
                 best_index = i - 1;
                 if (best_index < 0)
                 {
--- a/basic_autonomy/src/helper_functions.cpp
+++ b/basic_autonomy/src/helper_functions.cpp
@@ -67,12 +67,10 @@ namespace waypoint_generation
         int best_index = points.size() - 1;
         for(int i = 0;i < points.size(); i++){
             double downtrack = wm->routeTrackPos(points[i]).downtrack;
-            if(downtrack > target_downtrack + epsilon_){
-                best_index = i - 1;
-                if (best_index < 0)
-                {
-                throw std::invalid_argument("Given points are beyond the ending_downtrack");
-                }
+            if(downtrack > target_downtrack){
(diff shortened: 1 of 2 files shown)


msmcconnell:
@adev4a Is this meant to be pointed at develop? If so can you put up a zephyr->develop PR first so we do not rereview old code

adev4a:
> Switched the base to zephyr
> 
> @adev4a Is this meant to be pointed at develop? If so can you put up a zephyr->develop PR first so we do not rereview old code

basic_autonomy/src/helper_functions.cpp
```diff
@@ -62,12 +62,12 @@ namespace waypoint_generation
...
 
-    int get_nearest_index_by_downtrack(const std::vector<lanelet::BasicPoint2d>& points, const carma_wm::WorldModelConstPtr& wm, double ending_downtrack)
+    int get_nearest_index_by_downtrack(const std::vector<lanelet::BasicPoint2d>& points, const carma_wm::WorldModelConstPtr& wm, double target_downtrack)
     {
         int best_index = points.size() - 1;
-        for(int i=0;i < points.size();i++){
+        for(int i = 0;i < points.size(); i++){
             double downtrack = wm->routeTrackPos(points[i]).downtrack;
-            if(downtrack > ending_downtrack){
+            if(downtrack > target_downtrack + epsilon_){
```
msmcconnell, line 70: Can you provide a description of what error this change is fixing? Seems you are accounting for floating point arithmetic errors, but since this doesn't have an == check I'm a bit unclear on why not having the epsilon is an issue

adev4a: This was resulting in an exception when the downtrack was really small (to the order of 10e-13) and still going through this loop because the target downtrack was exactly 0.0

msmcconnell: I guess my preference in that case would be for us to have the system return 0 when no match is found rather than throwing an exception and needing this epsilon. A point located before the input would give negative downtrack and should return 0 index, which makes sense as that would be the nearest downtrack point. Seems like that could be easily implemented by changing the default value of best_index to 0. The comments should be updated to reflect that and the fact this is not a nearest function but actually a nearest less than method.

adev4a: The way the function is written, shouldn't the best index be initialized to point.size() -1 like it currently is?
We are checking against downtrack>target_downtrack and the only time best_index will not be changed is if we go through the entire list without meeting that condition. So the nearest point would be the last one.

This would be in addition to replacing the exception for negative values with a return 0.

msmcconnell: Yeah I think you are right.

msmcconnell (approved):
@adev4a I'm good with this, but please have @MishkaMN take a look before merging so he can approve and test it.

MishkaMN (approved):

Issue #1157, opened 2021-03-17
Route following params misloaded


adev4a:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
The config params from route following are being put into the guidance namespace instead of a route_following namespace.
This means that the variables loading these params are only using default values.
See comment here,
https://github.com/usdot-fhwa-stol/carma-platform/pull/1156#discussion_r595980898

### Expected behavior
The params should be mapped to route following's namespace and changing params in the config file should result in a change 
the loaded value.
### Steps to reproduce the actual behavior

1.Change the value of minimal_maneuver_duration to 20
2. Print the value of variable mvr_duration (which is supposed to read this param)

msmcconnell:
Resolved by above PR

PR #1323, opened 2021-06-24
Fix/route following parameters

adev4a:
# PR Details
Fix for Issue #1157 
## How Has This Been Tested?
Ros test and integration test

## Types of changes

## Checklist:


route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -47,7 +47,7 @@ namespace route_following_plugin
 
         // read ros parameters
         pnh_->param<double>("minimal_plan_duration", min_plan_duration_, 16.0);
-        pnh_->param<std::string>("lane_change_plugin", lane_change_plugin_);
+        pnh_->param<std::string>("lane_change_plugin", lane_change_plugin_, "CooperativeLaneChangePlugin");
```
msmcconnell, line 50: If this wasn't being loaded properly before and had no default value how was it ever working?

adev4a: I dont think it was. The variable is initialized in the header, so it was using that value as a constant

msmcconnell (changes requested):
Per @snallamothu we will hold off on merging this change for zephyr and will target it to develop as we do not have time to verify before Friday and its not use case critical functionality.

msmcconnell (approved):

PR #1348, opened 2021-07-15
Feature/add intersect trans

lewisid:
# PR Details
## Description

 This PR documents the implementation of the Work Zone use case node to handle Intersection Transit Maneuver messages. New files and folders have been added to carma-platform to assist in acquiring these messages, converting them into Lane Following maneuvers, and passing them to the In-Lane Cruising Plugin. 
## Related Issue

https://github.com/usdot-fhwa-stol/carma-platform/issues/1366#

## Motivation and Context

The Work Zone use case is in need of a node that can process intersection transit maneuvering messages sent from Carma Cloud. This plugin will be able to discern when such messages are being used and will convert those messages into lane following maneuvers to be passed into in-lane cruising. From there, the plugin will be able to provide trajectory calculations for the vehicle to follow.

## How Has This Been Tested?

This has been unit tested using gtest/gmock.

## Types of changes

## Checklist:


intersection_transit_maneuvering/src/intersection_transit_maneuvering.cpp
```diff
@@ -0,0 +1,214 @@
...
+                maneuver.type != cav_msgs::Maneuver::INTERSECTION_TRANSIT_LEFT_TURN ||
+                maneuver.type != cav_msgs::Maneuver::INTERSECTION_TRANSIT_RIGHT_TURN ||)
+                {
+                    throw std::invalid_argument("Intersection transit maneuvering does not support this maneuver type");
+                }
+            
+            /*Convert IT Straight*/
+            if (maneuver.type == cav_msgs::Maneuver::INTERSECTION_TRANSIT_STRAIGHT)
+            {
+                new_maneuver.parameters = maneuver.intersection_transit_straight_maneuver.parameters;
```
msmcconnell, line 152: probably best to clear the presence_vector and meta data fields as these may not be applicable to inlane-cruising.

I'm going back and forth on whether we should set a new maneuver id as well, it would help for the id to be persistent to support logging, but could potentially lead to issues if misinterpreted. Check with @kjrush if the forwarded lane follow maneuver should have its own maneuver id.

intersection_transit_maneuvering/src/intersection_transit_maneuvering.cpp
```diff
@@ -0,0 +1,214 @@
...
+        pnh_->param<double>("max_jerk_limit", max_jerk_limit_);
+        pnh_->param<double>("min_timestep",min_timestep_);
+        pnh_->param<double>("min_jerk", min_jerk_limit_);
+
+        discovery_pub_timer_ = pnh_->createTimer(
+            ros::Duration(ros::Rate(10.0)),
+            [this](const auto&) { 
+                plugin_discovery_pub_.publish(plugin_discovery_msg_);
+                std_msgs::Float64 jerk_msg;
+                jerk_msg.data = jerk_;
```
msmcconnell, line 76: This jerk message is not needed and probably a copy paste error.

intersection_transit_maneuvering/package.xml
```diff
@@ -0,0 +1,41 @@
...
+  <depend>roscpp</depend>
+  <depend>std_msgs</depend>
+  <depend>cav_srvs</depend>
+  <depend>cav_msgs</depend>
+  <depend>carma_utils</depend>
+  <depend>autoware_msgs</depend>
+  <depend>carma_wm</depend>
+  <depend>tf</depend>
+  <depend>tf2</depend>
+  <depend>tf2_geometry_msgs</depend>
```
msmcconnell, line 36: Seems most of these dependencies are not needed

msmcconnell (changes requested):
I think the overall approach looks good, however, there are numerous specific errors, including build errors. Please get this to the point where is compiles, has the ROS service call implemented, is commented cleanly, and the numerous copy paste errors are removed and then I'll take another look.

intersection_transit_maneuvering/src/intersection_transit_maneuvering.cpp
```diff
@@ -0,0 +1,233 @@
...
+        auto new_plan = convert_maneuver_plan(maneuver_plan);
+
+        for(const auto& man : new_plan)
+            {
+                req2.maneuver_plan.push_back(man);
+            }
+
+        /*Once the maneuvers have been successfully converted, call the inlanecruising plugin to calculate the trajectory*/
+
+        if (trajectory_client_ && trajectory_client_.exists() && trajectory_client_.isValid())
```
msmcconnell, line 81: This logic cannot be unit tested because it requires an active ros connection. Looks like this is based on the yield call from inlane-cruising which as we discussed is not a good approach. You want to follow the pattern used for publishers in that class. This means using dependency injection, you want to pass in an object to your IntersectionTransitManeuver constructor that could provide a call() method. There are two common ways to do this, either use std::function which matches the ros::Service::call api or define a c++ interface with a call method that you can create different implementations for.

intersection_transit_maneuvering/test/test_intersection_transit_maneuvering.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    maneuvers.push_back(man1);
+    maneuvers.push_back(man2);
+    maneuvers.push_back(man3);
+
+    auto converted2 = itm_node.convert_maneuver_plan(maneuvers);
+
+    for(auto i: maneuvers)
+    {
+        ASSERT_EQ(true, i.maneuver.type == cav_msgs::Maneuver::LANE_FOLLOWING); //Test that each maneuver has been converted to LANE_FOLLOWING
```
msmcconnell, line 140: The tests here should be a bit more robust. Please add lines to check the other fields in the output message. Since your node is primarily a message converter this is the core thing that needs to be validated.

msmcconnell (changes requested):

intersection_transit_maneuvering/CMakeLists.txt
```diff
@@ -0,0 +1,119 @@
...
+  roscpp
+  std_msgs
+  cav_msgs
+  cav_srvs
+  carma_utils
+  trajectory_utils
+  autoware_msgs
+  carma_wm
+  carma_debug_msgs
+  basic_autonomy
```
msmcconnell, line 35: Any deps which are not used should be removed

intersection_transit_maneuvering/CMakeLists.txt
```diff
@@ -0,0 +1,119 @@
...
+    test/test_intersection_transit_maneuvering.cpp
+    test/test_main.cpp
+)
+#catkin_add_gtest(${PROJECT_NAME}-test test/test_intersection_transit_maneuvering.cpp)
+if(TARGET ${PROJECT_NAME}-test)
+ target_link_libraries(${PROJECT_NAME}-test intersection_transit_maneuvering_library ${catkin_LIBRARIES})
+endif()
+
+## Add folders to be run by python nosetests
+# catkin_add_nosetests(test)
```
msmcconnell, line 119: No need for this unused comment block

intersection_transit_maneuvering/include/call_test.h
```diff
@@ -0,0 +1,76 @@
...
+         */
+        CallTest();
+
+        /**
+         * @brief Test call() function for unit testing
+         * 
+         * @param req Incoming PlanTrajectory service request 
+         * @param resp Incoming PlanTrajectory service response
+         * @return true 
+         * @return false
```
msmcconnell, line 49: For @return comments you would normal describe the return behavior not list the possible values. 
So something like 
@return true if method completes successfully, false otherwise

intersection_transit_maneuvering/include/intersection_transit_maneuvering.h
```diff
@@ -0,0 +1,96 @@
...
+#include <cav_msgs/IntersectionTransitRightTurnManeuver.h>
+#include <cav_msgs/IntersectionTransitStraightManeuver.h>
+#include <cav_srvs/PlanTrajectory.h>
+#include <carma_wm/WMListener.h>
+#include <functional>
+#include <unordered_set>
+#include <autoware_msgs/Lane.h>
+#include <ros/ros.h>
+#include <itm_helper.h>
+#include <basic_autonomy/helper_functions.h>
```
msmcconnell, line 38: Any includes which are not used should be removed

intersection_transit_maneuvering/include/intersection_transit_maneuvering.h
```diff
@@ -0,0 +1,96 @@
...
+    {
+        public:
+        /**
+        * \brief Constructor
+        * 
+        * \param wm Pointer to intialized instance of the carma world model for accessing semantic map data
+        * \param plugin_discovery_publisher Callback which will publish the current plugin discovery state
+        * \param obj Interface object to initialize ros::Service::call
+        */ 
+        IntersectionTransitManeuvering(carma_wm::WorldModelConstPtr& wm, PublishPluginDiscoveryCB plugin_discovery_publisher, std::shared_ptr<Interface> obj);
```
msmcconnell, line 56: The carma_wm::WorldModelConstPtr& should not be passed as reference (drop the &) because this is a smart pointer (shared_ptr) and they should be passed directly to ensure proper reference counting occurs. 
Additionally, passing a pointer by reference has basically no efficiency benefit as a pointer is already a very small element to pass into a method.

intersection_transit_maneuvering/include/intersection_transit_maneuvering.h
```diff
@@ -0,0 +1,96 @@
...
+        /**
+         *  \brief Converts a sequence of INTERSECTION_TRANSIT maneuvers to LANE_FOLLOWING maneuvers
+         * 
+         * \param maneuvers The list of maneuvers to convert
+         * 
+         * \return The new list of converted maneuvers
+        */
+        std::vector<cav_msgs::Maneuver> convert_maneuver_plan(const std::vector<cav_msgs::Maneuver>& maneuvers);
+
+        bool onSpin();
```
msmcconnell, line 77: Add method comment

intersection_transit_maneuvering/include/intersection_transit_maneuvering.h
```diff
@@ -0,0 +1,96 @@
...
+
+    private:
+
+        std::shared_ptr<Interface> object;
+        std::vector<cav_msgs::Maneuver> converted_maneuvers;
+        cav_msgs::TrajectoryPlan trajectory_plan;
+        cav_msgs::VehicleState vehicle_state_;
+        cav_msgs::Plugin plugin_discovery_msg_;
+        PublishPluginDiscoveryCB plugin_discovery_publisher_;
```
msmcconnell, line 86: Some of these member variables have trailing "_" to denote class member status while others do not. Please be consistent.

intersection_transit_maneuvering/include/intersection_transit_maneuvering_node.h
```diff
@@ -0,0 +1,82 @@
...
+
+            // ROS service servers
+            ros::ServiceServer trajectory_srv_;
+
+            //ROS publishers and subscribers
+            ros::Publisher plugin_discovery_pub_;
+            ros::Subscriber pose_sub_;
+            ros::Subscriber twist_sub_;
+            ros::Publisher jerk_pub_;
+            ros::Timer discovery_pub_timer_;
```
msmcconnell, line 52: Seems there are a number of unused variables in this file that can be removed

intersection_transit_maneuvering/include/intersection_transit_maneuvering_node.h
```diff
@@ -0,0 +1,82 @@
...
+  
+            plugin_discovery_pub_ = nh_.advertise<cav_msgs::Plugin>("plugin_discovery",1);
+            carma_wm::WMListener wml_;
+
+            carma_wm::WorldModelConstPtr wm_ = wml_.getWorldModel();
+
+            std::shared_ptr<itm_servicer::Servicer> srv;
+            IntersectionTransitManeuvering worker(wm_,[&plugin_discovery_pub_](const auto& msg) {plugin_discovery_pub_.publish(msg);}, srv);
+            ros::ServiceClient trajectory_client = nh_.serviceClient<cav_srvs::PlanTrajectory>("plugin/InlaneCruisingPlugin/plan_trajectory");
+            srv->set_client(trajectory_client);
```
msmcconnell, line 68: For clarity can you put this call before the worker constructor, so that srv is fully initialized before it is passed in.

intersection_transit_maneuvering/include/itm_helper.h
```diff
@@ -0,0 +1,38 @@
...
+#include <trajectory_utils/trajectory_utils.h>
+#include <trajectory_utils/conversions/conversions.h>
+#include <sstream>
+#include <carma_utils/containers/containers.h>
+#include <Eigen/Core>
+#include <Eigen/Geometry>
+#include <Eigen/LU>
+#include <Eigen/SVD>
+
+class Interface
```
msmcconnell, line 34: This class name is not descriptive. Please name the interface something meaningful.

msmcconnell: Also name the file the same as the interface to continue the pattern of one class per file

intersection_transit_maneuvering/include/itm_service.h
```diff
@@ -0,0 +1,57 @@
...
+#include <cav_msgs/Plugin.h>
+#include <boost/shared_ptr.hpp>
+#include <carma_utils/CARMAUtils.h>
+#include <boost/geometry.hpp>
+#include <carma_wm/Geometry.h>
+#include <cav_srvs/PlanTrajectory.h>
+#include <itm_helper.h>
+
+
+namespace itm_servicer
```
msmcconnell, line 31: It's not really an issue, but there is no need to have different namespaces between files in the same package unless you specifically want to convey meaning with the extra name. You could have all your files use the intersection_transit_maneuvering namespace

intersection_transit_maneuvering/package.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0"?>
+
+<!--  
+ Copyright (C) 2018-2020 LEIDOS.
```
msmcconnell, line 4: All new file copyrights should be 2021

intersection_transit_maneuvering/src/call_test.cpp
```diff
@@ -0,0 +1,29 @@
+#include "call_test.h"
+#include <ros/ros.h>
+
+namespace call_test
+{
+
+    CallTest::CallTest(){};
```
msmcconnell, line 7: When you have an empty constructor you can actually just have CallTest() = default; in the header file.

intersection_transit_maneuvering/src/intersection_transit_maneuvering.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+namespace intersection_transit_maneuvering
+{
+    IntersectionTransitManeuvering::IntersectionTransitManeuvering(carma_wm::WorldModelConstPtr& wm, PublishPluginDiscoveryCB plugin_discovery_publisher,
+                                                                     std::shared_ptr<Interface> obj)
+    {        
+        plugin_discovery_msg_.name = "IntersectionTransitManeuvering";
+        plugin_discovery_msg_.versionId = "v1.0";
+        plugin_discovery_msg_.available = true;
+        plugin_discovery_msg_.activated = false;
```
msmcconnell, line 45: This flag should be set to true when your plan_trajectory callback is triggered.

intersection_transit_maneuvering/include/itm_helper.h
```diff
@@ -0,0 +1,38 @@
...
+{
+    public:
+        virtual bool call(cav_srvs::PlanTrajectoryRequest& req, cav_srvs::PlanTrajectoryResponse& resp) = 0;
```
msmcconnell, line 37: Whenever you have a class with only pure virtual methods in c++ it is important to define an implementation for the destructor even if it is empty. This enforces proper sequence of destructor calls in child classes. 
In this case it would look like
virtual ~Interface() {};

intersection_transit_maneuvering/src/intersection_transit_maneuvering.cpp
```diff
@@ -0,0 +1,192 @@
...
+            if(req.maneuver_plan.maneuvers[i].type == cav_msgs::Maneuver::INTERSECTION_TRANSIT_STRAIGHT ||
+            req.maneuver_plan.maneuvers[i].type == cav_msgs::Maneuver::INTERSECTION_TRANSIT_LEFT_TURN || 
+            req.maneuver_plan.maneuvers[i].type == cav_msgs::Maneuver::INTERSECTION_TRANSIT_RIGHT_TURN )
+            {
+                maneuver_plan.push_back(req.maneuver_plan.maneuvers[i]);
+                resp.related_maneuvers.push_back(i);
+            }
+            else
+                {
+                    throw std::invalid_argument("Invalid Maneuver Type");
```
msmcconnell, line 68: You should break here instead of throw an exception, because there may be planned maneuvers the intersection which will be passed to different plugins after your plugin completes. So you want to break out of this loop once you have accounted for your related maneuvers. Throwing an exception by contrast would shutdown the whole system. 

It would be appropriate to throw an exception only if req.maneuver_plan.maneuvers[req.maneuver_index_to_plan] was not an intersection maneuver.

intersection_transit_maneuvering/src/intersection_transit_maneuvering.cpp
```diff
@@ -0,0 +1,192 @@
...
+                }
+        }
+
+        converted_maneuvers = convert_maneuver_plan(maneuver_plan);
+        cav_srvs::PlanTrajectoryRequest req2;
```
msmcconnell, line 73: variable name could be more meaningful

intersection_transit_maneuvering/src/itm_service.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include <carma_wm/Geometry.h>
+#include <cav_srvs/PlanTrajectory.h>
+namespace itm_servicer
+{
+
+    Servicer::Servicer(){};
+
+    bool Servicer::call(cav_srvs::PlanTrajectoryRequest& req, cav_srvs::PlanTrajectoryResponse& resp)
+    {
+        client.call(req,resp);
```
msmcconnell, line 19: This method returns bool but no return call is made. Presumably you want 
return client.call

msmcconnell (changes requested):

intersection_transit_maneuvering/include/intersection_transit_maneuvering.h
```diff
@@ -0,0 +1,97 @@
...
+         * 
+         * @return true if successful, otherwise false
+         */
+        bool onSpin();
+
+    private:
+
+        std::shared_ptr<CallInterface> object_;
+        std::vector<cav_msgs::Maneuver> converted_maneuvers_;
+        //cav_msgs::TrajectoryPlan trajectory_plan_;
```
msmcconnell, line 84: If a variable is unused it can just be deleted, no need to comment it out.

intersection_transit_maneuvering/src/intersection_transit_maneuvering.cpp
```diff
@@ -113,7 +113,11 @@ namespace intersection_transit_maneuvering
             /*Convert IT Straight*/
             if (maneuver.type == cav_msgs::Maneuver::INTERSECTION_TRANSIT_STRAIGHT)
             {
-                new_maneuver.lane_following_maneuver.parameters = maneuver.intersection_transit_straight_maneuver.parameters;
+                new_maneuver.lane_following_maneuver.parameters.maneuver_id = maneuver.intersection_transit_straight_maneuver.parameters.maneuver_id;
+                new_maneuver.lane_following_maneuver.parameters.planning_strategic_plugin = maneuver.intersection_transit_straight_maneuver.parameters.planning_strategic_plugin;
+                new_maneuver.lane_following_maneuver.parameters.planning_tactical_plugin = maneuver.intersection_transit_straight_maneuver.parameters.planning_tactical_plugin;
+                new_maneuver.lane_following_maneuver.parameters.presence_vector = maneuver.intersection_transit_straight_maneuver.parameters.presence_vector;
```
msmcconnell: Presence vector can be set to cav_msgs::ManeuverParameters::HAS_TACTICAL_PLUGIN to ensure the meta data fields are not accessed. Same for the other maneuver cases in this file.

intersection_transit_maneuvering/src/intersection_transit_maneuvering.cpp
```diff
@@ -123,6 +127,9 @@ namespace intersection_transit_maneuvering
                 new_maneuver.lane_following_maneuver.end_speed = maneuver.intersection_transit_straight_maneuver.end_speed;
                 new_maneuver.lane_following_maneuver.end_time = maneuver.intersection_transit_straight_maneuver.end_time;
 
+                new_maneuver.lane_following_maneuver.parameters.bool_valued_meta_data.clear();
+                new_maneuver.lane_following_maneuver.parameters.float_valued_meta_data.clear();
```
msmcconnell, line 130: Since you made a new message, the vectors will already be empty. No need to clear them unless you copied the data over like you were doing before.

msmcconnell (changes requested):

msmcconnell (approved):

Issue #1351, opened 2021-07-16
Route Following doesn't update speed limit if route not updated


adev4a:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

The current implementation of route following creates a maneuver plan when the route is set. Before passing the maneuver to the response the starting speed is updated but the ending speed is not changed.

If a geofence message is published to update the speed limit without changing the route, the route_cb will not be triggered and the speed limit will remain unchanged from before. 

### Expected behavior
The speed limit in the maneuver should reflect the current speed limit at all times.
### Steps to reproduce the actual behavior

1. Publish a geofence message to update only speed limit of lanelet without changing route
2. Check speed limit on UI and compare with actual speed of vehicle when passing through.

msmcconnell:
Resolved by above PR

PR #1352, opened 2021-07-16
Fix/regenerate maneuvers on map update

msmcconnell:
# PR Details
## Description

This PR does two things. 
First, it fixes a bug in carma_wm where the user defined map update callback would only trigger when an entirely new map was received. Now the callback will be triggered either when a new map is received or an update is applied successfully. 

Second, it uses this callback to trigger a maneuver regeneration in the route_following_plugin if a route is also available. The previous temporary logic to handle this case was also removed per discussion with @adev4a 

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1351

## Motivation and Context
Working basic travel use case

## How Has This Been Tested?
build and unit testing

## Types of changes

## Checklist:

### DIFF
--- a/route_following_plugin/src/route_following_plugin.cpp
+++ b/route_following_plugin/src/route_following_plugin.cpp
@@ -329,35 +329,6 @@ namespace route_following_plugin
         }
     }
 
-    void RouteFollowingPlugin::updateEndingSpeed(cav_msgs::Maneuver &maneuver, double end_speed) const
-    {
-        ROS_DEBUG_STREAM("end_speed: " << end_speed);
-        switch (maneuver.type)
-        {
-        case cav_msgs::Maneuver::LANE_FOLLOWING:
-            maneuver.lane_following_maneuver.end_speed = end_speed;
-            ROS_DEBUG_STREAM("end_speed updated for lane following: " << end_speed);
-            break;
-        case cav_msgs::Maneuver::LANE_CHANGE:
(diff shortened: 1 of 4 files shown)


adev4a (commented):
UpdateEndingSpeed and findLaneletIndexFromPath functions are unused after these changes and should be removed. findLaneletIndexFromPath should also be removed from unit tests

msmcconnell:
> UpdateEndingSpeed and findLaneletIndexFromPath functions are unused after these changes and should be removed. findLaneletIndexFromPath should also be removed from unit tests

Updated and unit tests are passing

route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -59,6 +59,12 @@ namespace route_following_plugin
             this->latest_maneuver_plan_ = routeCb(wm_->getRoute()->shortestPath());
         });
 
+        wml_->setMapCallback([this]() {
+            if (wm_->getRoute()) { // If this map update occured after a route was provided we need to regenerate maneuvers
+                this->latest_maneuver_plan_ = routeCb(wm_->getRoute()->shortestPath());
```
adev4a, line 64: Do we want a log statement here as well? To indicate what callback is causing the maneuver plan to be replanned

msmcconnell: Done

PR #1355, opened 2021-07-19
Feature/traffic signal

aswath1:
# PR Details
## Description
The DSRC spat message is picked up by the WM interface and the map is updated with the traffic light data with real time update. The map should be updated at a higher frequency to accurately track the traffic light. The map is updated and then send to the guidance. When the guidance sees the stop sign. The stop and wait manure is invoked with the help of a strategic plugin. When the light turns green, the intersection transit manure is invoked. Note: we already have the lane keeping straight which just needs to be replaced with intersection transit. At the exit, the guidance will just switch to lane following or any plugin based on the situation. 

## Types of changes

## Checklist:


carma_wm/src/WMListener.cpp
```diff
@@ -35,6 +36,7 @@ WMListener::WMListener(bool multi_thread) : worker_(std::unique_ptr<WMListenerWo
   map_sub_ = nh_.subscribe("semantic_map", 2, &WMListenerWorker::mapCallback, worker_.get());
   route_sub_ = nh_.subscribe("route", 1, &WMListenerWorker::routeCallback, worker_.get());
   roadway_objects_sub_ = nh_.subscribe("roadway_objects", 1, &WMListenerWorker::roadwayObjectListCallback, worker_.get());
+  traffic_spat_sub_ = nh_.subscribe("incoming_spat", 1, &WMListenerWorker::incomingSpatCallback, worker_.get());
```
msmcconnell, line 39: Since the system might operate in the presence of multiple intersections in the future, you should use a larger queue here to account for the possibility of multiple SPaT sources. I think 10 would probably be sufficient as I doubt we will encounter more unique intersections than that at any point.

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -93,6 +95,44 @@ void WMListenerWorker::mapCallback(const autoware_lanelet2_msgs::MapBinConstPtr&
...
+                           ", and signal_group_id: " << current_movement_state.signal_group);
+        continue;
+      }
+      auto general_regem = world_model_->getMutableMap()->regulatoryElementLayer.get(curr_light_id);
+      
+      // get ptr to the regem in the map
+      lanelet::CarmaTrafficLightPtr curr_light = world_model_->getMutableMap()->laneletLayer.findUsages(general_regem)[0].regulatoryElementsAs<lanelet::CarmaTrafficLight>()[0];
+
+      // check if we have processed this already or not
+      if (curr_light->revision_ == curr_intersection.revision)
```
msmcconnell, line 122: I don't think the revision field works this way. I believe it only changes to the intersection structure not a new spat message. Per what we had discussed before you might need to see multiple Spat messages before the full phase is described.

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -93,6 +95,44 @@ void WMListenerWorker::mapCallback(const autoware_lanelet2_msgs::MapBinConstPtr&
...
+      if (curr_light_id == lanelet::InvalId)
+      {
+        ROS_WARN_STREAM("Received a SPAT message for traffic light that is not in the map with intersection_id: " << curr_intersection.id.id << 
+                           ", and signal_group_id: " << current_movement_state.signal_group);
+        continue;
+      }
+      auto general_regem = world_model_->getMutableMap()->regulatoryElementLayer.get(curr_light_id);
+      
+      // get ptr to the regem in the map
+      lanelet::CarmaTrafficLightPtr curr_light = world_model_->getMutableMap()->laneletLayer.findUsages(general_regem)[0].regulatoryElementsAs<lanelet::CarmaTrafficLight>()[0];
```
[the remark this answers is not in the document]
msmcconnell: In such cases, the Spat message should be dropped until the regulatory element is available.

msmcconnell (changes requested):
For this initial review, I have ignored minor issues, like comments or method names etc.

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -93,6 +95,44 @@ void WMListenerWorker::mapCallback(const autoware_lanelet2_msgs::MapBinConstPtr&
...
+    for (auto current_movement_state : curr_intersection.movement_list) 
+    {
+      lanelet::Id curr_light_id = world_model_->getTrafficLightId(curr_intersection.id.id, current_movement_state.signal_group);
+      if (curr_light_id == lanelet::InvalId)
+      {
+        ROS_WARN_STREAM("Received a SPAT message for traffic light that is not in the map with intersection_id: " << curr_intersection.id.id << 
+                           ", and signal_group_id: " << current_movement_state.signal_group);
+        continue;
+      }
+      auto general_regem = world_model_->getMutableMap()->regulatoryElementLayer.get(curr_light_id);
```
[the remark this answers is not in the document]
aswath1: it is expected that all ids are registered in the map when updates are applied in the map. so no new ids are generated at this stage, but just lookup of what is there. Also, Misheel branch creates the lookups.

carma_wm/src/WMListenerWorker.cpp
msmcconnell: @aswath1 This is the disconnect. 
You are treating MovementEvent's as if each MovementEvent corresponds to one phase state. However, movements refer to something like what I have in the picture below
image

Each movement is one possible direction of travel from a lane through the intersection. So the SPaT message is actually only communicating the end time of the current state. You need to therefore listen to the message for the duration of a full cycle in order to populate the total set of phase transitions.

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -1118,5 +1118,78 @@ std::vector<lanelet::CarmaTrafficLightPtr> CARMAWorldModel::predictTrafficLight(
...
+      {
+        ROS_INFO_STREAM("movement_event_list is empty . intersection_id: " << (int)curr_intersection.id.id << 
+                           ", and signal_group_id: " << (int)current_movement_state.signal_group);
+        continue;
+      }
+
+      traffic_light_states_[curr_intersection.id.id][current_movement_state.signal_group].push_back(std::make_pair(ros::Time(current_movement_state.movement_event_list[0].timing.min_end_time), 
+                              static_cast<lanelet::CarmaTrafficLightState>(current_movement_state.movement_event_list[0].event_state.movement_phase_state)));
+      
+      if (traffic_light_states_[curr_intersection.id.id][current_movement_state.signal_group].size() >= 2
```
msmcconnell, line 1185: I think you probably can't wait for the full phase cycle to be received before the CarmaTrafficLight is updated. It seems unlikely that the vehicle would have like 40-60s to approach the light before needing this information. I think you probably need to add in a default phase state duration which can be used until the actual time is received. That way guidance can do its best to plan with the current state until the future state is also received.

msmcconnell (changes requested):

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -1118,5 +1118,78 @@ std::vector<lanelet::CarmaTrafficLightPtr> CARMAWorldModel::predictTrafficLight(
...
+      
+      // reset states if the intersection's geometry changed 
+      if (curr_light->revision_ != curr_intersection.revision)
+      {
+        ROS_INFO_STREAM("Received a new intersection geometry. intersection_id: " << (int)curr_intersection.id.id << 
+                           ", and signal_group_id: " << (int)current_movement_state.signal_group);
+        traffic_light_states_[curr_intersection.id.id].clear();
+      }
+
+      // currently expecting 1 lane, so 0th index is used only when setting states
```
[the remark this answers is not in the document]
msmcconnell: Sudhakar said the single movement assumption is fine for now. However, please add an enhancement issue on github to develop a more robust approach and add a link to the enhancement issue in your code comments about the assumption.

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -1061,4 +1081,126 @@ void CARMAWorldModel::setConfigSpeedLimit(double config_lim)
...
+      else if (curr_light->recorded_time_stamps.empty()) // if it was never initialized, do its best to plan with the current state until the future state is also received.
+      {
+        std::vector<std::pair<ros::Time, lanelet::CarmaTrafficLightState>> default_state;
+        // green 20sec, yellow 3sec, red 20sec, back to green 20sec etc...
+        default_state.push_back(std::make_pair<ros::Time, lanelet::CarmaTrafficLightState>(ros::Time(0), lanelet::CarmaTrafficLightState::PERMISSIVE_MOVEMENT_ALLOWED));
+        default_state.push_back(std::make_pair<ros::Time, lanelet::CarmaTrafficLightState>(ros::Time(3), lanelet::CarmaTrafficLightState::PRE_MOVEMENT));
+        default_state.push_back(std::make_pair<ros::Time, lanelet::CarmaTrafficLightState>(ros::Time(23), lanelet::CarmaTrafficLightState::STOP_AND_REMAIN));
+        default_state.push_back(std::make_pair<ros::Time, lanelet::CarmaTrafficLightState>(ros::Time(43), lanelet::CarmaTrafficLightState::PERMISSIVE_MOVEMENT_ALLOWED));
+        curr_light->setStates(default_state, curr_light->revision_);
+      }
```
msmcconnell, line 1201: There is a missing case here. 
When you have the full cycle you are setting it. When you have no cycle you are using defaults, but we also need the case where you set the partial cycle. 
For example if you default is (in length of phase) 30s g, 5s y, 10s r
Then when you receive a Spat with green phase of 20s duration you should update the cycle to have durations of 
20s g, 5s y, 10s r

When the yellow is then received as a duration of 8s you would update the known phases to 
20s g, 8s y, 10s r

Ditto for red

PR #1367, opened 2021-07-26
Feature/wz

itabrl:
# PR Details
## Description
This PR adds a Work Zone Strategic Plugin to carma-platform which is meant to support the WorkZone use case demonstration for CARMA3. The plugin is structured as a state machine containing and UNAVAILABLE, APPROACHING, WAITING, DEPARTING states. Each state generates its own maneuvers and the vehicle transitions through these states based off the light timing plan. 

## Related Issue
Supports https://github.com/usdot-fhwa-stol/carma-platform/issues/1388

## Motivation and Context
Working WorkZone use case

## How Has This Been Tested?
Unit Testing

## Types of changes

## Checklist:


carma_wm/include/carma_wm/CARMAWorldModel.h
```diff
@@ -105,6 +105,10 @@ class CARMAWorldModel : public WorldModel
   /*! \brief Set endpoint of the route
    */
   void setRouteEndPoint(const lanelet::BasicPoint3d& end_point);
+ 
+  /*! \brief helper for traffic light Id
+   */
```
adev4a, line 110: Can you add a description of what this function is doing? And the param and return descriptions?

wz_strategic_plugin/include/wz_strategic_plugin/wz_strategic_plugin.h
```diff
@@ -0,0 +1,114 @@
...
+
+            /**
+             * \brief Initialize ROS publishers, subscribers, service servers and service clients
+             */
+            void initialize();
+
+
+            bool planManeuverCb(cav_srvs::PlanManeuversRequest &req, cav_srvs::PlanManeuversResponse &resp);
+
+            ros::Publisher plugin_discovery_pub_;
```
adev4a, line 58: Can the class members defined from Lines 58 to 82 to private? Or are they being called elsewhere

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,282 @@
...
+        lanelet::BasicPoint2d curr_loc(pose_msg_->pose.position.x, pose_msg_->pose.position.y);
+        current_loc_ = curr_loc;
+    }
+
+    void WzStrategicPlugin::twist_cb(const geometry_msgs::TwistStampedConstPtr &msg)
+    {
+        current_speed_ = msg->twist.linear.x;
+    }
+
+    std::string WzStrategicPlugin::traffic_light_interpreter(lanelet::CarmaTrafficLightState& state)
```
adev4a, line 60: Looks like this method is returning the same string for each state. Can you add a To Do comment if this method needs to be revised later?

adev4a: Also can we add a default case to the switch statement

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,282 @@
...
+    bool WzStrategicPlugin::planManeuverCb(cav_srvs::PlanManeuversRequest &req, cav_srvs::PlanManeuversResponse &resp)
+    {
+        ROS_DEBUG("Finding car infomrmation");
+
+        // find car_down_track
+        double current_car_down_track = wm_->routeTrackPos(current_loc_).downtrack;
+        ROS_DEBUG("current_car_down_track %d", current_car_down_track);
+
+        auto current_lanelets = lanelet::geometry::findNearest(wm_->getMap()->laneletLayer, current_loc_, 1);
+        auto current_lanelet = current_lanelets[0];
```
adev4a, line 109: There is a check in the old implementation for route following that ensures that the current lanelet is on the shortest path. There is a chance here that the current lanelet picked up is not one in the route, i think it would be useful to have that check

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+        if(!traffic_list.empty()) {
+
+            auto nearest_traffic_light = traffic_list.front();
+
+            auto point = lanelet::traits::to2D(nearest_traffic_light->stopLine());
+            auto pos = carma_wm::geometry::trackPos(current_lanelet.second, point);
+            double traffic_light_down_track = pos.downtrack;
+
+            // double traffic_light_down_track = 0;
```
adev4a, line 122: Can we remove this comment?

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+            double distance_remaining_to_traffic_light  = traffic_light_down_track - current_car_down_track;
+            ROS_DEBUG("distance_remaining_to_traffic_light %d", distance_remaining_to_traffic_light);
+
+            ROS_DEBUG("min_distance_to_traffic_light %d", min_distance_to_traffic_light);
+
+            ros::Time current_time = ros::Time::now();
+            ros::Time end_time = ros::Time::now();
+
+            if(distance_remaining_to_traffic_light >= min_distance_to_traffic_light ) {
```
adev4a, line 133: Can we add an else condition to this as well? Maybe just a log statement that shows we're not entering this loop. If that means the strategic plugin cannot handle this condition, we should return a false and indicate failure

wz_strategic_plugin/src/wz_strategic_plugin.cpp
itabrl: default case does not exist in this condition.

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,375 @@
...
+                                                                                                  workzone_end_point_down_track, 
+                                                                                                  current_speed_, 
+                                                                                                  current_speed_,
+                                                                                                  start_time + ros::Duration(std::max(15.0,estimated_time_to_stop)),
+                                                                                                  start_lane_id);
+                        resp.new_plan.maneuvers.push_back(workzone_tactical);
+
+                    }
+
+                    if(traffic_light_next_predicted_state == 1) {
```
adev4a, line 221: Can we define this 0 and 1 as constant class members that specify what this state means? Just to make it easier to read.
For example if 0 is supposed to show red light state, lets create a const int member in the header RED_LIGHT_STATE_ = 0.
We should also change this second if condition to an "else if" since it is part of the same loop.
There should be an else condition following that which results in a "not valid state" warning.

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,375 @@
...
+                                                                                                  workzone_end_point_down_track, 
+                                                                                                  current_speed_, 
+                                                                                                  current_speed_, 
+                                                                                                  start_time + ros::Duration(std::max(15.0,estimated_time_to_stop)), 
+                                                                                                  start_lane_id);
+                        resp.new_plan.maneuvers.push_back(workzone_tactical);
+
+                    }
+
+                    if(traffic_light_current_state == 1) {
```
adev4a, line 275: Same note as above

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,508 @@
...
+        planWhenDEPARTING(req, resp, current_state, intersection_end_downtrack_.get(),
+                          intersection_speed_.get());
+        break;
+
+      default:
+        throw std::invalid_argument("WorkZone Strategic Plugin entered unknown state");
+        break;
+    }
+
+  } while (transition_table_.getState() != prev_state);  // If the state has changed then continue looping
```
adev4a, line 402: Is the state going to change multiple times within one planManeuver call? 
If we are within range of the intersection (lets say 50m away) will we ever exit this loop? 
Maybe i'm thinking of this wrong but prev state and current_state will be always be the same unless we are really close to moving from one state to another.

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,508 @@
...
+
+WzStrategicPlugin::VehicleState WzStrategicPlugin::extractInitialState(const cav_srvs::PlanManeuversRequest& req) const
+{
+  VehicleState state;
+  if (!req.prior_plan.maneuvers.empty())
+  {
+    state.stamp = GET_MANEUVER_PROPERTY(req.prior_plan.maneuvers.back(), end_time);
+    state.downtrack = GET_MANEUVER_PROPERTY(req.prior_plan.maneuvers.back(), end_dist);
+    state.speed = GET_MANEUVER_PROPERTY(req.prior_plan.maneuvers.back(), end_speed);
+    state.lane_id = getLaneletsBetweenWithException(state.downtrack, state.downtrack, true).front().id();
```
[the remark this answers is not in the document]
msmcconnell: TODO for me to very this under this PR before we merge

wz_strategic_plugin/include/wz_strategic_plugin/wz_states.h
```diff
@@ -0,0 +1,51 @@
...
+std::ostream& operator<<(std::ostream& os, TransitState s);
+
+
+//! @brief Enum describing the possible signals to change the current TransitState
+enum class TransitEvent
+{
+  IN_STOPPING_RANGE, // Transition event representing that the vehicle is in the stopping range of a light
+  STOPPED, // Transition event representing that the vehicle has come to a full stop
+  CROSSED_STOP_BAR, // Transition event representing that the vehicle has crossed the stop bar
+  RED_TO_GREEN_LIGHT, // Transition event representing that the current light of intereset has changed from red to green
```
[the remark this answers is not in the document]
msmcconnell: fixed

wz_strategic_plugin/src/wz_strategic_plugin.cpp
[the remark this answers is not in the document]
msmcconnell: veh_lane_id Is explicitly for the start of the whole maneuver plan not the end state, so the prior_plan must still be interpreted in this way when it is populated.

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,508 @@
...
+  VehicleState current_state = extractInitialState(req);
+
+  // Get current traffic light information
+  ROS_DEBUG("\n\nFinding traffic_light information");
+
+  auto traffic_list = wm_->getLightsAlongRoute({ req.veh_x, req.veh_y });
+
+  if (traffic_list.empty())  // If there were no lights found on the route then this plugin is not applicable
+  {
+    ROS_DEBUG("No lights found along route. Returning maneuver plan unchanged");
```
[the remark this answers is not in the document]
msmcconnell: fixed

wz_strategic_plugin/src/wz_strategic_plugin.cpp
msmcconnell: The state can change multiple times on one loop if the vehicle has satisfied multiple events within the 1s timing period of maneuver planning. For example:
- leaving an intersection and immediately approaching another intersection that is within stopping range

I elected to allow all the states to fully transition in this way instead of having to wait a whole 1s for the next plan before allowing a new state transition. 

I'm not sure I see where the risk of not exiting the loop comes from, can you elaborate?

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,508 @@
...
+
+  ROS_DEBUG("Finding car information");
+
+  // Extract vehicle data from request
+  VehicleState current_state = extractInitialState(req);
+
+  // Get current traffic light information
+  ROS_DEBUG("\n\nFinding traffic_light information");
+
+  auto traffic_list = wm_->getLightsAlongRoute({ req.veh_x, req.veh_y });
```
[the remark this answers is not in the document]
msmcconnell: break out of which loop? 
If you mean the state transition loop I would rather allow the states to stop transitioning before breaking out to ensure the current state is consistent with the planning environment on planning completion.

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,508 @@
...
+void WzStrategicPlugin::planWhenWAITING(const cav_srvs::PlanManeuversRequest& req,
+                                        cav_srvs::PlanManeuversResponse& resp, const VehicleState& current_state,
+                                        const std::vector<lanelet::CarmaTrafficLightPtr>& traffic_lights)
+{
+  if (traffic_lights.empty())
+  {
+    throw std::invalid_argument("While in WAITING state, the traffic lights disappeared.");
+  }
+
+  auto nearest_traffic_light = traffic_lights.front();
```
adev4a, line 289: Same concern as above about using front()

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,508 @@
...
+  if (current_light_state_optional.get() == lanelet::CarmaTrafficLightState::PERMISSIVE_MOVEMENT_ALLOWED)
+  {
+    transition_table_.signal(TransitEvent::RED_TO_GREEN_LIGHT);  // If the light is green send the light transition
+                                                                 // signal
+    return;
+  }
+
+  // If the light is not green then continue waiting by creating a stop and wait maneuver ontop of the vehicle
+  resp.new_plan.maneuvers.push_back(
+      composeStopAndWaitManeuverMessage(current_state.downtrack - 10.0, traffic_light_down_track, current_state.speed,
```
adev4a, line 310: What is the purpose of "-10" here?

wz_strategic_plugin/src/wz_strategic_plugin.cpp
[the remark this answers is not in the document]
adev4a: I meant the loop in the getLightsAlongRoute function in carma world model, instead of checking the complete shortest path, we should be only checking till the current downtrack.

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,508 @@
...
+  intersection_end_downtrack_ = boost::none;
+
+  if (traffic_lights.empty())
+  {
+    ROS_DEBUG("No lights found along route. Returning maneuver plan unchanged");
+    return;
+  }
+
+  double traffic_light_down_track =
+      wm_->routeTrackPos(traffic_lights.front()->stopLine().front().front().basicPoint2d()).downtrack;
```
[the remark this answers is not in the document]
msmcconnell: According to the comments for getLightsAlongRoute it returns 
"a sorted list (by downtrack distance) of traffic lights on the route ahead of us thus eliminating those behind the vehicle."

So the first light would be the one in front of the provided point on the route and therefore the one of interest. No?

wz_strategic_plugin/src/wz_strategic_plugin.cpp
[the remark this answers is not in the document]
adev4a: Looks like the intention of the function is to return a sorted list (by downtrack distance) of traffic lights on the route ahead of us, which it is doing

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,508 @@
...
+  double distance_remaining_to_traffic_light = traffic_light_down_track - current_state.downtrack;
+
+  ROS_DEBUG("distance_remaining_to_traffic_light %f", distance_remaining_to_traffic_light);
+
+  double stopping_dist = estimate_distance_to_stop(current_state.speed, config_.vehicle_decel_limit_multiplier *
+                                                                            config_.vehicle_decel_limit);
+
+  double plugin_activation_distance = std::max(stopping_dist, config_.min_approach_distance);
+
+  if (current_state.downtrack >= plugin_activation_distance)
```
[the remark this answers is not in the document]
msmcconnell: done

wz_strategic_plugin/src/wz_strategic_plugin.cpp
```diff
@@ -0,0 +1,508 @@
...
+  {
+    transition_table_.signal(TransitEvent::IN_STOPPING_RANGE);  // Evaluate Signal
+  }
+}
+// TODO should we handle when the vehicle is not going to make the light but doesn't have space to stop?
+void WzStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversRequest& req,
+                                            cav_srvs::PlanManeuversResponse& resp, const VehicleState& current_state,
+                                            const std::vector<lanelet::CarmaTrafficLightPtr>& traffic_lights)
+{
+  if (traffic_lights.empty())  // If are are in the approaching state and there is no longer any lights ahead of us then
```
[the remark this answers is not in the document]
msmcconnell: done

wz_strategic_plugin/src/wz_strategic_plugin.cpp
msmcconnell: It's a fixed buffer to add to stopping maneuvers once the vehicle is already stopped to ensure that they can have their
trajectory planned. I made it a constant in code with comment

wz_strategic_plugin/src/wz_strategic_plugin.cpp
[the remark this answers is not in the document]
msmcconnell: Added in most recent commit

carma_wm/include/carma_wm/WMTestLibForGuidance.h
```diff
@@ -476,6 +477,64 @@ inline void setRouteByIds (std::vector<lanelet::Id> lanelet_ids, std::shared_ptr
...
+    if (iterator == cmw->getMutableMap()->laneletLayer.end())
+      throw std::invalid_argument("Provided with lanelet id not in map: " + std::to_string(id));
+
+    controlled_lanelets.push_back(*iterator);
+  }
+
+  // Get owning lanelet by id
+  auto iterator = cmw->getMutableMap()->laneletLayer.find(owning_lanelet_id);
+    
+  if (iterator == cmw->getMutableMap()->laneletLayer.end())
```
adev4a, line 517: I wasn't able to find the definition of this find, but would it work better to initialize this iterator to -1 and check against that here?

msmcconnell: find is defined here and uses iterator and const_iterator the same way std::vector works. 
https://github.com/usdot-fhwa-stol/autoware.ai/blob/c51ccf194c704cc2335a37d1a8235837e024fad1/lanelet2/lanelet2_core/include/lanelet2_core/LaneletMap.h#L90

I do not agree with initializing it to -1. The correct practice with iterators is to compare them with container.end() as an iterator is an object and is not equivalent to an index though the index can be extracted from it.

wz_strategic_plugin/src/wz_strategic_plugin.cpp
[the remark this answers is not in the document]
msmcconnell: Done

carma_wm/include/carma_wm/WMTestLibForGuidance.h
adev4a: Ok that sounds good then. Thanks for the explanation!

adev4a (approved):

Issue #121, opened 2018-06-25
Light bar not turning on for merging vehicle when in platoon


msmcconnell:
In TO26 I-95 testing, The light bar does not turn-on reliably for platooning after merge. This is because the RSU is sending commands to do a lane change as the route geometry does not match the road (rsu thinks vehicle is in the wrong lane). The requested steering command causes the lane change plugin to request the light bar as shown below. This prevents from platooning getting access till much later. 

```
2018-06-25 13:32:35:074 | INFO | LightBarManager | NONE | Platooning Plugin failed to set the LightBarIndicator YELLOW as this was already controlled by Lane Change Plugin 
```

darrelld05:
This issue was transferred from the original CARMAPlatform: Old issue # 955

msmcconnell:
Closed as OBE in carma3 and noted on the CARMA2 tracker spreadsheet

PR #1368, opened 2021-07-26
Guidance Plugin Validator Node

JonSmet:
# PR Details
## Description

This PR includes the implementation of a new Guidance Plugin Validator Node, which is intended to conduct a basic validation check of Guidance Strategic, Tactical, and Control-wrapper plugins in order to provide confidence that the particular plugin is capable of properly participating in the CARMA Guidance Subsystem. At a basic level, this node verifies that each plugin's node has the correct subscribers, publishers, and advertised services. Additionally, it checks that each plugin's node publishes the correct information to the /guidance/plugin_discovery topic. 

Since this node's intended usage is to write all validation information to its log file, it has a large number of ros-logging statements. If a required component from a plugin is missing or incorrect, an "ERROR" line will be included in the log file accordingly; similarly, if an optional component from a plugin is missing or incorrect, a "WARNING" line will be included in the log file.

Finally, this node is enabled by default from the 'enable_guidance_plugin_validator' argument included in the carma_docker.launch file for the development carma config, with the expectation that this node will primarily be used by a developer when running the CARMA system with mock drivers at their workstation. These applicable changes are included in the following related PR for the carma-config repository:

- [carma-config PR #121: Add flag for enabling Guidance Plugin Validator Node](https://github.com/usdot-fhwa-stol/carma-config/pull/121)

## How Has This Been Tested?

This has been tested locally by running the CARMA system locally with mock drivers. Tests were conducted with correct and incorrect guidance plugins.

## Types of changes

## Checklist:


msmcconnell:
Can you guys make sure this python node is Python3 compatible so that we don't need to go back and update for ROS Noetic?

JonSmet:
> Can you guys make sure this python node is Python3 compatible so that we don't need to go back and update for ROS Noetic?

I went through the node and found that everything was Python3 compatible, but I did make some updates to setup.py and CMakeLists.txt in order to avoid build errors after the migration. One change to setup.py would have created a build error for ROS Kinetic, so I left a TODO statement instead.

guidance_plugin_validator/config/params.yaml
```diff
@@ -0,0 +1,37 @@
...
+# Double: The time allocated for system startup. Validation begins after this much time has passed.
+# Units: s
+startup_duration: 60.0
+
+# Double: The time allocated for validation; once startup is complete, validation is complete after this much time has passed.
+# Units: s
+validation_duration: 60.0
+
+# List of String: Guidance Strategic Plugins that will be validated by this node
+guidance_strategic_plugins:
```
kjrush, line 24: Doe it make sense to make this an argument to the launch file more than a param? Or possibly both somehow (arg overriding param file?)? Just thinking, because that way we could set which plugins will be validated from the carma-config image - via the docker-compose.yml - which might be handy/convenient.

guidance_plugin_validator/config/params.yaml
```diff
@@ -0,0 +1,37 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
+# Double: The time allocated for system startup. Validation begins after this much time has passed.
+# Units: s
+startup_duration: 60.0
```
kjrush, line 17: Not sure of the pros/cons of a fixed startup time, vs waiting for drivers ready from the health monitor

guidance_plugin_validator/package.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0"?>
+<package format="2">
```
kjrush, line 2: Copyright comment

guidance_plugin_validator/src/guidance_plugin_validator/guidance_plugin_components.py
```diff
@@ -0,0 +1,402 @@
...
+    """
+
+    def __init__(self, plugin_name):
+        """Constructor for StrategicPluginResults"""
+
+        # Set the plugin's name, as used by the plugin_discovery topics and the plugin's advertised service(s)
+        self.plugin_name = plugin_name
+        
+        # Set the plugin's node name
+        if plugin_name == "RouteFollowing":
```
kjrush, line 33: Can you explain a bit more about why this special case for RFP is required?

kjrush: Or does this currently only support validating RFP?

kjrush: If we simply need the node name based on the plugin name it should be "/guidance/plugins/<PLUGIN_NAME>" by design, though I'm unsure if our current implementation is correct for that.

guidance_plugin_validator/src/guidance_plugin_validator/guidance_plugin_components.py
```diff
@@ -0,0 +1,402 @@
...
+    """
+
+    def __init__(self, plugin_name):
+        """Constructor for TacticalPluginResults"""
+
+        # Set the plugin's name, as used by the plugin_discovery topics and the plugin's advertised service(s)
+        self.plugin_name = plugin_name
+        
+        # Set the plugin's node name
+        if plugin_name == "InLaneCruisingPlugin":
```
kjrush, line 160: Like above, I'd prefer to use the intended topic patterns instead of specific casing on the plugin names where possible.

guidance_plugin_validator/src/guidance_plugin_validator/guidance_plugin_components.py
```diff
@@ -0,0 +1,402 @@
...
+        # Validation result indicating whether tactical plugin's node advertises required service
+        self.plan_trajectory_service = "/guidance/plugins/" + str(plugin_name) + "/plan_trajectory"
+        self.has_plan_trajectory_service = False
+
+        # Validation result indicating whether tactical plugin's node publishes to required topic
+        self.plugin_discovery_topic = "/guidance/plugin_discovery"
+        self.has_plugin_discovery_pub = False
+
+        # Validation results indicating whether tactical plugin's node publishes required information to the plugin_discovery topic
+        self.correct_plugin_discovery_type = 2
```
kjrush, line 256: Not sure what this integer is representing exactly, same as the above usage for strategic. If possible can we replace this with an IntEnum that provides some more clarity?

kjrush (commented):
Still working my way through, but wanted to release a few comments now.

guidance_plugin_validator/src/guidance_plugin_validator/guidance_plugin_validator.py
```diff
@@ -0,0 +1,479 @@
...
+        self.control_plugin_results_index_by_name = {}   # Key is plugin's name; Value is plugin's index in its respective 'validation results' list
+
+        # Call member function to populate the 'validation results' lists and 'index by name' dicts
+        self.populate_results_lists_and_dicts(self.strategic_plugin_names, self.tactical_plugin_names, self.control_plugin_names)
+
+    def populate_results_lists_and_dicts(self, strategic_plugin_names, tactical_plugin_names, control_plugin_names):
+        """Initialize the the 'validation results' lists and 'index by name' dicts for this object"""
+
+        # Populate list and dict for Strategic Plugins
+        for i in range(0, len(strategic_plugin_names)):
```
kjrush, line 75: I don't have a concrete comment here, but I feel there's got to be a more pythonic way to do this, not that I can think of it at the moment. Maybe ```for plugin in strategic_plugin_names:...``` and then index everything by name? Not sure, and doesn't have to be changed, but if wanted to put the thought out there into the ether.

guidance_plugin_validator/src/guidance_plugin_validator/guidance_plugin_validator.py
```diff
@@ -0,0 +1,479 @@
...
+        
+        # Populate list and dict for Control Plugins
+        for i in range(0, len(control_plugin_names)):
+            plugin_name = control_plugin_names[i]
+            self.control_plugin_validation_results.append(guidance_plugin_components.ControlPluginResults(plugin_name))
+            self.control_plugin_results_index_by_name[plugin_name] = i
+
+        return
+
+    def wait_for_startup_to_complete(self):
```
kjrush, line 94: This might be where we could wait for DRIVERS_READY system status?

guidance_plugin_validator/config/params.yaml
JonSmet: Your suggestion definitely makes sense. I was unable to figure out a way to provide these arguments from the docker-compose.yml file, and instead added these as args within the carma_docker.launch file (in the related carma-config PR), so that a user can still add them within the carma-config image without re-building the image. Is this approach acceptable? 

Also, I have removed the parameters from the params.yaml file since I thought it may get confusing having two locations to input the lists of plugin names.

guidance_plugin_validator/config/params.yaml
JonSmet: I agree. I updated the implementation to wait for the 'DRIVERS_READY' message from the health monitor.

guidance_plugin_validator/package.xml
JonSmet: Added.

guidance_plugin_validator/src/guidance_plugin_validator/guidance_plugin_components.py
JonSmet: I saw that you created a new Github Issue #1400 to address this. This issue exists for all current strategic, tactical, and control plugins from what I have seen so far. Once it is addressed, I can update this casing that exists in the Guidance Plugin Validator.

guidance_plugin_validator/src/guidance_plugin_validator/guidance_plugin_components.py
JonSmet: Fixed.

guidance_plugin_validator/src/guidance_plugin_validator/guidance_plugin_validator.py
JonSmet: I have updated the implementation for this. The plugin now uses dicts with each key being the plugin's name, and the corresponding value being the plugin's <plugin-type>PluginResults object.

kjrush (approved):

kjrush (approved):

Issue #1284, opened 2021-05-28
CARMA Needs lanelet2 participant type definition for the vehicle running the system


msmcconnell:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

CARMA Uses the Lanelet2 library to define road geometries and road rules. This library defines rules using a hierarchical participant framework (for example "vehicle" would refer to all vehicles while "vehicle:car" refers to only cars). At the moment the carma-platform has no way for the user to specify the participant type currently in use(car or truck, etc). This means that the its possible for some traffic rules to be misinterpreted by the system (truck drives in cars only lane for example). A configuration parameter should be added to carma config to let the user set the participant type. This parameter then needs to be fed into all world model instances so that the default returned traffic rules have the correct participant type associated with them. 

### Carma version where this issue was discovered

develop after 3.5
### Expected behavior
See description
### Actual behavior

See description
### Steps to reproduce the actual behavior

N/A

msmcconnell:
Resolved by above PRs

PR #1415, opened 2021-08-23
Feature/lanelet2 participant type

lewisid:
# PR Details
## Description

The carma_wm repository has been modified to support processing the vehicle participation type parameter. Each relevant header file has been updated to include methods to obtain this value and process it throughout all carma world model instances.
## Related Issue

https://github.com/usdot-fhwa-stol/carma-platform/issues/1284
## Motivation and Context

CARMA Uses the Lanelet2 library to define road geometries and road rules. This library defines rules using a hierarchical participant framework (for example "vehicle" would refer to all vehicles while "vehicle:car" refers to only cars). At the moment the carma-platform has no way for the user to specify the participant type currently in use(car or truck, etc). This means that the its possible for some traffic rules to be misinterpreted by the system (truck drives in cars only lane for example). A configuration parameter should be added to carma config to let the user set the participant type. This parameter then needs to be fed into all world model instances so that the default returned traffic rules have the correct participant type associated with them.
## Types of changes

## Checklist:


carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -509,7 +509,15 @@ void CARMAWorldModel::setMap(lanelet::LaneletMapPtr map, size_t map_version, boo
...
-    TrafficRulesConstPtr traffic_rules = *(getTrafficRules(lanelet::Participants::Vehicle));
+    TrafficRulesConstPtr traffic_rules;
+    if(participant_type_ == lanelet::Participants::VehicleCar)
+    {
+      traffic_rules = *(getTrafficRules(lanelet::Participants::VehicleCar));
+    }
+    else if(participant_type_ == lanelet::Participants::VehicleTruck)
+    {
+      traffic_rules = *(getTrafficRules(lanelet::Participants::VehicleTruck));
+    }
```
msmcconnell, line 520: What is the point for these if checks? Seems like this whole block is equivalent to the one liner 
```
traffic_rules = *(getTrafficRules(participant_type_ ));
```

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -572,7 +580,15 @@ void CARMAWorldModel::computeDowntrackReferenceLine()
...
-  TrafficRulesConstPtr traffic_rules = *(getTrafficRules(lanelet::Participants::Vehicle));
+  TrafficRulesConstPtr traffic_rules;
+  if(participant_type_ == lanelet::Participants::VehicleCar)
+  {
+     traffic_rules = *(getTrafficRules(lanelet::Participants::VehicleCar));
+  }
+  else if(participant_type_ == lanelet::Participants::VehicleTruck)
+  {
+     traffic_rules = *(getTrafficRules(lanelet::Participants::VehicleTruck));
+  }
```
msmcconnell, line 591: Same note as above

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -1073,6 +1090,18 @@ void CARMAWorldModel::setConfigSpeedLimit(double config_lim)
...
+void CARMAWorldModel::setVehicleParticipationType(const std::string& participant)
+{
+  if(participant == "vehicle:car")
+  {
+    participant_type_ = lanelet::Participants::VehicleCar;
+  }
+  if(participant == "vehicle:truck")
+  {
+    participant_type_ = lanelet::Participants::VehicleTruck;
+  }
```
msmcconnell, line 1102: Again not seeing the purpose of these checks. Those are type defs of strings so this should be equivalent to 
```
participant_type_ = participant;
```

carma_wm/src/WMListener.cpp
```diff
@@ -41,7 +41,11 @@ WMListener::WMListener(bool multi_thread) : worker_(std::unique_ptr<WMListenerWo
   double cL;
   nh2_.getParam("/config_speed_limit", cL);
   setConfigSpeedLimit(cL);
-  
+
+  std::string part_;
+  nh2_.getParam("/participation_type", part_);
```
msmcconnell, line 46: param name should be /vehicle_participant_type

carma_wm/src/WMListener.cpp
```diff
@@ -111,6 +115,11 @@ void WMListener::setConfigSpeedLimit(double config_lim) const
   worker_->setConfigSpeedLimit(config_lim);
 }
 
+void WMListener::setVehicleParticipationType(std::string participant) const
```
msmcconnell, line 118: Does the WMListener really need a setter for this? If the intent is to load it from the parameter in this class, I don't see why this function is necessary. It could lead to confusion for the user to have a situation where they set the value but it gets overwritten by a parameter anyway.

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -425,6 +425,19 @@ double WMListenerWorker::getConfigSpeedLimit() const
...
+void WMListenerWorker::setVehicleParticipationType(std::string participant)
+{
+  if(participant == "vehicle:car")
+  {
+    participant_ = lanelet::Participants::VehicleCar;
+  }
+  if(participant == "vehicle:truck")
+  {
+    participant_ = lanelet::Participants::VehicleTruck;
+  }
```
msmcconnell, line 437: Logic block can be simplified as above

carma_wm/src/WMListenerWorker.h
```diff
@@ -115,6 +126,7 @@ class WMListenerWorker
   std::function<void()> route_callback_;
   void newRegemUpdateHelper(lanelet::Lanelet parent_llt, lanelet::RegulatoryElement* regem) const;
   double config_speed_limit_;
+  std::string participant_;
```
msmcconnell, line 129: I don't think you need to actually store the participant_ in the WMListenerWorker just forward it to the world model object.

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -1032,6 +1032,18 @@ void WMBroadcaster::setConfigSpeedLimit(double cL)
...
+void WMBroadcaster::setVehicleParticipationType(std::string participant)
+{
+  if(participant == "vehicle:car")
+  {
+    participant_ = lanelet::Participants::VehicleCar;
+  }
+  if(participant == "vehicle:truck")
+  {
+    participant_ = lanelet::Participants::VehicleTruck;
+  }
```
msmcconnell, line 1044: Can be simplified

carma_wm_ctrl/src/WMBroadcasterNode.cpp
```diff
@@ -85,6 +85,9 @@ int WMBroadcasterNode::run()
   pnh2_.getParam("/config_speed_limit", config_limit);
   wmb_.setConfigSpeedLimit(config_limit);
 
+  std::string participant;
+  pnh2_.getParam("/participation_type", participant);
```
msmcconnell, line 89: Same note about parameter name

carma_wm/include/carma_wm/CARMAWorldModel.h
```diff
@@ -176,7 +180,7 @@ class CARMAWorldModel : public WorldModel
   LaneletRoutingGraphConstPtr getMapRoutingGraph() const override;
 
   lanelet::Optional<TrafficRulesConstPtr>
-  getTrafficRules(const std::string& participant = lanelet::Participants::Vehicle) const override;
+  getTrafficRules(const std::string& participant) const override;//Participant type is sent from WMListenerWorker
```
msmcconnell, line 183: I think you ought to have two versions of this method. 
This one where the user can provide the participant type in question and therefore not use the WMListener set value.
The other should be a zero argument method which uses the default WMListener set value in the world model. 
getTrafficRules()

That will maintain API compatibility with the existing code that relied on the default parameter.

msmcconnell (changes requested):

msmcconnell:
@lewisid The build is failing on CI with error
```
/opt/carma/src/CARMAPlatform/platooning_strategic/src/platoon_strategic.cpp: In member function 'double platoon_strategic::PlatoonStrategicPlugin::findSpeedLimit(const lanelet::ConstLanelet&)':
/opt/carma/src/CARMAPlatform/platooning_strategic/src/platoon_strategic.cpp:262:96: error: call of overloaded 'getTrafficRules()' is ambiguous
                                                                                                ^
In file included from /opt/carma/src/CARMAPlatform/carma_wm/include/carma_wm/WMListener.h:23:0,
                 from /opt/carma/src/CARMAPlatform/platooning_strategic/include/platoon_strategic.h:39,
                 from /opt/carma/src/CARMAPlatform/platooning_strategic/src/platoon_strategic.cpp:20:
/opt/carma/src/CARMAPlatform/carma_wm/include/carma_wm/WorldModel.h:217:3: note: candidate: virtual lanelet::Optional<std::shared_ptr<const lanelet::traffic_rules::TrafficRules> > carma_wm::WorldModel::getTrafficRules(const string&) const
   getTrafficRules(const std::string& participant = lanelet::Participants::Vehicle) const = 0;
   ^
/opt/carma/src/CARMAPlatform/carma_wm/include/carma_wm/WorldModel.h:225:3: note: candidate: virtual lanelet::Optional<std::shared_ptr<const lanelet::traffic_rules::TrafficRules> > carma_wm::WorldModel::getTrafficRules() const
```
Looks like you need to remove the default string as it makes the function call ambiguous. Also since this is only appearing on CI and you didn't run into this while testing that makes me think you only tested one of the these two methods. Make sure both have a unit test.

PR #1416, opened 2021-08-25
Feature/freight pickup

JonSmet:
# PR Details
## Description

This PR includes updates to port_drayage_plugin that enable it to parse a received port drayage MobilityOperation message from V2XHub that instructs the system to drive to a 'PICKUP' location. Additionally, when the system arrives at the specified 'PICKUP' location, it will broadcast an arrival message to V2XHub. 

To facilitate proper port drayage state machine flow, the following updates have been made:

- Port Drayage State Machine transitions from INACTIVE to EN_ROUTE when the CARMA System is engaged for the first time (if the port_drayage_plugin is enabled from its config parameter).
- Port Drayage State Machine transitions from EN_ROUTE to AWAITING_DIRECTION (and broadcasts an arrival message to V2XHub) when a 'ROUTE_LOADED' route event is published immediately after a 'ROUTE_COMPLETED' route event. The 'ROUTE_LOADED' event is required since this indicates the previously completed route is no longer active, so the system is ready to receive a new route from V2XHub. 

## How Has This Been Tested?

Integration tested on Black Pacifica and unit tested.

## Types of changes

## Checklist:


port_drayage_plugin/include/port_drayage_plugin/port_drayage_worker.h
```diff
@@ -29,17 +31,17 @@
...
      * strategy_params field with strategy "carma/port_drayage"
      */
     struct PortDrayageMobilityOperationMsg
     {
         boost::optional<std::string> cargo_id;
-        std::string operation;
+        std::string operation = "";
         PortDrayageEvent port_drayage_event_type; // PortDrayageEvent associated with this message
-        bool has_cargo; // Flag to indicate whether vehicle has cargo during this action
+        bool has_cargo = false; // Flag to indicate whether vehicle has cargo during this action
```
kjrush, line 44: Is `has_cargo` a necessary data element to track if we have `cargo_id` as an optional element? Is there ever a case where `cargo_id` exists and `has_cargo` would be false or vice-versa?

port_drayage_plugin/src/port_drayage_worker.cpp
```diff
@@ -117,26 +117,51 @@ namespace port_drayage_plugin
...
-            pt.put("cargo", true);
-            pt.put("cargo_id", _cargo_id);
-        }
-        else {
-            pt.put("cargo", false);
+        // Add flag to indicate whether CMV is carring cargo
+        pt.put("cargo", _has_cargo);
+
+        // If CMV has not received any port drayage messages yet, assign 'operation' field for its initial arrival message
+        if (!_has_received_first_mobility_operation_msg) {
```
kjrush, line 131: Seems like maybe this kind of value could be moved into the state machine (I see references to a _pdsm in here, which I'm assume is Port Drayage State Machine) as the ultimate keeper of plugin state. And that this should be checking if we're in a state where we need to encode this data more so than the indirect property that we've received a mobility message, if that makes sense.

port_drayage_plugin/src/port_drayage_worker.cpp
```diff
@@ -182,20 +212,44 @@ namespace port_drayage_plugin
...
         _latest_mobility_operation_msg.operation = pt.get<std::string>("operation");
         ROS_DEBUG_STREAM("operation: " << _latest_mobility_operation_msg.operation);
-        ROS_DEBUG_STREAM("cargo flag: " << _latest_mobility_operation_msg.has_cargo);
-
-        if(_latest_mobility_operation_msg.operation == "MOVING_TO_LOADING_AREA") {
+        if (_latest_mobility_operation_msg.operation == PORT_DRAYAGE_PICKUP_OPERATION_ID) {
             _latest_mobility_operation_msg.port_drayage_event_type = PortDrayageEvent::RECEIVED_NEW_DESTINATION;
+            if (_has_cargo) {
+                ROS_DEBUG_STREAM("Received 'PICKUP' operation, but CMV is already carrying cargo.");
+                throw std::invalid_argument("Received 'PICKUP' operation, but CMV is already carrying cargo.");
```
kjrush, line 251: Are we guaranteed that CARMA streets will only ever broadcast this message once? I think otherwise we're liable to throw an uncaught exception here. Is this filtered by vehicle mobility ID as well? Or would we throw this exception if another vehicle was executing the pickup operation while we had cargo?

port_drayage_plugin/src/port_drayage_worker.cpp
```diff
@@ -204,6 +258,12 @@ namespace port_drayage_plugin
             ROS_DEBUG_STREAM("cargo id: " << *_latest_mobility_operation_msg.cargo_id);
         }
         else{
+            // If this message is for 'PICKUP', then the 'cargo_id' field is required
+            if(_latest_mobility_operation_msg.operation == PORT_DRAYAGE_PICKUP_OPERATION_ID) {
+                ROS_DEBUG_STREAM("Received 'PICKUP' operation, but no cargo_id was included.");
+                throw std::invalid_argument("Received 'PICKUP' operation, but no cargo_id was included.");
```
kjrush, line 264: Same as above. To be honest, might be more reasonable to simply ROS_LOG these at WARN or something. Anomalous conditions for sure, but does it prevent vehicle operation entirely?

port_drayage_plugin/config/params.yml
```diff
@@ -1,4 +1,5 @@
 stop_speed_epsilon: 1.0
 cmv_id: "123" # Test value; CARMA Streets expects cmv_id to be an unsigned long (32-bit unsigned integer)
-cargo_id: "321" # Test value (Empty string indicates the vehicle is not carrying cargo); CARMA Streets expects cargo_id to be a string
-host_id: HOST_ID
+cargo_id: "" # Test value (Empty string indicates the vehicle is not carrying cargo); CARMA Streets expects cargo_id to be a string
+enable_port_drayage: false # bool: Flag to activate and deactivate port drayage operations. If false, state machine will remain INACTIVE.
```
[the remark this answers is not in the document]
JonSmet: I think those are both good approaches to pursue in the future for this.

port_drayage_plugin/include/port_drayage_plugin/port_drayage_worker.h
JonSmet: When I initially implemented this, I thought it was cleaner to have an explicit `has_cargo` flag for this. I completely agree though that it is not necessary to track both, and a better solution is to only track the `_cargo_id` element (there is never a case where `_cargo_id` and `has_cargo` conflict with each other). I have removed the `has_cargo` element.

JonSmet: Good catch. Paul Bourelly and I had updated the design to remove the `has_cargo` flag from messages sent from V2XHub to carma-platform, so I have removed it.

port_drayage_plugin/include/port_drayage_plugin/port_drayage_worker.h
```diff
@@ -109,22 +116,33 @@ namespace port_drayage_plugin
...
                 _cmv_id(cmv_id),
                 _cargo_id(cargo_id),
                 _host_id(host_id),
                 _publish_mobility_operation(mobility_operations_publisher),
                 _set_active_route(call_set_active_route_client),
-                _stop_speed_epsilon(stop_speed_epsilon) {
-                    initialize();
+                _stop_speed_epsilon(stop_speed_epsilon),
+                _enable_port_drayage(enable_port_drayage) {
                     _has_cargo = (_cargo_id == "") ? false : true;
```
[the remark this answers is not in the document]
JonSmet: I have removed the `_has_cargo` element from PortDrayageWorker, since the information can already be inferred from `_cargo_id`. As a result, this `_has_cargo = (_cargo_id == "") ? false : true;` statement has been removed.

port_drayage_plugin/src/port_drayage_worker.cpp
JonSmet: I agree it is much cleaner to update the state machine for this logic. I have made an update to account for this.

port_drayage_plugin/src/port_drayage_worker.cpp
JonSmet: Yes, CARMA Streets will only ever broadcast this message once, and PortDrayageWorker filters these messages by cmv_id, so a PICKUP message broadcasted to another vehicle would not cause an exception to be thrown here. Additionally, a message will not be processed if its strategy_params match the previously processed message's strategy_params.

kjrush (approved):

PR #1422, opened 2021-09-01
Add vehicle status generator

itabrl:
# PR Details
## Description
As CARMA Platform on a vehicle entering an unsignalized intersection, I want to broadcast my status as an Entering Vehicle so that other vehicles can estimate and plan their stopping trajectories.

## Types of changes

## Checklist:

### DIFF
--- a/vehicle_status_generator/include/vehicle_status_generator_worker.h
+++ b/vehicle_status_generator/include/vehicle_status_generator_worker.h
@@ -1,0 +1,42 @@
+#pragma once
+
+/*
+ * Copyright (C) 2019-2021 LEIDOS.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+#include <vector>
+#include <ros/ros.h>
+#include <stdint.h>
+#include <algorithm>
+
+namespace vehicle_status_generator
+{
+    class VehicleStatusGeneratorWorker
+    {
+        public:
+            
+            VehicleStatusGeneratorWorker();
+            uint8_t getNextMsgCount();
+            std::vector<uint8_t> getMsgId(const ros::Time now);
+            uint16_t getSecMark(const ros::Time now);
+
+        private:
+
+            // variables for BSM generation
+            uint8_t msg_count_ {0};
+            int random_id_ {0};
+            ros::Time last_id_generation_time_;
+    };
+}
--- a/vehicle_status_generator/package.xml
+++ b/vehicle_status_generator/package.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0"?>
+
+<!--  
+ Copyright (C) 2018-2021 LEIDOS.
+
+ Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ use this file except in compliance with the License. You may obtain a copy of
+ the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ License for the specific language governing permissions and limitations under
+ the License.
+-->
+
+<package format="3">
+  <name>vehicle_status_generator</name>
+  <version>3.3.0</version>
+  <description>This generates BSM according to J2735 standard.</description>
+  <maintainer email="carma@todo.todo">carma</maintainer>
+  <license>Apache License 2.0</license>
+  <author email="carma@todo.todo">carma</author>
+  <buildtool_depend>catkin</buildtool_depend>
+  <depend>roscpp</depend>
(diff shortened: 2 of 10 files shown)


vehicle_status_generator/include/vehicle_status_generator_worker.h
```diff
@@ -0,0 +1,42 @@
...
+            
+            VehicleStatusGeneratorWorker();
+            uint8_t getNextMsgCount();
+            std::vector<uint8_t> getMsgId(const ros::Time now);
+            uint16_t getSecMark(const ros::Time now);
+
+        private:
+
+            // variables for BSM generation
+            uint8_t msg_count_ {0};
```
kjrush, line 38: Is this generating BSMs or just referencing data from the most recent BSM the vehicle broadcast?

vehicle_status_generator/package.xml
```diff
@@ -0,0 +1,39 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ License for the specific language governing permissions and limitations under
+ the License.
+-->
+
+<package format="3">
+  <name>vehicle_status_generator</name>
+  <version>3.3.0</version>
+  <description>This generates BSM according to J2735 standard.</description>
```
kjrush, line 22: Same with this description, this should only be generating the mobility operations message needed for the all-way-stop use case right?

vehicle_status_generator/src/vehicle_status_generator.cpp
```diff
@@ -0,0 +1,62 @@
...
+        nh_.reset(new ros::CARMANodeHandle());
+        pnh_.reset(new ros::CARMANodeHandle("~"));
+        
+        pnh_->param<double>("vehicle_status_generation_frequency_", vehicle_status_generation_frequency_, 10.0);
+        nh_->param<double>("vehicle_length", vehicle_length_, 5.0);
+        nh_->param<double>("vehicle_width", vehicle_width_, 2.0);
+
+        nh_->param<double>("/vehicle_acceleration_limit", vehicle_acceleration_limit_, 2.0);
+        nh_->param<double>("/vehicle_deceleration_limit", vehicle_deceleration_limit, 2.0);
+        nh_->param<double>("/guidance/yield_plugin/x_gap", x_gap, 2.0);
```
kjrush, line 35: I think at least this param would be better served being read from `x_gap`, and then remapped to `/guidance/yield_plugin/x_gap` in the launch file. The others are okay as they're more general, vehicle-wide parameters, but this is reading another plugins plugin-specific config so we should handle it with a bit more abstraction.

vehicle_status_generator/src/vehicle_status_generator_worker.cpp
```diff
@@ -0,0 +1,58 @@
...
+        uint8_t old_msg_count = msg_count_;
+        msg_count_++;
+        if(msg_count_ == 128)
+        {
+            msg_count_ = 0;
+        }
+        return old_msg_count;
+    }
+
+    std::vector<uint8_t> VehicleStatusGeneratorWorker::getMsgId(const ros::Time now)
```
kjrush, line 37: I'm not sure I understand the purpose of this method. I'm assuming this is related to the BSM ID field in the structure of the mobility operations message needed by CARMA Streets? If so, I'm not sure why we're generating a random ID here instead of actually reading the one we're currently broadcasting in the BSM? We have the BSM generator package already handling the generation and rotation of the BSM ID when it publishes the BSM for broadcast. We want to putting that data in the mobility message so that CARMA Streets can correlate the mobility message we broadcast at some timestep `t` with the BSM that was broadcast at time `t` as well.

vehicle_status_generator/src/vehicle_status_generator_worker.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+namespace vehicle_status_generator
+{
+    VehicleStatusGeneratorWorker::VehicleStatusGeneratorWorker() {}
+    
+    uint8_t VehicleStatusGeneratorWorker::getNextMsgCount()
+    {
+        uint8_t old_msg_count = msg_count_;
+        msg_count_++;
+        if(msg_count_ == 128)
```
kjrush, line 30: See comment below, but for the record the cleanest way to implement this would be with the modulus operator ` msg_count = ++msg_count % 128` handles the rollover cleanly.

This also only ever returns the previous message count despite the method name being `getNext...` which is slightly confusing, but not a big deal functionally.

vehicle_status_generator/src/vehicle_status_generator.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+    void VehicleStatusGenerator::speedCallback(const std_msgs::Float64ConstPtr& msg)
+    {
+        current_speed_ = msg->data;
+    }
+
+    void VehicleStatusGenerator::generateMobilityOperation(const ros::TimerEvent& event)
+    {
+        mo_.header.timestamp = ros::Time::now().toNSec();
+        mo_.header.sender_bsm_id = std::string(reinterpret_cast<const char*>(&worker.getMsgId(ros::Time::now())[0]), worker.getMsgId(ros::Time::now()).size());
```
kjrush, line 57: Having read further down the code, I think we need to instead be subscribing to the host vehicle's outbound BSM topic, and scraping the most recent BSM ID from there for this data. That's where we'd also want to get the msgCnt value from too.

vehicle_status_generator/src/vehicle_status_generator.cpp
```diff
@@ -0,0 +1,62 @@
...
+        mo_.strategy_params = std::to_string(x_gap) + "," + std::to_string(vehicle_acceleration_limit_) + "," + std::to_string(vehicle_deceleration_limit);
```
kjrush, line 58: Should we be encoding the msgCnt in here anywhere?

kjrush (changes requested):

Issue #1420, opened 2021-08-31
CARMA-WEB-UI: traffic signal intersection id and group id  are not available


dan-du-car:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

There is one traffic signal controller respectively at each side of the road to signalize the vehicle whether the vehicle should stop at the intersection or not. In another word, there are two traffic signal controllers in total for the workzone use case. 

The traffic signal controllers are constantly sending signal to v2xhub. V2xhub forward the corresponding SPAT messages to the SV vehicle. However, vehicle OBU receives all the SPAT messages from both traffic signal controllers. The Platform and UI need to process/display the SPAT correctly and distinguish which SPAT is from the approaching traffic signal controller for the SV. 
### Carma version where this issue was discovered

carma-3.7.0

### Expected behavior
UI should only display the SPAT that  is from the approaching traffic signal controller for the SV. 
### Actual behavior

Currently, the UI subscribe to all incoming SPAT messages and cannot distinguish the SPAT sources in terms of which traffic signal controller is sending the signal.

### Steps to reproduce the actual behavior

N/A

MishkaMN:
closed with above PRs

PR #1425, opened 2021-09-01
Add logic to publish group id and intersection ids

dan-du-car:
# PR Details
## Description
Added logic to get the traffic light regulatory element that is relevant to the SV. Publishing the relevant intersection and group ids for SV web UI to display the correct traffic signal phase.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1420

## Motivation and Context
WorkZone use case

## How Has This Been Tested?
N/A

## Types of changes

## Checklist:

### DIFF
--- a/carma_wm/src/WMListenerWorker.cpp
+++ b/carma_wm/src/WMListenerWorker.cpp
@@ -429,6 +429,30 @@ double WMListenerWorker::getConfigSpeedLimit() const
   return config_speed_limit_;
 }
 
+void WMListenerWorker::currentLocationCallback(const geometry_msgs::PoseStamped& current_pos)
+{  
+  const lanelet::BasicPoint2d cur_loc;
+  cur_loc.x() = current_pos.pose.position.x;
+  cur_loc.y() = current_pos.pose.position.y;
 
+  std::vector<lanelet::CarmaTrafficLightPtr> light_v = world_model_->getLightsAlongRoute(cur_loc);
+  lanelet::CarmaTrafficLightPtr first_light = light_v.front();
+  ROS_DEBUG_STREAM("Get First TF_LIGHT of Id: " << first_light->id());
+  
+  for( const auto& ligth_id_m : traffic_light_ids_ ) 
+  {
+        std::cout << "Key:[" << ligth_id_m.first << "] Value:[" << ligth_id_m.second << "]\n";
+        if(ligth_id_m.second == first_light->id() )
+        {
+            ROS_DEBUG_STREAM("CURRENT Key:[" << ligth_id_m.first << "] Value:[" << ligth_id_m.second << "]"));
+            ROS_DEBUG_STREAM("Get First TF_LIGHT of Id: " << first_light->id());
+            uint32_t temp = ligth_id_m.first;
+            //get intersection id and group id
+            unit8_t group_id = (uint8_t)(temp >> 0);
+            unit16_t intersection_id = (unit16_t)(temp >> 8);
+            ROS_DEBUG_STREAM("Intersection id:[" << intersection_id << "] Groupd id:[" << group_id << "]"));
+        }
+    }
+}
 
 }  // namespace carma_wm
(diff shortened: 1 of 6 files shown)


carma_wm/src/WMListenerWorker.cpp
```diff
@@ -429,6 +429,42 @@ double WMListenerWorker::getConfigSpeedLimit() const
...
+            ROS_DEBUG_STREAM("CURRENT Key:" << ligth_id_m.first << "Value:" << ligth_id_m.second);
+            ROS_DEBUG_STREAM("Get First TF_LIGHT of Id: " << first_light->id());
+            uint32_t temp = ligth_id_m.first;
+            //get intersection id and group id
+            unsigned group_id = (temp & 0xFF);
+            unsigned intersection_id = (temp >> 8);
+            ROS_DEBUG_STREAM("Intersection id: " << intersection_id << " Groupd id:" << group_id);
+            
+            std_msgs::Int32MultiArray intersection_group_ids;
+            intersection_group_ids.data[0] = intersection_id;
```
MishkaMN, line 460: one of these should be `group_id`

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -429,6 +429,42 @@ double WMListenerWorker::getConfigSpeedLimit() const
...
 }
 
+std_msgs::Int32MultiArray WMListenerWorker::getIntersectionGroupIdsByCurLoc(const geometry_msgs::PoseStamped& current_pos)
+{  
+  //Get all traffic light regulatory elements along the route based on current SV position
+  lanelet::BasicPoint2d cur_loc;
+  cur_loc.x() = current_pos.pose.position.x;
+  cur_loc.y() = current_pos.pose.position.y;
+  std::vector<lanelet::CarmaTrafficLightPtr> light_v = world_model_->getLightsAlongRoute(cur_loc);
+  lanelet::CarmaTrafficLightPtr first_light = light_v.front();
```
MishkaMN, line 439: we should check if the returned list is empty or not before getting front() element, otherwise it might segfault

MishkaMN (changes requested):

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -429,6 +429,42 @@ double WMListenerWorker::getConfigSpeedLimit() const
...
+  {
+        std::cout << "Key:[" << ligth_id_m.first << "] Value:[" << ligth_id_m.second << "]\n";
+        if( ligth_id_m.second == first_light->id() )
+        {
+            ROS_DEBUG_STREAM("CURRENT Key:" << ligth_id_m.first << "Value:" << ligth_id_m.second);
+            ROS_DEBUG_STREAM("Get First TF_LIGHT of Id: " << first_light->id());
+            uint32_t temp = ligth_id_m.first;
+            //get intersection id and group id
+            unsigned group_id = (temp & 0xFF);
+            unsigned intersection_id = (temp >> 8);
```
MishkaMN, line 456: can we factor out this small method into a different method like here https://github.com/usdot-fhwa-stol/carma-platform/blob/develop/carma_wm/include/carma_wm/WorldModelUtils.h#L116 (where we pass in the group_id and intersection_ids as a reference to be changed) 
This is just to be consistent

carma_wm/src/WMListenerWorker.cpp
dan-du-car: Addressed https://github.com/usdot-fhwa-stol/carma-platform/pull/1425/commits/ca7ac52778215882aedfe5cabc8f7803fea80f98

msmcconnell (changes requested):
My main concern is the location of this functionality in carma_wm. Each node using the world model contains its own instance, so this logic would result in 10+ publications being added to the system and the UI receiving duplicate information. This would be better placed in carma_wm_ctrl so that the data is only published by one node

carma_wm/src/WMListenerWorker.cpp
dan-du-car: addressed

MishkaMN:
resolved for now with https://github.com/usdot-fhwa-stol/carma-platform/pull/1439
accounting for Mike's comment.

PR #1437, opened 2021-09-15
CARMA-WEB-UI: traffic signal intersection id and group id are not available

aswath1:
# PR Details
## Description
There is one traffic signal controller respectively at each side of the road to signalize the vehicle whether the vehicle should stop at the intersection or not. In another word, there are two traffic signal controllers in total for the workzone use case.

The traffic signal controllers are constantly sending signal to v2xhub. V2xhub forward the corresponding SPAT messages to the SV vehicle. However, vehicle OBU receives all the SPAT messages from both traffic signal controllers. The Platform and UI need to process/display the SPAT correctly and distinguish which SPAT is from the approaching traffic signal controller for the SV.

## Related Issue

https://github.com/usdot-fhwa-stol/carma-platform/issues/1420
## Types of changes

## Checklist:

### DIFF
--- a/carma_wm_ctrl/src/WMBroadcaster.cpp
+++ b/carma_wm_ctrl/src/WMBroadcaster.cpp
@@ -1878,8 +1878,8 @@ void WMBroadcaster::publishLightId()
       if (convert_success)
       {
         ROS_DEBUG_STREAM("Found Traffic Light with Intersection id: " << intersection_id << " Groupd id:" << group_id);
-        upcoming_intersection_ids_.data.push_back(std::static_cast<int>(intersection_id));
-        upcoming_intersection_ids_.data.push_back(std::static_cast<int>(group_id));
+        upcoming_intersection_ids_.data.push_back(static_cast<int>(intersection_id));
+        upcoming_intersection_ids_.data.push_back(static_cast<int>(group_id));
       }
 
     }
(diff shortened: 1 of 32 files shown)


msmcconnell (changes requested):

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -1844,7 +1844,47 @@ void WMBroadcaster::currentLocationCallback(const geometry_msgs::PoseStamped& cu
...
+      if (!traffic_lights.empty())
+      {
+        ROS_DEBUG_STREAM("Found Traffic Light Regulatory Element id: " << traffic_lights.front()->id());
+        convert_success = convertLightIdToInterGroupId(intersection_id,group_id,  traffic_lights.front()->id());
+      }
+      if (convert_success)
+      {
+        ROS_DEBUG_STREAM("Found Traffic Light with Intersection id: " << intersection_id << " Groupd id:" << group_id);
+        upcoming_intersection_ids_.data.push_back(intersection_id);
+        upcoming_intersection_ids_.data.push_back(group_id);
```
dan-du-car, line 1881: using  static_cast<int> to convert this char/int to int before publishing for both intersection_id and group_id

Issue #118, opened 2019-01-10
Platooning plugin should use common headway parameter


qswawrq:
Ideally, the platoon leader should send out a desired headway to its platoon member and the platoon member should maintain the headway based on the receiving headway value. Currently their are keeping a configurable headway with their leader based on its own platooning parameter.

darrelld05:
This issue was transferred from the original CARMAPlatform: Old issue # 1048

PR #1438, opened 2021-09-15
Update Port Drayage Plugin to publish UI Instructions to the Web UI when a route to a newly received destination has been generated.

JonSmet:
# PR Details
## Description

This PR updates the Port Drayage Plugin by enabling it to publish a cav_msgs/UIInstructions.msg message to the /ui/ui_instructions topic, which results in the Web UI displaying a pop-up to the user. 

More specifically, after the Port Drayage Plugin processes a received Mobility Operation message with strategy 'carma/port_drayage' that instructs the CMV to drive to a new destination, the plugin will first call the /guidance/set_active_route service to set a new active route to the received destination. If this route generation is successful, the updates from this PR will cause the Port Drayage Plugin to publish a message to the /ui/ui_instructions topic, which results in a popup on the Web UI that notifies the user that a route to the newly-received destination has been received, and that the user can select to engage the system on that route, or select to remain disengaged (exiting out of the popup will also result in the system remaining disengaged).

Example message in Web UI popup: "A new Port Drayage route with operation type 'PICKUP' has been received. Select YES to engage the system on the route, or select NO to remain disengaged."

## Motivation and Context

After a CARMA Freight vehicle receives a new destination from V2X-Hub and generates a new active route to that destination, the user must be notified that that system can be engaged on the new route.

## How Has This Been Tested?

Integration tested on Ford Fusion along with [CARMA Web UI PR #118](https://github.com/usdot-fhwa-stol/carma-web-ui/pull/118)

## Types of changes

## Checklist:


port_drayage_plugin/src/port_drayage_worker.cpp
```diff
@@ -102,6 +107,33 @@ namespace port_drayage_plugin
         return route_req;
     }
 
+    cav_msgs::UIInstructions PortDrayageWorker::compose_ui_instructions(const PortDrayageMobilityOperationMsg& msg, const std::string& previous_operation) {
```
kjrush, line 110: Looks like the only bit of the operations message that's required here is the operation type? Can we make this just accept that as a parameter instead of the full message?

JonSmet: Fixed.

PR #1443, opened 2021-09-20
a1 release verification testing fixes

MishkaMN:
# PR Details
## Description
Fixes encountered duing a1 verification testing, which includes:
- mobilitypath_visualizer no longer runs into race condition where it fails before map is available
- stop_and_wait_plugin no longer creates "steering right" trajectory due to bug in the way points were being sampled from route. it was comparing traveled downtrack bs lanelet's downtrack percentage to get the nearest point, which is incorrect. It directly compares downtrack using binary search now.
- inlanecruising_plugin as plugin name field
- platooning stack no longer crashes during workzone test run.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1444

## Motivation and Context
See above.

## How Has This Been Tested?

## Types of changes

## Checklist:

### DIFF
--- a/carma_wm/include/carma_wm/IndexedDistanceMap.h
+++ b/carma_wm/include/carma_wm/IndexedDistanceMap.h
@@ -149,7 +149,7 @@ public:
    *        NOTE: Unlike the rest of this class, this method runs in O(log n) where n is this.size() for accessing linestring index
    *              If accessing point index (get_point=true) then it runs an addition O(log m) where m is the linestring index.size();
    *              This means the max complexity of this function is O(log n) + O(log m)
-   *  TODO: Add info about complexity and returned data
+   *  
    * \throw std::invalid_argument if distance does not fit within bounds [0, totalLength()]
    * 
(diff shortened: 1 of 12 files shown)


carma_wm/include/carma_wm/IndexedDistanceMap.h
```diff
@@ -146,11 +146,19 @@ class IndexedDistanceMap
 
   /*!
    * \brief Returns index of the linestring which the provided distance is within.
-   *        NOTE: Unlike the rest of this class, this method runs in O(log n) where n is this.size()
-   *
+   *        NOTE: Unlike the rest of this class, this method runs in O(log n) where n is this.size() for accessing linestring index
+   *              If accessing point index (get_point=true) then it runs an addition O(log m) where m is the linestring index.size();
+   *              This means the max complexity of this function is O(log n) + O(log m)
+   *  TODO: Add info about complexity and returned data
```
MishkaMN, line 152: I think we added this, so we can remove

PR #1445, opened 2021-09-21
Commit changes to fix/wz_plugin

lewisid:
# PR Details
## Description

In order to accomodate the work zone use case testing, I modified the node to read the /vehicle_length parameter and pass it to the worker class. Using this parameter, I ensured that the target stopping location would be shifted backward by vehicle length.
## Related Issue

https://github.com/usdot-fhwa-stol/carma-platform/issues/1446
## Motivation and Context

At the moment the vehicle is trying to stop such that its back axel is on the stop bar. What we really want is for the front of the vehicle to be at the stop bar. The easiest way to do this is to update the wz_strategic_plugin to account for this delta when creating stop and wait maneuvers.
## How Has This Been Tested?
This has been unit tested using gtest/gmock.

## Types of changes

## Checklist:

### DIFF
--- a/wz_strategic_plugin/src/wz_strategic_plugin.cpp
+++ b/wz_strategic_plugin/src/wz_strategic_plugin.cpp
@@ -290,7 +290,7 @@ void WzStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversRequest
 
     // TODO do we need to check for anything before pushing onto the plan?
     resp.new_plan.maneuvers.push_back(composeLaneFollowingManeuverMessage(
-        current_state.downtrack, traffic_light_down_track, current_state.speed, intersection_speed_.get(),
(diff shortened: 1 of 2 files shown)


wz_strategic_plugin/include/wz_strategic_plugin/wz_strategic_plugin_config.h
```diff
@@ -28,6 +28,9 @@ struct WzStrategicPluginConfig
   //! A multiplier to apply to the maximum allowable vehicle deceleration limit so we plan under our capabilities
   double vehicle_decel_limit_multiplier = 0.75;
 
+  //! The length parameter of the participant vehicle used to help calculate the distance before stopping at a red traffic signal
+  double vehicle_length = 10.0;
```
msmcconnell, line 32: Set this default to 5 to match the pacifica config

msmcconnell (commented):
I think the unit tests need to be updated, but i don't see that in the pr

msmcconnell (changes requested):

Issue #1507, opened 2021-10-29
Input vectors empty exception from cooperative lane change


adev4a:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Sometimes towards the end of lane change the cooperative lane change plugin throws a Input Vectors are empty exception.

![CLC_exception](https://user-images.githubusercontent.com/51463994/139490373-eac22c2e-8de5-4094-bb46-c48ef3a7dd65.png)

### Carma version where this issue was discovered
3.8.2

### Expected behavior
Lane change should complete without exceptions
### Actual behavior

Close to the end of a lanechange carma crashes with this exception.
### Steps to reproduce the actual behavior

1. Run carma 3.8.2 on a route that requires a lane change.

msmcconnell:
Seemingly resolved by above PR

PR #1511, opened 2021-11-01
add additional points to lc geom

adev4a:
# PR Details
Fix for input vector is empty exception thrown at the end of a lanechange. 
## Description
The method proposed adds centerline points from the lanelet following the last lanelet in lane change. 

## Related Issue
Issue #1507 

## Types of changes

## Checklist:

### DIFF
--- a/basic_autonomy/src/basic_autonomy.cpp
+++ b/basic_autonomy/src/basic_autonomy.cpp
@@ -870,6 +870,14 @@ namespace basic_autonomy
             lanelet::BasicLineString2d new_points = create_lanechange_path(lanelets_in_path[lane_change_iteration],lanelets_in_path[lane_change_iteration + 1]);
             centerline_points.insert(centerline_points.end(), new_points.begin(), new_points.end());
 
+            //Add points from following lanelet to provide sufficient distance for adding buffer
+            auto following_lanelets = wm->getMapRoutingGraph()->following(lanelets_in_path[lane_change_iteration + 1], false);
+            if(!following_lanelets.empty()){
+                //Arbitrarily choosing first following lanelet for buffer since points are only being used to fit spline
+                centerline_points.insert(centerline_points.end(), following_lanelets.front().centerline2d().basicLineString().begin(), 
+                                                                            following_lanelets[0].centerline2d().basicLineString().end());
+            }
+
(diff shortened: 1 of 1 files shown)


basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -870,6 +870,14 @@ namespace basic_autonomy
             lanelet::BasicLineString2d new_points = create_lanechange_path(lanelets_in_path[lane_change_iteration],lanelets_in_path[lane_change_iteration + 1]);
             centerline_points.insert(centerline_points.end(), new_points.begin(), new_points.end());
 
+            //Add points from following lanelet to provide sufficient distance for adding buffer
+            auto following_lanelets = wm->getMapRoutingGraph()->following(lanelets_in_path[lane_change_iteration + 1], false);
```
msmcconnell, line 874: Does lanelets_in_path[lane_change_iteration + 1 give us the straight lanelet following the lanechange?

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -870,6 +870,14 @@ namespace basic_autonomy
...
+            if(!following_lanelets.empty()){
+                //Arbitrarily choosing first following lanelet for buffer since points are only being used to fit spline
+                centerline_points.insert(centerline_points.end(), following_lanelets.front().centerline2d().basicLineString().begin(), 
+                                                                            following_lanelets[0].centerline2d().basicLineString().end());
```
msmcconnell, line 878: I'm a bit concerned about this call
following_lanelets.front().centerline2d().basicLineString()
When this is called is the returned vector always at the same address or is it copied out as a new vector. If the latter then these iterators (begin(), end()) might not point to the same data set. I think it would be safer to store the basicLineString first then access the iterators

msmcconnell (changes requested):
Also I think we want this pointed at Battista and Elise as two duplicate PRs

basic_autonomy/src/basic_autonomy.cpp
adev4a: lanelets_in_path[lane_change_iteration + 1, gives the last lanelet involved in lane change. We are using the lanelet following() that which is the straight lanelet following the lanelet

msmcconnell (approved):
LGTM, @JonSmet should make the call on merging it

msmcconnell:
Closing in favor of #1533 which contains this change

PR #1518, opened 2021-11-03
Commit edits to bsm_generator node

lewisid:
# PR Details
## Description

Updated CARMA to keep a fixed BSM ID. The initial BSM ID may be random, but persistent through the course of the test cycle. This is required for the simulation to know which vehicle is which.  
## Related Issue

https://github.com/usdot-fhwa-stol/carma-platform/issues/1525
## Motivation and Context

This change is required for simulation testing to keep track of vehicle bsm messages through their id rather than lose data with constantly shifting ids.
## Types of changes

## Checklist:


bsm_generator/src/bsm_generator.cpp
```diff
@@ -124,7 +125,11 @@ namespace bsm_generator
     {
         bsm_.header.stamp = ros::Time::now();
         bsm_.core_data.msg_count = worker.getNextMsgCount();
-        bsm_.core_data.id = worker.getMsgId(ros::Time::now());
+        
+        if(bsm_id_rotation_enabled_)
+        {
+            bsm_.core_data.id = worker.getMsgId(ros::Time::now());
```
msmcconnell, line 131: How does this value get set in he non-rotating case? Seems a default value should be generated on node initialization or loaded from a parameter

msmcconnell (changes requested):

bsm_generator/src/bsm_generator.cpp
```diff
@@ -23,9 +23,17 @@ namespace bsm_generator
...
     {
+        int bsmid;
         nh_.reset(new ros::CARMANodeHandle());
         pnh_.reset(new ros::CARMANodeHandle("~"));
         pnh_->param<double>("bsm_generation_frequency", bsm_generation_frequency_, 10.0);
+        pnh_->param<bool>("bsm_id_rotation_enabled", bsm_id_rotation_enabled_, true);
+        pnh_->param<int>("bsm_message_id", bsmid);
+        for(size_t i = 0; i < 4; ++i )
+        {
+            bsm_message_id_[i] = bsmid >> (8 * i);
```
msmcconnell, line 34: This looks like a guaranteed segfault as you are access element i of bsm_message_id_ before it is set. 
Since you are just doing a regular iteration you can just do on this line instead
```
bsm_message_id_.emplace_back( bsmid >> (8 * i) );
```

bsm_generator/src/bsm_generator.cpp
```diff
@@ -23,9 +23,17 @@ namespace bsm_generator
 
     void BSMGenerator::initialize()
     {
+        int bsmid;
         nh_.reset(new ros::CARMANodeHandle());
         pnh_.reset(new ros::CARMANodeHandle("~"));
         pnh_->param<double>("bsm_generation_frequency", bsm_generation_frequency_, 10.0);
+        pnh_->param<bool>("bsm_id_rotation_enabled", bsm_id_rotation_enabled_, true);
+        pnh_->param<int>("bsm_message_id", bsmid);
+        for(size_t i = 0; i < 4; ++i )
```
msmcconnell, line 32: Comment here explaining why 4 is the endpoint would be good for clarity

msmcconnell (changes requested):

bsm_generator/src/bsm_generator.cpp
```diff
@@ -23,9 +23,17 @@ namespace bsm_generator
 
     void BSMGenerator::initialize()
     {
+        int bsmid;
         nh_.reset(new ros::CARMANodeHandle());
         pnh_.reset(new ros::CARMANodeHandle("~"));
         pnh_->param<double>("bsm_generation_frequency", bsm_generation_frequency_, 10.0);
+        pnh_->param<bool>("bsm_id_rotation_enabled", bsm_id_rotation_enabled_, true);
+        pnh_->param<int>("bsm_message_id", bsmid);
```
msmcconnell, line 31: Lets provide a default value of 0 here

msmcconnell: Actually I take that back. This is better since it will enforce the parameter is set by the user

bsm_generator/config/parameters.yaml
```diff
@@ -1,2 +1,5 @@
 # The desired frequency for BSM generation
 bsm_generation_frequency: 10.0
+
+#Enable/disable rotation of BSM ID during vehicle operations
+bsm_id_rotation_enabled: true
```
msmcconnell, line 5: This can be removed from here if it is set in one of the vehicle config files

msmcconnell (changes requested):

bsm_generator/src/bsm_generator.cpp
```diff
@@ -23,9 +23,17 @@ namespace bsm_generator
 
     void BSMGenerator::initialize()
     {
+        int bsmid;
         nh_.reset(new ros::CARMANodeHandle());
         pnh_.reset(new ros::CARMANodeHandle("~"));
         pnh_->param<double>("bsm_generation_frequency", bsm_generation_frequency_, 10.0);
+        pnh_->param<bool>("bsm_id_rotation_enabled", bsm_id_rotation_enabled_, true);
+        pnh_->param<int>("bsm_message_id", bsmid, 0);
```
msmcconnell, line 31: If these parameters are being loaded from global scope then they need a leading slash.

msmcconnell (approved):

PR #1520, opened 2021-11-03
Potential ILC 32-bit range exception fix correlation to lanechange geometry creation

MishkaMN:
# PR Details
## Description

At ATEF, we were running into this "duration is out of 32-bit range" (from vehicle state having a speed of -nan in ILC logs) ~1-2 seconds after starting the lane change. From our experience, this could actually come from lane change's faulty logic as the state with erroneous speed can be sent to ILC through plan_delegator. This could be related to a small bug in CLC where it is creating extra lane_follow geometry behind the lane change to avoid running out of points that is creating duplicate points when attaching lanelets (last and first points of the two lanelets).

This is only a potential fix, as the duplicate points very well be removed through downsampling or getting the nearest point of end_dist. However, using +1 point is definitely the correct approach even if unrelated

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/pull/1515
https://github.com/usdot-fhwa-stol/carma-platform/issues/1521

## Motivation and Context
See above.

## How Has This Been Tested?
Not yet.

## Types of changes

## Checklist:

### DIFF
--- a/basic_autonomy/src/basic_autonomy.cpp
+++ b/basic_autonomy/src/basic_autonomy.cpp
@@ -902,7 +902,7 @@ namespace basic_autonomy
             if(!following_lanelets.empty()){
                 //Arbitrarily choosing first following lanelet for buffer since points are only being used to fit spline
                 auto following_lanelet_centerline = following_lanelets.front().centerline2d().basicLineString();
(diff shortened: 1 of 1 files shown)


msmcconnell (approved):

msmcconnell:
Closing in favor of #1533 which contains this change

PR #1526, opened 2021-11-04
Feature/update lane change trajectory

adev4a:
# PR Details
This PR changes the lane change logic to remove the constrain that lane change should be over a complete lanelet. With this change lanechange can also span multiple lanelets.

## Types of changes

## Checklist:


basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+                    shared_boundary_found = true;
+                }
+
+                else if(current_lanelet.rightBound() == ending_lanelet.leftBound()){
+                    shared_boundary_found = true;
+                }
+
+                else{
+                    //In case there are no following lanelets on route, lanechange should be completing before reaching it
+                    if(wm->getMapRoutingGraph()->following(current_lanelet, false).empty())
```
msmcconnell, line 271: I think a bit more description on this check would be helpful. As I understand it we could have a lanechange that touches the final lanelet on the route, but presumable the above two checks would trigger on this iteration and so we wouldn't hit this?

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+                    {
+                        throw(std::invalid_argument("No following lanelets from current lanelet reachable without a lane change, incorrectly chosen end lanelet"));
+                    }
+
+                    current_lanelet = wm->getMapRoutingGraph()->following(current_lanelet, false).front(); 
+                    auto current_lanelet_linestring = current_lanelet.centerline2d().basicLineString();    
+                    reference_centerline.insert(reference_centerline.end(), current_lanelet_linestring.begin(), current_lanelet_linestring.end());
```
msmcconnell, line 278: We need the +1 for overlap

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+                    if(wm->getMapRoutingGraph()->right(starting_lane[0])){
+                        curr_end_lanelet = wm->getMapRoutingGraph()->right(starting_lane[0]).get();
+                    }
+                    else{
+                        curr_end_lanelet = wm->getMapRoutingGraph()->adjacentRight(starting_lane[0]).get();
+                    }
+                }
+                
+                auto ending_lane_linestring = curr_end_lanelet.centerline2d().basicLineString();
+                ending_lane_centerline.insert(ending_lane_centerline.end(), ending_lane_linestring.begin(), ending_lane_linestring.end());
```
msmcconnell, line 310: Always remember +1

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+                
+            }
```
msmcconnell, line 312: Under the current logic, I think there is a case missing in this loop for when we are already in the ending lane. This plugin will still be getting called until the lane change is actually complete. However, if we did not rely on vehicle state and used the start, end lanelets from the maneuver as mentioned in my earlier comment this would not be an issue.

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+
+            cav_msgs::VehicleState ending_state = waypoint_generation::get_nearest_state_to_downtrack(ending_lane_centerline, ending_downtrack, wm);
+
+            int reference_line_start_index = waypoint_generation::get_nearest_point_index(reference_centerline, state);
+            int ending_line_end_index = waypoint_generation::get_nearest_point_index(ending_lane_centerline, ending_state);
+
+            lanelet::BasicLineString2d starting_centerline(reference_centerline.begin() + reference_line_start_index, reference_centerline.end());
+            lanelet::BasicLineString2d ending_centerline(ending_lane_centerline.begin(), ending_lane_centerline.begin() + ending_line_end_index);
```
msmcconnell, line 320: I find the mismatch of the index's between start and end centerline a bit strange. Lets says we had the lanechange crossing a lanelet boundary. The ending centerline would start from the start of the earliest lanelet on the target side then end partway through the target lanelet. While the starting centerline would start partway through the starting lanelet and go till the end of the final starting side lanelet. I'm not exactly sure if it would cause issues, but what's the advantage of doing it that way over either keeping them approximately aligned?

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+
+            //Convert to vector of BasicPoint2d
+
+                //Downsample centerlines
+            lanelet::BasicLineString2d downsampled_start_centerline;
+            downsampled_start_centerline.reserve(400);
+            downsampled_start_centerline = carma_utils::containers::downsample_vector(starting_centerline, 5);
```
msmcconnell, line 327: Hardcoded downsample should be made a parameter. Additionally, an note/assumption should be placed in the method comments that the user has selected a down sample which gives reasonable distances for spline fitting based on the map in use.

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+            std::vector<lanelet::BasicPoint2d> start_centerline;
+            for(int i = 0;i<starting_centerline.size();++i){
+                start_centerline.push_back(starting_centerline[i]);
```
msmcconnell, line 330: The names start_centerline and starting_centerline are hard to distinguish when looking at the code and the difference in meaning is not clear. Consider renaming.

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+            }
+
+            lanelet::BasicLineString2d downsampled_end_centerline;
+            downsampled_end_centerline.reserve(400);
+            downsampled_end_centerline = carma_utils::containers::downsample_vector(ending_centerline, 5);
+            std::vector<lanelet::BasicPoint2d> end_centerline;
+            for(int i = 0; i<ending_centerline.size();++i){
+                end_centerline.push_back(ending_centerline[i]);
```
msmcconnell, line 338: Same note on ending vs end centerline name

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+                auto centerlines = resample_linestrings_to_same_length(start_centerline, end_centerline);
+                start_centerline = centerlines.front();
+                end_centerline = centerlines.back();
+
+            }
+
+            //Create Trajectory geometry
+            double delta_step = 1.0 / start_centerline.size();
+
+            for (int i = 0; i < start_centerline.size(); i++)
```
msmcconnell, line 353: For all these positive iterators use size_t instead of int to avoid compiler/sonar cloud warnings

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+
+                centerline_points.push_back(current_position);
+            }
+
+            //Add points from following lanelet to provide sufficient distance for adding buffer
+            auto following_lanelets = wm->getMapRoutingGraph()->following(ending_lanelet, false);
+            if(!following_lanelets.empty()){
+            //Arbitrarily choosing first following lanelet for buffer since points are only being used to fit spline
+            auto following_lanelet_centerline = following_lanelets.front().centerline2d().basicLineString();
+            centerline_points.insert(centerline_points.end(), following_lanelet_centerline.begin(),
```
msmcconnell, line 370: Indent block for readability

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+                double delta = delta_step * i;
+                current_position.x() = end_lane_pt.x() * delta + (1 - delta) * start_lane_pt.x();
+                current_position.y() = end_lane_pt.y() * delta + (1 - delta) * start_lane_pt.y();
+
+                centerline_points.push_back(current_position);
+            }
+
+            //Add points from following lanelet to provide sufficient distance for adding buffer
+            auto following_lanelets = wm->getMapRoutingGraph()->following(ending_lanelet, false);
+            if(!following_lanelets.empty()){
```
msmcconnell, line 367: Seems like this check is in the wrong place. The only purpose of the buffers is to support spline fitting, but the spline was already fit in the resample_linestrings_to_same_length method

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+            //Arbitrarily choosing first following lanelet for buffer since points are only being used to fit spline
+            auto following_lanelet_centerline = following_lanelets.front().centerline2d().basicLineString();
+            centerline_points.insert(centerline_points.end(), following_lanelet_centerline.begin(), 
+                                                                        following_lanelet_centerline.end());
+            }
+
+            return centerline_points;
+        }
+
+        std::vector<std::vector<lanelet::BasicPoint2d>> resample_linestrings_to_same_length(std::vector<lanelet::BasicPoint2d> line_1, std::vector<lanelet::BasicPoint2d> line_2){
```
msmcconnell, line 377: This is a pretty costly function since both the spline fitting occurs twice and the compute_arc_lengths method is called twice. I think we ought to log the execution time as a debug log.

basic_autonomy/src/helper_functions.cpp
```diff
@@ -112,5 +112,28 @@ namespace waypoint_generation
         double ending_downtrack = wm->routeTrackPos(state_pos).downtrack;
         return get_nearest_index_by_downtrack(points, wm, ending_downtrack);
     }
+
+    cav_msgs::VehicleState get_nearest_state_to_downtrack(const std::vector<lanelet::BasicPoint2d>& points, double target_downtrack, const carma_wm::WorldModelConstPtr &wm)
```
msmcconnell, line 116: Doesn't a method in the world model do basically this, like get nearest index by downtrack or something like that?

msmcconnell (changes requested):

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
             return constrained_points;
         }
 
+        std::vector<lanelet::BasicPoint2d> create_lanechange_route(int ending_lane_id, double starting_downtrack, double ending_downtrack,
+                                                                   const carma_wm::WorldModelConstPtr &wm,const cav_msgs::VehicleState &state)
+        {
+            std::vector<lanelet::BasicPoint2d> centerline_points = {};
+
+            //Get starting lanelet and ending lanelets
+            auto lanelets = wm->getLaneletsBetween(starting_downtrack, ending_downtrack,true);
```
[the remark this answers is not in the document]
adev4a: Should we still be using the starting lanelet id from the maneuver if the tactical plugin is being asked to plan before it?
We do update the starting downtrack to be the current downtrack in case we are behind. Would that still be valid?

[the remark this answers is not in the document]
msmcconnell: I think if the maneuver's starting lanelet does not include the starting downtrack than it is an invalid maneuver.

basic_autonomy/src/basic_autonomy.cpp
adev4a: There are 3 instances of spline fitting here. The first 2 happen on the starting and target centerlines only to resample them. At this point we use a downsampled centerline which should always be able to fit through the spline.
If the centerlines are the same length, the spline fitting wouldn't happen.

At a later point, we resize the geometry points we are using for the trajectory, to be from the current state to the end.
Here we add a buffer, and then fit the spline through the actual lane change path/geometry points which the vehicle will use while lane changing.
I'm guessing a buffer would still be needed here?

msmcconnell: Can you point me to where this 3rd fit occurs?

adev4a: Just realized the third fit got overwritten, i'm adding it in now. It will happen in compose_lanechange_trajectory_from_path

basic_autonomy/include/basic_autonomy/basic_autonomy.h
```diff
@@ -263,6 +263,14 @@ namespace basic_autonomy
...
                                                                    const carma_wm::WorldModelConstPtr &wm, cav_msgs::VehicleState &ending_state_before_buffer,
                                                                    const cav_msgs::VehicleState &state, const DetailedTrajConfig &detailed_config);
+     
+     std::vector<lanelet::BasicPoint2d> create_lanechange_route(int ending_lane_id, double starting_downtrack, double ending_downtrack,
+                                                            const carma_wm::WorldModelConstPtr &wm,const cav_msgs::VehicleState &state);
+
+     cav_msgs::VehicleState get_nearest_state_to_downtrack(const std::vector<lanelet::BasicPoint2d>& points, double target_downtrack, 
+                                                            const carma_wm::WorldModelConstPtr &wm);   
+
+     std::vector<std::vector<lanelet::BasicPoint2d>> resample_linestrings_to_same_length(std::vector<lanelet::BasicPoint2d> line_1, std::vector<lanelet::BasicPoint2d> line_2);
```
kjrush, line 273: This is also kind of unclear in terms of what it refers to, is length spatial (meters) or list size (number of elements). I'd recommend clarifying by renaming to "...same_num_points" or something if that's the case.

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+            lanelet::ConstLanelet starting_lanelet =lanelets.front();
```
kjrush, line 244: Worth either checking for, or documenting an assumption or pre-condition, but this is an unchecked access into a vector that's not 100% guaranteed to have elements in it. If the above call fails for any reason (bad map, invalid maneuver) this call will segfault for accessing lanelets[0].

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+            lanelet::ConstLanelet ending_lanelet = wm->getMap()->laneletLayer.get(ending_lane_id);
+            
+            lanelet::ConstLanelet current_lanelet = starting_lanelet;
+            lanelet::ConstLanelets starting_lane;
```
kjrush, line 248: Consider renaming to starting_lanelet for consistency

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+            starting_lane.push_back(starting_lanelet);
+
+            std::vector<lanelet::BasicPoint2d> reference_centerline;
+            bool shared_boundary_found = false;
+            bool is_lanechange_left = false;
+            
+            lanelet::BasicLineString2d current_lanelet_centerline = starting_lanelet.centerline2d().basicLineString();
+            reference_centerline.insert(reference_centerline.end(), current_lanelet_centerline.begin(), current_lanelet_centerline.end());
+
+            while(!shared_boundary_found){
```
kjrush, line 258: Is it possible that this infinite loops? If we had a circular track but solid/unpassable boundaries between all left/right lanelets, would we spin here forever? Unlikely to happen that obviously, but graph cycles can often be non-obvious. It may be worth adding some kind of hard upper bound or perhaps a way to track if we've made a loop of some kind.

kjrush: Looking at it further, if we're only looking at lanelets on the route, this may not be an issue.

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+
+            //Sample spline to get centerlines of equal size
+            std::vector<lanelet::BasicPoint2d> all_sampling_points_line1;
+            std::vector<lanelet::BasicPoint2d> all_sampling_points_line2;
+
+            size_t total_point_size = std::min(line_1.size(), line_2.size());
+
+            all_sampling_points_line1.reserve(1 + total_point_size * 2);
+            std::vector<double> downtracks_raw_line1 = carma_wm::geometry::compute_arc_lengths(line_1);
+            int total_step_along_curve1 = static_cast<int>(downtracks_raw_line1.back() / 2.0);
```
kjrush, line 401: Is static_cast what we want here? Or is math::round more appropriate? Static cast here is equivalent to math::floor().

kjrush, line 401: Also if we're accessing `*.back()` it's probably worth documenting the assumption that the input lists are non-empty and/or actually checking and enforcing that ourselves. We should do what we can to ensure that bad input doesn't segfault nodes but is instead captured and logged, then handled.

basic_autonomy/src/helper_functions.cpp
kjrush: And this doesn't seem to be fully populating the VehicleState message either, so that may not be the correct data structure to use for just the raw X, Y data. I'd be inclined to use the WM function if it fits the right API.

basic_autonomy/include/basic_autonomy/helper_functions.h
```diff
@@ -95,7 +95,18 @@ namespace waypoint_generation
...
+    /**
+     * \brief Downsamples an input lanelet::BasicPoint2d vector by saving only each nth element.
+     * For example, given an input vector of { 0, 1, 2, 3, 4, 5 } and n = 2
+     * The output vector will be {0, 2, 4}
+     * 
+     * \param input The input lanelet::BasicPoint2d vector to downsample
+     * \param n The count of the elements to save
+     * \return The downsampled vector
+     */ 
+    std::vector<lanelet::BasicPoint2d> downsample_basicpoint2d_vector(const std::vector<lanelet::BasicPoint2d>& input, unsigned int n);
```
msmcconnell, line 109: What does this method do that the downsample method from carma-utils does not? The utils method is templated and therefore can support any type of vector, this seems like an unnecessary specialization?

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,7 +234,208 @@ namespace basic_autonomy
...
+            reference_centerline.reserve(400);
+            bool shared_boundary_found = false;
+            bool is_lanechange_left = false;
+            
+            lanelet::BasicLineString2d current_lanelet_centerline = starting_lanelet.centerline2d().basicLineString();
+            lanelet::ConstLanelet current_lanelet = starting_lanelet;
+            reference_centerline.insert(reference_centerline.end(), current_lanelet_centerline.begin(), current_lanelet_centerline.end());
+
+            while(!shared_boundary_found){
+                //Assumption here is for lane change to happen between two adjacent lanelets, they must share a lane boundary (linestring)
```
msmcconnell, line 261: Good to list the assumption here, but its really needed at the interface comment rather than in the implementation. Assumptions need to be understood by the method user.

msmcconnell (changes requested):

msmcconnell (approved):
No more comments assuming all of Kyle's get addressed.

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -234,6 +234,198 @@ namespace basic_autonomy
...
+            std::vector<lanelet::BasicPoint2d> reference_centerline;
+            bool shared_boundary_found = false;
+            bool is_lanechange_left = false;
+            
+            lanelet::BasicLineString2d current_lanelet_centerline = starting_lanelet.centerline2d().basicLineString();
+            reference_centerline.insert(reference_centerline.end(), current_lanelet_centerline.begin(), current_lanelet_centerline.end());
+
+            while(!shared_boundary_found){
+
+                if(current_lanelet.leftBound() == ending_lanelet.rightBound()){
```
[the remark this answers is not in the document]
kjrush: I think this comment might be better suited for the method doc comment in the definition, doesn't hurt to also have it here however. The doc comments bear extra weight as they're normally what would pop-up when hovering over a method in an IDE or would be converted into method documentation by a tool like Doxygen or similar. Having the important details someone would need to know about how to use a method there is extra important.

basic_autonomy/src/basic_autonomy.cpp
kjrush: Has this been addressed?

basic_autonomy/src/basic_autonomy.cpp
kjrush: Any plans to address this?

basic_autonomy/src/basic_autonomy.cpp
kjrush: Still think it would be a good idea to check at the top of this method to catch if either of the input lists are empty and gracefully fail. We've seen several cases where 0 length maneuvers get generated via some weird geometry or other edge cases, and having a proper exception for those cases is substantially easier to trace and debug than a silent segfault or other form of undefined behavior.

In general validating inputs like that as a form of defensive programming is something I'd advocate for, if a method depends on a particular property of it's inputs (non-emptiness, non-nullity, etc.) it should validate those properties itself so that errors can be caught. If most or all functions do this then it means that errors can be caught (and hopefully handled) closer to their actual origin instead of deeper in the callstack.

basic_autonomy/src/basic_autonomy.cpp
adev4a: Yes, we are using the starting downtrack for the maneuver to plan the lane change trajectory, like suggested in the second part of the comment

basic_autonomy/src/basic_autonomy.cpp
adev4a: This was updated so we query the lanelet directly from the world model based on the lanelet id

basic_autonomy/src/basic_autonomy.cpp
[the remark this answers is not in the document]
adev4a: The comment was added to the method doc comment and here as well

basic_autonomy/src/basic_autonomy.cpp
adev4a: For this one, there is a curve fitting that happens right before this which requires there be atleast 4 points in the vector. There is an exception being thrown if that fails, would that be a sufficient check?

kjrush (approved):

msmcconnell:
Local build testing confirmed validity. CI may be misconfigured. Will merge this PR

PR #1531, opened 2021-11-05
tsmo use case one plugins

saina-ramyar:
# PR Details

## Description
In this PR, Strategic and Tactical Stop-Controlled Intersection Plugins are updated based on integration testing of TSMO Use Case #1 on CARMA PLatform using mock schedule messages.

## How Has This Been Tested?
Unit test and integration test on Black and white pacificas.

## Types of changes

## Checklist:


sci_strategic_plugin/src/sci_strategic_plugin.cpp
```diff
@@ -86,63 +86,90 @@ void SCIStrategicPlugin::mobilityOperationCb(const cav_msgs::MobilityOperationCo
...
-      scheduled_latest_depart_time_ = p.second.get<uint32_t>("latest_depart_p");
-      ROS_DEBUG_STREAM("scheduled_latest_depart_time_: " << scheduled_latest_depart_time_);
+  // sample strategy_params: "st:1634067044,et:1634067059, dt:1634067062.3256602,dp:2,,access: 0"
+  std::string params = strategy_params;
+  std::vector<std::string> inputsParams;
+  boost::algorithm::split(inputsParams, params, boost::is_any_of(","));
+
+  std::vector<std::string> st_parsed;
+  boost::algorithm::split(st_parsed, inputsParams[0], boost::is_any_of(":"));
+  uint32_t st = std::stoi(st_parsed[1]);
```
kjrush, line 145: Is there a reason for this intermediate variable in all of these parsings, instead of just:
`scheduled_stop_time_ = std::stoi(st_parsed[1])`?

sci_strategic_plugin/src/sci_strategic_plugin.cpp
```diff
@@ -226,29 +257,71 @@ bool SCIStrategicPlugin::planManeuverCb(cav_srvs::PlanManeuversRequest& req, cav
...
 
   auto stop_intersection_list = wm_->getIntersectionsAlongRoute({ req.veh_x, req.veh_y });
   auto nearest_stop_intersection = stop_intersection_list.front();
-  double stop_intersection_down_track =
-  wm_->routeTrackPos(nearest_stop_intersection->stopLines().front().front().basicPoint2d()).downtrack;
+  
+  // extract the intersection stop line information
+  std::vector<double> stop_lines;
+
+  for (auto l:nearest_stop_intersection->stopLines())
```
kjrush, line 287: Whitespace near the `:` could use some cleanup.

sci_strategic_plugin/src/sci_strategic_plugin.cpp
```diff
@@ -226,29 +257,71 @@ bool SCIStrategicPlugin::planManeuverCb(cav_srvs::PlanManeuversRequest& req, cav
...
+  {
+    //TODO: temp use veh_length here until planning stack is updated with front bumper pos
+    double stop_bar_dtd = wm_->routeTrackPos(l.front().basicPoint2d()).downtrack - config_.veh_length;
+    stop_lines.push_back(stop_bar_dtd);
+  }
+  std::sort(stop_lines.begin(), stop_lines.end());
+  
+  double stop_intersection_down_track = stop_lines.front();
```
kjrush, line 295: I'd recommend a check that stop_lines isn't empty (and throwing an exception if it is for some reason, or just logging it and returning no maneuvers) prior to dereferencing an element of it.

sci_strategic_plugin/src/sci_strategic_plugin.cpp
```diff
@@ -226,29 +257,71 @@ bool SCIStrategicPlugin::planManeuverCb(cav_srvs::PlanManeuversRequest& req, cav
...
+  std::vector<lanelet::ConstLanelet> intersection_lanelets =
+          getLaneletsBetweenWithException(stop_intersection_down_track, intersection_end_downtrack, true, true);
+
+  // find the turn direction at intersection:
+
+  intersection_turn_direction_ = getTurnDirectionatIntersection(intersection_lanelets);
+
+  uint32_t base_time = street_msg_timestamp_;
+  
+  bool time_to_approach_int = (int(scheduled_stop_time_) - int(street_msg_timestamp_))>0;
```
kjrush, line 315: I think this whole clause can be brought down a few lines to be:

`if (scheduled_stop_time_ - street_msg_timestamp_ > 0) {...`

sci_strategic_plugin/src/sci_strategic_plugin.cpp
```diff
@@ -257,59 +330,153 @@ bool SCIStrategicPlugin::planManeuverCb(cav_srvs::PlanManeuversRequest& req, cav
...
-      // at the stop line
-      // stop and wait maneuver
-      auto stop_line_lanelet = nearest_stop_intersection->lanelets().front();
-      double stop_duration = (scheduled_enter_time_ - scheduled_stop_time_)*1000.0;
-      ROS_DEBUG_STREAM("Planning stop and wait maneuver");
-      resp.new_plan.maneuvers.push_back(composeStopAndWaitManeuverMessage(
-        current_downtrack_, stop_intersection_down_track, current_state.speed, stop_line_lanelet.id(),
-        stop_line_lanelet.id(), current_state.stamp,
-        current_state.stamp + ros::Duration(stop_duration)));
+  bool cond1 = (int(street_msg_timestamp_) - int(scheduled_stop_time_)) > 0;
```
kjrush, line 369: Is there a reason for this int casts? I'm not sure what role they serve, if we're looking for some kind of an epsilon factor I'd rather be explicit about it `if (street_msg_timestamp - scheduled_stop_time > EPSLION) { ...` for clarity's sake than relying on datatype conversion to do it.

Also cond1 and cond2 could be named better, to make it more clear what's being checked and why it relates to having enough time to stop. If not renaming them, at least a comment would be helpful to describe what this block of logic does.

sci_strategic_plugin/src/sci_strategic_plugin.cpp
```diff
@@ -257,59 +330,153 @@ bool SCIStrategicPlugin::planManeuverCb(cav_srvs::PlanManeuversRequest& req, cav
...
+  ROS_DEBUG_STREAM("time_to_stop_at_stopline: " << time_to_stop_at_stopline);
 
-    }
+  if (time_to_stop_at_stopline)
+  {
+    base_time = std::max(scheduled_stop_time_, street_msg_timestamp_);
+    double stop_duration = (scheduled_enter_time_ - base_time)/1000;
+    ROS_DEBUG_STREAM("stop_duration:  " << stop_duration);
+    ROS_DEBUG_STREAM("Planning stop and wait maneuver");
+    double stopping_accel = 1.5;
```
kjrush, line 380: Should this be parameterized? Either for this plugin or based on the vehicle's decelearation limits?

sci_strategic_plugin/src/sci_strategic_plugin.cpp
```diff
@@ -257,59 +330,153 @@ bool SCIStrategicPlugin::planManeuverCb(cav_srvs::PlanManeuversRequest& req, cav
...
+            current_state.stamp + ros::Duration(stop_duration)));
+  }
+
+  if (!is_allowed_int_)
+  {
+    ROS_DEBUG_STREAM("Vehicle at stop line, waiting for access to intersection");
+    // vehicle is at the intersection but it is not allowed access, so it must wait
+    // arbitrary parameters
+    double stop_acc = 1.5;
+    double stop_duration = 99999;
```
kjrush, line 394: Recommend again parameterizing the accel, and using `std::numeric_limits<double>::max()` or `std::numer_limits<double>::infinity()` for indefinite values.

sci_strategic_plugin/src/sci_strategic_plugin.cpp
```diff
@@ -257,59 +330,153 @@ bool SCIStrategicPlugin::planManeuverCb(cav_srvs::PlanManeuversRequest& req, cav
...
     resp.new_plan.maneuvers.push_back(composeIntersectionTransitMessage(
-      current_downtrack_, intersection_end_downtrack, current_state.speed, intersection_speed_limit,
-      current_state.stamp, req.header.stamp + ros::Duration(intersection_transit_time), crossed_lanelets.front().id(), crossed_lanelets.back().id()));
+      current_state.downtrack, intersection_end_downtrack+20, current_state.speed, intersection_speed_limit,
+      current_state.stamp, req.header.stamp + ros::Duration(intersection_transit_time), intersection_turn_direction_, crossed_lanelets.front().id(), crossed_lanelets.back().id()));
+    
     // when passing intersection, set the flag to false
-    // TODO: Another option, check the time on mobilityoperation
-    approaching_stop_controlled_interction_ = false;
+    // TODO: Another option, check the time on mobilityoperation to see if it passes intersection
```
kjrush, line 442: Is this todo tracked somewhere as a github issue or JIRA story?

sci_strategic_plugin/src/sci_strategic_plugin.cpp
```diff
@@ -394,19 +571,19 @@ cav_msgs::Maneuver SCIStrategicPlugin::composeLaneFollowingManeuverMessage(int c
...
   {
     case 1:
       caseOneSpeedProfile(speed_before_decel, start_speed, time_to_stop, &float_metadata_list);
       break;
     case 2:
-      caseTwoSpeedProfile(end_dist, speed_before_decel, start_speed, time_to_stop, speed_limit_, &float_metadata_list);
-      break;
-    case 3:
-      caseThreeSpeedProfile(end_dist, start_speed, time_to_stop, &float_metadata_list);
+      caseTwoSpeedProfile(end_dist-start_dist, speed_before_decel, start_speed, time_to_stop, speed_limit_, &float_metadata_list);
```
kjrush, line 582: Is case 3 intended to be running the case 2 speed profile? It's unclear, and the dead code below indicates possibly otherwise?

sci_strategic_plugin/src/sci_strategic_plugin.cpp
```diff
@@ -433,28 +610,45 @@ double SCIStrategicPlugin::findSpeedLimit(const lanelet::ConstLanelet& llt) cons
...
-
-    std::string id(bsm_id.begin(), bsm_id.end());
-    mo_.header.sender_bsm_id = id;
+    mo_.header.timestamp = ros::Time::now().toNSec()/1000000;
+    mo_.header.sender_id = config_.vehicle_id;
+    mo_.header.sender_bsm_id = bsm_id_;
+    mo_.strategy = stop_controlled_intersection_strategy_;
 
-    int flag = (approaching_stop_controlled_interction_ ? 1 : 0);
+    int flag = (is_allowed_int_ ? 1 : 0);
```
kjrush, line 621: Should just be able to assign a bool value to an int for an implicit conversion, and vice-versa.

arbitrator/config/arbitrator_params.yaml
```diff
@@ -26,4 +26,4 @@ use_fixed_costs: true
 # Map: The priorities associated with each plugin during the planning 
 # process, values will be normalized at runtime and inverted into costs
 # Unit: N/a
-plugin_priorities: {RouteFollowingPlugin: 4.0, WorkZonePlugin: 8.0, PlatooningStrategicPlugin: 1.0}
+plugin_priorities: {SCIStrategicPlugin: 2.0, RouteFollowingPlugin: 1.0, WorkZonePlugin: 1.0, PlatooningStrategicPlugin: 1.0}
```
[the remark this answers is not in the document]
saina-ramyar: SCI is short for stop-controlled intersection. I used the acronym mainly for the namespace and class names to avoid long lines. But I can change the plugin name if needed.

sci_strategic_plugin/config/parameters.yaml
```diff
@@ -5,11 +5,23 @@ vehicle_decel_limit_multiplier : 0.50
...
+stop_line_buffer : 3.0
 
 # Double: Update time interval of carma streets. Units in s
 delta_t : 1.0
 
+# Double: Update min_gap. Units in m
+min_gap : 10.0
+
+# Double: Length od the vehicle
+veh_length : 4.0
```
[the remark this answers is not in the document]
saina-ramyar: yes, this is a temporary parameter

sci_strategic_plugin/config/parameters.yaml
```diff
@@ -5,11 +5,23 @@ vehicle_decel_limit_multiplier : 0.50
...
+
+# Double: Minimum vehicle reaction time. Units in s
+reaction_time : 5.0
```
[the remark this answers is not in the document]
saina-ramyar: This is a parameter we dont use in carma, but the scheduler needs it. It is the time it takes the vehicle to react to or execute the schedule message. They said this should be a value in the range of a few seconds. I just made it configurable so it is easier to change

sci_strategic_plugin/config/parameters.yaml
```diff
@@ -5,11 +5,23 @@ vehicle_decel_limit_multiplier : 0.50
...
+
+# Double: The distance vehicle drives to be considered out of intersections
+intersection_transit_length : 15.0
```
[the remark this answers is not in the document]
saina-ramyar: changed

sci_strategic_plugin/include/sci_strategic_plugin.h
```diff
@@ -258,37 +268,52 @@ class SCIStrategicPlugin
...
+
+  /**
+   * \brief Determine the turn direction at intersection
+   *
+   * \param lanelets_list List of lanelets crossed around the intersection area
+   *
+   * \return turn direction in format of straight, left, right
+   *
+   */
+  std::string getTurnDirectionatIntersection(std::vector<lanelet::ConstLanelet> lanelets_list);
```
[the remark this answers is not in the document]
saina-ramyar: The turn direction extracted from lanelets is a string, so I used the same type. But I updated it based on your suggestion

sci_strategic_plugin/src/sci_strategic_plugin.cpp
saina-ramyar: since it will only start planning once it receives an empty schedule message with "carma/stop_controlled_intersection" strategy, I didnt check the intersection. But made the update.

sci_strategic_plugin/src/sci_strategic_plugin.cpp
saina-ramyar: I think because of the type of these variables, the difference would always come out positive and activate the logic at the wrong time. But if you have a solution for this issue I can make the change.

sci_strategic_plugin/src/sci_strategic_plugin.cpp
saina-ramyar: As mentioned above, the regular difference is not giving the result I expect. But rename the conditions to be more clear. Basically, we are using the schedule timestamps to know what to do at the intersection, and when to do it.

sci_strategic_plugin/src/sci_strategic_plugin.cpp
saina-ramyar: At this point, the vehicle is already stopped so we just need a valid acceleration value. The value doesnt affect the performance. But made it configurable.

sci_strategic_plugin/src/sci_strategic_plugin.cpp
saina-ramyar: updated

sci_strategic_plugin/src/sci_strategic_plugin.cpp
saina-ramyar: Not needed. It was just an idea I had during the implementation

sci_strategic_plugin/src/sci_strategic_plugin.cpp
saina-ramyar: No case 3 is moved and is planned by the stop and wait plugin since it is only deceleration to stop. I think github is showing the lines incorrectly.

sci_strategic_plugin/src/sci_strategic_plugin.cpp
saina-ramyar: Updated.

arbitrator/config/arbitrator_params.yaml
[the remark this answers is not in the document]
kjrush: More I'm just wondering if there is an industry standard name for this operation, if not SCI is okay, but also a common acronym in the gov't space, so might be a bit confusing. If SCI is the industry standard name that's just fine though, just thought it might be wise to check with the transportation engineers to see if they had any input.

sci_strategic_plugin/config/parameters.yaml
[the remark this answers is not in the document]
kjrush: So this is just for communication with the scheduler? Might be a good idea to document that here then, as is it looks a lot like a control parameter of some kind.

kjrush (approved):

msmcconnell:
@saina-ramyar Is this PR active still? If not lets close it

msmcconnell:
This PR is replaced by #1563

PR #1534, opened 2021-11-11
Set cmv_id in port_drayage_plugin to vehicle_id parameter

JonSmet:
# PR Details
## Description

This PR update port_drayage_plugin to set its cmv_id using the 'vehicle_id' global parameter, which includes the vehicle's license plate information. As part of this update, cmv_id within port_drayage_plugin has been changed from an unsigned long to a string. Similarly, V2XHub expects cmv_id to be a string value.

## How Has This Been Tested?

Built and tested locally with development config; also unit tested.

## Types of changes

## Checklist:

### DIFF
--- a/port_drayage_plugin/config/params.yml
+++ b/port_drayage_plugin/config/params.yml
@@ -1,5 +1,4 @@
 stop_speed_epsilon: 1.0
-cmv_id: "123" # Test value; CARMA Streets expects cmv_id to be an unsigned long (32-bit unsigned integer)
 cargo_id: "" # Test value (Empty string indicates the vehicle is not carrying cargo); CARMA Streets expects cargo_id to be a string
 enable_port_drayage: false # bool: Flag to activate and deactivate port drayage operations. If false, state machine will remain INACTIVE.
 starting_at_staging_area: true # bool: Flag indicating CMV's first destination; 'true' indicates Staging Area Entrance; 'false' indicates Port Entrance.
--- a/port_drayage_plugin/src/port_drayage_plugin.cpp
+++ b/port_drayage_plugin/src/port_drayage_plugin.cpp
@@ -39,7 +39,7 @@ namespace port_drayage_plugin
         bool enable_port_drayage;
         _pnh->param<bool>("enable_port_drayage", enable_port_drayage, false);
         std::string cmv_id;
-        _pnh->getParam("vehicle_id", cmv_id);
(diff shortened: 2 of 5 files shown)


port_drayage_plugin/src/port_drayage_plugin.cpp
```diff
@@ -38,12 +38,8 @@ namespace port_drayage_plugin
...
         bool enable_port_drayage;
         _pnh->param<bool>("enable_port_drayage", enable_port_drayage, false);
-
-        // Read in 'cmv_id' parameter as a string, then convert to an unsigned long before initializing the PortDrayageWorker object
-        std::string cmv_id_string;
-        _pnh->param<std::string>("cmv_id", cmv_id_string, "0");
-        unsigned long cmv_id = std::stoul(cmv_id_string);
-
+        std::string cmv_id;
+        _pnh->getParam("vehicle_id", cmv_id);
```
msmcconnell, line 42: This needs to be "/vehicle_id" in order to be loaded from global scope.

port_drayage_plugin/config/params.yml
```diff
@@ -1,5 +1,4 @@
 stop_speed_epsilon: 1.0
-cmv_id: "123" # Test value; CARMA Streets expects cmv_id to be an unsigned long (32-bit unsigned integer)
```
msmcconnell, line 2: CARMA Streets expects a 32bit unsigned int. That limits the id string to at most 4 characters. All of our vehicle Ids are longer than 4 characters. Either some use case specific encoding needs to be happening or the message updated to support a larger value.

msmcconnell (changes requested):

port_drayage_plugin/src/port_drayage_plugin.cpp
JonSmet: Fixed

port_drayage_plugin/config/params.yml
JonSmet: This comment was from an earlier design decision that required cmv_id to be set to a 32-bit unsigned integer. Later in testing, it was detemrined that the cmv_id should match the vehicle's vehicle_id parameter. V2XHub/CARMA Streets has already been updated to expect and treat the cmv_id value as a string in received MobilityOperation messages for Port Drayage.

msmcconnell (approved):
LGTM

PR #1538, opened 2021-11-29
case_1_2_unit

aswath1:
# PR Details
## Description
As CARMA Platform operating on vehicle approaching a fixed-time traffic signal, I want to follow Trajectory Case 2 specified in CRMTFC-257: Trajectory Plan: Case Determination (UC#2 - All Classes)DONE to plan a trajectory such that the ego vehicle does not have to stop prior to the light turning green so that traffic flow behind the vehicle does not suffer a shockwave.    

This case will assume that the target departure time is soon enough (but not sooner than what is required for CASE 1) based on the vehicle’s present location and speed, so that the vehicle should first accelerate and then decelerate to leave the intersection at the target speed. In this case, the trajectory profile include two sections: acceleration and deceleration (with respect to the order).

please refer to the Trajectory Smoothing (TS) section of the algorithm design for more details.

## Motivation and Context
TSMO CASE 2

## How Has This Been Tested?
Unit Tested

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,7 @@ jobs:
           name: Build CARMA
           command: |
             source ${INIT_ENV}
-            export ROS_PARALLEL_JOBS='-j2 -l2' # Try to reduce memory consumption on build
+            export ROS_PARALLEL_JOBS='-j1 -l1' # Try to reduce memory consumption on build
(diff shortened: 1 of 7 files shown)


MishkaMN:
rerunning autoware with manually pushed clean image.

PR #1551, opened 2021-12-06
PR for review of usf driver

msmcconnell:
# PR Details
## Types of changes

## Checklist:

### DIFF
--- a/mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
+++ b/mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
@@ -1,0 +1,249 @@
+/*
+ * Copyright (C) 2020-2021 LEIDOS.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+//head files required by USF controller driver
+#include <ros/ros.h>
+#include "rosbag_mock_drivers/VehicleCmd.h"
+
+#include <iostream>
+#include <string>
+#include <cstdio>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <pthread.h>
+//head files required by USF controller driver
+
+#include "rosbag_mock_drivers/MockControllerDriver.h"
+
+namespace mock_drivers
+{
+std::vector<DriverType> MockControllerDriver::getDriverTypes()
+{
+  return { DriverType::CONTROLLER };
+}
+
+uint8_t MockControllerDriver::getDriverStatus()
+{
+  return cav_msgs::DriverStatus::OPERATIONAL;
+}
+
+//have a new vehicle command callback function below
+/*void MockControllerDriver::vehicleCmdCallback(const autoware_msgs::VehicleCmd::ConstPtr& msg)
+{
+  robot_enabled_ = true;  // If a command was received set the robot enabled status to true
+}*/
+
+bool MockControllerDriver::enableRoboticSrv(const cav_srvs::SetEnableRobotic::Request& req,
+                                            cav_srvs::SetEnableRobotic::Response& res)
+{
+  if (robot_enabled_ && req.set == cav_srvs::SetEnableRobotic::Request::ENABLE)
+  {
+    robot_active_ = true;
+  }
+  else
+  {
+    robot_active_ = false;
+  }
+
+  return true;
+}
+
+MockControllerDriver::MockControllerDriver(bool dummy)
+{
+  mock_driver_node_ = MockDriverNode(dummy);
+
+  robot_status_ptr_ =
+      boost::make_shared<ROSComms<cav_msgs::RobotEnabled>>(CommTypes::pub, false, 10, robot_status_topic_);
+  
+  ////vehicle command topic is subscribed by the vehiclesender node.
+
+  /*vehicle_cmd_ptr_ = boost::make_shared<ConstPtrRefROSComms<autoware_msgs::VehicleCmd>>(
+      std::bind(&MockControllerDriver::vehicleCmdCallback, this, std::placeholders::_1), CommTypes::sub, false, 10,
+      vehicle_cmd_topic_);*/
+
+  enable_robotic_ptr_ =
+      boost::make_shared<ROSComms<cav_srvs::SetEnableRobotic::Request&, cav_srvs::SetEnableRobotic::Response&>>(
+          std::bind(&MockControllerDriver::enableRoboticSrv, this, std::placeholders::_1, std::placeholders::_2),
+          CommTypes::srv, enable_robotic_srv_);
+}
+
+bool MockControllerDriver::onSpin()
+{
+  cav_msgs::RobotEnabled robot_status;
+  robot_status.robot_active = robot_active_;
+  robot_status.robot_enabled = robot_enabled_;
+
+  mock_driver_node_.publishDataNoHeader<cav_msgs::RobotEnabled>(robot_status_topic_, robot_status);
+
+  return true;
+}
+
+unsigned int MockControllerDriver::getRate()
+{
+  return 20;  // 20 Hz as default spin rate to match expected controller data rate
+}
+
+//USF controller message structure
+struct CommandData {
+  double linear_x;
+  double angular_z;
+
+  void reset();
+};
+
+void CommandData::reset()
+{
+  linear_x      = 0;
+  angular_z     = 0;
+}
+
+static CommandData command_data;
+
+static void vehicleCmdCallback(const autoware_msgs::VehicleCmd& msg)
+{
+  command_data.linear_x = msg.twist_cmd.twist.linear.x;
+  command_data.angular_z = msg.twist_cmd.twist.angular.z;
+}
+
+static void *sendCommand(void *arg)
+{
+  int *client_sockp = static_cast<int*>(arg);
+  int client_sock = *client_sockp;
+  delete client_sockp;
+
+  std::ostringstream oss;
+  oss << command_data.linear_x << ",";
+  oss << command_data.angular_z;
+
+  std::string cmd(oss.str());
+  ssize_t n = write(client_sock, cmd.c_str(), cmd.size());
+  if(n < 0){
+    std::perror("write");
+    return nullptr;
+  }
+
+  if(close(client_sock) == -1){
+    std::perror("close");
+    return nullptr;
+  }
+
+  std::cout << "cmd: " << cmd << ", size: " << cmd.size() << std::endl;
+  return nullptr;
+}
+
+static void* receiverCaller(void *unused)
+{
+  constexpr int listen_port = 10001;
+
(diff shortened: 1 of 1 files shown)


mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
```diff
@@ -0,0 +1,249 @@
...
+    std::perror("bind");
+    goto error;
+  }
+
+  ret = listen(sock, 20);
+  if(ret == -1){
+    std::perror("listen");
+    goto error;
+  }
+  while(true){
```
msmcconnell, line 180: I'm not overly familiar with socket programming in c++, but I'm not seeing any sort of periodic delay in this loop which means it spins as fast as possible. This could potentially eat up a lot of cpu which might have a negative impact on carma's performance. Looping at 120Hz or something like that would still minimize latency on the data output with respect to the rest of the system without locking other processes out as much.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
```diff
@@ -0,0 +1,249 @@
...
+    *client_sock = accept(sock, reinterpret_cast<sockaddr*>(&client), &len);
+    if(*client_sock == -1){
+      std::perror("accept");
+      break;
+    }
+
+    std::cout << "get connect." << std::endl;
+
+    pthread_t th;
+    if(pthread_create(&th, nullptr, sendCommand, static_cast<void*>(client_sock)) != 0){
```
msmcconnell, line 193: Starting a new thread to send this data every time seems a bit excessive and could potentially result in data being sent in an incorrect order (though pretty unlikely).

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
```diff
@@ -0,0 +1,249 @@
...
+      break;
+    }
+
+    if(pthread_detach(th) != 0){
+      std::perror("pthread_detach");
+      break;
+    }
+  }
+
+  error:
```
msmcconnell, line 204: Is this a syntax error?

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
```diff
@@ -0,0 +1,249 @@
...
+  if(pthread_create(&th, nullptr, receiverCaller, nullptr) != 0){
+    std::perror("pthread_create");
+    std::exit(1);
+  }
+
+  if (pthread_detach(th) != 0){
+    std::perror("pthread_detach");
+    std::exit(1);
+  }
+  ros::spin();
```
msmcconnell, line 243: don't call spin() here. This is called from the parent class

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
```diff
@@ -0,0 +1,249 @@
...
+  //USF customized DBW control codes
+
+  return 0;
```
msmcconnell, line 246: Return true. The std::exit lines in this method could also be return false instead of std::exit.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
```diff
@@ -0,0 +1,249 @@
...
+{
+  robot_enabled_ = true;  // If a command was received set the robot enabled status to true
+}*/
+
+bool MockControllerDriver::enableRoboticSrv(const cav_srvs::SetEnableRobotic::Request& req,
+                                            cav_srvs::SetEnableRobotic::Response& res)
+{
+  if (robot_enabled_ && req.set == cav_srvs::SetEnableRobotic::Request::ENABLE)
+  {
+    robot_active_ = true;
```
msmcconnell, line 57: On a real vehicle this is meant to come from feedback from the hardware based on the systems automation control status. From a logical level though this will work.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+  robot_status_ptr_ =
+      boost::make_shared<ROSComms<cav_msgs::RobotEnabled>>(CommTypes::pub, false, 10, robot_status_topic_);
+  
+  ////vehicle command topic is subscribed by the vehiclesender node.
+
+  /*vehicle_cmd_ptr_ = boost::make_shared<ConstPtrRefROSComms<autoware_msgs::VehicleCmd>>(
+      std::bind(&MockControllerDriver::vehicleCmdCallback, this, std::placeholders::_1), CommTypes::sub, false, 10,
+      vehicle_cmd_topic_);*/
+
```
msmcconnell, line 79: There needs to also be a "vehicle/twist" topic (geometry_msgs/TwistStamped) published which exposes the current vehicle speed.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
kjrush: It reads to me as a goto label for error handling, this section seems very C-style to me as this might normally be handled by destructors in C++.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
```diff
@@ -0,0 +1,249 @@
...
+  //mock_driver_node_.addSub(vehicle_cmd_ptr_); //vehicle command topic will be subscribed later.
+  mock_driver_node_.addSrv(enable_robotic_ptr_);
+  MockControllerDriver::robot_enabled_ = true; //always set robot_enabled to be true
+  
+  
+  //USF customized DBW control codes
+  int argc;
+  char **argv;
+
+  ros::init(argc, argv, "vehicle_sender"); //define a node to receive control command and send it to DBW using TCP/IP
```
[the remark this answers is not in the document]
kjrush: I think ultimately we'd want it to be it's own standalone node right? The mock_drivers aspect here is just a template since they wanted a node that had a lot of the topics already setup. Ideally this'd just be another driver in our system similar to the ssc_interface_wrapper.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+  ret = listen(sock, 20);
+  if(ret == -1){
+    std::perror("listen");
+    goto error;
+  }
+  while(true){
+    std::cout << "Waiting access..." << std::endl;
+
+    int *client_sock = new int();
```
kjrush, line 183: Is there a reason this is initialized as an int* on the stack? It looks like it could be an int just fine without needing to be a pointer, especially considering i don't see it get `delete`'d anywhere.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+static CommandData command_data;
+
+static void vehicleCmdCallback(const autoware_msgs::VehicleCmd& msg)
+{
+  command_data.linear_x = msg.twist_cmd.twist.linear.x;
+  command_data.angular_z = msg.twist_cmd.twist.angular.z;
+}
+
+static void *sendCommand(void *arg)
```
kjrush, line 124: Is there a reason this accepts a `void *`, then forces a cast immediately to an `int*`, then dereferences that into an int? Would it not be simpler to simply accept the int directly as a parameter?

Also, what is the purpose of the return type here? It appears this function always returns nullptr, even on success? You could simply omit the need to return anything by changing the return type from `void *` to `void`. Or, if an error status should be returned you could return a bool for a simple pass/fail or an int if differentiation between the error codes matters.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
```diff
@@ -0,0 +1,249 @@
...
+  if(close(client_sock) == -1){
+    std::perror("close");
+    return nullptr;
+  }
+
+  std::cout << "cmd: " << cmd << ", size: " << cmd.size() << std::endl;
+  return nullptr;
+}
+
+static void* receiverCaller(void *unused)
```
kjrush, line 150: Same as above, if it's unused there's no need to accept any input at all just define it as `static void receiverCaller();`. Similar comment as above re: return value as well.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
kjrush: We'll likely need to map this and a few other data streams in from other devices on their vehicles.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
```diff
@@ -0,0 +1,249 @@
...
+  robot_status.robot_enabled = robot_enabled_;
+
+  mock_driver_node_.publishDataNoHeader<cav_msgs::RobotEnabled>(robot_status_topic_, robot_status);
+
+  return true;
+}
+
+unsigned int MockControllerDriver::getRate()
+{
+  return 20;  // 20 Hz as default spin rate to match expected controller data rate
```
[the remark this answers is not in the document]
Tom-ABCD-sxw: Revised.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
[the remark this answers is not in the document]
Tom-ABCD-sxw: Has been removed.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
Tom-ABCD-sxw: Has been removed.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
Tom-ABCD-sxw: Revised.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
Tom-ABCD-sxw: Shall I revise anything here?

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
```diff
@@ -0,0 +1,249 @@
...
+  linear_x      = 0;
+  angular_z     = 0;
+}
+
+static CommandData command_data;
+
+static void vehicleCmdCallback(const autoware_msgs::VehicleCmd& msg)
+{
+  command_data.linear_x = msg.twist_cmd.twist.linear.x;
+  command_data.angular_z = msg.twist_cmd.twist.angular.z;
```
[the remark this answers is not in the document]
Tom-ABCD-sxw: Has been revised to:
command_data.linear_x = msg.ctrl_cmd.linear_velocity;
command_data.angular_z = msg.ctrl_cmd.steering_angle;

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
Tom-ABCD-sxw: Let's keep it to see what will happen.

mock_drivers/rosbag_mock_drivers/src/USFControllerDriver.cpp
Tom-ABCD-sxw: Yes, it is a goto label. Once an error happens, the socket will be closed.

error:
  close(sock);
  return nullptr;

msmcconnell:
Closing this for now as it has served its purpose.

PR #1560, opened 2021-12-14
The new commit to feature cut-in-front. Code difference is updated.

xuhan417:
# PR Details
## Description

This is UCLA team's commit of the IHP2 platoon plugin which includes the function of cut-in front join.
## Related Issue

There is no related issues.
## Motivation and Context

This is the development for IHP2 project.
It adds necessary functions to cut-in join from front.
## How Has This Been Tested?

- Add unit test to test additional state transitions.

- Compile and ran on UCLA's vehicle platform (single vehicle only).

## Types of changes

## Checklist:

### DIFF
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -99,8 +99,6 @@ namespace platoon_strategic_ihp
     {
         // Static ID is permanent ID for each vehicle
         std::string staticId;
-        // Current BSM Id for each CAV
-        std::string bsmId;
         // Vehicle real time command speed in m/s
         double commandSpeed;
         // Actual vehicle speed in m/s
@@ -305,29 +303,9 @@ namespace platoon_strategic_ihp
         std::string previousFunctionalDynamicLeaderID_ = "";
         int previousFunctionalDynamicLeaderIndex_ = -1;
 
-        // Thresholds to determine violation.
-        double maxAllowableHeadaway_ = 4.0; // s
-        double minAllowableHeadaway_ = 1.6; // s
-
-        // Thresholds to determine gap stable (i.e., Determines whether to switch dynamic leader for APF).
-        double headawayStableUpperBond_ = 3.9; // s
-        double headawayStableLowerBond_ = 1.7 ; // s
-
         double vehicleLength_ = 5.0;  // m
         double gapWithPred_ = 0.0; // time headway with predecessor, in s.
         double downtrack_progress_ = 0; // m
-
-        // ---------------------- UCLA: parameters for IHP platoon trajectory regulation ----------------
-        /**
-        * \brief Parameter sets for IHP platoon trajectory regulation algorithm. 
-        * Please refer to the updated design doc for detailed parameter description.
-        */
-        double ss_theta = 4.0; // Stanstill determining threshold, in m/s.
-        double standstill = 2.0; // Stanstill reaction time adjuster, in s.
-        double inter_tau = 1.5; // Inter-platoon time gap, in s.
-        double intra_tau = 0.6; // Intra-platoon time gao, in s.
-        double gap_weight = 0.9; // Weighted ratio for time-gap based calculation, unitless.
-        //--------------------------------------------------------------------------------
 
         std::string algorithmType_ = "APF_ALGORITHM";
 
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -300,7 +300,9 @@ namespace platoon_strategic_ihp
         // UCLA: add params for frontal join
         std::string JOIN_FROM_FRONT_PARAMS = "SIZE:%1%,SPEED:%2%,DTD:%3%";
         std::string  MOBILITY_STRATEGY = "Carma/Platooning";
-        
+
+        double minGap_ = 22.0;                                  // m
+        double maxGap_ = 32.0;                                  // m
         std::string previousFunctionalDynamicLeaderID_ = "";
         int previousFunctionalDynamicLeaderIndex_ = -1;
 
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -90,7 +90,7 @@ namespace platoon_strategic_ihp
         LEADWITHOPERATION,          // 7;
         //UCLA: CUT-IN JOIN STATE
         PREPARETOJOIN               // 8;
-    };z
+    };
 
     /**
     * \brief Platoon States 
@@ -241,7 +241,7 @@ namespace platoon_strategic_ihp
         /**
         * \brief Returns current crosstrack distance, in m.
         */
-        double getCurrentCrosstrackDistace() const;
+        double getCurrentCrosstrackDistance() const;
 
         /**
         * \brief Returns current host static ID as a string.
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -35,7 +35,7 @@
 #include <boost/uuid/uuid_generators.hpp>
 #include <boost/uuid/uuid_io.hpp>
 #include <autoware_msgs/ControlCommandStamped.h>
-
+#include "platoon_config_ihp.h"
 
 
 
@@ -137,7 +137,6 @@ namespace platoon_strategic_ihp
         * 
         * \param senderId static id of the broadcasting vehicle
         * \param platoonId platoon id
-        * \param senderBsmId bsm id of the broadcasting vehicle
         * \param params strategy parameters
         * \param Dtd downtrack distance
         */
@@ -254,15 +253,22 @@ namespace platoon_strategic_ihp
         double getIHPDesPosFollower(double dt);
 
         /**
-        \brief UCLA: Return joiner's desired position
-        to cut into the platoon 
-        */
+         * \brief UCLA: Return joiner's desired position in terms of platoon index to cut into the platoon. 
+         * 
+         * \params joinerDtD: The current downtrack distance (with regards to host vehicle) of the joiner vehicle.
+         * 
+         * \return: cut-in index: The index of the gap-leading vehicle within the platoon. If front join, return -1.
+         */
         int getClosestIndex(double joinerDtD);
 
         /**
-        \brief UCLA: Return the desired gap size
-        for cut-in join 
-        */
+         * \brief UCLA: Return the desired gap size for cut-in join, in m.
+         * 
+         * \params gap_leading_index: The platoon index of the  gap-leading vehicle. 
+         *         joinerDtD: The current downtrack distance (with regards to host vehicle) of the joiner vehicle.
+         * 
+         * \return: cut-in gap: The desired gap size for cut-in join, in m.
+         */
         double getCutInGap(int gap_leading_index, double joinerDtD);
 
         // Member variables
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -298,16 +298,20 @@ namespace platoon_strategic_ihp
         std::string JOIN_FROM_FRONT_PARAMS = "SIZE:%1%,SPEED:%2%,DTD:%3%";
         std::string  MOBILITY_STRATEGY = "Carma/Platooning";
 
-        double minGap_ = 22.0; // m
-        double maxGap_ = 32.0; // m
+        double minGap_ = 22.0;                                  // m
+        double maxGap_ = 32.0;                                  // m
         std::string previousFunctionalDynamicLeaderID_ = "";
         int previousFunctionalDynamicLeaderIndex_ = -1;
 
-        double vehicleLength_ = 5.0;  // m
-        double gapWithPred_ = 0.0; // time headway with predecessor, in s.
-        double downtrack_progress_ = 0; // m
+        // note: APF related parameters are moved to config.h.
 
-        std::string algorithmType_ = "APF_ALGORITHM";
+        double vehicleLength_ = 5.0;                            // the length of the vehicle, in m.
+        double gapWithPred_ = 0.0;                              // time headway with predecessor, in s.
+        double downtrack_progress_ = 0;                         // current downtrack distance, in m.
+
+        // Note: Parameters for IHP platoon trajectory regulation are moved to config.h. 
+
+        std::string algorithmType_ = "APF_ALGORITHM";           // a string that defines the current algorithm to determine the dynamic leader.
 
         /**
         * \brief Check the gap with the predecessor vehicle. 
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -99,13 +99,13 @@ namespace platoon_strategic_ihp
     {
         // Static ID is permanent ID for each vehicle
         std::string staticId;
-        // Current BSM Id for each CAV
+        // Vehicle real time command speed in m/s.
         double commandSpeed;
-        // Actual vehicle speed in m/s
+        // Actual vehicle speed in m/s.
         double vehicleSpeed;
-        // Vehicle current down track distance on the current route in m
+        // Vehicle current down track distance on the current route in m.
         double vehiclePosition;
-        // The local time stamp when the host vehicle update any informations of this member
+        // The local time stamp when the host vehicle update any informations of this member.
         long   timestamp;
         PlatoonMember(): staticId(""), commandSpeed(0.0), vehicleSpeed(0.0), vehiclePosition(0.0), timestamp(0) {} 
         PlatoonMember(std::string staticId, double commandSpeed, double vehicleSpeed, double vehiclePosition, long timestamp): staticId(staticId),
@@ -315,19 +315,6 @@ namespace platoon_strategic_ihp
         // Note: Parameters for IHP platoon trajectory regulation are moved to config.h. 
 
         std::string algorithmType_ = "APF_ALGORITHM";           // a string that defines the current algorithm to determine the dynamic leader.
-
-        /**
-        * \brief Parameter sets for IHP platoon trajectory regulation algorithm. 
-        * Please refer to the updated design doc for detailed parameter description.
-        */
-        double ss_theta = 4.0; // Stanstill determining threshold, in m/s.
-        double standstill = 2.0; // Stanstill reaction time adjuster, in s.
-        double inter_tau = 1.5; // Inter-platoon time gap, in s.
-        double intra_tau = 0.6; // Intra-platoon time gao, in s.
-        double gap_weight = 0.9; // Weighted ratio for time-gap based calculation, unitless.
-        //--------------------------------------------------------------------------------
-
-        std::string algorithmType_ = "APF_ALGORITHM";
 
         /**
         * \brief Check the gap with the predecessor vehicle. 
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -90,10 +90,11 @@ namespace platoon_strategic_ihp
         LEADWITHOPERATION,          // 7;
         //UCLA: CUT-IN JOIN STATE
         PREPARETOJOIN               // 8;
-    };
+    };z
 
     /**
-    * \brief Platoon States
+    * \brief Platoon States 
+    * Note: Include cross track as one of the class variables.
     */
     struct PlatoonMember
     {
@@ -103,13 +104,15 @@ namespace platoon_strategic_ihp
         double commandSpeed;
         // Actual vehicle speed in m/s.
         double vehicleSpeed;
-        // Vehicle current down track distance on the current route in m.
+        // Vehicle current downtrack distance on the current route in m.
         double vehiclePosition;
+        // Vehicle current crosstrack distance on the current route in m.
+        double vehicleCrossTrack;
         // The local time stamp when the host vehicle update any informations of this member.
         long   timestamp;
-        PlatoonMember(): staticId(""), commandSpeed(0.0), vehicleSpeed(0.0), vehiclePosition(0.0), timestamp(0) {} 
-        PlatoonMember(std::string staticId, double commandSpeed, double vehicleSpeed, double vehiclePosition, long timestamp): staticId(staticId),
-            commandSpeed(commandSpeed), vehicleSpeed(vehicleSpeed), vehiclePosition(vehiclePosition), timestamp(timestamp) {}
+        PlatoonMember(): staticId(""), commandSpeed(0.0), vehicleSpeed(0.0), vehiclePosition(0.0), vehicleCrossTrack(0.0), timestamp(0) {} 
+        PlatoonMember(std::string staticId, double commandSpeed, double vehicleSpeed, double vehiclePosition, double vehicleCrossTrack, long timestamp): staticId(staticId),
+            commandSpeed(commandSpeed), vehicleSpeed(vehicleSpeed), vehiclePosition(vehiclePosition), vehicleCrossTrack(vehicleCrossTrack), timestamp(timestamp) {}
     };
         
 
@@ -129,8 +132,11 @@ namespace platoon_strategic_ihp
         // Current vehicle pose in map
         geometry_msgs::PoseStamped pose_msg_;
 
-        // Current vehicle downtrack
+        // Current vehicle downtrack, in m.
         double current_downtrack_distance_ = 0;
+
+        // Current vehcile crosstrack, in m.
+        double current_crosstrack_distance_ = 0;
 
         /**
         * \brief Update platoon members information
@@ -140,7 +146,7 @@ namespace platoon_strategic_ihp
         * \param params strategy parameters
         * \param Dtd downtrack distance
         */
-        void memberUpdates(const std::string& senderId, const std::string& platoonId, const std::string& params, const double& DtD);
+        void memberUpdates(const std::string& senderId, const std::string& platoonId, const std::string& params, const double& DtD, const double& CtD);
 
         /**
          * \brief Given any valid platooning mobility STATUS operation parameters and sender staticId,
@@ -152,7 +158,7 @@ namespace platoon_strategic_ihp
          * \param platoonId sender platoon id
          * \param params strategy params from STATUS message in the format of "CMDSPEED:xx,DOWNTRACK:xx,SPEED:xx"
          **/
-        void updatesOrAddMemberInfo(std::string senderId, double cmdSpeed, double dtDistance, double curSpeed);
+        void updatesOrAddMemberInfo(std::string senderId, double cmdSpeed, double dtDistance, double ctDistance, double curSpeed);
         
         /**
         * \brief Returns total size of the platoon , in number of vehicles.
@@ -233,6 +239,11 @@ namespace platoon_strategic_ihp
         double getCurrentDowntrackDistance() const;
 
         /**
+        * \brief Returns current crosstrack distance, in m.
+        */
+        double getCurrentCrosstrackDistace() const;
+
+        /**
         * \brief Returns current host static ID as a string.
         */
         std::string getHostStaticID() const;
@@ -263,6 +274,7 @@ namespace platoon_strategic_ihp
 
         /**
          * \brief UCLA: Return the desired gap size for cut-in join, in m.
+         *        Note: The origin of the vehicle (for downtrack distance calculation) is located at the rear axle. 
          * 
          * \params gap_leading_index: The platoon index of the  gap-leading vehicle. 
          *         joinerDtD: The current downtrack distance (with regards to host vehicle) of the joiner vehicle.
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -290,6 +290,9 @@ namespace platoon_strategic_ihp
         
     private:
 
+        // local copy of configuration file
+        PlatoonPluginConfig config_;
+
         std::string targetPlatoonId;
         std::string OPERATION_INFO_TYPE = "INFO";
         std::string OPERATION_STATUS_TYPE = "STATUS";
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -35,7 +35,6 @@
 #include <boost/uuid/uuid_generators.hpp>
 #include <boost/uuid/uuid_io.hpp>
 #include <autoware_msgs/ControlCommandStamped.h>
-#include "platoon_config_ihp.h"
 
 
 
@@ -71,19 +70,26 @@ namespace platoon_strategic_ihp
     };
 
     /**
-    * \brief Platoon States. UCLA: Added two additional states 
+    * \brief Platoon States. UCLA: Added additional states 
     * (i.e., LEADERABORTING and CANDIDATELEADER) for same-lane front join.
+    * (i.e., LEADWITHOPERATION and PREPARETOJOIN) for cut-in front join.
     *  
     */
     enum PlatoonState
     {
-        STANDBY,
-        LEADERWAITING,
-        LEADER,
-        CANDIDATEFOLLOWER,
-        FOLLOWER,
-        LEADERABORTING,
-        CANDIDATELEADER
+        STANDBY,                    // 0;
+        LEADERWAITING,              // 1;
+        LEADER,                     // 2;
+        CANDIDATEFOLLOWER,          // 3;
+        FOLLOWER,                   // 4;
+        //UCLA: FRONTAL JOIN STATE
+        LEADERABORTING,             // 5;
+        //UCLA: FRONTAL JOIN STATE
+        CANDIDATELEADER,            // 6;
+        //UCLA: CUT-IN JOIN STATE
+        LEADWITHOPERATION,          // 7;
+        //UCLA: CUT-IN JOIN STATE
+        PREPARETOJOIN               // 8;
     };
 
     /**
@@ -93,6 +99,8 @@ namespace platoon_strategic_ihp
     {
         // Static ID is permanent ID for each vehicle
         std::string staticId;
+        // Current BSM Id for each CAV
+        std::string bsmId;
         // Vehicle real time command speed in m/s
         double commandSpeed;
         // Actual vehicle speed in m/s
@@ -131,6 +139,7 @@ namespace platoon_strategic_ihp
         * 
         * \param senderId static id of the broadcasting vehicle
         * \param platoonId platoon id
+        * \param senderBsmId bsm id of the broadcasting vehicle
         * \param params strategy parameters
         * \param Dtd downtrack distance
         */
@@ -158,6 +167,7 @@ namespace platoon_strategic_ihp
         * \brief Returns dynamic leader of the host vehicle.
         */
         PlatoonMember getDynamicLeader();
+
 
         /**
          * \brief This is the implementation of all predecessor following (APF) algorithm for leader
@@ -243,6 +253,18 @@ namespace platoon_strategic_ihp
         */
         double getIHPDesPosFollower(double dt);
 
+        /**
+        \brief UCLA: Return joiner's desired position
+        to cut into the platoon 
+        */
+        int getClosestIndex(double joinerDtD);
+
+        /**
+        \brief UCLA: Return the desired gap size
+        for cut-in join 
+        */
+        double getCutInGap(int gap_leading_index, double joinerDtD);
+
         // Member variables
         int platoonSize = 2;
         std::string leaderID = "default_leader_id";
@@ -262,11 +284,13 @@ namespace platoon_strategic_ihp
 
         // host vehicle's static ID 
         std::string HostMobilityId = "default_host_id";
-    
-    private:
 
-        // local copy of configuration file
-        PlatoonPluginConfig config_;
+        // UCLA: add indicator of creating gap
+        bool isCreateGap = false;
+        // UCLA: add indicator of lane change 
+        bool safeToLaneChange = false;
+        
+    private:
 
         std::string targetPlatoonId;
         std::string OPERATION_INFO_TYPE = "INFO";
@@ -282,12 +306,12 @@ namespace platoon_strategic_ihp
         int previousFunctionalDynamicLeaderIndex_ = -1;
 
         // Thresholds to determine violation.
-        // double maxAllowableHeadaway_ = 4.0; // s
-        // double minAllowableHeadaway_ = 1.6; // s
+        double maxAllowableHeadaway_ = 4.0; // s
+        double minAllowableHeadaway_ = 1.6; // s
 
         // Thresholds to determine gap stable (i.e., Determines whether to switch dynamic leader for APF).
-        // double headawayStableUpperBond_ = 3.9; // s
-        // double headawayStableLowerBond_ = 1.7 ; // s
+        double headawayStableUpperBond_ = 3.9; // s
+        double headawayStableLowerBond_ = 1.7 ; // s
 
         double vehicleLength_ = 5.0;  // m
         double gapWithPred_ = 0.0; // time headway with predecessor, in s.
@@ -298,11 +322,11 @@ namespace platoon_strategic_ihp
         * \brief Parameter sets for IHP platoon trajectory regulation algorithm. 
         * Please refer to the updated design doc for detailed parameter description.
         */
-        // double ss_theta = 4.0; // Stanstill determining threshold, in m/s.
-        // double standstill = 2.0; // Stanstill reaction time adjuster, in s.
-        // double inter_tau = 1.5; // Inter-platoon time gap, in s.
-        // double intra_tau = 0.6; // Intra-platoon time gao, in s.
-        // double gap_weight = 0.9; // Weighted ratio for time-gap based calculation, unitless.
+        double ss_theta = 4.0; // Stanstill determining threshold, in m/s.
+        double standstill = 2.0; // Stanstill reaction time adjuster, in s.
+        double inter_tau = 1.5; // Inter-platoon time gap, in s.
+        double intra_tau = 0.6; // Intra-platoon time gao, in s.
+        double gap_weight = 0.9; // Weighted ratio for time-gap based calculation, unitless.
         //--------------------------------------------------------------------------------
 
         std::string algorithmType_ = "APF_ALGORITHM";
@@ -315,7 +339,7 @@ namespace platoon_strategic_ihp
         * \return (bool) if the predecessor is to close.
         */
         bool insufficientGapWithPredecessor(double distanceToPredVehicle);
-        
+                
         /**
         * \brief Calculate the time headaway of each platoon member and save as a vector.
         * 
@@ -325,7 +349,7 @@ namespace platoon_strategic_ihp
         * \return A vector containing the time headaway of all platoon members.
         */
         std::vector<double> calculateTimeHeadway(std::vector<double> downtrackDistance, std::vector<double> speed) const;
-
+        
         /**
         * \brief Determine the proper vehicle to follow based the time headway of each member. 
         * Note that the host will always choose the closest violator (i.e. time headaway too small or too large) to follow.
@@ -343,7 +367,7 @@ namespace platoon_strategic_ihp
         * 
         * \return An index indicating the closest violating vehicle. If no violator, return -1.
         */
-        int findLowerBoundaryViolationClosestToTheHostVehicle(std::vector<double> timeHeadways) const;
+       int findLowerBoundaryViolationClosestToTheHostVehicle(std::vector<double> timeHeadways) const;
 
         /**
         * \brief Find the closest vehicle to the host vehicle that violates the (time headaway) maximum spacing condition. 
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -99,7 +99,7 @@ namespace platoon_strategic_ihp
     {
         // Static ID is permanent ID for each vehicle
         std::string staticId;
-        // Vehicle real time command speed in m/s
+        // Current BSM Id for each CAV
         double commandSpeed;
         // Actual vehicle speed in m/s
         double vehicleSpeed;
@@ -315,6 +315,19 @@ namespace platoon_strategic_ihp
         // Note: Parameters for IHP platoon trajectory regulation are moved to config.h. 
 
         std::string algorithmType_ = "APF_ALGORITHM";           // a string that defines the current algorithm to determine the dynamic leader.
+
+        /**
+        * \brief Parameter sets for IHP platoon trajectory regulation algorithm. 
+        * Please refer to the updated design doc for detailed parameter description.
+        */
+        double ss_theta = 4.0; // Stanstill determining threshold, in m/s.
+        double standstill = 2.0; // Stanstill reaction time adjuster, in s.
+        double inter_tau = 1.5; // Inter-platoon time gap, in s.
+        double intra_tau = 0.6; // Intra-platoon time gao, in s.
+        double gap_weight = 0.9; // Weighted ratio for time-gap based calculation, unitless.
+        //--------------------------------------------------------------------------------
+
+        std::string algorithmType_ = "APF_ALGORITHM";
 
         /**
         * \brief Check the gap with the predecessor vehicle. 
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -300,9 +300,7 @@ namespace platoon_strategic_ihp
         // UCLA: add params for frontal join
         std::string JOIN_FROM_FRONT_PARAMS = "SIZE:%1%,SPEED:%2%,DTD:%3%";
         std::string  MOBILITY_STRATEGY = "Carma/Platooning";
-
-        double minGap_ = 22.0;                                  // m
-        double maxGap_ = 32.0;                                  // m
+        
         std::string previousFunctionalDynamicLeaderID_ = "";
         int previousFunctionalDynamicLeaderIndex_ = -1;
 
--- a/platooning_strategic_IHP/include/platoon_strategic_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_strategic_ihp.h
@@ -785,8 +785,12 @@ namespace platoon_strategic_ihp
             // Plugin discovery message
             cav_msgs::Plugin plugin_discovery_msg_;
 
+            // Add condition for leader aborting state, to prevent sending repeated requests 
+            bool isFirstLeaderAbortRequest_ = true;
+
+            // Internal FSM parameters 
             int infoMessageInterval_ = 200;             // ms
-            long prevHeartBeatTime_ = 0.0;               // ms
+            long prevHeartBeatTime_ = 0.0;              // ms
             int statusMessageInterval_ = 100;           // ms
             int NEGOTIATION_TIMEOUT = 5000;             // ms
             int LANE_CHANGE_TIMEOUT = 300000;           // ms (5 min)
@@ -795,7 +799,7 @@ namespace platoon_strategic_ihp
             double waitingStateTimeout = 25.0;          // s
             
             // Threshold for determining if vehicle is stopped; non-zero allows for measurement noise
-            const double STOPPED_SPEED = 0.05; // m/s
+            const double STOPPED_SPEED = 0.05;          // m/s
 
             // UCLA: add member variable for state prepare to join
             int target_join_index_;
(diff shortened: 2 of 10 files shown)


platooning_strategic_IHP/include/platoon_strategic_plugin_node_ihp.h
```diff
@@ -77,10 +77,10 @@ class PlatoonStrategicIHPPluginNode
     pnh.param<double>("desiredJoinGap", config.desiredJoinGap, config.desiredJoinGap);
     pnh.param<double>("waitingStateTimeout", config.waitingStateTimeout, config.waitingStateTimeout);
     pnh.param<double>("cmdSpeedMaxAdjustment", config.cmdSpeedMaxAdjustment, config.cmdSpeedMaxAdjustment);
-    pnh.param<double>("minAllowableHeadaway", config.minAllowableHeadaway, config.minAllowableHeadaway);
-    pnh.param<double>("maxAllowableHeadaway", config.maxAllowableHeadaway, config.maxAllowableHeadaway);
-    pnh.param<double>("headawayStableLowerBond", config.headawayStableLowerBond, config.headawayStableLowerBond);
-    pnh.param<double>("headawayStableUpperBond", config.headawayStableUpperBond, config.headawayStableUpperBond);
+    pnh.param<double>("lowerBoundary", config.lowerBoundary, config.lowerBoundary);
```
TonysCousin, line 80: It looks like all the changes here are obsolete?  The names don't match anything in the platoon_config_ihp.h.  Not sure how it compiles.  In any case, names like "lowerBoundary" are too generic - there could be many physical quantities that have a lower boundary, so something more descriptive is needed, especially for a parameter file that needs to stand alone.

platooning_strategic_IHP/include/platoon_manager_ihp.h
```diff
@@ -131,6 +137,7 @@ namespace platoon_strategic_ihp
         * 
         * \param senderId static id of the broadcasting vehicle
         * \param platoonId platoon id
+        * \param senderBsmId bsm id of the broadcasting vehicle
```
TonysCousin, line 140: Description of BSM ID has been added, but I don't see an actual parameter that this refers to.  We don't really want this, do we?

platooning_strategic_IHP/include/platoon_manager_ihp.h
```diff
@@ -245,6 +253,18 @@ namespace platoon_strategic_ihp
         */
         double getIHPDesPosFollower(double dt);
 
+        /**
+        \brief UCLA: Return joiner's desired position
```
TonysCousin, line 257: Need description of the parameter, incl units of measure.
Also, for return value, need more clarity on what the integer means.  It seems this would be the position # that it will cut in front of.  Are the numbers 1-indexed or 0-indexed?

platooning_strategic_IHP/include/platoon_manager_ihp.h
```diff
@@ -245,6 +253,18 @@ namespace platoon_strategic_ihp
...
+        to cut into the platoon 
+        */
+        int getClosestIndex(double joinerDtD);
+
+        /**
+        \brief UCLA: Return the desired gap size
```
TonysCousin, line 263: Units of measure?
Need descriptions of both parameters.  don't assume I know what "leading" means :)

xuhan417:
My bad. I compiled the old code. :(  I will fix it and Re-commit tonight.
Sorry for the confusion. I will take all your blame.

XH



John Stark ***@***.***> 于 2021年12月14日周二 下午6:52写道：

platooning_strategic_IHP/src/platoon_manager_ihp.cpp
```diff
@@ -361,13 +361,13 @@ namespace platoon_strategic_ihp
     bool PlatoonManager::insufficientGapWithPredecessor(double distanceToPredVehicle) {
         
         // For normal operation, gap > minGap is necessary. 
-        bool frontGapIsTooSmall = distanceToPredVehicle < minGap_; 
+        bool frontGapIsTooSmall = distanceToPredVehicle < config_.minGap;
```
TonysCousin, line 364: It feels like a silly question, but I don't see where the config_ member variable is getting populated.  It is declared in the .h file, but there is an empty constructor here, so it seems something is missing.

platooning_strategic_IHP/src/platoon_manager_ihp.cpp
```diff
@@ -631,4 +631,63 @@ namespace platoon_strategic_ihp
         // ---> 4.4 return IHP calculated desired speed
         return pos_des;
     }
+
+    // UCLA: find the clsest index to the cut-in joining vehicle.
+    // Note: The index pointing to the gap-leading vehicle's index.
```
TonysCousin, line 636: This Note is not clear at all.  It has no verb.  It seems to say that an index is pointing to an index.

platooning_strategic_IHP/src/platoon_manager_ihp.cpp
```diff
@@ -631,4 +631,63 @@ namespace platoon_strategic_ihp
...
+        */
+       // cut-in from front 
+       if (gap_leading_index == -1)
+       {
+            double gap_rear_dtd = platoon[0].vehiclePosition;
+            double gap_size = joinerDtD - gap_rear_dtd;
+            return gap_size;
+       }
+       // cut-in from behind 
+       else if (gap_leading_index == platoon.size() + 1)
```
TonysCousin, line 677: This needs to be tested against platoon.size() - 1 since that will be the index of the final vehicle in the platoon (it's where the loop in getClosestIndex() stops).

platooning_strategic_IHP/src/platoon_manager_ihp.cpp
```diff
@@ -631,4 +631,63 @@ namespace platoon_strategic_ihp
...
+                {
+                    min_diff = current_member_dtd;
+                    cut_in_index = i;
+                }
+        }
+        
+        return cut_in_index;
+    }
+
+    // UCLA: find the cut-in join target gap size in downtrack distance (m).
```
TonysCousin, line 663: General approach to this method sometimes considers host vehicle length and sometimes not.  What assumption is being made about the origin of the vehicle coordinate frame?  That is, to what point on the vehicle is DTD measured?  This assumption should be stated in a comment.  Elsewhere in Carma, the vehicle's origin is at the rear axle, and I imagine that applies here also.

Use of vehicle length is appropriate in the middle cut-in case.  However, it seems it should apply to at least one of the other two cases as well.  For example, if the origin is at front bumper, then the front cut-in case should subtract vehicle length from the DTD diff.  Unfortunately, having an origin somewhere in the middle of the vehicle makes it difficult to do this math.  But Maybe it could be ignored for the front join case (origin near the rear), but it should be subtracted in the rear join case.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
```diff
@@ -149,6 +150,21 @@ namespace platoon_strategic_ihp
             */
             cav_msgs::Maneuver composeManeuverMessage(double current_dist, double end_dist, double current_speed, double target_speed, int lane_id, ros::Time& current_time);
 
+            /**
+            * \brief Find start(current) and target(end) lanelet index from path to generate lane change maneuver message.
+            * 
+            * \param current_dist current downtrack distance
```
TonysCousin, line 156: Please add units of measure to all these param descriptions.  Speed, especially, can be uncertain since Americans tend to talk in mph.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
```diff
@@ -149,6 +150,21 @@ namespace platoon_strategic_ihp
...
+            * \param end_dist ending downtrack distance
+            * \param current_speed current speed
+            * \param target_speed target speed
+            * \param starting_lane_id current lanelet id which serves as the starting lanlet id
```
TonysCousin, line 160: Using current distance and current lanelet ID implies that we are constraining ourselves to planning something that must happen immediately.  Is there any chance that we may want to plan a maneuver that starts sometime in the future, in which case the host vehicle will be on a different lanelet than the current one?  Strategic plugins will be called on to make plans several seconds in the future, so it needs to handle such a situation.  I haven't studied the code enough to know if this particular method is affected, but I guess it would be.  Even if it is intended to be for "immediate" use only, consider the case where this method is called just before the vehicle rolls off the end of the current lanelet, and by the time processing time is complete the wheels are on a new lanelet.  Therefore, I think that use of "current" in these param names is wrong.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
```diff
@@ -252,7 +274,22 @@ namespace platoon_strategic_ihp
             */
             void run_candidate_leader();
 
-        
+            // -------------- UCLA: add two states for cut-in join ------------------
```
TonysCousin, line 277: Above, on lines 241 & 264, member variables pm_ and pose_ecef_point_ are declared as public.  As a rule, these should be in the private section.  If they really need to be public, then there should be a comment explaining why.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
```diff
@@ -252,7 +274,22 @@ namespace platoon_strategic_ihp
...
+            
+            /**
+            * \brief UCLA Run prepare to join State
+            */
+            void run_prepare_to_join();
+
+            /**
+             * \brief UCLA find platoon leader bsmID
+             */
+            void front_bsm_cb(const cav_msgs::BSMConstPtr& msg);
```
TonysCousin, line 291: Didn't we decide to ignore BSM IDs?  Why is this being added?

platooning_strategic_IHP/include/platoon_strategic_ihp.h
```diff
@@ -267,7 +304,7 @@ namespace platoon_strategic_ihp
             std::shared_ptr<carma_wm::WMListener> wml_;
             carma_wm::WorldModelConstPtr wm_;
 
-            // local copy of configuration file
+            // local copy of pose
```
TonysCousin, line 307: I don't understand this change.  It seems the old comment was quite accurate.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
```diff
@@ -497,7 +534,7 @@ namespace platoon_strategic_ihp
             */
             lanelet::BasicPoint2d ecef_to_map_point(cav_msgs::LocationECEF ecef_point);
 
-            // -------------- UCLA: front join function headers -----------------------
+            // -------------- UCLA implemented functions -----------------------
```
TonysCousin, line 537: Nice organization!

platooning_strategic_IHP/include/platoon_strategic_ihp.h
```diff
@@ -510,6 +547,26 @@ namespace platoon_strategic_ihp
             */
             bool isVehicleRightBehind(double downtrack);
             
+            /**
+            * \brief Function to determine if the target platoon leader is in an adjacent lane (used for cut-in from front) 
+            *
+            * \param type boolean
```
TonysCousin, line 553: Params are not described.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
```diff
@@ -510,6 +547,26 @@ namespace platoon_strategic_ihp
...
+            *
+            * \return if target vehicle in front
+            */
+            bool isVehicleNextLane(double frontVehicleDtd, double rearVehicleDtd, double frontVehicleCtD);
+
+            /**
+            * \brief Function to find the starting and ending lanelet ID for lane change in a two-lane scenario (used for cut-in from front) 
+            * 
+            * Note: This is a temporary function for internal test only. The scenario is not genrealized and further development is needed.
```
TonysCousin, line 562: For temp code, please use a TODO comment so that it's easy to search for.  These things have a way of being forgotten about.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
```diff
@@ -510,6 +547,26 @@ namespace platoon_strategic_ihp
             */
             bool isVehicleRightBehind(double downtrack);
             
+            /**
```
TonysCousin, line 550: These two new methods are described as being used for front cut-in only.  But their names and params don't indicate anything that would be restricted to a front join.  It would be nice if they be generalized enough to use for any cut-in.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
```diff
@@ -532,13 +589,28 @@ namespace platoon_strategic_ihp
...
             */
             cav_msgs::MobilityOperation composeMobilityOperationLeaderAborting();
-            
             /**
             * \brief Function to compose mobility operation in CandidateLeader.
             *
             * \return mobility operation msg.
             */
             cav_msgs::MobilityOperation composeMobilityOperationCandidateLeader();
+            /**
```
TonysCousin, line 598: In this whole section, please add a blank line after each method declaration (the compilable line) to make better visual grouping of declarations and their descriptions.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
```diff
@@ -532,13 +589,28 @@ namespace platoon_strategic_ihp
...
+            * \brief Function to compose mobility operation in LeadWithOperation (cut-in join)
```
TonysCousin, line 599: In these two methods, the input string has a risk of being malformed.  How will the methods handle garbage input?  If it throws and exception or returns null, etc, this condition needs to be documented here.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
```diff
@@ -614,6 +728,11 @@ namespace platoon_strategic_ihp
             // flag to check if map is loaded
             bool map_loaded_ = false;
 
+            // UCLA: flag to indicate the join type
```
TonysCousin, line 731: I suggest considering an enumerated variable instead of three bools.  What happens when none of these bools is true?  What happens when more than one is true?  Both situations are undefined states.  An enum will avoid that possibility.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -74,7 +74,7 @@ namespace platoon_strategic_ihp
      *
```
TonysCousin, line 74: Are all the notes here still relevant?  That is, do these actions still need to be done?  If so, please add a "TODO" tag so that it can easily be identified in a search.  And confirm for me that these will be done soon, right?

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -298,7 +303,105 @@ namespace platoon_strategic_ihp
...
+
+    // Generate a maneuver ID for new maneuver plans (only used for compose lane change maneuver)
+    std::string PlatoonStrategicIHPPlugin::getNewManeuverId() const {
+        /**
+         * @brief Function to generate new maneuver ID. (Use for genreating new lane change maneuver)
+         * 
+         * @return A newly generated uuid as int.
+         */
+
+        static auto gen = boost::uuids::random_generator(); // Initialize uuid generator
```
TonysCousin, line 316: Normally a static random number generator is initialized exactly once during a program's lifetime in order to maximize randomness.  Code as is will re-create the "gen" object and initialize on every call.  gen should be made a member variable and this statement moved to the constructor.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -298,7 +303,105 @@ namespace platoon_strategic_ihp
...
+        lanelet::ConstLanelet target_lanelet;
+        int target_lanelet_id;
+        if (!lanelets_in_path.empty())
+        {
+            for (int i = 0; i < lanelets_in_path.size(); ++i)
+            {
+                lanelet::BasicPoint2d p1=lanelets_in_path[i].centerline2d().basicLineString().back();
+                double downtrack=wm_->routeTrackPos(p1).downtrack;
+                double crosstrack=wm_->routeTrackPos(p1).crosstrack;
+                if (abs(crosstrack)>1)//1 is a threshold, the assumption for this condition is that there are two lanes on the road
```
TonysCousin, line 347: I realize this is only test code, but still some concern.  Don't restrict the code to assuming exactly two lanes.  It needs to handle 3 or more lanes.  I believe the crosstrack value ismeters from the roadway centerline (across all lanes).  Since you're using an integer here, it's not clear if you believe that or if you think it is number of lanes.  Therefore, crosstrack may be a negative number, and two adjacent lanes may both have a value > 1.  So I feel like this code is fragile, at best, and probably incorrect.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -298,7 +303,105 @@ namespace platoon_strategic_ihp
...
+         * 
+         * @return A newly generated uuid as int.
+         */
+
+        static auto gen = boost::uuids::random_generator(); // Initialize uuid generator
+        
+        return boost::lexical_cast<std::string>(gen()); // generate uuid and convert to string
+    }
+
+    // Find target lane indexes based on downtracks for lane change (used for testing in a two-lane map).
```
TonysCousin, line 321: Is this method going to disappear completely?  If so, when?  If so, it should have a "TODO" tag on it to make sure we can easily find anything that needs to be cleaned up.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -298,7 +303,105 @@ namespace platoon_strategic_ihp
...
+         *   (double) rear downtrack: the downtrack of the last target platoon member.
+         *   (double) front crosstrack: the crosstrack of the target platoon leader. 
+         * @return
+         *   (bool): if the target leader behind
+         */
+
+       double currentDtd = current_downtrack_;
+       double currentCtd = current_crosstrack_;
+       // lateral error for two lanes 
+       double two_lane_cross_error = 2*config_.maxCrosstrackError + 3.5; // todo: 3.5m is lane width, consider read it from map
```
TonysCousin, line 384: Need to complete the todos in this and next line before we call this PR complete.  These are good ideas.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -298,7 +303,105 @@ namespace platoon_strategic_ihp
...
+       // todo: use maxgap (35m) as longitudinal threshold, consider change 
+       bool longitudinalCheck = ((currentDtd >= rearVehicleDtd - maxAllowedJoinGap_) || (currentDtd <= frontVehicleDtd + 10));
```
TonysCousin, line 386: Due to operator precedence, these two statements don't need any parentheses.  One could argue that the inner parens give you some comfort, but the outer set adds nothing but visual clutter.

TonysCousin: The value 10 seems quite arbitrary, and needs to be explained.  Better to get it from a config parameter.  Be sure the code can accommodate vehicles of various lengths (including heavy trucks).

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -298,7 +303,105 @@ namespace platoon_strategic_ihp
...
+       bool lateralCheck = ((currentCtd >= frontVehicleCtd - two_lane_cross_error) || (currentCtd <= frontVehicleCtd + two_lane_cross_error));
```
TonysCousin, line 387: Clever test!  I suppose it may result in false sometimes if vehicles are both to the far sides of their respective lanes AND there is some measurement error.  Worth thinking about just a little bit more error allowance for that, or accepting that sometimes it will just fail.  At least you have logging for a failure.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -298,7 +303,105 @@ namespace platoon_strategic_ihp
...
+                int maxPosition = max_element(downtracks.begin(),downtracks.end()) - downtracks.begin();
+                target_lanelet=lanelets_sidelane[maxPosition];
+                target_lanelet_id=target_lanelet.id();
+            }
+        }
+        return target_lanelet_id;
+    } 
+
+    // Check if the platoon close to the platoon for cut-in join.
+    bool PlatoonStrategicIHPPlugin::isVehicleNextLane(double frontVehicleDtd, double rearVehicleDtd, double frontVehicleCtd)
```
TonysCousin, line 368: The method name doesn't match the stated stated meaning of the return variable.  I suggest renaming the method isVehicleNearPlatoon(), and updating the description of the return value accordingly.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -643,8 +809,8 @@ namespace platoon_strategic_ihp
     {
```
TonysCousin, line 809: It's still really confusing when this method will be used and what the situation might be. This will be called when host is in the LEADER state, which I believe says it is either leading a platoon or driving solo.  Does the logic fully accommodate both of those situations?  The comments within the method seem to assume the host is a solo vehicle listening to a platoon leader.  Will a solo vehicle transmit INFO messages?  At a minimum, these things need to be explained in a comment block.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -659,18 +825,21 @@ namespace platoon_strategic_ihp
...
-            
+
+            // use ecef_loc to calculate front Ctd in m.
+            double frontVehicleCtd = wm_->routeTrackPos(incoming_pose).crosstrack;
+
             // Use pm_ to find platoon end vehicle and its downtrack in m.
             int rearVehicleIndex = pm_.getTotalPlatooningSize()- 1;
             double rearVehicleDtd = pm_.platoon[rearVehicleIndex].vehiclePosition; 
             
             // downtrack of the platoon leader --> used for frontal join
```
TonysCousin, line 840: This comment doesn't match the compilable code. One is talking about leader, one is talking about rear vehicle.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -173,16 +174,19 @@ namespace platoon_strategic_ihp
     bool PlatoonStrategicIHPPlugin::isVehicleRightInFront(double downtrack)
```
TonysCousin, line 174: Now that we have cut-in capability, I think both isVehicleRightInFront() and isVehicleRightBehind() need to test CTD also to ensure that host is in the same lane as the platoon.  This allows the simple front or rear join.  If in different lanes, they will need to go through the cut-in logic, yes?

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -688,6 +857,8 @@ namespace platoon_strategic_ihp
                 request.location = pose_to_ecef(pose_msg_);
                 request.plan_type.type = cav_msgs::PlanType::JOIN_PLATOON_AT_REAR;
                 request.strategy = MOBILITY_STRATEGY;
+                // UCLA: Desired joining index for cut-in join, indicate the index of gap-leading vehicle. -1 indicate cut-in from front.
+                request.join_index = -1; // Note: not used by same-lane functions.
```
TonysCousin, line 861: I don't understand why this is here, if it is not used for same-lane join, which is what this code block is doing.  In any case, -1 indicates front join and this is the rear join logic.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -732,14 +903,16 @@ namespace platoon_strategic_ihp
                 request.header.sender_id = config_.vehicleID;
                 request.header.timestamp = ros::Time::now().toNSec() / 1000000;
                 request.location = pose_to_ecef(pose_msg_);
+                // UCLA: Desired joining index for cut-in join, indicate the index of gap-leading vehicle. -1 indicate cut-in from front.
+                request.join_index = -1; // Note: not used by same-lane functions.
```
TonysCousin, line 907: Don't understand why this is here for same-lane front join.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -688,6 +857,8 @@ namespace platoon_strategic_ihp
                 request.location = pose_to_ecef(pose_msg_);
```
TonysCousin, line 857: Assignment of plan_id above should use the singleton "gen" member variable I suggested above.  Look for other random generation statements elsewhere to make sure they all do it the same way.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -760,7 +933,57 @@ namespace platoon_strategic_ihp
...
+                fmter %pose_ecef_point_.ecef_z;     //  index = 4
+
+                request.strategy_params = fmter.str();
+                request.urgency = 50;
+
+                // UCLA: generate new platoonplan for front join.
+                pm_.current_plan = PlatoonPlan(true, request.header.timestamp, request.header.plan_id, request.header.sender_id);
+                mobility_request_publisher_(request);
+                ROS_DEBUG_STREAM("Publishing request to the leader " << senderId << " with params " << request.strategy_params << " and plan id = " << request.header.plan_id);
+                potentialNewPlatoonId_front = newLeaderPlatoonId;
```
TonysCousin, line 984: Would be nice if potentialNewPlatoonId_front would use consistent style: either camel case or snake case throughout, and end with a _ to indicate it is a member variable.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -760,7 +933,57 @@ namespace platoon_strategic_ihp
...
+                
+                // compose request
+                cav_msgs::MobilityRequest request;
+                request.header.plan_id = boost::uuids::to_string(boost::uuids::random_generator()());
+                request.header.recipient_id = senderId;
+                request.header.sender_id = config_.vehicleID;
+                request.header.timestamp = ros::Time::now().toNSec() / 1000000;
+                request.location = pose_to_ecef(pose_msg_);
+                // UCLA: Desired joining index for cut-in join, indicate the index of gap-leading vehicle. -1 indicate cut-in from front.
+                request.join_index = -1; // Note: not used by same-lane functions.
```
TonysCousin, line 956: The isVehicleNextLane() method that got us into this block returns true if host is anywhere beside the platoon. not necessarily near the front. So how can we now assume that this will be a front join?

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -760,7 +933,57 @@ namespace platoon_strategic_ihp
...
+                
+                // UCLA: change plan type --> ADD THIS TO PLANTYPE.MSG !!
+                request.plan_type.type = cav_msgs::PlanType::CUT_IN_FROM_SIDE; // UCLA: --> add this in plantype.msg
```
TonysCousin, line 959: This needs a TODO tag.  When will it be done?  We will need one or more PRs for the msgs repo to keep them in sync with these platform changes.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -814,8 +1038,123 @@ namespace platoon_strategic_ihp
...
+         * Otherwise, just send status params.
+         * Note: leader in state 'leading with operation' also send out INFO msg
+         */
+
+        // read parameters
+        std::string strategyParams = msg.strategy_params;
+        std::string senderId = msg.header.sender_id; 
+        std::string platoonId = msg.header.plan_id;  
+        std::string newLeaderID = msg.header.sender_id; 
+        std::string newLeaderPlatoonId = pm_.currentPlatoonID;
```
TonysCousin, line 1071: It's not clear why the platoon ID would be changing when we are doing a middle cut-in join.  I suggest a comment block explaining it.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -814,8 +1038,123 @@ namespace platoon_strategic_ihp
...
+    // UCLA: Mobility operation callback for prepare to join state (cut-in join).
+    void PlatoonStrategicIHPPlugin::mob_op_cb_preparetojoin(const cav_msgs::MobilityOperation& msg)
+    {
+        /*
+         * If same lane with leader, then send request to do same lane join. 
+         * Otherwise, just send status params.
+         * Note: leader in state 'leading with operation' also send out INFO msg
+         */
+
+        // read parameters
```
TonysCousin, line 1066: I am seeing lots of repetition of code like this.  I suggest looking for opportunities to refactor and pull these lines out into one or more helper functions that could be called by all of these mob_op or mob_req methods and make the class more compact.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -814,8 +1038,123 @@ namespace platoon_strategic_ihp
...
+
+            // Use pm_ to find platoon end vehicle and its downtrack in m.
+            int rearVehicleIndex = pm_.platoon.size() - 1;
+            double rearVehicleDtd = pm_.platoon[rearVehicleIndex].vehiclePosition; 
+            
+            // downtrack of the platoon leader --> used for frontal join
+            ROS_DEBUG_STREAM("rearVehicleDtd from ecef: " << rearVehicleDtd);
+
+            // determine joining vehicle position 
+            bool isSameLaneWithPlatoon = (frontVehicleCtd - current_crosstrack_ > -1 || frontVehicleCtd - current_crosstrack_ < 1);
```
TonysCousin, line 1098: Why not just use isVehicleNextLane()?  This is slightly different logic than in there, but it seems it should all be the same.  If it really needs to be different, it would be worth a comment here explaining why.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -814,8 +1038,123 @@ namespace platoon_strategic_ihp
...
+            bool isCutInFront = (frontVehicleDtd < current_downtrack_);
```
TonysCousin, line 1099: If isCutInFront is false, shouldn't we be testing for rear join also?

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -814,8 +1038,123 @@ namespace platoon_strategic_ihp
...
+
+            if (isSameLaneWithPlatoon)
```
TonysCousin, line 1101: According to definition of PREPARETOJOIN state, which determines that this method is called, the vehicle is not in the same lane because it is a cut-in join.  Therefore, it seems this test will always be false.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -814,8 +1038,123 @@ namespace platoon_strategic_ihp
...
+                request.header.sender_id = config_.vehicleID;
+                request.header.timestamp = ros::Time::now().toNSec()/1000000;
+                request.location = pose_to_ecef(pose_msg_);
+
+                // UCLA: send request based on cut-in type
+                if (isCutInFront) 
+                {
+                    request.plan_type.type = cav_msgs::PlanType::CUT_IN_FRONT_DONE;
+                }
+                else
```
TonysCousin, line 1119: No rear cut-in possible?  It may still need to change lanes to join from rear.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -814,8 +1038,123 @@ namespace platoon_strategic_ihp
...
+            bool isCutInFront = (frontVehicleDtd < current_downtrack_);
+
+            if (isSameLaneWithPlatoon)
+            {
+                // request 1. reset the safeToChangLane indicators  
+                pm_.safeToLaneChange = false;
+                
+                // request 2. change to same lane operation states (determine based on DTD differences)
+                cav_msgs::MobilityRequest request;
+                request.header.plan_id = boost::uuids::to_string(boost::uuids::random_generator()());
```
TonysCousin, line 1108: Extra parens?

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -921,6 +1270,7 @@ namespace platoon_strategic_ihp
         // Check joining plan type.
         bool isFrontJoin = (plan_type.type == cav_msgs::PlanType::JOIN_PLATOON_FROM_FRONT);
         bool isRearJoin = (plan_type.type == cav_msgs::PlanType::JOIN_PLATOON_AT_REAR);
+        bool isCutinJoin = (plan_type.type == cav_msgs::PlanType::CUT_IN_FROM_SIDE);
 
         // rear join; platoon leader --> leader waiting
         if (isRearJoin)
```
TonysCousin, line 1276: There is a TODO on line 1281 that I believe needs to be resolved before this plugin is complete.  Do you agree?

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -921,6 +1270,7 @@ namespace platoon_strategic_ihp
         // Check joining plan type.
         bool isFrontJoin = (plan_type.type == cav_msgs::PlanType::JOIN_PLATOON_FROM_FRONT);
         bool isRearJoin = (plan_type.type == cav_msgs::PlanType::JOIN_PLATOON_AT_REAR);
+        bool isCutinJoin = (plan_type.type == cav_msgs::PlanType::CUT_IN_FROM_SIDE);
```
TonysCousin, line 1273: Presumably all cut-ins are from the side.  So setting this could still indicate a front or rear join (from adjacent lane), right?

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1127,18 +1477,115 @@ namespace platoon_strategic_ihp
...
+        // UCLA: conditions for cut-in join; platoon leader --> leading with operation
+        else if (isCutinJoin)
+        {
+            /* We are currently checking two basic JOIN conditions:
+             *     1. The size limitation on current platoon based on the plugin's parameters.
+             *     2. Calculate how far away the joining vehicle to the platoon (longitudinal & lateral),  
+             *        Determine if it was in a reasonable distance for cut-in join.
+             */ 
+            
+            cav_msgs::MobilityHeader msgHeader = msg.header;
```
TonysCousin, line 1489: This block of ~40 lines beginning here is repeated in each of these 3 major if/else blocks.  It should be pulled out and executed above the if statement at the top of the method, thus substantially shrinking the code size and making it more maintainable.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1127,18 +1477,115 @@ namespace platoon_strategic_ihp
...
+            double applicantCurrentDtd = wm_->routeTrackPos(incoming_pose).downtrack;
+            ROS_DEBUG_STREAM("applicantCurrentmemberUpdates from ecef pose: " << applicantCurrentDtd);
+
+            // UCLA: Use crosstrack to determine condition to check if joining vehicle is close by
+            double applicant_crosstrack = wm_->routeTrackPos(incoming_pose).crosstrack;
+            ROS_DEBUG_STREAM("applicant_crosstrack from ecef pose: " << applicant_crosstrack);
+            double two_lane_cross_error = 2*config_.maxCrosstrackError + 3.5;
+
+            // cut-in vehicle and platoon are in adjacent lanes.
+            bool isLateralNearBy = (abs(applicant_crosstrack - current_crosstrack_) < two_lane_cross_error);
```
TonysCousin, line 1525: Why not call isVehicleNextLane() for consistency and code reuse (better maintainability).  This is the third different logic used for such a test.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1127,18 +1477,115 @@ namespace platoon_strategic_ihp
...
+            if (hasEnoughRoomInPlatoon && isLateralNearBy) 
+            {
+                ROS_DEBUG_STREAM("The current platoon has enough room for the applicant with size " << applicantSize);
+
+                // UCLA: change to read platoon front info
+                double currentFrontDtd = pm_.getPlatoonFrontDowntrackDistance();
+                double currentRearDtd = pm_.getPlatoonRearDowntrackDistance();
+                ROS_DEBUG_STREAM("The target cut-in platoon leader downtrack is " << currentFrontDtd << " m,  and the platoon rear downtrack is " << currentRearDtd << "m");
+                
+                // UCLA: check if cut-in vehicle is too far away from the platoon logitudinally
```
TonysCousin, line 1542: If you determine isLateralNearby by using isVehicleNextLane(), then a value of true guarantees that host is close enough, so this whole check is not necessary here.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1168,14 +1615,95 @@ namespace platoon_strategic_ihp
...
+                request. 
+                request.
+                ROS_DEBUG_STREAM("");
+                return MobilityRequestResponse::ACK;
+            }
+            */
+        }
+
+        // task 2: For cut-in from front, the leader need to stop creating gap when gap is big enough
+        if ((plan_type.type == cav_msgs::PlanType::STOP_CREAT_GAP) && (pm_.isCreateGap) )
```
TonysCousin, line 1671: CREATE is misspelled in this type name.  Is it correct in the cav_msgs?

TonysCousin, line 1671: Extraneous parentheses.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1168,14 +1615,95 @@ namespace platoon_strategic_ihp
...
+        {
+            // reset creat gap indicator
+            pm_.isCreateGap = false;
+            // no need to response, simple reset the indicator
+            return MobilityRequestResponse::NO_RESPONSE;
+
+        }
+
+        // task 3.1 cut-in front: After creating gap, revert back to same-lane operation 
+        if ((plan_type.type == cav_msgs::PlanType::CUT_IN_FRONT_DONE))
```
TonysCousin, line 1681: Extraneous parentheses.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1168,14 +1615,95 @@ namespace platoon_strategic_ihp
...
+        {
+            ROS_DEBUG_STREAM("Cut-in from front lane change finished, leader revert to same-lane maneuver.");
+            pm_.current_platoon_state = PlatoonState::LEADERABORTING;
+            return MobilityRequestResponse::ACK;
+        }
+
+        // todo: task 3 cut-in from middle/rear
```
TonysCousin, line 1688: Isn't this PR supposed to cover side cut-in?  Did we overlook this block?

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1168,14 +1615,95 @@ namespace platoon_strategic_ihp
...
+        */
+
+        // Check request plan type  
+        cav_msgs::PlanType plan_type = msg.plan_type;
+        if (plan_type.type == cav_msgs::PlanType::PLATOON_CUTIN_JOIN) 
+        {
+            // task 1. slow down to create gap 
+            int join_index = msg.join_index;
+            // determine if joining vehicle in position
+            double cutinDtdDifference = current_downtrack_ - pm_.platoon[0].vehiclePosition - config_.vehicleLength;
```
TonysCousin, line 1633: Not clear what is being computed. This method is for when host is platoon leader that someone else is joining, right?  Then current_downtrack_ is the host's DTD, which is the same as pm_.platoon[0].vehicle_position, right?  Therefore, cutinDtDDifference will always be -vehicleLength.  I may be seeing something wront.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1168,14 +1615,95 @@ namespace platoon_strategic_ihp
...
+            bool isFrontJoinerInPosition = (cutinDtdDifference <= 1.5*config_.vehicleLength) || (cutinDtdDifference >= -1.5*config_.vehicleLength);
```
TonysCousin, line 1634: General joining philosophy question: we can join a platoon from many car lengths away, right (e.g. a rear join)?  Can't we also join from many car lengths in front of the platoon, by just having the joiner gradually slow down to close the gap?  If these are true, then why does a cut-in need to be almost exactly at the front of the platoon whereas a same-lane front join may be from a lot farther in front?

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1168,14 +1615,95 @@ namespace platoon_strategic_ihp
...
+            // todo: reserved for cut-in from middle/rear
+            // bool isMidJoinerInPosition = ((pm_.platoon[join_index].vehiclePosition - current_downtrack_ <= 3*config_.vehicleLength)); 
+            // bool isRearJoinerInPosition = (pm_.platoon[pm_.platoon.size()-1].vehiclePosition - current_downtrack_ <= 3*config_.vehicleLength);
+
+            if (join_index == -1 && isFrontJoinerInPosition )
+            {
+                // slow down leader to allow joiner
```
TonysCousin, line 1641: If joiner is farther away than 1 vehicle length, then host should not have to slow down and create a gap.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1322,7 +1860,8 @@ namespace platoon_strategic_ihp
 
     // UCLA: add conditions to account for frontal join states (candidate follower) 
     void PlatoonStrategicIHPPlugin::mob_resp_cb_leader(const cav_msgs::MobilityResponse& msg)
-    {   /**  
+    {   
+        /**
```
TonysCousin, line 1864: I like this description of the functionality. But what if the host is the current platoon leader?  What mob_op_req would this be responding to, and what would the msg.plan_type be?  My guess is that all 3 of the bools below would be false, but it would be nice to see a comment confirming that.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1376,6 +1916,17 @@ namespace platoon_strategic_ihp
...
+                {
+                    ROS_DEBUG_STREAM("Received positive response for plan id = " << pm_.current_plan.planId);
+                    ROS_DEBUG_STREAM("Change to Prepare to join state and prepare to change lane. ");
+
+                    // Change to candidate leader and idle
+                    pm_.current_platoon_state = PlatoonState::PREPARETOJOIN;
+                    candidatestateStartTime = ros::Time::now().toNSec() / 1000000;
+                    // do not need to update platoon leader and platoon ID for candidate leader
+                }
                 else
```
TonysCousin, line 1930: This else block has been misinterpreted as only happening if the response was NACK (according to the debug stream).  This block will also get executed if host is the platoon leader, and (I think) all 3 boolean flags are false.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1458,20 +2008,90 @@ namespace platoon_strategic_ihp
...
+    // UCLA: response callback for prepare to join (inherited from leader waiting)
+    void PlatoonStrategicIHPPlugin::mob_resp_cb_preparetojoin(const cav_msgs::MobilityResponse& msg)
+    {
+        /*
+            If leader notify the member to slow down and ACK the request,
+            start to check the gap and change lane when gap is large enough 
+        */
+
+        cav_msgs::PlanType plan_type = msg.plan_type;
+        bool isCreatingGap = (plan_type.type == cav_msgs::PlanType::PLATOON_CUTIN_JOIN);
```
TonysCousin, line 2025: Extraneous parentheses.  Operator precedence makes them unnecessary.  This is typical of several places in this file.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1458,20 +2008,90 @@ namespace platoon_strategic_ihp
...
+        bool isFinishLaneChangeFront = (plan_type.type == cav_msgs::PlanType::CUT_IN_FRONT_DONE); 
+        // todo: add cut-in from middle and rear 
+        // bool isFinishLaneChangeMid
+        // bool isFinishLaneChangeRear
+
+        // UCLA: Creat Gap
+        if (msg.is_accepted && isCreatingGap)
```
TonysCousin, line 2032: It might be simpler to put this entire method inside a block that begins with "if (msg.is_accepted)", then it won't have to be tested multiple times.  It also makes the final else condition more clear.  See my comment above in mob_resp_cb_leader(), which would also benefit from such a structure.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1458,20 +2008,90 @@ namespace platoon_strategic_ihp
...
+        {
+            // task 1: check gap 
+            double joinerDtD = current_downtrack_;
+            double cut_in_gap = pm_.getCutInGap(target_join_index_, joinerDtD);   
+            ROS_DEBUG_STREAM("Start loop to check cut-in gap, start lane change when gap allows");
+            while (cut_in_gap < config_.maxGap)  // todo: use max gap as "safe to cut-in" gap, may need to adjust change later
+            {
+                ROS_DEBUG_STREAM("The target gap is not safe for lane change, wait for a larger gap");
+                cut_in_gap = pm_.getCutInGap(target_join_index_, joinerDtD);
```
TonysCousin, line 2041: Need a sleep statement in this loop, else it will run out of control.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1458,20 +2008,90 @@ namespace platoon_strategic_ihp
...
+            request.header.timestamp = ros::Time::now().toNSec() / 1000000;;
+            // UCLA: add new plan type to stop creat gap, add this in cav_msgs/plan_type
+            request.plan_type.type = cav_msgs::PlanType::STOP_CREAT_GAP;
+            request.strategy = MOBILITY_STRATEGY;
+            request.strategy_params = "";
+            request.urgency = 50;
+            request.location = pose_to_ecef(pose_msg_);
+            // UCLA: add joining index for cut-in join --> add this in cav_msgs/MobilityRequests
+            // note: for rear join, cut-in index == platoon.size()-1; for join from front, index == -1
+            //       for cut-in in middle, index indicate the gap leading vehicle's index
```
TonysCousin, line 2061: Please add clarification that this index is zero-based.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1643,7 +2271,7 @@ namespace platoon_strategic_ihp
         ROS_DEBUG_STREAM("Since we have max allowed gap as " << desiredJoinGap << " m then max join gap became " << maxJoinGap << " m");
         ROS_DEBUG_STREAM("The current gap from radar is " << currentGap << " m");
         // TODO: temporary
-        if (true)//(currentGap <= maxJoinGap && pm_.current_plan.valid == false) {
+        if (currentGap <= maxJoinGap && pm_.current_plan.valid == false)
```
TonysCousin, line 2274: Is this still temporary?  If not, please remove the TODO above.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1658,6 +2286,10 @@ namespace platoon_strategic_ihp
             request.strategy_params = "";
             request.urgency = 50;
             request.location = pose_to_ecef(pose_msg_);
+            
+            // UCLA added for cut-in, not used for other cases
+            request.join_index = -1; // Note: not used by same-lane functions.
```
TonysCousin, line 2291: Seems to indicate that this will always be a front join.  How does that work when plan_type implies host will be a follower (in fact, this method name implies that this is only for followers, not a future leader)?

platooning_strategic_IHP/include/platoon_strategic_ihp.h
xuhan417: bad version control, updated comment.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
xuhan417: Bad description, they are used for all cut-in cases. Updated comments accordingly.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
xuhan417: Updated.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
xuhan417: Added TODO mark.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
xuhan417: no problem.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
xuhan417: Updated.

platooning_strategic_IHP/include/platoon_strategic_ihp.h
xuhan417: Indeed, will update this.

platooning_strategic_IHP/include/platoon_strategic_plugin_node_ihp.h
xuhan417: Bad version control. Updated and recompiled.

xuhan417 (commented):
I address most of the comments. I will put together the updated request-response communication chart soon and also address the comments in the emails.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1142,48 +1562,149 @@ namespace platoon_strategic_ihp
...
+            joiningID_ = "";
+            pm_.current_platoon_state = PlatoonState::LEADERABORTING;
             return MobilityRequestResponse::ACK;
         }
-        else
+
+        // todo: task 3 cut-in from middle/rear
+
+        // task 4: if other joining vehicle send joning request, NACK it since there is already a cut-in join going on.
+        if (plan_type.type == cav_msgs::PlanType::CUT_IN_FROM_SIDE)
```
saina-ramyar: why do you need to specify the CUT_IN_FROM_SIDE? cant you just use the "else" to reject other requests?
What if you get like a join_from_rear req by mistake?

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1854,16 +2666,58 @@ namespace platoon_strategic_ihp
...
+        {
+            maneuver_msg.lane_change_maneuver.end_time = current_time + ros::Duration((end_dist - current_dist) / (0.5 * cur_plus_target));
+        }
+
+        // UCLA: need both start laneID and end laneID  for lane change
+        maneuver_msg.lane_change_maneuver.starting_lane_id = { std::to_string(starting_lane_id) };
+        maneuver_msg.lane_change_maneuver.ending_lane_id = { std::to_string(ending_lane_id) };
+
+        current_time = maneuver_msg.lane_change_maneuver.end_time;
+        ROS_DEBUG_STREAM("Creating lane follow start dist:"<<current_dist<<" end dist:"<<end_dist);
```
saina-ramyar: you are creating "lane change"

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1945,35 +2806,154 @@ namespace platoon_strategic_ihp
...
+                    auto p = shortest_path[last_lanelet_index].centerline2d().back(); // change to lane change path
+                    // set to next lane destination, consider sending ecef instead of dtd 
+                    double end_dist = wm_->routeTrackPos(shortest_path[last_lanelet_index].centerline2d().back()).downtrack;
+                    end_dist = std::min(end_dist, total_maneuver_length);
+                    ROS_DEBUG_STREAM("end_dist: " << end_dist);
+                    // consider calculate dtd_diff and ctd_diff
+                    double dist_diff = end_dist - current_progress;
+                    ROS_DEBUG_STREAM("dist_diff: " << dist_diff);
+                    // UCLA: start and end lane IDs
+                    int current_lanelet_id = shortest_path[last_lanelet_index].id();
```
saina-ramyar: your current lanelet id should be determined using the pose, not from list of shortest path

saina-ramyar (changes requested):

platooning_strategic_IHP/src/platoon_manager_ihp.cpp
```diff
@@ -617,7 +631,7 @@ namespace platoon_strategic_ihp
         else
         {   
             // ---> 4.1 pos_g 
-            pos_g = (pred_pos - vehicleLength_ + ego_pos*(timeGapAndStepRatio)) / (1 + timeGapAndStepRatio);
+            pos_g = (pred_pos - config_.vehicleLength + ego_pos*(timeGapAndStepRatio)) / (1 + timeGapAndStepRatio);
```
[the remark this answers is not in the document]
xuhan417: yes. I will re-organize this variable.

platooning_strategic_IHP/src/platoon_manager_ihp.cpp
```diff
@@ -631,4 +645,63 @@ namespace platoon_strategic_ihp
...
+    {
+        /*
+            Locate the target cut-in join gap size based on the index.   
+        */
+       // cut-in from front 
+       if (gap_leading_index == -1)
+       {
+            double gap_rear_dtd = platoon[0].vehiclePosition;
+            double gap_size = joinerDtD - gap_rear_dtd - config_.vehicleLength;
+            return gap_size;
```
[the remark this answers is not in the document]
xuhan417: Updated.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -298,7 +350,52 @@ namespace platoon_strategic_ihp
         return msg;
     }
 
-    // compose platoon info msg for all states
+    // ----------------------------- UCLA: helper functions for cut-in from front -------------------------------//
+
+    // Note: The function "find_target_lanelet_id" was used to test the IHP platooning logic and is only a pre-written scenario. 
+    // The IHP platooning should provide necessary data in a maneuver plan for the arbitrary lane change module.  
+
+    // Check if the platoon close to the platoon for cut-in join.
```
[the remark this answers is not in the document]
xuhan417: Yes !  Updated the comment.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -298,7 +350,52 @@ namespace platoon_strategic_ihp
...
+         */
+
+        // platoon leader and rear positions
+        int lastVehicleIndex = pm_.platoon.size()-1;
+        double rearVehicleDtd = pm_.platoon[lastVehicleIndex].vehiclePosition;
+        double frontVehicleDtd = pm_.platoon[0].vehiclePosition;
+        double frontVehicleCtd = pm_.platoon[0].vehiclePosition;
+
+        // lateral error for two lanes 
+        double two_lane_cross_error = 2*config_.maxCrosstrackError + 3.5; // todo: 3.5m is lane width, consider read it from map
```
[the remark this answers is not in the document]
xuhan417: Indeed. Updated.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -298,7 +350,52 @@ namespace platoon_strategic_ihp
...
+        // todo: use maxgap (35m) as longitudinal threshold, consider change 
+        bool longitudinalCheck = (currentDtd >= rearVehicleDtd - config_.maxAllowedJoinGap) || (currentDtd <= frontVehicleDtd + config_.maxGap);
```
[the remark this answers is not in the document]
xuhan417: Log added for both longitudinal and lateral checks.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
xuhan417: Indeed. Updated to else. This should definitely apply to all requests.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
xuhan417: Updated

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
xuhan417: Updated. We can use current_lanelets list (which was determined based on vehicle pose) to read current lanelet_ID.

xuhan417:
In addition to address the comments in this PR. I also addressed the concerns John raised in previous emails:

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
xuhan417: This TODO comment is actually carried over from the previous platooning code. The lane information is measured using cross track. I will delete this obsolete comment.

Issue #1462, opened 2021-09-24
Improve Stopping Behavior Reliability


msmcconnell:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

For the WorkZone use case the vehicle's need to come to a stop at a Red light. This works semi reliably, however, if the vehicle overshoots it results in an error. Additionally, the location of stopping is very inconsistent (0-10 ft of variability). The stop and wait plugin and or the relevant strategic plugins should be updated to stop more reliably. Previous investigations pointed to the trajectory replanning behavior to be at least partly to blame as is created small bumps in command speed which impacted the speed control loop. To help analyze this, AStuff sent Leidos a ROS Bag containing a stable and repeatable stopping behavior from Autoware.ai. This bag should be analyzed and compared with our own performance to identify the source of the issues and a plan developed to resolve them in support of TSMO Use Cases 1-3. 
### Carma version where this issue was discovered

3.8.0
### Expected behavior
Vehicle stops reliably within 5ft of the same point each run. Vehicle control commands are smooth. Vehicle can repeatedly stop 9/10 without operator override. 
### Actual behavior

See description
### Steps to reproduce the actual behavior

Run workzone use case at TFHRC. Stopping behavior can also be observed at the end of the route.

PR #1565, opened 2021-12-16
Feature/plan frontbumper2

saina-ramyar:
# PR Details
## Description

Currently, planning plugins plan based on vehicle pose which is located on its back axle. The arbitrator and strategic plugins are updated to consider the front bumper of the vehicle for route and maneuver planning.

## Related Issue
#1462 

## Motivation and Context

To improve stopping behavior, especially when the vehicle is required to stop behind a certain line.

## How Has This Been Tested?
Integration tested with White Pacifica.

## Types of changes

## Checklist:


arbitrator/src/arbitrator.cpp
```diff
@@ -194,4 +197,19 @@ namespace arbitrator
     {
         vehicle_state_.velocity = msg->twist.linear.x;
     }
+
+    void Arbitrator::lookupFrontBumperTransform() 
+    {
+        tf2_listener_.reset(new tf2_ros::TransformListener(tf2_buffer_));
+        tf2_buffer_.setUsingDedicatedThread(true);
```
kjrush, line 204: Does this create a whole new thread for doing the transform listening? Given that this function is called every pose_cb, it seems a bit excessive to re-initialize the listener every time, especially if it spins up a new thread. I've not worked too much with the tf2 server clients, so if this is standard practice I may be off base. But I think it could also make sense to just look up the transform once and cache it, as the transform itself should be a constant value.

plan_delegator/src/plan_delegator.cpp
```diff
@@ -172,6 +187,8 @@ namespace plan_delegator
                 ++current_maneuver_index;
                 continue;
             }
+
+            updateManeuverDistances(maneuver);
```
kjrush, line 191: I think we need a check here to delete the maneuver if the new end distance is less than the vehicle's current location after adjustment. In that case it should be removed from the list of the maneuvers to be planned as it's already complete.

arbitrator/src/arbitrator.cpp
saina-ramyar: Since this transform represents the location of front bumper in the map, it needs to be updated with localization. I tried looking it up once, but it was not able to get the current location of the vehicle.

plan_delegator/src/plan_delegator.cpp
saina-ramyar: updated

wz_strategic_plugin/include/wz_strategic_plugin/wz_strategic_plugin_config.h
```diff
@@ -29,7 +29,7 @@ struct WzStrategicPluginConfig
   double vehicle_decel_limit_multiplier = 0.75;
 
   //! The length parameter of the participant vehicle used to help calculate the distance before stopping at a red traffic signal
-  double vehicle_length = 5.0;
+  double vehicle_length = 0.0;
```
msmcconnell, line 32: It would be better to remove this parameter in its entirety then have it function as a no-op.

arbitrator/src/arbitrator.cpp
kjrush: Circling back on this, maybe I'm misunderstanding the whole transform tree system, but shouldn't it just be a relative value based on the urdf file? As in, the transform from rear-center axle to front bumper should be some fixed offset based on the vehicle length, and given a rear axle pose we only need to apply that fixed offset transform  (which could be looked up once and cached) to get the pose of the front bumper?

Am I misunderstanding the way the transforms are applied or stored?

route/src/route_generator_worker.cpp
```diff
@@ -469,12 +469,36 @@ namespace route {
         return true;
     }
 
+    void RouteGeneratorWorker::lookupFrontBumperTransform() 
+    {
+        tf2_listener_.reset(new tf2_ros::TransformListener(tf2_buffer_));
+        tf2_buffer_.setUsingDedicatedThread(true);
+        try
+        {
+            tf_ = tf2_buffer_.lookupTransform("map", "vehicle_front", ros::Time(0), ros::Duration(0.10)); //save to local copy of transform 20 sec timeout
```
kjrush, line 478: I think I understand now, my question from above. This is the transform from map -> vehicle front that's being queried, which is why it needs to be requeried every timestep, basically not using the pose topic data but instead using this transform value (that was computed by TF2 based on this transform).

I think what I was suggesting was more along the lines of just looking up the base_link -> vehicle_front transform once, and then applying that transform to the pose data at each timestep. I'm not sure what the overhead on a transform look up is, but if we do the transform look up once I think that's probably more efficient than doing it at 10+Hz.

saina-ramyar: I understand what you mean, and that is what I initially tried. However, I realized the baselink->bumper transform is in the base link frame, and the pose is in the map frame. So they don't work together. For example, if the current pose = (3, 3, 0) then when  transformed, bumper pose = (7, 3, 0). However, the 4 m added to x, does not necessarily mean the pose is shifted longitudinally to the front, it could be any direction, which is incorrect and not to mention causes route generation to fail.

kjrush: Do the transforms themselves encode the rotation data as well? Such that the rotation encoded in the map -> base_link transform, the vehicle's orientation w.r.t. the map, should already be accounted for prior to us adding the front bumper offset into the mix. 

With map as our baseline, we have the 4x4 transform matrix describing the position and orientation of the vehicle rear axle reference w.r.t. to map frame T<sub>1</sub> and another 4x4 transform matrix describing the position and orientation of the front bumper reference point w.r.t. rear axle T<sub>2</sub>, I think the pose of the front bumper w.r.t to map frame should be T<sub>1</sub>T<sub>2</sub> right?

arbitrator/src/arbitrator.cpp
```diff
@@ -165,12 +165,18 @@ namespace arbitrator
...
-    void Arbitrator::pose_cb(const geometry_msgs::PoseStampedConstPtr& msg) 
+    void Arbitrator::bumper_pose_cb()
     {
-        vehicle_state_.stamp = msg->header.stamp;
-        lookupFrontBumperTransform();   
-        
-        
+        try
+        {
+            tf_ = tf2_buffer_.lookupTransform("map", "vehicle_front", ros::Time(0), ros::Duration(1.0)); //save to local copy of transform 20 sec timeout
```
kjrush, line 172: Where is this 20s timeout defined? It looks like 1.0s timeout on the lookup itself, is there somewhere else a 20 second timeout (on data validity?) is set?

route/src/route.cpp
```diff
@@ -32,7 +32,7 @@ namespace route {
         route_state_pub_ = nh_->advertise<cav_msgs::RouteState>("route_state", 1, true);
         route_marker_pub_= nh_->advertise<visualization_msgs::Marker>("route_marker", 1, true);
         // init subscribers
-        pose_sub_ = nh_->subscribe("current_pose", 1, &RouteGeneratorWorker::pose_cb, &rg_worker_);
+        // pose_sub_ = nh_->subscribe("current_pose", 1, &RouteGeneratorWorker::pose_cb, &rg_worker_);
```
kjrush, line 35: If these are no longer needed due to the front_bumper transform, let's delete these lines entirely.

route/test/test_route_generator.cpp
```diff
@@ -484,8 +486,8 @@ TEST(RouteGeneratorTest, test_set_active_route_cb)
     route::RouteGeneratorWorker worker2;
     worker2.georeference_cb(msg_ptr); 
     worker2.setWorldModelPtr(cmw);
-    worker2.set_route_file_path("../resource/route/");
-    worker2.pose_cb(mpt);
+    worker2.set_route_file_path("/workspaces/carma_ws/src/carma-platform/route/resource/route/");
```
kjrush, line 489: Is this normal to use an absolute path instead of relative? Might cause breakages on other machines right?

kjrush (approved):
Generally looks good to me, a few minor cleanup items but then I think we're okay!

route/test/test_route_generator.cpp
saina-ramyar: No I tried changing it maybe the map can be loaded properly. which didn't help. revert back.

route/src/route.cpp
saina-ramyar: removed

arbitrator/src/arbitrator.cpp
saina-ramyar: It is arbitrary, not required.

kjrush (approved):

PR #1573, opened 2021-12-30
fix stopping issue

adev4a:
# PR Details
This pr fixes issues seen with stopping while testing tsmo use case 1.
Updates were made to parameters and logic was added to stop_and_wait plugin to prevent acceleration when the vehicle was trying to slow down.
## Types of changes

## Checklist:

### DIFF
--- a/stop_and_wait_plugin/config/parameters.yaml
+++ b/stop_and_wait_plugin/config/parameters.yaml
@@ -16,7 +16,7 @@ accel_limit_multiplier : 0.95
 
 # Double: Minimum speed the vehicle can command before being ready to stop
 # Units: m/s
-crawl_speed : 0.89
+crawl_speed : 0.91
 
 # Double: The gap in meters between points sampled from the lanelet centerlines for planning trajectory positions
 # Units: m
--- a/stop_and_wait_plugin/config/parameters.yaml
+++ b/stop_and_wait_plugin/config/parameters.yaml
@@ -16,7 +16,7 @@ accel_limit_multiplier : 0.95
 
 # Double: Minimum speed the vehicle can command before being ready to stop
 # Units: m/s
-crawl_speed : 2.2352
+crawl_speed : 0.89
 
 # Double: The gap in meters between points sampled from the lanelet centerlines for planning trajectory positions
 # Units: m
--- a/stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
+++ b/stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
@@ -237,7 +237,10 @@ std::vector<cav_msgs::TrajectoryPlanPoint> StopandWait::compose_trajectory_from_
 
   if (req_dist > remaining_distance)
   {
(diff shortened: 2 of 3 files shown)


saina-ramyar (approved):

stop_and_wait_plugin/config/parameters.yaml
```diff
@@ -16,7 +16,7 @@ accel_limit_multiplier : 0.95
 
 # Double: Minimum speed the vehicle can command before being ready to stop
 # Units: m/s
-crawl_speed : 2.2352
+crawl_speed : 0.89
```
msmcconnell, line 19: Looks like this has been set to 2mph. What is the motivation for that change? 2mph is the threshold used for commanding zero speed in at least some of the SSC modules. Choosing it as the crawl_speed could potentially lead to unexpected braking due to noise in the system. 

Also I don't believe the trucks are capable of commanding only 2mph so this could potentially interfere with port drayage. 

The previous value of 5mph should be trivial to stop from when a 0 is sent, so what is the benefit of this change?

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -237,6 +237,10 @@ std::vector<cav_msgs::TrajectoryPlanPoint> StopandWait::compose_trajectory_from_
 
   if (req_dist > remaining_distance)
   {
+    while (remaining_distance <= 0.0 && remaining_distance <= (stop_location - starting_downtrack)){
+      remaining_distance += 0.2;
+    }
```
msmcconnell, line 242: This bit of logic is pretty unintuitive. It needs a detailed comment describing what edge case it is addressing and why this is a good solution.

stop_and_wait_plugin/config/parameters.yaml
adev4a: Based on testing on the Black Pacifica, the 5mph value was too high. The vehicle is not able to stop successfully when the speed command goes from 5 to 0. It can be tuned to a higher value after testing
The way the crawl_speed is working currently the vehicle is being asked to stay at crawl speed halfway through the buffer. During testing this meant there was only 1 point at the end with a commanded speed of zero, with the buffer set at 10m.
A work around that should work as well is setting really high buffers (~20m) so that there are more points commanding the zero speed.

msmcconnell: We used the Black Pacifica extensively during workzone testing and were able to stop with the 5mph parameter. Why would this be a critical factor now?

adev4a: Could be that the map is sampled at a lower rate. It does make sense to me to not change this param, but we would have to increase the stopping buffer

msmcconnell: If we are targeting the midpoint of the buffer though, shouldn't there be a few meters of extra space which we can command zero in if that point is overshot? 

Either way, Jon looked into the truck SSC file and thinks 0.9 is our minimum there (at least as a parameter), so if we could up this to 0.91 I would be comfortable with merging it.

msmcconnell (approved):

PR #1577, opened 2022-01-06
Feature/platoon in left lane

saina-ramyar:
# PR Details
## Description
Update platooning strategic plugin to not accept platoon join requests in the rightmost lane.

## How Has This Been Tested?
Unit test

## Types of changes

## Checklist:


platooning_strategic/src/platoon_strategic.cpp
```diff
@@ -129,6 +131,36 @@ namespace platoon_strategic
         }
     }
 
+    void PlatoonStrategicPlugin::checkForRightMostLane(const lanelet::BasicPoint2d& current_location)
+    {
+        auto current_lanelet = wm_->getLaneletsFromPoint(current_location, 1);
```
kjrush, line 136: I'd recommend checking for size on this output, the getLaneletsFromPoint method is not guaranteed to return a non-empty list.

kjrush (approved):
Let's add a size check where mentioned but otherwise this seems solid to me. I think in review of this I might have found a potential bug in world model utils (as it relates to the lanelet geometry API) but it's pretty minor and unlike to arise I'm not concerned about fixing it here. I'll write up a Github issue for it though.

platooning_strategic/src/platoon_strategic.cpp
saina-ramyar: added

platooning_strategic/src/platoon_strategic.cpp
```diff
@@ -134,6 +134,10 @@ namespace platoon_strategic
     void PlatoonStrategicPlugin::checkForRightMostLane(const lanelet::BasicPoint2d& current_location)
     {
         auto current_lanelet = wm_->getLaneletsFromPoint(current_location, 1);
+        if (current_lanelet.size()<1)
+        {
+            std::invalid_argument("There are no lanelets in the current location.");
```
kjrush, line 139: I think you're missing the `throw` here, but otherwise this looks good to me. Quick tweak and then we should be good to merge, though Sonar Scanner appears to be broken somehow.

PR #1585, opened 2022-01-12
Feature/ns 3 adapter

lewisid:
# PR Details
## Description

With the recently added additions, this NS-3 adapter should be able perform the following functionalities within the context of the simulation:
1. Configure and connect to the NS-3 CARMA application within the MOSAIC environment, ensuring that the NS-3 simulation is aware and ready to respond to the CARMA platform instance.
2. During runtime: any outbound binary messages from the CARMA Platform should be forwarded to the NS-3 CARMA Application per the interface defined by ODU. This will submit a request to the NS-3 simulator to begin simulation of that packet broadcast.
3. During runtime: Listen for any inbound messages from the same connection to the NS-3 CARMA Application. Upon receipt of such a message, process the wrapper defined by ODU's interface and publish it to the appropriate inbound_binary topic for further processing by the CARMA platform.
4. Provide any other system-level status information that would be required by the comms driver.

## Motivation and Context

In order to integrate with the NS-3 network simulation functionality developed by ODU it is necessary to create a new ROS node in the CARMA Platform hardware interface layer. This ROS node should operate similarly to (and replace for the purposes of simulation) the current Cohda DSRC OBU driver. 
## Types of changes

## Checklist:


ns-3_adapter/include/ns-3_adapter.h
```diff
@@ -0,0 +1,196 @@
...
+        NS3Adapter(int argc, char** argv);
+
+        ~NS3Adapter() { shutdown(); }
+
+    private:
+
+        std::vector<std::string> api_;
+
+        //ROS
+        ros::Publisher comms_pub_;
```
kjrush, line 64: We'll likely need a few other data streams here as well, I believe. We should be subscribing to the vehicle pose, and we'll also need the vehicle ID (from VehicleConfigParams should be fine) to uniquely identify which vehicle this NS-3 message is associated with (and the pose to say where it's broadcast from). I'd recommend adding subscribers for those whereever you think is most appropriate so that the data can get into that publish NS-3 message function call.

ns-3_adapter/package.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0"?>
```
kjrush, line 1: Copyright notice, I'd recommend just doing a pass on all relevant files to check for this.

ns-3_adapter/package.xml
```diff
@@ -0,0 +1,49 @@
...
+<package>
+  <name>ns-3_adapter</name>
+  <version>0.0.0</version>
+  <description>The ns-3_adapter package</description>
+
+  <!-- One maintainer tag required, multiple allowed, one person per tag -->
+  <!-- Example:  -->
+  <!-- <maintainer email="jane.doe@example.com">Jane Doe</maintainer> -->
+  <maintainer email="carma@todo.todo">carma</maintainer>
```
kjrush, line 10: We can replace with carma@dot.gov

ns-3_adapter/package.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0"?>
+<package>
+  <name>ns-3_adapter</name>
+  <version>0.0.0</version>
```
kjrush, line 4: Recommend version 0.0.1, and adding to the description that this is for use with CARMA XIL.

ns-3_adapter/package.xml
```diff
@@ -0,0 +1,49 @@
...
+  <!-- One maintainer tag required, multiple allowed, one person per tag -->
+  <!-- Example:  -->
+  <!-- <maintainer email="jane.doe@example.com">Jane Doe</maintainer> -->
+  <maintainer email="carma@todo.todo">carma</maintainer>
+
+
+  <!-- One license tag required, multiple allowed, one license per tag -->
+  <!-- Commonly used license strings: -->
+  <!--   BSD, MIT, Boost Software License, GPLv2, GPLv3, LGPLv2.1, LGPLv3 -->
+  <license>TODO</license>
```
kjrush, line 16: We can specify Apache 2.0 here.

kjrush (approved):

PR #1586, opened 2022-01-13
Feature/wm intersection

MishkaMN:
# PR Details
## Description

This PR is to 
- integrate SignalizedIntersectionManager into the world model. This would mean MAP and SPAT msgs should be using it in world model. And map_update should also take it into account
   - any timer related variables in CARMAWorldModel.cpp were moved to SignalizedIntersectionManager
- processSPAT function supports dynamic timers. 
   - It starts from default timers as usual. Then sets individual light colors until full cycle. 
   - Do nothing until new signal's prediction doesn't match with that of existing. If different, shift end of current cycle to the received min_end time (retaining previous cycle duration, to be changed again). And start setting individual signal colors again until full cycle again. 
   - This existed before, but another note of setting SPAT signal timer is that it sets received signal if the next received signal is different. e.g:
   Received: R  Y  G  R  Y
   Set: ......... N  R  Y  G  R -> (full cycle) 
   Received: R  Y  Y  G  R  R  R  R  Y
   Set: ......... N  R  N  Y  G  N  N  N  R -> (full cycle) 
- MapUpdateCallback supports signalized intersection.
   - since binary map_update was not able to retain weakLanelet memory addresses after deserializing, fromBinMsg also uses existing map's primitives to overwrite parameters with same id to avoid letting the weak_ptr (weakLanelet for example) expire.
   - carma_wm_ctrl sends SignalizedIntersectionManager info in the last update of all updates to avoid duplicate information.
   - each carma_traffic_signal and carma_signalized_intersection has its own map_update. So 1 intersection with 4 signal_groups for example would send 5 map_updates.
 

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1587

## Motivation and Context
This is to support TSMO USE CASE 2

## How Has This Been Tested?
Local integration tested with real Map msg.

## Types of changes

## Checklist:


carma/launch/message.launch
```diff
@@ -47,6 +47,10 @@
   </group>
 
   <!-- Message Encoder/Decoder Node -->
-  <include file="$(find cpp_message)/launch/cpp_message.launch"/>
+  <group>
+    <remap from="incoming_j2375_map" to="incoming_j2735_map"/>
```
msmcconnell, line 51: Could you put a small pr up on the carma-messenger repo to correct this typo. I know this remap would work, but since its a typo in the actual base topic name it would probably be best to resolve it directly.

carma_wm/src/SignalizedIntersectionManager.cpp
```diff
@@ -143,14 +154,14 @@ namespace carma_wm
...
-        ROS_DEBUG_STREAM("Adding  entry_lane id: " << entry_lane);
+        ROS_DEBUG_STREAM("Adding entry_lane id: " << entry_lane);
         if (entry.find(entry_lane) != entry.end())
         {
           signal_group_to_entry_lanelet_ids_[iter->first].insert(entry[entry_lane]);
         }
         else
         {
-          throw std::invalid_argument(std::string("Unable to convert entry lane Id: " + std::to_string((int)entry_lane) + ", to lanelet id using the given MAP.msg!").c_str());
+          ROS_WARN_STREAM("Unable to convert entry lane Id: "  + std::to_string((int)entry_lane) + ", to lanelet id using the given MAP.msg!");
```
msmcconnell, line 164: Seems that many exceptions have been replaced with warning logs. 
What is the impact of this? If these events occur is the vehicle still safely able to operate?

carma_wm/src/SignalizedIntersectionManager.cpp
```diff
@@ -287,7 +298,6 @@ namespace carma_wm
     // check if it already exists
     for (auto sig_grp_pair : signal_group_to_exit_lanelet_ids_)
     {
-      ROS_DEBUG_STREAM("Creating signal for: " << (int)sig_grp_pair.first);
```
msmcconnell, line 290: Any particular reason this and the above debug log were removed?

carma_wm/src/TrafficControl.cpp
```diff
@@ -51,7 +51,22 @@ void fromBinMsg(const autoware_lanelet2_msgs::MapBin& msg, std::shared_ptr<carma
   std::stringstream ss;
   ss << data_str;
   boost::archive::binary_iarchive oa(ss);
+
   oa >> *gf_ptr;
+
+  if (!lanelet_map)
+    return;
```
msmcconnell, line 58: Are there two use cases for this function now? It looks like some of the logic gets run when there is no-map provided. Are both cases being used?

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -198,7 +200,31 @@ void WMListenerWorker::mapUpdateCallback(const autoware_lanelet2_msgs::MapBinPtr
 
   ROS_DEBUG_STREAM("Geofence id" << gf_ptr->id_ << " sends record of intersections size: " << gf_ptr->sim_.intersection_id_to_regem_id_.size());
   if (gf_ptr->sim_.intersection_id_to_regem_id_.size() > 0)
+  {
     world_model_->sim_ = gf_ptr->sim_;
+    for (auto pair : world_model_->sim_.intersection_id_to_regem_id_)
```
msmcconnell, line 205: Seems like these loops are entirely for printing, can they be moved to a separate debug print function to isolate the printing logic from the business logic in the method?

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -274,67 +296,67 @@ void WMListenerWorker::mapUpdateCallback(const autoware_lanelet2_msgs::MapBinPtr
...
                                                             std::const_pointer_cast<lanelet::RegulatoryElementData>(regem->constData()));
 
   // we should extract general regem to specific type of regem the geofence specifies
   switch(resolveGeofenceType(regem->attribute(lanelet::AttributeName::Subtype).value()))
   {
     case GeofenceType::PASSING_CONTROL_LINE:
     {
-
-      lanelet::PassingControlLinePtr control_line = std::dynamic_pointer_cast<lanelet::PassingControlLine>(factory_pcl);
+      lanelet::PassingControlLinePtr control_line = std::dynamic_pointer_cast<lanelet::PassingControlLine>(factory_regem);
```
msmcconnell, line 307: Should have caught this earlier, but if a dynamic_pointer_cast is being used you should check the resulting pointer is populated as dynamic_cast returns a nullptr if the cast failed.

carma_wm_ctrl/src/WMBroadcasterNode.cpp
```diff
@@ -68,8 +68,11 @@ int WMBroadcasterNode::run()
   georef_sub_ = cnh_.subscribe("georeference", 1, &WMBroadcaster::geoReferenceCallback, &wmb_);
   // Geofence Sub
   geofence_sub_ = cnh_.subscribe("geofence", 200, &WMBroadcaster::geofenceCallback, &wmb_);
+  // External Map Msg Sub
+  incoming_map_sub_ = cnh_.subscribe("incoming_map", 200, &WMBroadcaster::externalMapMsgCallback, &wmb_);
```
msmcconnell, line 72: queue size can be smaller, map messages publish at 1 Hz, since we would probably never be in range of more than ~5 intersections that's 5 hz max. I think a queue size of 10 or 20 would be sufficient.

msmcconnell (changes requested):

msmcconnell (changes requested):
Realized I missed CARMAWorldModel.cpp in my last PR.
It has a massive change set most of which appears to be nothing but formatting. Please confirm and specify exactly which methods have actually changed so I can review them.

carma/launch/message.launch
MishkaMN: There is an open PR fix for that here
https://github.com/usdot-fhwa-stol/carma-messenger/pull/118

carma_wm/src/SignalizedIntersectionManager.cpp
MishkaMN: I deleted it by mistake. I added it back

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -36,1360 +36,1487 @@
...
+    return signal_id;
   }
 
-}
-
-TrackPos CARMAWorldModel::routeTrackPos(const lanelet::ConstLanelet& lanelet) const
-{
-  // Check if the route was loaded yet
-  if (!route_)
+  TrackPos CARMAWorldModel::routeTrackPos(const lanelet::ConstLanelet& lanelet) const
```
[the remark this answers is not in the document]
MishkaMN: This function is only formatting

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -36,1360 +36,1487 @@
...
-      else if (tp.downtrack < minPos.downtrack)
+    }
+    return std::make_pair(minPos, maxPos);
+  }
+
+  lanelet::Id CARMAWorldModel::getTrafficSignalId(uint16_t intersection_id, uint8_t signal_group_id)
+  {
+    if (!traffic_light_ids_.empty())
+    {
+      // TODO: Old logic that needs be removed when workzone is connected
```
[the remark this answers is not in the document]
MishkaMN: Added the link of open issue

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -36,1360 +36,1487 @@
...
+    lanelet::ConstLanelets path_lanelets(route_->shortestPath().begin(), route_->shortestPath().end());
+    shortest_path_view_ = lanelet::utils::createConstSubmap(path_lanelets, {});
+    computeDowntrackReferenceLine();
+    // NOTE: Setting the route_length_ field here will likely result in the final lanelets final point being used. Call setRouteEndPoint to use the destination point value
+    route_length_ = routeTrackPos(route_->getEndPoint().basicPoint2d()).downtrack;  // Cache the route length with
+                                                                                   // consideration for endpoint
   }
 
-  lanelet::ConstPoint3d const_end_point{ lanelet::utils::getId(), end_point.x(), end_point.y(), end_point.z() };
+  void CARMAWorldModel::setRouteEndPoint(const lanelet::BasicPoint3d& end_point)
```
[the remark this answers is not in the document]
MishkaMN: Yes

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -36,1360 +36,1487 @@
 namespace carma_wm
 {
 
-std::pair<TrackPos, TrackPos> CARMAWorldModel::routeTrackPos(const lanelet::ConstArea& area) const
-{
-  // Check if the route was loaded yet
-  if (!route_)
+  std::pair<TrackPos, TrackPos> CARMAWorldModel::routeTrackPos(const lanelet::ConstArea& area) const
```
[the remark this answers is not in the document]
MishkaMN: Yup!

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -36,1360 +36,1487 @@
...
-}
+    return sorted_output;
+  }
 
-std::vector<lanelet::BasicPoint2d> CARMAWorldModel::sampleRoutePoints(double start_downtrack, double end_downtrack,
-                                                                      double step_size) const
-{
-  std::vector<lanelet::BasicPoint2d> output;
-  if (!route_)
+  std::vector<lanelet::BasicPoint2d> CARMAWorldModel::sampleRoutePoints(double start_downtrack, double end_downtrack,
```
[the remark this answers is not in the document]
MishkaMN: Yup!

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -36,1360 +36,1487 @@
...
-  for (auto basic_p : line.basicLineString())
+  std::string CARMAWorldModel::getRouteName() const
   {
-    coppied_points.push_back(lanelet::Point3d(lanelet::utils::getId(), basic_p));
+    return route_name_;
   }
 
-  return lanelet::LineString3d(lanelet::utils::getId(), coppied_points);
-}
+  lanelet::LineString3d CARMAWorldModel::copyConstructLineString(const lanelet::ConstLineString3d& line) const
```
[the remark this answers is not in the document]
MishkaMN: Yup!

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -36,1360 +36,1487 @@
...
-  // NOTE: Setting the route_length_ field here will likely result in the final lanelets final point being used. Call setRouteEndPoint to use the destination point value
-  route_length_ = routeTrackPos(route_->getEndPoint().basicPoint2d()).downtrack;  // Cache the route length with
-                                                                                  // consideration for endpoint
-}
-
-void CARMAWorldModel::setRouteEndPoint(const lanelet::BasicPoint3d& end_point)
-{
-  if (!route_) {
-    throw std::invalid_argument("Route endpoint set before route was available.");
+  void CARMAWorldModel::setRoute(LaneletRoutePtr route)
```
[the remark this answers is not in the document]
MishkaMN: Yup1

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -36,1360 +36,1487 @@
...
+    }
+    bool operator>(const LaneletDowntrackPair& pair) const
+    {
+      return this->downtrack_ > pair.downtrack_;
+    }
+  };
 
-  auto lanelet_map = route_->laneletMap();
-  for (lanelet::ConstLanelet lanelet : lanelet_map->laneletLayer)
+  std::vector<lanelet::ConstLanelet> CARMAWorldModel::getLaneletsBetween(double start, double end, bool shortest_path_only,
```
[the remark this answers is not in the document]
MishkaMN: yup!

carma_wm/src/TrafficControl.cpp
MishkaMN: First case where the map is not required is easy for unit test and the map_update_debugger.
Otherwise there is no solid reason not to enforce it at the start.

MishkaMN:
> Realized I missed CARMAWorldModel.cpp in my last PR. It has a massive change set most of which appears to be nothing but formatting. Please confirm and specify exactly which methods have actually changed so I can review them.

Main functions that had change from this file are:
- getTrafficSignalId
- getTrafficSignal
- getSignalizedIntersectionsAlongRoute
- getSignalsAlongRoute
- processSpatFromMsg (biggest)

carma_wm/src/SignalizedIntersectionManager.cpp
MishkaMN: Having `throw exception` there made it really hard to integration test as it would automatically shut every thing down. In worst case scenario, the car will drive through red as it did not detect it as entry into an intersection. But again, for continuing testing it might be better as this.

carma_wm_ctrl/src/WMBroadcasterNode.cpp
MishkaMN: Reduced to 20

carma_wm/src/WMListenerWorker.cpp
MishkaMN: added guards for every one of them

carma_wm/src/WMListenerWorker.cpp
MishkaMN: factored it out to a different helper function

carma_wm/src/TrafficControl.cpp
msmcconnell: This should be specified in the function comments

carma_wm/include/carma_wm/SignalizedIntersectionManager.h
```diff
@@ -36,8 +36,9 @@ namespace carma_wm
 
 struct LANE_DIRECTION
 {
-  static const uint8_t INGRESS = 1;
-  static const uint8_t EGRESS = 2;
+  // Integration testing found reverse of j2735_msgs as below. TODO
```
msmcconnell, line 39: TODO mean anything here?

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -36,1360 +36,1488 @@
...
-        ROS_DEBUG_STREAM("Updated time for state: " << static_cast<lanelet::CarmaTrafficSignalState>(current_movement_state.movement_event_list[0].event_state.movement_phase_state)  << ", with time: " 
-                          << std::to_string(lanelet::time::toSec(min_end_time)));
-        traffic_light_states_[curr_intersection.id.id][current_movement_state.signal_group].back().first = min_end_time;
-        continue;
-      }
+        if (curr_intersection.moy_exists) //account for minute of the year
+        {
+          auto inception_boost(boost::posix_time::time_from_string("1970-01-01 00:00:00.000")); // inception of epoch
+          auto duration_since_inception(lanelet::time::durationFromSec(ros::Time::now().toSec()));
+          auto curr_time_boost = inception_boost + duration_since_inception;
```
msmcconnell: Why is it necessary to add the time to the current epoch? If the time is already UTC shouldn't this always give an equal value?

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -36,1360 +36,1488 @@
...
-                        ", time: " << ros::Time::fromBoost(min_end_time));
+          auto curr_day_boost(boost::posix_time::time_from_string(std::to_string(curr_year) + "/" + std::to_string(curr_month) + "/" + std::to_string(curr_day) +" 00:00:00.000")); // GMT is the standard
+          auto curr_hour_boost = curr_day_boost + boost::posix_time::hours(hours_of_day);
 
+          min_end_time += lanelet::time::durationFromSec(lanelet::time::toSec(curr_hour_boost));
+          ROS_DEBUG_STREAM("New min_end_time: " << std::to_string(lanelet::time::toSec(min_end_time)));
+        }
+
+        auto last_time_difference = sim_.last_seen_state_[curr_intersection.id.id][current_movement_state.signal_group].first - min_end_time;  
+        bool is_duplicate = last_time_difference.total_milliseconds() >= -100 && last_time_difference.total_milliseconds() <= 100;
```
msmcconnell: Can you provide some reasoning as to why >= -100 and <= 100 is the check for duplicate time?

carma_wm/src/SignalizedIntersectionManager.cpp
msmcconnell: Can you put up an enhancement issue to review this and add a TODO with the link here as something we should come back to review.

docker/checkout.bash
```diff
@@ -43,11 +43,11 @@ cd ${dir}/src
 if [[ "$BRANCH" = "develop" ]]; then
       git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch $BRANCH
       git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-utils.git --branch $BRANCH
-      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-messenger.git --branch $BRANCH
+      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-messenger.git --branch fix/map_msg_name
```
msmcconnell, line 46: Reminder to clean this up before we merge this PR

msmcconnell (changes requested):

carma_wm/src/SignalizedIntersectionManager.cpp
MishkaMN: https://github.com/usdot-fhwa-stol/carma-platform/issues/1593
Enhancement issue added.

carma_wm/src/TrafficControl.cpp
MishkaMN: Added to function comments

carma_wm/include/carma_wm/SignalizedIntersectionManager.h
MishkaMN: I removed the TODO comment as there is nothing to do from carma-platform side, but I was thinking maybe Carma-cloud has a bug where it switched the types of lanes. 
From my understanding INGRESS is entry lane, while EGRESS is exit lane: 
https://github.com/usdot-fhwa-stol/carma-msgs/blob/develop/j2735_msgs/msg/LaneDirection.msg
But integration testing found it is sending the opposite

carma_wm/src/CARMAWorldModel.cpp
MishkaMN: its goal is merely a conversion from `posix_duration` to `posix_time`. 
Inception_boost is `posix_time`
ros::Time::now().toSec() is converted to duration easiest. So we add time + duration to get time again.

carma_wm/src/CARMAWorldModel.cpp
MishkaMN: There seem to have slight drift or change in the timers sent, so sometimes exact `==` doesn't work. So I thought it is safer to check if new value is within 0.1 seconds boundary

docker/checkout.bash
MishkaMN: changed it back

carma_wm/include/carma_wm/SignalizedIntersectionManager.h
msmcconnell: I believe you have the correct interpretation? Can you follow up with Sudhakar about this? Might be an issue in v2x hub

carma_wm/src/CARMAWorldModel.cpp
msmcconnell: Kinda strange way to go about it, but in the interest of time I think we can leave it as is.

carma_wm/src/CARMAWorldModel.cpp
msmcconnell: Since 100ms is the period of spat messages in the ideal case, maybe a lower value would be better. Like 30 to avoid potentially ignoring a message as a false duplicate?

MishkaMN: reduced it to 30ms

carma_wm/include/carma_wm/SignalizedIntersectionManager.h
MishkaMN: Apparently the ingress/egress is just a static file that can be edited. I have a general idea from Will on how to edit that XML file, so I reversed the INGRESS/EGRESS to match j2735

msmcconnell (approved):

PR #1592, opened 2022-01-20
Improve Stopping behavior for Fiesta release

saina-ramyar:
# PR Details
## Description
During verification of Fiesta release, it was noticed that the deceleration of the vehicle to stop at the intersection is very jerky. The parameters of involved plugins are tuned to minimize jerkiness and also ensure the vehicle stops.

## How Has This Been Tested?
The changes are tested on Blue Lexus and Ford Fusion.

## Types of changes

## Checklist:

### DIFF
--- a/arbitrator/config/arbitrator_params.yaml
+++ b/arbitrator/config/arbitrator_params.yaml
@@ -26,4 +26,4 @@ use_fixed_costs: true
 # Map: The priorities associated with each plugin during the planning 
 # process, values will be normalized at runtime and inverted into costs
 # Unit: N/a
-plugin_priorities: {SCIStrategicPlugin: 3.0, RouteFollowingPlugin: 2.0, WorkZonePlugin: 4.0, PlatooningStrategicPlugin: 1.0}
+plugin_priorities: {SCIStrategicPlugin: 4.0, RouteFollowingPlugin: 2.0, WorkZonePlugin: 4.0, PlatooningStrategicPlugin: 1.0}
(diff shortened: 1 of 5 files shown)


arbitrator/config/arbitrator_params.yaml
```diff
@@ -26,4 +26,4 @@ use_fixed_costs: true
 # Map: The priorities associated with each plugin during the planning 
 # process, values will be normalized at runtime and inverted into costs
 # Unit: N/a
-plugin_priorities: {SCIStrategicPlugin: 3.0, RouteFollowingPlugin: 2.0, WorkZonePlugin: 4.0, PlatooningStrategicPlugin: 1.0}
+plugin_priorities: {SCIStrategicPlugin: 4.0, RouteFollowingPlugin: 2.0, WorkZonePlugin: 4.0, PlatooningStrategicPlugin: 1.0}
```
msmcconnell, line 29: Maybe its not causing an issue but SCIStrategicPlugin and WorkZonePlugin both have 4.0. Maybe setting SCIStrategicPlugin to 5.0 to ensure priority would be preferable?

sci_strategic_plugin/src/sci_strategic_plugin.cpp
```diff
@@ -347,10 +347,10 @@ bool SCIStrategicPlugin::planManeuverCb(cav_srvs::PlanManeuversRequest& req, cav
       double desired_deceleration = config_.vehicle_decel_limit * config_.vehicle_decel_limit_multiplier;
 
 
-      double safe_distance = pow(current_state.speed, 2)/(2*desired_deceleration) + config_.min_gap;
+      double safe_distance = pow(current_state.speed, 2)/(2*desired_deceleration);
```
msmcconnell, line 350: Is the min_gap parameter no longer used? Should it be removed from the node?

arbitrator/config/arbitrator_params.yaml
saina-ramyar: I think since the geofence and spat msgs are not received workzone isn't activated. Workzone wasn't causing any problems, but SCI weight was close to RouteFollowing, that's why I increased it.

sci_strategic_plugin/src/sci_strategic_plugin.cpp
saina-ramyar: it is still used by the status and intent msg sent to carma streets. It was also used as extra space for the safe distance, but that is not needed anymore.

PR #1594, opened 2022-01-25
Feature/platoon lane negotiation

JonSmet:
# PR Details
## Description

This PR adds the communication of a target lane index from a front vehicle to a rear vehicle when conducting an initial platoon negotiation. 

If a front vehicle is in the rightmost lane with multiple lanes of the current travel direction, and it receives a 'JOIN_PLATOON_AT_REAR' MobilityRequest from a rear vehicle, the front vehicle will set an internal flag indicating that a lane change into the left adjacent lane is required. After performing the lane change, the front vehicle will respond 'ACK' to the rear vehicle, and will broadcast a 'JOIN_REQUIREMENTS' MobilityOperation messages to that rear vehicle, indicating the target lane index in which platooning will be performed. If the rear vehicle is not currently located in that target lane index, an internal flag will be set to indicate that a lane change into the target lane index is required. Once the rear vehicle (a CandidateFollower) is positioned in the target lane index provided by the target leader, it will broadcast a 'PLATOON_FOLLOWER_JOIN' MobilityRequest to the target leader.

Note 1: Lane index is defined as 0 being the rightmost lane of the current travel direction; 1 = second rightmost, etc.
Note 2: This PR does not include changes to plan a lane change for either the front or the rear vehicle, it only includes updates to the communication between the two vehicles.

## Motivation and Context

An initial platoon cannot be formed in the rightmost lane when there are multiple lanes of the current travel direction. This change enables the communication of a target lane index between two vehicles forming a platoon, so that both vehicles can plan maneuvers and ultimately form a platoon in a suitable lane.

## How Has This Been Tested?

Unit Tested

## Types of changes

## Checklist:


platooning_strategic/src/platoon_strategic.cpp
```diff
@@ -169,6 +169,9 @@ namespace platoon_strategic
             single_lane_road_ = false;
             ROS_DEBUG_STREAM("Vehicle is NOT in a single-lane road");
         }
+
+        // Store the current local lane index of the host vehicle (0 is rightmost lane, 1 is second rightmost, etc.); considers only the current travel direction
+        current_lane_index_ = (routing_graph->rights(current_lanelet[0])).size();
```
kjrush, line 174: Seems a bit tacked on here, especially considering that it's a bit strange for a function named `check...` or `get...` to be updating object state under the hood. I could imagine that when making edits to this class later someone might not ensure that `checkForRightMostLane()` is called before the message publication to ensure that the value of this variable is set correctly. 

It's here though because all the needed data elements are already present in this function right? Would it be a big lift to extract this to a separate function `getCurrentLaneIndex()` or something?

platooning_strategic/include/platoon_strategic.h
```diff
@@ -512,12 +530,11 @@ namespace platoon_strategic
             const std::string MOBILITY_STRATEGY = "Carma/Platooning";
             const std::string OPERATION_INFO_TYPE = "INFO";
             const std::string OPERATION_STATUS_TYPE = "STATUS";
+            const std::string JOIN_REQUIREMENTS_TYPE = "JOIN_REQUIREMENTS";
```
kjrush, line 533: Is this another `MobilityOperation` sub-type? Let's label it as `OPERATION_JOIN_REQUIREMENTS_TYPE` for consistency with the elements above it if so.

platooning_strategic/src/platoon_strategic.cpp
```diff
@@ -696,7 +739,15 @@ namespace platoon_strategic
                     waitingStartTime = ros::Time::now().toNSec()/1000000;
                     lw_applicantId_ = msg.header.sender_id;
                     return MobilityRequestResponse::ACK;
-                } else {
+                } 
+                else if(isDistanceCloseEnough && !laneConditionsSatisfied) {
```
kjrush, line 743: Minor formatting quibble, let's move this else-if up to the line with the brace `} else if ... {`

platooning_strategic/src/platoon_strategic.cpp
```diff
@@ -974,6 +1028,27 @@ namespace platoon_strategic
...
+            boost::algorithm::split(inputsParams, strategyParams, boost::is_any_of(","));
+
+            std::vector<std::string> target_lane_index_parsed;
+            boost::algorithm::split(target_lane_index_parsed, inputsParams[0], boost::is_any_of(":"));
+            int target_lane_index = std::stoi(target_lane_index_parsed[1]);
+
+            ROS_DEBUG_STREAM("Received JOIN_REQUIREMENTS MobilityOperation with target lane index: " << target_lane_index);
+
+            // Store the target CandidateFollower platoon lane index provided by the target Leader
+            cf_target_lane_index_ = target_lane_index;
```
kjrush, line 1045: We're modifying state here, and then referencing it (potentially at a different timestep) above. Is there a possibility for issues with multiple negotiations going on simultaneously? Either multiple leaders negotiating with a single follower or multiple followers negotiating with the same leader? If so, it'd be good to add some kind of lock - almost like a mutex - where we record the ID of the vehicle we're negotiating with (if any) and don't process incoming messages from any other ID until that negotiation is completed (or maybe a timeout).

Does that sound like a realistic concern as we approach supporting 3+ vehicle platoons?

platooning_strategic/src/platoon_strategic.cpp
```diff
@@ -974,6 +1028,27 @@ namespace platoon_strategic
...
+
+            // Set flag to indicate CandidateFollower must change lanes if it is not currently in the target lane platoon lane index
+            if (current_lane_index_ != cf_target_lane_index_) {
```
kjrush, line 1048: I don't have a perfect understanding of the lanelet 2 library, but is it possible (if the two negotiating vehicles are on different lanelets for example, where the number of lanes might change) for an index based comparison to have issues?

platooning_strategic/src/platoon_strategic.cpp
JonSmet: Agreed, I've gone ahead and extracted this into its own function `getCurrentLaneIndex()`.

platooning_strategic/include/platoon_strategic.h
JonSmet: Fixed.

platooning_strategic/src/platoon_strategic.cpp
JonSmet: Fixed.

kjrush (approved):

platooning_strategic/src/platoon_strategic.cpp
JonSmet: I did find that multiple negotiations could occur simultaneously for both scenarios: (1) a leader negotiating with multiple followers, and (2) a follower receiving JOIN_REQUIREMENTS from multiple leaders. As you've suggested, I've provided checks to ensure that the full negotiation only occurs is one other vehicle (based on its ID).

platooning_strategic/src/platoon_strategic.cpp
JonSmet: It is possible for an issue to arise in situations where the two negotiating vehicles are on different lanelets, with both lanelets belonging to a lane group of nonequal sizes. I've added a note in the code to document this, and also added a WARN log statement in order to capture when a rear vehicle receives a target lane index from a front vehicle that is located in a different sized lane group than the rear vehicle's current lane group.

Issue #1608, opened 2022-02-08
The Web UI does not notify the user when the vehicle has left the route


JonSmet:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

When conducting CARMA System 3.10.0 regression/verification testing on trucks at ATEF, it was discovered that the Web UI does not notify the user when the vehicle has left the route. For this testing, the system was engaged on a route, and then after manually overriding the system, the vehicle was driven ~50 meters off of the lanelet map without a popup occurring.

### Carma version where this issue was discovered

CARMA System 3.10.0 on multiple trucks at ATEF

### Expected behavior
If the user engages the system, and then manually overrides the system and drives the vehicle off of the route (Note: the route can include more than just the lanelets in the shortest path), then a popup will be displayed on the Web UI to notify the user that the vehicle has left the route.

### Actual behavior

When the user engages the system, and then manually overrides the system and drives the vehicle off of the route (in this test case, the vehicle was driven completely off of the lanelet map), no popup is ever displayed on the Web UI to indicate that the vehicle has left the route.

### Steps to reproduce the actual behavior

1. Engage the CARMA System on a route
2. After a few seconds, manually override the system and drive the vehicle completely off of the lanelet map
3. Check the Web UI to see whether a popup is displayed that notifies the user that the vehicle has left the route.

JonSmet:
Closed by PR #1615

PR #1615, opened 2022-02-10
Fix name for Route package's 'cte_max_count' parameter

JonSmet:
# PR Details
## Description

This PR fixes a mis-match between the 'cte_max_count' parameter in the Route package's configuration file, and the parameter name used to access it in the package's route.cpp file. With this change, the Route package now uses the config file parameter value, rather than the value of an uninitialized integer.

## Related Issue

Issue #1608 

## How Has This Been Tested?

Tested on the Red Truck at ATEF.

## Types of changes

## Checklist:


msmcconnell:
Wow good catch. The change looks good to me, but I will wait to merge till I'm done with getting the ROS2 changes into develop which should be by the end of today

PR #1649, opened 2022-03-04
Increase buffer ending downtracks for lane change plugins

JonSmet:
# PR Details
## Description

Since plan_delegator now decreases the starting and ending downtracks of each maneuver by the distance between base_link and vehicle_front (PR #1565), lane change maneuvers are only planned up until <distance-between-base_link-and-vehicle-front> meters before the end of the target lane change lanelet. On the trucks, this distance is roughly 7 meters, so the original buffer_ending_downtrack values of 5 meters caused no buffer points to be added to the end of the lane change, resulting in an exception near the end of the lane change since a spline could not be fit to the lane change points.

## How Has This Been Tested?

Tested on the White Truck at ATEF.

## Types of changes

## Checklist:

### DIFF
--- a/cooperative_lanechange/config/parameters.yaml
+++ b/cooperative_lanechange/config/parameters.yaml
@@ -97,4 +97,4 @@ curve_resample_step_size: 1.0
 back_distance: 0.0
 
 #Double: 
-buffer_ending_downtrack: 5.0
+buffer_ending_downtrack: 15.0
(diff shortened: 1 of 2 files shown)


msmcconnell (approved):

msmcconnell:
Per discussion on slack. There is a different root cause which should be addressed instead of this fix which resolves only the symptom.

PR #1647, opened 2022-03-03
Update baseline and recompiled cut_in_mid_rear

xuhan417:
# PR Details
## Description
This is UCLA team's commit of the IHP2 platoon plugin cut-in join from middle and rear.

## Related Issue
There is no related issues.

## Motivation and Context
This is the development for IHP2 project cut-in join from middle and rear.

PR context: 
- Mid/rear join functions.
- Include "join index" in STATUS params, and keep tracking it as a property of platoonmember class.
- Update plan type name and platoon manager APF function.

## How Has This Been Tested?
- Add unit test to test additional state transitions.
- Compile and ran on UCLA's vehicle platform (single vehicle only).

## Types of changes

- [ ] Defect fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that cause existing functionality to change)

## Checklist:

- [x] I have added any new packages to the sonar-scanner.properties file
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
[CARMA Contributing Guide](https://github.com/usdot-fhwa-stol/carma-platform/blob/develop/Contributing.md) 
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1656, opened 2022-03-08
Platooning control plugin handles control switchover

TonysCousin:
# PR Details
## Description
Enhanced the platooning control plugin to cooperate with other control plugins to accommodate the fact that all control plugins publish their output on a common ros topic, and this topic is only intended to carry data from a single controller at a time.  When trajectory executor switches from one control plugin to another, we want the old one to stop publishing.  However, it needs to wait some amount of time since it is spinning faster than the producer of its input trajectory data. Since this input data is not guaranteed to arrive at a precise interval, the control plugin needs to wait for a conservative timeout period before deciding that the input stream has dried up.

Note: architecture documentation is to be updated at a later time.

## Related Issue

PR #1647 

## Motivation and Context

Ensures a control plugin stops publishing output data when another control plugin takes over.

## How Has This Been Tested?

Not yet tested - this PR is for integration testing.

## Types of changes

## Checklist:

### DIFF
--- a/platooning_control/src/platoon_control.cpp
+++ b/platooning_control/src/platoon_control.cpp
@@ -106,18 +106,21 @@ namespace platoon_control
         // If it has been a long time since input data has arrived then reset the input counter and return
         // Note: this quiets the controller after its input stream stops, which is necessary to allow 
         // the replacement controller to publish on the same output topic after this one is done.
-        long current_time = ros::Time::now().toNsec() / 1000000;
+        long current_time = ros::Time::now().toNSec() / 1000000;
+        ROS_DEBUG_STREAM("current_time = " << current_time << ", prev_input_time_ = " << prev_input_time_ << ", input counter = " << consecutive_input_counter_);
         if (current_time - prev_input_time_ > config_.shutdownTimeout)
         {
+            ROS_DEBUG_STREAM("returning due to timeout.");
             consecutive_input_counter_ = 0;
-            return;
+            return false;
         }
 
         // If there have not been enough consecutive timely inputs then return (waiting for
         // previous control plugin to time out and stop publishing, since it uses same output topic)
         if (consecutive_input_counter_ <= config_.ignoreInitialInputs)
         {
-            return;
+            ROS_DEBUG_STREAM("returning due to first data input");
+            return false;
         }
 
         cav_msgs::TrajectoryPlanPoint second_trajectory_point = latest_trajectory_.trajectory_points[1]; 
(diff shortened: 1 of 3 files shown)


platooning_control/src/platoon_control.cpp
```diff
@@ -99,6 +103,22 @@ namespace platoon_control
 
     bool PlatoonControlPlugin::controlTimerCb()
     {
+        // If it has been a long time since input data has arrived then reset the input counter and return
+        // Note: this quiets the controller after its input stream stops, which is necessary to allow 
+        // the replacement controller to publish on the same output topic after this one is done.
+        long current_time = ros::Time::now().toNsec() / 1000000;
+        if (current_time - prev_input_time_ > config_.shutdownTimeout)
```
saina-ramyar, line 110: I suggest we add some log statements to see what is value of "current_time - prev_input_time_" so we can tune the parameter

platooning_control/src/platoon_control.cpp
```diff
@@ -99,6 +103,22 @@ namespace platoon_control
...
+        {
+            consecutive_input_counter_ = 0;
+            return;
+        }
+
+        // If there have not been enough consecutive timely inputs then return (waiting for
+        // previous control plugin to time out and stop publishing, since it uses same output topic)
+        if (consecutive_input_counter_ <= config_.ignoreInitialInputs)
+        {
+            return;
```
saina-ramyar, line 120: again, a log statement could be useful

platooning_control/src/platoon_control.cpp
```diff
@@ -109,17 +129,17 @@ namespace platoon_control
...
-        
+    void  PlatoonControlPlugin::trajectoryPlan_cb(const cav_msgs::TrajectoryPlan::ConstPtr& tp)
+    {
         if (tp->trajectory_points.size() < 2) {
             ROS_WARN_STREAM("PlatoonControlPlugin cannot execute trajectory as only 1 point was provided");
             return;
         }
 
         latest_trajectory_ = *tp;
+        prev_input_time_ = ros::Time::now().toNsec() / 1000000;
```
saina-ramyar, line 141: again Log statemet saying not received a new trajectory plan
Also, can you add log statements printing "prev_input_time_" and "tp->header.stamp.toNsec() / 1000000"
I think its better to use the trajectory plan's own timestamp, but I remember those didn't have reliable values in the past, so want to make sure.

platooning_control/src/platoon_control.cpp
TonysCousin: Good idea.

platooning_control/src/platoon_control.cpp
TonysCousin: Yep

platooning_control/src/platoon_control.cpp
TonysCousin: Sure.

Issue #1433, opened 2021-09-10
3rd vehicle in platoon may not be using correct target gap to leader


kjrush:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Steps to reproduce the actual behavior

1. Do this
1. Then do this...

PR #1657, opened 2022-03-09
Gap calculation for multi-vehicle platoon

saina-ramyar:
# PR Details
## Description
The gap calculation in Platooning Strategic Plugin is updated to be able to handle platoons with more than two vehicles.

## Related Issue
#1433 

## How Has This Been Tested?
Unit test

## Types of changes

## Checklist:

### DIFF
--- a/platooning_strategic/src/platoon_manager.cpp
+++ b/platooning_strategic/src/platoon_manager.cpp
@@ -76,6 +76,7 @@ namespace platoon_strategic
                 updatesOrAddMemberInfo(senderId, senderBsmId, cmdSpeed, dtDistance, curSpeed);
                 leaderID = (platoon.size()==0) ? HostMobilityId : platoon[0].staticId;
                 ROS_DEBUG_STREAM("The first vehicle in our list is now " << leaderID);
+                
             } 
             else
             {
@@ -133,7 +134,7 @@ namespace platoon_strategic
 
     int PlatoonManager::getTotalPlatooningSize() const{
         if(isFollower) {
-            return platoonSize;
+            return std::max((int)(platoon.size() + 1), platoonSize);
         }
         return platoon.size() + 1;
     }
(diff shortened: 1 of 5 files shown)


platooning_strategic/src/platoon_manager.cpp
```diff
@@ -133,7 +134,7 @@ namespace platoon_strategic
 
     int PlatoonManager::getTotalPlatooningSize() const{
         if(isFollower) {
-            return platoonSize;
+            return std::max((int)(platoon.size() + 1), platoonSize);
```
kjrush, line 137: I'm not sure I follow what this is accomplishing, platoonSize here is the number of vehicles in the platoon? And platoon is the list of those vehicles? I'm not sure i understand how they would be a different number.

More minor: Is the (int) cast needed simply because std::max can't work on a size_t (from platoon.size()) and an int?

platooning_strategic/src/platoon_strategic.cpp
```diff
@@ -842,7 +842,12 @@ namespace platoon_strategic
                 status_msg.leader_cmd_speed = platoon_leader.commandSpeed;
                 status_msg.host_platoon_position = pm_.getNumberOfVehicleInFront();
 
-                status_msg.desired_gap = std::max(config_.standStillHeadway, config_.timeHeadway *  current_speed_);
+                int numOfVehiclesGaps = pm_.getNumberOfVehicleInFront() - pm_.dynamic_leader_index_;
```
kjrush, line 845: Just to make sure I understand the math, this `dynamic_leader_index_` is a 0-based index right?

platooning_strategic/src/platoon_manager.cpp
saina-ramyar: platoonSize is the size of the full platoon. However, platoon vector includes only the vehicles in front of the host vehicle. Therefore, unless the host is the last follower, the platoon.size() will not show the correct platoon size.

kjrush (approved):

PR #1671, opened 2022-03-17
Fix lane changes spanning multiple lanelets

JonSmet:
# PR Details
## Description

This PR includes updates to plan_delegator and basic_autonomy (note: only lanechange-related functions in basic_autonomy have been updated) to address three issues with trajectory planning for lane changes:

- Carma-platform PR #1565 updated plan_delegator to shift each maneuver's start_dist and end_dist parameters forward by the distance between 'base_link' and 'vehicle_front'. This PR includes updates to also adjust the lanelet-specific parameter(s) associated with each maneuver based on these adjusted start_dist and end_dist locations. Specifically, the `lane_ids` parameter for lane following maneuvers is updated, as well as the `starting_lane_id` and `ending_lane_id` parameters for all other maneuver types.
- As a result of the previously mentioned carma-platform PR #1565, lane change maneuvers can now end several meters before the end of the ending_lane_id; however, basic_autonomy only adds buffer points from the next lane following lanelet. This PR ensures that the rest of the ending_lane_id's centerline points are added as buffer points as well.
- This PR also includes an update to basic_autonomy's create_lanechange_geometry() so that the entire target lane's centerline is obtained between the lane change's start_dist and end_dist when the lane change spans multiple lanelets. Previously, the function only sampled the target lane's centerline from start_dist until the downtrack of the end of the first lanelet included in the lane change.

## How Has This Been Tested?

Tested on the Black Pacifica at TFHRC.

## Types of changes

## Checklist:


basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -397,7 +400,10 @@ namespace basic_autonomy
                 centerline_points.push_back(current_position);
             }
 
-            //Add points from following lanelet to provide sufficient distance for adding buffer
+            // Add points from the remaining length of the target lanelet to provide sufficient distance for adding buffer
+            centerline_points.insert(centerline_points.end(), downsampled_target_centerline.begin() + end_index_target_centerline, downsampled_target_centerline.end());
+            
+            // Also add points from the lanelet following the target lanelet to provide sufficient distance for adding buffer
             auto following_lanelets = wm->getMapRoutingGraph()->following(ending_lanelet, false);
```
adev4a, line 407: Seems like this additional buffer would not be needed if we already have the required length from line 404.
I guess its not entirely required to have this check, since the buffer would be removed anyway but we can probably avoid some cost with looking up a lanelet and extracting points in cases we dont need them

plan_delegator/src/plan_delegator.cpp
```diff
@@ -20,6 +20,42 @@
...
+    namespace 
+    {
+        /**
+         * \brief Anonymous function to set the lanelet ids for all maneuver types except lane following. These
+         * maneuver parameters cannot be set with SET_MANEUVER_PROPERTY calls since they are not included in
+         * LANE_FOLLOW maneuvers.
+         */ 
+        void setManeuverLaneletIds(cav_msgs::Maneuver& mvr, lanelet::Id start_id, lanelet::Id end_id) {
+
+            switch(mvr.type) {
```
adev4a, line 33: We can replace this loop with a macro
 https://github.com/usdot-fhwa-stol/carma-platform/blob/23a79861f1e6b22f02a98b574f5f745db38e73a2/route_following_plugin/include/route_following_plugin.h#L48

JonSmet: This macro only works for properties that are shared by all maneuver types. Since only lane_following maneuvers have the `lane_ids` parameter, and all other maneuver types have `starting_lane_id` and `ending_lane_id` parameters, these parameters require separate get/set methods. A similar function is included in route_following_plugin [here.](https://github.com/usdot-fhwa-stol/carma-platform/blob/23a79861f1e6b22f02a98b574f5f745db38e73a2/route_following_plugin/src/route_following_plugin.cpp#L55)

plan_delegator/src/plan_delegator.cpp
```diff
@@ -167,6 +275,84 @@ namespace plan_delegator
...
+            lanelet::ConstLanelet original_starting_lanelet = wm_->getMap()->laneletLayer.get(original_starting_lanelet_id);
+
+            // Check whether the lanelet prior to the original starting lanelet is crossed by the updated maneuver
+            for(auto lanelet : adjusted_crossed_lanelets) {
+                auto routing_relation = wm_->getMapRoutingGraph()->routingRelation(lanelet, original_starting_lanelet);
+
+                // Lanelet prior to the original starting lanelet is crossed by the updated maneuver, so it is added to the beginning of lane_ids
+                if (routing_relation == lanelet::routing::RelationType::Successor) {
+                    ROS_DEBUG_STREAM("Lanelet " << lanelet.id() << " added to front of maneuver's lane_ids");
+                    maneuver.lane_following_maneuver.lane_ids.insert(maneuver.lane_following_maneuver.lane_ids.begin(), std::to_string(lanelet.id()));
```
msmcconnell, line 304: I think as a final step the lane_ids need to be sorted by downtrack distance. 
It doesn't specify it in the message spec but I think somewhere I've seen an assumption that this list is ordered.

msmcconnell (changes requested):

plan_delegator/src/plan_delegator.cpp
JonSmet: It is assumed that strategic plugins populate the lane_ids array with lanelet ids in order of increasing downtrack distance, so the first element corresponds to the starting lanelet, and the last element corresponds to the ending lanelet. I have added comments to the code to state this.

msmcconnell (approved):

Issue #1677, opened 2022-03-23
Compiler warnings in carma_wm


TonysCousin:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

During routine compiles of carma_wm there are lots of warnings generated.  Many of the problems identified are innocuous (e.g. extra semi-colon), but they clutter the output making real problems difficult to see.  As a result, real problem are being overlooked.  Some of these warnings point out legitimate code defects that need to be fixed.

### Carma version where this issue was discovered

Develop after version 3.11.0

### Expected behavior
Code compiles with no warning or info messages.

### Actual behavior

A large number of warnings appear during compile.

### Steps to reproduce the actual behavior

1. Run the build script for carma-platform

msmcconnell:
Resolved by above PR

PR #1678, opened 2022-03-23
Fix/compile defects1

TonysCousin:
# PR Details
## Description

Addressed all compiler warning messages in carma_wm.  I was able to make fixes directly, for those I was confident wouldn't break some functionality.  In other places, I simply left a TODO comment for someone more knowledgeable about carma_wm to figure out the best path forward.  Some of these problems pose legitimate defects for functionality or correct test results.

## Related Issue

#1677 

## Motivation and Context

Need to clean up compiler outputs!

## How Has This Been Tested?

Build carma_wm.

## Types of changes

## Checklist:


carma_wm/include/carma_wm/WMTestLibForGuidance.h
```diff
@@ -323,8 +323,11 @@ inline void addObstacle(double x, double y, std::shared_ptr<carma_wm::CARMAWorld
  * \param length length of roadway object, default 3 meters
  * NOTE: This assumes a similar simple shape of the GuidanceTestMap and does not populate cartesian components of the roadway object.
  */
-inline void addObstacle(carma_wm::TrackPos tp, lanelet::Id lanelet_id, std::shared_ptr<carma_wm::CARMAWorldModel> cmw, std::vector<carma_wm::TrackPos> pred_trackpos_list = {}, int time_step = 100, double width = 3, double length = 3)
+inline void addObstacle(carma_wm::TrackPos tp, lanelet::Id lanelet_id, std::shared_ptr<carma_wm::CARMAWorldModel> cmw, 
+                        std::vector<carma_wm::TrackPos> pred_trackpos_list = {}, int time_step = 100, double width = 3, double length = 3)
 {
+  //TODO: width & length are not used; if there are no plans to use them soon, remove them from param list
```
msmcconnell, line 329: These are definitely meant to be used to allow for different sized objects in the test cases. So actually seems like a bug to have them unused

carma_wm/src/SignalizedIntersectionManager.cpp
```diff
@@ -191,7 +191,10 @@ namespace carma_wm
     }
   }
 
-  lanelet::Id SignalizedIntersectionManager::matchSignalizedIntersection(const lanelet::Lanelets& entry_llts, const lanelet::Lanelets& exit_llts, const std::shared_ptr<lanelet::LaneletMap>& map)
+  lanelet::Id SignalizedIntersectionManager::matchSignalizedIntersection(const lanelet::Lanelets& entry_llts, const lanelet::Lanelets& exit_llts, 
+                                                                        const std::shared_ptr<lanelet::LaneletMap>& map)
+  //TODO: if map is not going to be needed, remove it from the param list
```
msmcconnell, line 196: Seems like it can be removed

carma_wm/src/SignalizedIntersectionManager.cpp
```diff
@@ -381,4 +384,12 @@ namespace carma_wm
 
     return *this;
   }
+
+  SignalizedIntersectionManager::SignalizedIntersectionManager(const SignalizedIntersectionManager& other)
```
msmcconnell, line 388: Out of curiosity was there a place where the copy constructor was being used?

carma_wm/src/collision_detection.cpp
```diff
@@ -22,13 +22,15 @@ namespace carma_wm {
 
     namespace collision_detection {
 
-        std::vector<cav_msgs::RoadwayObstacle> WorldCollisionDetection(const cav_msgs::RoadwayObstacleList& rwol, const cav_msgs::TrajectoryPlan& tp, const geometry_msgs::Vector3& size, const geometry_msgs::Twist& veloctiy, const __uint64_t target_time) {
+        std::vector<cav_msgs::RoadwayObstacle> WorldCollisionDetection(const cav_msgs::RoadwayObstacleList& rwol, const cav_msgs::TrajectoryPlan& tp, 
+                                                                        const geometry_msgs::Vector3& size, const geometry_msgs::Twist& velocity, 
+                                                                        const __uint64_t target_time) {
 
-            ROS_DEBUG_STREAM("WorldCollisionDetection");
+            ROS_DEBUG_STREAM("WorldCollisionDetection at target_time: " << target_time); //TODO: if target_time won't be needed, remove it from param list
```
msmcconnell, line 29: Based on the function comments it seems like target_time is meant to let you sample only a portion [0, target_time] of the provided trajectory for collision which is a reasonable optimization to have, but this is a pretty inefficient function already as it has 
object_count * prediction_count * trajectory_point
square root calls. 

Adjusting target time would reduce only 1 of those 3 terms.

Probably could be removed and the function slated for rework.

carma_wm/src/collision_detection.cpp
```diff
@@ -268,25 +270,25 @@ namespace carma_wm {
...
 
             polygon_t union_polygon;  
-            boost::geometry::assign_points(union_polygon, unioin_future_polygon_points);
+            boost::geometry::assign_points(union_polygon, union_future_polygon_points);
 
             polygon_t hull_polygon;
             boost::geometry::convex_hull(union_polygon, hull_polygon);
 
-            collision_detection::MovingObject output_object = {hull_polygon, op.linear_velocity};
+            collision_detection::MovingObject output_object = {hull_polygon, op.linear_velocity}; //TODO: fix missing initializer for fp element
```
msmcconnell, line 288: Not quire sure what this TODO comment is referring too

carma_wm/test/CARMAWorldModelTest.cpp
```diff
@@ -39,6 +39,9 @@ namespace carma_wm
 void createTestingWorld(carma_wm::CARMAWorldModel& cmw, std::vector<lanelet::Lanelet>& llts,
                                           lanelet::LaneletMapPtr& map,  std::vector<cav_msgs::ExternalObject>& obstacles)
 {
+  //TODO: cmw is not used; if there are no plans to use it soon, remove it from param list
+  //TODO: map is overwritten before it is used, so it has no value as an input param; is it intended to be an output?
```
msmcconnell, line 43: I think all 4 parameters are output parameters here which is why all of them are non-const references. 
cmw missing seems like an oversight. If its not used it can be removed.

carma_wm/test/SignalizedIntersectionManagerTest.cpp
```diff
@@ -69,7 +69,12 @@ TEST(SignalizedIntersectionManger, convertLaneToLaneletId)
   intersection.id.id = 9001;
 
   cav_msgs::GenericLane lane;
-  lane.lane_id = 1210;
+  lane.lane_id = 1210;        //TODO: broken! implicit conversion from int to unsigned char makes this value 186. Is lane_id the wrong type???
```
msmcconnell, line 72: Since this is a test case, this isn't actually broken since the test scenario still seems self consistent. 
Not sure what the best comment would be here.

msmcconnell (changes requested):

carma_wm/include/carma_wm/WMTestLibForGuidance.h
TonysCousin: Added log.

carma_wm/src/SignalizedIntersectionManager.cpp
TonysCousin: Removed.

carma_wm/src/SignalizedIntersectionManager.cpp
TonysCousin: Yes, the compiler complained that the implicit copy constructor may have been doing something undesirable.  Let me check what that was.

carma_wm/src/collision_detection.cpp
TonysCousin: Removed & added TODO reminder to consider refactoring.

carma_wm/src/collision_detection.cpp
TonysCousin: The MovingObject contains 3 elements, and only two are initialized here.  fp also needs to be initialized.  This is a complex object itself, without any documentation, so I'm not sure what its purpose is.

carma_wm/test/SignalizedIntersectionManagerTest.cpp
TonysCousin: I guess you have a point.  However, there are comment "diagrams" throughout this code using these IDs as well, and the overall effect is to communicate to other developers that these are legitimate lanelet IDs.  When operational code chokes on such values, people could spend a long time wondering why, when the unit tests pass.  My bigger concern is that if lanelet IDs really do take on values like this then our GenericLane message type (and maybe others?) is inadequate to represent them.  I've changed all the values to be in the 170-200 range so they are legal.  If you decide we do need to accommodate values > 256, then it would be wise to put large values back into at least one of these tests.

carma_wm/test/CARMAWorldModelTest.cpp
TonysCousin: cmw removed.

carma_wm/src/collision_detection.cpp
TonysCousin: I tracked down the meaning of fp (and added description!) and made a few changes to fix this warning in several places.

msmcconnell (approved):

PR #1682, opened 2022-03-24
CMV broadcast acknowledgement after processing the incoming TCM

dan-du-car:
# PR Details
As an automated CMV, I will send a negative acknowledgment to CARMA Streets, upon receiving a geofence acknowledgement request from it, but I’m not able to process the geofence message.

As an automated CMV, I will send a positive acknowledgment to CARMA Streets, when I receive an geofence acknowledgement request from it, and after I process the geofence message .

## Description
CMV broadcast acknowledgement after processing the incoming TCM

## Related Issue
NA

## Motivation and Context
Freight Workzone

## How Has This Been Tested?
NA

## Types of changes

## Checklist:

### DIFF
--- a/carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcaster.h
+++ b/carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcaster.h
@@ -461,8 +461,10 @@ private:
   size_t update_count_ = 0; // Records the total number of sent map updates. Used as the set value for update.seq_id
 
   carma_wm::SignalizedIntersectionManager sim_;
-  const uint8_t acknowledgement_status__acknowledged = 1;
-  const uint8_t acknowledgement_status__rejected = 2;
+  enum class AcknowledgementStatus {
+    ACKNOWLEDGED = 1,
+    REJECTED = 2
+  };
   const std::string geofence_ack_strategy_ = "carma3/Geofence_Acknowledgement";
   int ack_pub_times_ = 1;
   std::string vehicle_id_;
--- a/carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcaster.h
+++ b/carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcaster.h
@@ -391,13 +391,13 @@ public:
   void setConfigVehicleId(std::string vehicle_id);
 
    /*!
-   * \brief Sets the TCM Aknowledgement publish times. 
-   @param ack_pub_times the number of times it publishes TCM Aknowledgement 
+   * \brief Sets the TCM Acknowledgement publish times. 
+   @param ack_pub_times the number of times it publishes TCM Acknowledgement 
    */
   void setConfigACKPubTimes(int ack_pub_times);
 
   /*!
-  * \brief Construct TCM aknowledgement object and populate it with params. Publish the object for a configured number of times.
+  * \brief Construct TCM acknowledgement object and populate it with params. Publish the object for a configured number of times.
   */
   void pubTCMACK(j2735_msgs::Id64b tcm_req_id, uint16_t msgnum, uint8_t ack_status, const std::string ack_reason);
 
--- a/carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcaster.h
+++ b/carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcaster.h
@@ -399,7 +399,7 @@ public:
   /*!
   * \brief Construct TCM acknowledgement object and populate it with params. Publish the object for a configured number of times.
   */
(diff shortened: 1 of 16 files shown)


carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcaster.h
```diff
@@ -440,6 +461,11 @@ class WMBroadcaster
   size_t update_count_ = 0; // Records the total number of sent map updates. Used as the set value for update.seq_id
 
   carma_wm::SignalizedIntersectionManager sim_;
+  const uint8_t acknowledgement_status__acknowledged = 1;
+  const uint8_t acknowledgement_status__rejected = 2;
```
msmcconnell, line 465: Are these two const effectively an enum?
Can it be reworked to ?
```
enum class AcknowledgementStatus {
  ACKNOWLEDGED=1
  REJECTED=2
}
```
Just good practice for this kind of thing

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -1046,16 +1046,22 @@ void WMBroadcaster::geofenceCallback(const cav_msgs::TrafficControlMessage& geof
 {
   
   std::lock_guard<std::mutex> guard(map_mutex_);
+  std::stringstream reason_ss;
   // quickly check if the id has been added
   if (geofence_msg.choice != cav_msgs::TrafficControlMessage::TCMV01) {
-    ROS_WARN_STREAM("Dropping received geofence for unsupported TrafficControl version: " << geofence_msg.choice);
+    reason_ss << "Dropping received geofence for unsupported TrafficControl version: " << geofence_msg.choice;
+    ROS_WARN_STREAM(reason_ss.str());
```
msmcconnell, line 1053: I could be wrong but I think since you are using the _STREAM logs you can just pass in the stream object directly
```
ROS_WARN_STREAM(reason_ss);
```
Which avoids the extra string conversion

carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcaster.h
```diff
@@ -381,6 +383,24 @@ class WMBroadcaster
...
+   /*!
+   * \brief Sets the TCM Acknowledgement publish times. 
+   @param ack_pub_times the number of times it publishes TCM Acknowledgement 
+   */
+  void setConfigACKPubTimes(int ack_pub_times);
+
+  /*!
+  * \brief Construct TCM acknowledgement object and populate it with params. Publish the object for a configured number of times.
+  */
+  void pubTCMACK(j2735_msgs::Id64b tcm_req_id, uint16_t msgnum, uint8_t ack_status, const std::string ack_reason);
```
msmcconnell, line 402: string should be passed as const reference to prevent unneeded copy
const std::string& ack_reason

carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcaster.h
```diff
@@ -381,6 +383,24 @@ class WMBroadcaster
   /*! \brief helps to populate upcoming_intersection_ids_ from local traffic lanelet ids
    */
   void publishLightId();
+  
+  /*!
+   * \brief Retrieve the vehicle ID from global vehicle parameters, and set instance memeber vehicle id
+   @param vehicle_id Vehicle ID from UniqueVehicleParams.yaml 
+   */
+  void setConfigVehicleId(std::string vehicle_id);
```
msmcconnell, line 391: string should be passed as const reference to prevent unneeded copy
const std::string& vehicle_id

carma_wm_ctrl/src/WMBroadcasterNode.cpp
```diff
@@ -62,6 +66,8 @@ int WMBroadcasterNode::run()
   control_msg_pub_= cnh_.advertise<cav_msgs::TrafficControlRequest>("outgoing_geofence_request", 1, true);
   //Check Active Geofence Publisher
   active_pub_ = cnh_.advertise<cav_msgs::CheckActiveGeofence>("active_geofence", 200, true);
+  //publish TCM acknowledgement after processing TCM
+  tcm_ack_pub_ = cnh_.advertise<cav_msgs::MobilityOperation>("outgoing_geofence_ack", 1, true);
```
msmcconnell, line 70: Probably a good idea to set the queue size here greater than 1 maybe some multiple of the publication number like 2*msgnum so that messages published in the loop don't get dropped

msmcconnell (changes requested):

carma_wm_ctrl/src/WMBroadcaster.cpp
dan-du-car: There seems to be some complains from CI.

msmcconnell (approved):

PR #1686, opened 2022-03-29
Feature/ihp identities

TonysCousin:
# PR Details
## Description

Logic fixes for rear & front join to consistently use identifiers for the various vehicles involved, platoon IDs and joining plan information for each interaction.

## Related Issue

PR #1647

## Motivation and Context

Problems completing a front join during testing on 3/24/22.

## How Has This Been Tested?

Code successfully built.  This is a branch off of an integration branch, so will be testing in vehicles later.

## Types of changes

## Checklist:

### DIFF
--- a/platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
+++ b/platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
@@ -989,6 +989,8 @@ namespace platoon_strategic_ihp
             request.location = pose_to_ecef(pose_msg_);
             request.strategy = PLATOONING_STRATEGY;
             request.urgency = 50;
+
+            int platoon_size = pm_.getTotalPlatooningSize();
             
             // step 3. Request Rear Join 
             if(isVehicleRightInFront(rearVehicleDtd, rearVehicleCtd)  &&  !config_.test_front_join)
@@ -1794,7 +1796,6 @@ namespace platoon_strategic_ihp
             ROS_DEBUG_STREAM("CUT-IN join maneuver is already in operation, NACK incoming join requests from other candidates.");
             return MobilityRequestResponse::NACK;
         }
-
     }
 
     // UCLA: add request call-back function for prepare to join (for cut-in join)
@@ -2673,7 +2674,7 @@ namespace platoon_strategic_ihp
             infoOperation = composeMobilityOperationLeadWithOperation(OPERATION_INFO_TYPE);
             mobility_operation_publisher_(infoOperation);
             prevHeartBeatTime_ = ros::Time::now().toNSec() / 1000000;
-            ROS_DEBUG_STREAM("Published heart beat platoon INFO mobility operatrion message");
+            ROS_DEBUG_STREAM("Published heart beat platoon INFO mobility operation message");
         }
         // Task 2
         // if (isTimeForHeartBeat) 
@@ -2836,6 +2837,7 @@ namespace platoon_strategic_ihp
         else if (pm_.current_platoon_state == PlatoonState::PREPARETOJOIN)
         {
             run_prepare_to_join();
+        }
         // coding oversight
         else
         {
--- a/platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
+++ b/platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
@@ -124,6 +124,19 @@ namespace platoon_strategic_ihp
         pm_.isFollower = true;
     }
 
+    // helper to store info on new action plan and platoon updates
+    void PlatoonStrategicIHPPlugin::create_join_action_plan(const cav_msgs::MobilityOperation& msg, const cav_msgs::MobilityRequest& request, 
+                                                            const int targetPlatoonSize)
+    {
+        pm_.currentPlatoonID = request.header.plan_id;
+        if (targetPlatoonSize > 1)
+        {
+            pm_.currentPlatoonID = msg.header.plan_id; //this will be empty if sender is a solo vehicle
+        }
+        pm_.platoonLeaderID = msg.header.sender_id;
+        pm_.current_plan = ActionPlan(true, request.header.timestamp, request.header.plan_id, msg.header.sender_id);
+    }
+
     // Callback to calculate downtrack based on pose message.
     void PlatoonStrategicIHPPlugin::pose_cb(const geometry_msgs::PoseStampedConstPtr& msg)
     {
@@ -300,7 +313,7 @@ namespace platoon_strategic_ihp
 
         // Extract data
         cav_msgs::MobilityOperation msg;
-        msg.header.plan_id = pm_.current_plan.planId;
+        msg.header.plan_id = pm_.currentPlatoonID;
         msg.header.recipient_id = "";
         std::string hostStaticId = config_.vehicleID;
         msg.header.sender_id = hostStaticId;
@@ -332,7 +345,7 @@ namespace platoon_strategic_ihp
          *           |-------0-----------1---------2--------3----------4----------5-------|
          */ 
         cav_msgs::MobilityOperation msg;
-        msg.header.plan_id = pm_.current_plan.planId; // msg.header.plan_id is the platoon ID of the request sender (rear join and frontal join). 
+        msg.header.plan_id = pm_.currentPlatoonID; // msg.header.plan_id is the platoon ID of the request sender (rear join and frontal join). 
         msg.header.recipient_id = "";
         std::string hostStaticId = config_.vehicleID;
         msg.header.sender_id = hostStaticId;
@@ -487,7 +500,7 @@ namespace platoon_strategic_ihp
         // compose msgs for anything other than standby
         if (pm_.current_platoon_state != PlatoonState::STANDBY)
         {
-            status_msg.platoon_id = pm_.current_plan.planId;
+            status_msg.platoon_id = pm_.currentPlatoonID;
             status_msg.size = pm_.getTotalPlatooningSize();
             status_msg.size_limit = config_.maxPlatoonSize;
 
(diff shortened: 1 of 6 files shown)


platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -2467,18 +2486,29 @@ namespace platoon_strategic_ihp
...
+                pm_.currentPlatoonID = pm_.targetPlatoonID;
+                pm_.targetPlatoonID = "";
+            }
+            pm_.platoonLeaderID = pm_.current_plan.peerId;
         }
 
-        ROS_DEBUG_STREAM("pm_.getTotalPlatooningSize()" << pm_.getTotalPlatooningSize());
         //Task 4: publish platoon status message (as single joiner)
-        if (pm_.getTotalPlatooningSize() > 1) {
+        if (pm_.current_plan.valid) //don't want to do this until iterations after MobReq message is delivered above
```
saina-ramyar, line 2504: I dont understand why we have this condition. I thought in this new version the current plan is only for joining. But according to this condition, we wont get any status msgs during regular operation

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -2501,52 +2531,75 @@ namespace platoon_strategic_ihp
...
+        // and thereby adds the joiner to the platoon record. This process requires host's mob_req_cb_leader() to ACK the join request, then 
+        // the remote vehicle to handle that ACK and broadcast its first op STATUS, then host to receive and process it. All that has to happen
+        // before the below code block runs, even though this method starts to spin immediately after we send out the mentioned ACK. To avoid
+        // a race condition, we must wait a few cycles to ensure the 2-way messaging has completed.
+
+        // Check if gap is big enough and if there is no currently active plan and this method has been called several times
         // Add a condition to prevent sending repeated requests (Note: This is a same-lane maneuver, so no need to consider lower bound of joining gap.)
-        if (currentGap <= maxJoinGap  &&  !pm_.current_plan.valid  &&  isFirstLeaderAbortRequest_) 
+        ++numLeaderAbortingCalls_;
+        if (currentGap <= maxJoinGap  &&  !pm_.current_plan.valid  &&  numLeaderAbortingCalls_ > 4)
```
saina-ramyar, line 2587: I dont understand how the request being sent late is going to prevent a race condition? since the request is the first step of frontal join

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -2501,52 +2531,75 @@ namespace platoon_strategic_ihp
...
+            //TODO: this is just a test - remove when satisfied that this var doesn't need to be in the above if statement
+            ROS_DEBUG_STREAM("Gap size good & no valid plan. isFirstLeaderAbortRequest = " << isFirstLeaderAbortRequest_);
+
             // compose frontal joining plan, senderID is the old leader 
             cav_msgs::MobilityRequest request;
             std::string planId = boost::uuids::to_string(boost::uuids::random_generator()());
             long currentTime = ros::Time::now().toNSec() / 1000000;
             request.header.plan_id = planId;
-            request.header.recipient_id = pm_.targetLeaderId;
+            request.header.recipient_id = pm_.platoonLeaderID; //the new joiner
```
saina-ramyar, line 2597: isnt this the front join request on the joiner's side? so why is the recipient's id the new joiner??

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1031,18 +1037,8 @@ namespace platoon_strategic_ihp
                  */
                 ROS_DEBUG_STREAM("Neighbor platoon leader is right behind us");
                 
-                // compose request
```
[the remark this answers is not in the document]
TonysCousin: Header info in each of these else if branches was consolidated above the first if, around line 984.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1075,26 +1082,15 @@ namespace platoon_strategic_ihp
...
-                ROS_DEBUG_STREAM("Host vehicle platoon id = " << newLeaderPlatoonId << "; join the rear platoon as the new leader (frontal join).");
-
-                // compose request
-                cav_msgs::MobilityRequest request;
-                request.header.plan_id = boost::uuids::to_string(boost::uuids::random_generator()());
-                request.header.recipient_id = senderId;
-                request.header.sender_id = config_.vehicleID;
-                request.header.timestamp = ros::Time::now().toNSec() / 1000000;
-                request.location = pose_to_ecef(pose_msg_);
+                // complete the request
```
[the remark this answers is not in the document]
TonysCousin: Consolidated above.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1372,32 +1369,52 @@ namespace platoon_strategic_ihp
...
         }
         else
         {
             ROS_DEBUG_STREAM("Received platoon request with vehicle id = " << msg.header.sender_id);
             ROS_DEBUG_STREAM("The request type is " << msg.plan_type.type << " and we choose to ignore");
-            return MobilityRequestResponse::NO_RESPONSE;
+            response = MobilityRequestResponse::NO_RESPONSE; //TODO: may be better to NACK of target vehicle is correct
         }
+
+        // Indicate the current join activity is complete
```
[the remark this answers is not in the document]
TonysCousin: Good catch. It should have been a call to clearActionPlan().  Also, this means that the creation of a new ActionPlan a few lines above is unnecessary.  I've removed that also.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1560,34 +1587,29 @@ namespace platoon_strategic_ihp
...
-                    // change platoon ID to candidate leader --> read it from current platoonplan, plan is send out by frontal joiner
-                    targetPlatoonId = potentialNewPlatoonId_front_;
-                    ROS_DEBUG_STREAM("targetPlatoonId = " << targetPlatoonId);
-                    // change leader to candidate leader --> read it from current platoonplan --> peerID is the frontal joiner vehID 
-                    pm_.targetLeaderId = pm_.current_plan.peerId;
-                    ROS_DEBUG_STREAM("pm_.targetLeaderId = " << pm_.targetLeaderId);
-                    // -----------------------------------------------------------------------------------------
-                    
-                    waitingStartTime = ros::Time::now().toNSec() / 1000000;
+                    // If we are testing with a single vehicle representing this platoon, then we don't yet have a platoon ID,
```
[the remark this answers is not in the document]
TonysCousin: In mob_op_cb_leader() around line 1063 it decides whether to use the existing platoon's ID, if one exists, or to propose a new one.  The logic here is supposed to complement that so that both sides of the conversation agree on where the platoon ID is coming from.  If the host (old leader) is in a real platoon already, we should keep its ID.  Therefore, I believe it is good as is.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1876,31 +1885,37 @@ namespace platoon_strategic_ihp
...
                 {
                     // We change to follower state and start to actually follow that leader
                     // The platoon manager also need to change the platoon Id to the one that the target leader is using 
-                    ROS_DEBUG_STREAM("The leader " << msg.header.sender_id << " agreed on our join. Change to follower state.");
                     pm_.current_platoon_state = PlatoonState::FOLLOWER;
-                    targetPlatoonId = msg.header.plan_id;
-                    // change platoon manager to follower state 
-                    pm_.changeFromLeaderToFollower(targetPlatoonId, msg.header.sender_id);
+                    pm_.changeFromLeaderToFollower(pm_.currentPlatoonID, msg.header.sender_id);
+                    ROS_DEBUG_STREAM("The leader " << msg.header.sender_id << " agreed on our join. Change to follower state.");
```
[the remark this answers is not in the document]
TonysCousin: Yes, the join plan is cleared on line ~1916 (my new edits have changed numbering a bit). But this makes me realize that targetPlatoonID needs to be cleared even on ACK, so I've moved it outside the else block.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
TonysCousin: There is a lot to keep track of with the layers of changing states! I may very well have overlooked something, but I think this is good. For the first several calls to run_candidate_follower(), current_plan.valid will be false and this block will not be executed. Presumably host is moving into position during that time. At some point gap is adequate, and the "task 3" block runs, which sends the MobReq and sets current_plan.valid. From then on, while still in CANDIDATEFOLLOWER state, host will send op STATUS messages here. Once an ACK is received, host transitions to FOLLOWER and starts sending op STATUS from run_follower().

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
TonysCousin: This is the second request in the frontal join process (near the bottom of the sequence diagram).  After the first request/ACK completes, then the host (former lead vehicle) must see at least one STATUS message from the joiner before forming this request message.  If that does not happen, the calculation of currentGap on line 2564 returns 0 since it doesn't see the joiner yet.  That result would allow the second request to go out prematurely and complete the join even if the vehicle isn't in position.  I've expanded the NOTE comment above to help clarify the situation.  Let me know if better/more wording is needed.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
TonysCousin: This is on the old leader vehicle (run_leader_aborting).  platoonLeaderID is set in mob_op_cb_leader() to represent the joiner.  This is correct.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -499,12 +490,8 @@ namespace platoon_strategic_ihp
 
             if (pm_.current_platoon_state == PlatoonState::FOLLOWER)
             {
-                ROS_DEBUG_STREAM("isFollower: " << pm_.isFollower);
                 ROS_DEBUG_STREAM("pm platoonsize: " << pm_.getTotalPlatooningSize());
 
-                //TODO: this method should not be changing the object's state; consider moving this stmt to run_follower()
-                pm_.isFollower = true;
```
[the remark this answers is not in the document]
TonysCousin: Looks okay. isFollower is being assigned appropriately for state changes to/from FOLLOWER state.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
saina-ramyar: now that you explain it, the comment above is clear. Another question, from what I see, unless the speed the joinGap on line 2564 is never zero?
Also, I am concerned that by the time numLeaderAbortingCalls_ is higher than 4, the leaderaborting state is timed out! so maybe we need to make that 4 a configurable param?

TonysCousin: Not sure what your first question is.  I don't see maxJoinGap ever being zero.  My original explanation referred to the wrong line, which should have been the calculation of currentGap, now on line 2570.  The concern is that currentGap will be zero unless we've seen a STATUS message from the joiner.
Agree, the call threshold should be a param.  I'll make that change.

PR #1696, opened 2022-03-31
return remaining points

adev4a:
# PR Details
This PR aims to resolve an issue where planning for lane follow maneuver starts close enough to the end of the maneuver such that there are only 1 one remaining point to be added in path.

## Types of changes

## Checklist:

### DIFF
--- a/basic_autonomy/src/basic_autonomy.cpp
+++ b/basic_autonomy/src/basic_autonomy.cpp
@@ -877,7 +877,23 @@ namespace basic_autonomy
             ROS_DEBUG_STREAM("Ending state's index before applying buffer (buffer_pt_index): " << buffer_pt_index);
             ROS_DEBUG_STREAM("Corresponding to point: x: " << all_sampling_points[buffer_pt_index].x() << ", y:" << all_sampling_points[buffer_pt_index].y());
 
-            if(nearest_pt_index + 1 >= buffer_pt_index){
+            if(nearest_pt_index + 1 = buffer_pt_index){
+                ROS_WARN_STREAM("Returning the two remaining points in the maneuver");
+
+                std::vector<lanelet::BasicPoint2d> remaining_traj_points = {all_sampling_points[nearest_pt_index], all_sampling_points[nearest_pt_index + 1]};
+
+                std::vector<double> downtracks = carma_wm::geometry::compute_arc_lengths(remaining_traj_points);
+                std::vector<double> speeds = {state.longitudinal_vel, state.longitudinal_vel};//Keep current speed
+                std::vector<double> times;
+                trajectory_utils::conversions::speed_to_time(downtracks, speeds, &times);
+                std::vector<double> yaw = {state.orientation, state.orientation}; //Keep current orientation
+                
+                std::vector<cav_msgs::TrajectoryPlanPoint> traj_points =
+                trajectory_from_points_times_orientations(remaining_traj_points, times, yaw, state_time);
+
+                return traj_points;
+            }
+            else if(nearest_pt_index + 1 > buffer_pt_index){
                 ROS_WARN_STREAM("Current state is at or past the planned end distance. Couldn't generate trajectory");
                 return {};
             }


basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -878,8 +878,34 @@ namespace basic_autonomy
...
 
             if(nearest_pt_index + 1 >= buffer_pt_index){
-                ROS_WARN_STREAM("Current state is at or past the planned end distance. Couldn't generate trajectory");
-                return {};
+                
+                lanelet::BasicPoint2d current_pos(state.x_pos_global, state.y_pos_global);
+                lanelet::BasicPoint2d ending_pos(ending_state.x_pos_global, ending_state.y_pos_global);
+
+                if( nearest_pt_index > buffer_pt_index ||
+                    (wm->routeTrackPos(ending_pos).downtrack -  wm->routeTrackPos(current_pos).downtrack) < epsilon_ ){
```
kjrush, line 886: I think this is still equivalent to `nearest_pt_idx >= buffer_pt_idx`, except we're now using an epsilon for the equals side of the comparison. Would it make more sense for this whole conditional to be:

`if (wm->routeTrackPos(ending_pos).downtrack < wm.routeTrackPos(current_pos).downtrack) { ... }`

That should cover all cases where we can't generate a trajectory right?

adev4a: Yes that makes sense, only thing is nearest_pt_index being greater than buffer_pt_index will cause an exception further down in this function on line 913. 
Do we want to have that check happen separately?

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -878,8 +878,33 @@ namespace basic_autonomy
             ROS_DEBUG_STREAM("Corresponding to point: x: " << all_sampling_points[buffer_pt_index].x() << ", y:" << all_sampling_points[buffer_pt_index].y());
 
             if(nearest_pt_index + 1 >= buffer_pt_index){
-                ROS_WARN_STREAM("Current state is at or past the planned end distance. Couldn't generate trajectory");
-                return {};
+                
+                lanelet::BasicPoint2d current_pos(state.x_pos_global, state.y_pos_global);
+                lanelet::BasicPoint2d ending_pos(ending_state.x_pos_global, ending_state.y_pos_global);
+
+                if(wm->routeTrackPos(ending_pos).downtrack -  wm->routeTrackPos(current_pos).downtrack < epsilon_ ){
```
kjrush, line 885: Is `epsilon_` defined here? I can't find it in the CPP file or the .h file. I wanted to check what it's value is, since potentially we'd see the same behavior as before (fail to plan) within epsilon of the final point. If epsilon is very small, that makes sense to me (as the data is effectively meaningless at such a small scale) but hypothetically there's nothing requiring the epsilon here, we could use a strict `if (!(current_dtd < end_dtd))...` check.

kjrush: I found it, nevermind: https://github.com/usdot-fhwa-stol/carma-platform/blob/23a79861f1e6b22f02a98b574f5f745db38e73a2/basic_autonomy/include/basic_autonomy/helper_functions.h#L24   

Certainly seems less likely to cause the issue than what we had before (effectively the same, but an epsilon of 1.0m). The question is if it's worth dropping the epsilon for a guarantee. If we're closer than to the end of the maneuver 0.0000001m I think there's two questions:
1. Would plan delegator send this maneuver to ILC for planning?
2. Is there a valid answer to that plan request?

My inclination is to lean towards yes for both? What are your thoughts?

adev4a: Yes epsilon is defined here https://github.com/usdot-fhwa-stol/carma-platform/blob/23a79861f1e6b22f02a98b574f5f745db38e73a2/basic_autonomy/include/basic_autonomy/helper_functions.h#L24

It is really small (0.1mm)

adev4a: The way it is currently set up the plan delegator would send a request to plan to ILC. Maneuvers are only treated as expired if the timestamps dont match. 
As far as the valid answer goes, i guess having this condition for specifically make a response to it invalid.
Without this change, unless the points are being sampled at high enough rate, 0.1mm before the end of maneuver might still lead to an invalid trajectory

adev4a: I was wondering if a fix for this should be in the tactical plugin, or the plan delegator itself?  
The trajectory from the next maneuver should be valid at this point, so shouldn't we be using that instead?

kjrush (approved):

PR #1704, opened 2022-04-06
fix py eq and inters var

MishkaMN:
# PR Details
## Description

When vehicle starts from STOPPING range, it hasn't saved the intersection variables so it throws exception. 
Also if condition still had python syntax which gives wrong results

## Motivation and Context

TSMO UC2 testing
## How Has This Been Tested?
not yet

## Types of changes

## Checklist:

### DIFF
--- a/lci_strategic_plugin/src/lci_strategic_plugin.cpp
+++ b/lci_strategic_plugin/src/lci_strategic_plugin.cpp
@@ -540,18 +540,6 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
     return;
   }
 
-   // At this point we know the vehicle is within the activation distance and we know the current and next light phases
-  // If the vehicle is at a stop trigger the stopped state
-  constexpr double HALF_MPH_IN_MPS = 0.22352;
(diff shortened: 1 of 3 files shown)


lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -585,6 +574,17 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
 
   ROS_DEBUG_STREAM("intersection_end_downtrack_: " << intersection_end_downtrack_.get());
 
+  // If the vehicle is at a stop trigger the stopped state
+  constexpr double HALF_MPH_IN_MPS = 0.22352;
+  if (current_state.speed < HALF_MPH_IN_MPS &&
+      fabs(distance_remaining_to_traffic_light) < config_.stopping_location_buffer + length_to_front_bumper_)
```
msmcconnell, line 580: Seems like a good opportunity to remove the length_to_front_bumper from this check. To get the same behavior is before you could correspondingly increase stopping_location_buffer parameter value.

msmcconnell (approved):

PR #1709, opened 2022-04-07
Fixes for TCM processing and applied map updates

JonSmet:
# PR Details
## Description

This PR includes a couple of fixes related to the processing of incoming TCMs:

- WMListenerWorker: An update has been made to enable queued map updates to trigger a recomputation of the routing graph and trigger a re-route (by the route node) if one of the applied queued map updates invalidated the route.
- WMBroadcaster: Converts the an incoming TCM speed limit from m/s to MPH. This value (in MPH) is used to set the speed limit in the map.

## Motivation and Context

During Freight Workzone testing, it was discovered that if a queued route-invalidation map update (for example, a closed lanelet map update) was applied to the map, a reroute was not triggered, and the routing graph was no re-computed.

## How Has This Been Tested?

Tested on the Silver Truck at Suntrax as part of the Freight Workzone use case.

## Types of changes

## Checklist:

### DIFF
--- a/carma_wm/src/WMListenerWorker.cpp
+++ b/carma_wm/src/WMListenerWorker.cpp
@@ -496,8 +496,8 @@ void WMListenerWorker::routeCallback(const cav_msgs::RouteConstPtr& route_msg)
     return;
   }
 
-  // If one of the applied queued map updates invalidated the route, then the routing graph must be updated again
-  if (route_invalidated_by_queued_map_update){
+  // If one of the applied queued map updates invalidated the route, then the routing graph must be updated again for the route node
+  if (route_invalidated_by_queued_map_update && route_node_flag_){
     ROS_DEBUG_STREAM("At least one applied queued map update has invalidated the route. Routing graph will be recomputed.");
     world_model_->setMap(world_model_->getMutableMap(), current_map_version_, route_invalidated_by_queued_map_update);
     ROS_DEBUG_STREAM("Finished recomputing the routing graph for the applied queued map update(s)");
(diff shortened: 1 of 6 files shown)


carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -281,7 +281,7 @@ void WMBroadcaster::geofenceFromMsg(std::shared_ptr<Geofence> gf_ptr, const cav_
   if (msg_detail.choice == cav_msgs::TrafficControlDetail::MAXSPEED_CHOICE) 
   {  
     //Acquire speed limit information from TafficControlDetail msg
-    sL = lanelet::Velocity(msg_detail.maxspeed * lanelet::units::MPH()); 
+    sL = lanelet::Velocity(msg_detail.maxspeed * MS_TO_MPH * lanelet::units::MPH());
```
msmcconnell, line 284: Rather than adding this multiplication you could have just changed the units I believe

```
lanelet::Velocity(msg_detail.maxspeed * lanelet::units::MPS()); 
```


Same below. Add you could remove the extra constant

msmcconnell (commented):
Quick comment while I look over the routing changes which will take more time

carma_wm_ctrl/src/WMBroadcaster.cpp
JonSmet: Fixed.

carma_wm/src/WMListenerWorker.cpp
```diff
@@ -483,37 +488,52 @@ void WMListenerWorker::routeCallback(const cav_msgs::RouteConstPtr& route_msg)
...
+    rerouting_flag_ = false; // Reset flag since no applied queued map updates invalidated the route for the route node
 
-  world_model_->setRouteEndPoint({route_msg->end_point.x,route_msg->end_point.y,route_msg->end_point.z});
+    auto path = lanelet::ConstLanelets();
+    for(auto id : route_msg->shortest_path_lanelet_ids)
+    {
+      auto ll = world_model_->getMap()->laneletLayer.get(id);
+      path.push_back(ll);
+    }
+    if(path.empty()) return;
```
msmcconnell, line 515: Is this a likely case? Should there be an error report here?
I would say that if we hit this case we should also be clearing the world model's route, such that a user node would no longer find a route available.

msmcconnell (changes requested):

carma_wm/src/WMListenerWorker.cpp
JonSmet: Good point, I don't think it is possible (or at least highly unlikely) for the route node to publish a route message without any lanelet IDs listed in the shortest path. This specific logic was carried over from the initial release/jetta-gillet branch, and I would be fine with removing the `if(path.empty()) return;` line completely.

JonSmet: Pushed a commit to remove this check.

msmcconnell (approved):

PR #1711, opened 2022-04-07
Fix/equation x x2

MishkaMN:
# PR Details
## Description

Added ability to choose early or late arrival for case when the car starts from WAITING state. Then it should only go if late arrival in green is certain.
Trying to fix negative duration error by adding max(x,0) guard.
And also found the actual reason why there is negative duration which is the v_hat was nan due to wrong equation

## How Has This Been Tested?
Tested on Blue Lexus with R30 left of the signal from EL1

## Types of changes

## Checklist:

### DIFF
--- a/lci_strategic_plugin/src/lci_strategic_plugin.cpp
+++ b/lci_strategic_plugin/src/lci_strategic_plugin.cpp
@@ -758,7 +758,7 @@ void LCIStrategicPlugin::planWhenWAITING(const cav_srvs::PlanManeuversRequest& r
   if (!bool_optional_late_certainty)
   {
     ROS_ERROR_STREAM("Unable to resolve green light...");
-    return ;
+    return;
   }
 
   if (current_light_state_optional.get().second == lanelet::CarmaTrafficSignalState::PROTECTED_MOVEMENT_ALLOWED &&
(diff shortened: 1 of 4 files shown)


lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -734,14 +746,22 @@ void LCIStrategicPlugin::planWhenWAITING(const cav_srvs::PlanManeuversRequest& r
 
   ROS_DEBUG("traffic_light_down_track %f", traffic_light_down_track);
   
+  auto bool_optional_late_certainty = canArriveAtGreenWithCertainty(current_state.stamp, traffic_light, true, false);
```
msmcconnell, line 749: This should be below the valid light state check?

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -734,14 +746,22 @@ void LCIStrategicPlugin::planWhenWAITING(const cav_srvs::PlanManeuversRequest& r
...
+  
+  if (!bool_optional_late_certainty)
+  {
+    throw std::invalid_argument("Unable to resolve the green signal...");
```
msmcconnell, line 753: Instead of throwing an exception just return another stop and wait maneuver to keep the vehicle in place while the light rolls over.

msmcconnell (changes requested):

aswath1 (approved):

snallamothu (approved):

snallamothu (approved):

MishkaMN:
created another duplicate PR here
https://github.com/usdot-fhwa-stol/carma-platform/pull/1713
and that one is merged earlier, so closing this

PR #1714, opened 2022-04-08
Add checks to ensure objects are in front of vehicle

### LINKED ISSUE #1708: Yield plugin does not disregard vehicles behind the ego vehicle
### Types of Issue

### Descriptive summary

The yield plugin does not properly check that the vehicle's added to the list of possible collisions are further downtrack than the ego vehicle, resulting in slowdowns to avoid vehicles behind the ego vehicle.
### Expected behavior
The yield plugin should disregard vehicles behind the ego vehicle.
### Actual behavior

The yield plugin attempts to avoid collisions from behind by slowing down.

kjrush:
Added checks to yield plugin to ignore objects at or behind the vehicle in downtrack distance.

# PR Details
## Description

Minor fixes for issue #1708 . This PR adds a check to the Yield Plugin to verify that a vehicle is in front of the ego vehicle (by comparing downtrack distances) prior to assessing time-to-collision. If the vehicle is not in front of the ego vehicle, it will not be added to the list of collisions for the ego vehicle to avoid. This resolves an issue observed in XIL testing. This PR also includes a handful of log message fixes, which were outputting incorrect data.

## Related Issue

Closes #1708 
## Motivation and Context

Without this check the Yield Plugin may yield to vehicles behind it, as the result of the TTC calculation will return a negative value due to the negative displacement between the two vehicles. This results in unnecessarily conservative behavior of commanding 0 or very low speeds for long periods of time.

## Types of changes

## Checklist:


yield_plugin/src/yield_plugin.cpp
```diff
@@ -457,11 +458,13 @@ namespace yield_plugin
         
           // std::abs might not be needed cause vehicles in the behind of vehicle to cause problem
```
saina-ramyar, line 459: since std::abs is not used anymore, this comment can be removed.

PR #1719, opened 2022-04-11
resolve merge conflicts platoon_id and cutin join

saina-ramyar:
# PR Details
## Types of changes

## Checklist:

### DIFF
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -265,6 +265,21 @@ namespace platoon_strategic_ihp
         double getPlatoonFrontDowntrackDistance();
 
         /**
+         * \brief UCLA: Return the time headway summation of all predecessors, in m.
+         */
+        double getPredecessorTimeHeadwaySum();
+        
+        /**
+         * \brief UCLA: Return the speed of the preceding vehicle, in m/s.
+         */
+        double getPredecessorSpeed();
+
+        /**
+         * \brief UCLA: Return the position of the preceding vehicle, in m.
+         */
+        double getPredecessorPosition();
+        
+        /**
         * \brief UCLA: Return follower's desired position (i.e., downtrack, in m) that maintains the desired 
         * intra-platoon time gap, based on IHP platoon trajectory regulation algorithm.
         * 
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -265,7 +265,7 @@ namespace platoon_strategic_ihp
         double getPlatoonFrontDowntrackDistance();
 
         /**
-         * \brief UCLA: Return the time headway summation of all predecessors, in m.
+         * \brief UCLA: Return the time headway summation of all predecessors, in s.
          */
         double getPredecessorTimeHeadwaySum();
         
--- a/platooning_strategic_IHP/src/platoon_manager_ihp.cpp
+++ b/platooning_strategic_IHP/src/platoon_manager_ihp.cpp
@@ -776,7 +776,7 @@ namespace platoon_strategic_ihp
         */
 
         double min_diff = 99999.000;
-        int cut_in_index = -2;
+        int cut_in_index = -1;
         // Loop through all platoon members  
         for(size_t i = 0; i < platoon.size(); i++) 
         {
--- a/platooning_strategic_IHP/src/platoon_manager_ihp.cpp
+++ b/platooning_strategic_IHP/src/platoon_manager_ihp.cpp
@@ -109,7 +109,7 @@ namespace platoon_strategic_ihp
                 ROS_DEBUG_STREAM("It seems that the current leader is joining another platoon.");
                 ROS_DEBUG_STREAM("So the platoon ID is changed from " << currentPlatoonID << " to " << platoonId);
                 currentPlatoonID = platoonId;
-                updatesOrAddMemberInfo(senderId, cmdSpeed, dtDistance, ctDistance, curSpeed);
+                updatesOrAddMemberInfo(senderId, cmdSpeed, dtDistance, ctDistance, curSpeed); 
             } 
             else if (currentPlatoonID == platoonId)
             {
@@ -129,7 +129,7 @@ namespace platoon_strategic_ihp
             if (currentPlatoonID == platoonId)
             {
                 ROS_DEBUG_STREAM("This STATUS messages is from our platoon. Updating the info...");
-                updatesOrAddMemberInfo(senderId, cmdSpeed, dtDistance, ctDistance, curSpeed);
+                updatesOrAddMemberInfo(senderId, cmdSpeed, dtDistance, ctDistance, curSpeed); 
             }
             else
             {
@@ -200,6 +200,8 @@ namespace platoon_strategic_ihp
             }
         }
     }
+    
+    // TODO: Place holder for delete member info due to dissovle operation.
 
     // Get the platoon size.
     int PlatoonManager::getTotalPlatooningSize() {
@@ -270,12 +272,18 @@ namespace platoon_strategic_ihp
     int PlatoonManager::allPredecessorFollowing(){
 
         ///***** Case Zero *****///
-        // If the host vehicle is the second vehicle in this platoon,we will always follow the platoon leader in front of host vehicle
-        if(hostPosInPlatoon_ == 1) {
-            ROS_DEBUG_STREAM("As the second vehicle in the platoon, it will always follow the leader. Case Zero");
-            return 0;
+        // If the host vehicle is the second follower of a platoon, it will always follow the platoon leader in the front 
+        if(getNumberOfVehicleInFront() == 1) {
+            if (isFollower){
+                // If the platoon length is 2 and the host is follower, then follow the leader.
+                ROS_DEBUG_STREAM("As the second vehicle in the platoon, it will always follow the leader. Case Zero");
+                return 0;
+            }
+            else{
+                ROS_DEBUG_STREAM("Function getDynamicLeader was called when host is leader, returning zero.");
+                return 0;
+            }
         }
-
         ///***** Case One *****///
         // If there weren't a leader in the previous time step, follow the first vehicle (i.e., the platoon leader) as default.
         if(previousFunctionalDynamicLeaderID_ == "") {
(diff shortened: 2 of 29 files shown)


platooning_strategic_IHP/src/platoon_manager_ihp.cpp
```diff
@@ -270,12 +272,18 @@ namespace platoon_strategic_ihp
...
-            ROS_DEBUG_STREAM("As the second vehicle in the platoon, it will always follow the leader. Case Zero");
-            return 0;
+        // If the host vehicle is the second follower of a platoon, it will always follow the platoon leader in the front 
+        if(getNumberOfVehicleInFront() == 1) {
+            if (isFollower){
+                // If the platoon length is 2 and the host is follower, then follow the leader.
+                ROS_DEBUG_STREAM("As the second vehicle in the platoon, it will always follow the leader. Case Zero");
+                return 0;
+            }
+            else{
```
TonysCousin, line 282: Continuing my rant, the else makes no sense.  Regardless of which branch we take, the return value is 0.  Why are we testing anything?  These are things to be ironed out after this PR.

platooning_strategic_IHP/src/platoon_manager_ihp.cpp
```diff
@@ -709,7 +776,7 @@ namespace platoon_strategic_ihp
         */
 
         double min_diff = 99999.000;
-        int cut_in_index = -1;
+        int cut_in_index = -2;
```
TonysCousin, line 779: This should remain -1.  The value -2 doesn't belong anywhere.  Even though the below loop is broken (a different problem), it will return this value if joiner is in front of everyone.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1087,11 +1109,18 @@ namespace platoon_strategic_ihp
...
+                std::string newLeaderPlatoonId = pm_.currentPlatoonID; // set self's (front join veh) platoon ID as the new platoon ID
+                ROS_DEBUG_STREAM("Host vehicle platoon id = " << newLeaderPlatoonId << "; join the rear platoon as the new leader (frontal join).");
+
+                // Note: Request conposed outside of if conditions
+                // UCLA: Desired joining index for cut-in join, indicate the index of gap-leading vehicle. -1 indicate cut-in from front.
+                // Note: remove join_index to info param.
                 request.plan_type.type = cav_msgs::PlanType::CUT_IN_FROM_SIDE; 
 
+                // note: At this step all cut-in types all start with this request, so the join_index at this point is set to default, -2.
+                // REVISIOn note: At this step a join index is really  not needed until later. The later index were changed to name "target_index". So it is ok to delete cut-in here.
```
TonysCousin, line 1122: Seems like this should be a TODO  that needs to be resolved before we are ready to test cut-ins.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1087,11 +1109,18 @@ namespace platoon_strategic_ihp
                 //  -ensure plan type corresponds to join_index
 
                 // complete the request
-
-                int join_index = -1;
                 // UCLA: this is a newly added plan type 
+                std::string newLeaderPlatoonId = pm_.currentPlatoonID; // set self's (front join veh) platoon ID as the new platoon ID
```
TonysCousin, line 1113: Hard to tell at this point if this is wrong, but I'm suspicious.  First, the host is imposing its own platoon ID on the existing platoon, which is different from how the front join is done above.  Second, if host is a solo joining vehicle, then its pm_.currentPlatoonID should be empty.  Third, the variable name and comments assume the joiner is doing a front join; however, this case 5 is supposed to cover joining in all 3 positions.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1440,10 +1551,12 @@ namespace platoon_strategic_ihp
          */
         bool isFrontJoin = plan_type.type == cav_msgs::PlanType::JOIN_PLATOON_FROM_FRONT;
         bool isRearJoin = plan_type.type == cav_msgs::PlanType::JOIN_PLATOON_AT_REAR;
-        bool isCutinJoin = plan_type.type == cav_msgs::PlanType::CUT_IN_FROM_SIDE;
+        bool isCutInJoin = plan_type.type == cav_msgs::PlanType::CUT_IN_FROM_SIDE;
+        bool isDepart = (plan_type.type == cav_msgs::PlanType::PLATOON_DEPARTURE);
+        // TODO: Place holder for departure plan type.
```
TonysCousin, line 1556: This TODO should be removed.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -1618,7 +1731,8 @@ namespace platoon_strategic_ihp
         }
         
         // UCLA: conditions for cut-in join; platoon leader --> leading with operation
-        else if (isCutinJoin)
+        // Use joningID_ to make sure only the leader only interact with the first requesting joining vehicle
```
TonysCousin, line 1734: This comment is bogus, as joiningID_ no longer exists.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -2768,6 +2939,7 @@ namespace platoon_strategic_ihp
         // UCLA: assign a new plan type
         request.plan_type.type = cav_msgs::PlanType::PLATOON_CUTIN_JOIN;
         request.strategy = PLATOONING_STRATEGY;
+        request.strategy_params = "";
```
TonysCousin, line 2942: This line does nothing.  strategy_params is overwritten a few lines below.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -2984,41 +3161,21 @@ namespace platoon_strategic_ihp
...
-        // {
-        //     // 3. Use CARMA planning method (control_length = step_time * speed limit).
-        //     target_speed = findSpeedLimit(current_lanelet);   //get Speed Limit TOTO update
-        //     total_maneuver_length = current_progress + config_.time_step * target_speed;
-        // }
-        // ----------------------------------------------------------------
-
-        target_speed = findSpeedLimit(current_lanelet);   //get Speed Limit TOTO update
+        // Note: The desired gap will be adjusted and send to control plugin (via platoon_info_msg) where gap creation will be handled.
+        double total_maneuver_length;
```
TonysCousin, line 3171: total_maneuver_length doesn't appear to be used below.  Does it need to be declared?

TonysCousin (changes requested):
A couple of my comments probably should involve changes. Others are just for consideration.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
xuhan417: Indeed. The depart plan type is already included.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
xuhan417: I guess this is also a  left-over piece of code. It is safe to delete this line and the following line, as the variable "newLeaderPlatoonId" was only used twice through out this cpp file.  

I also double checked this method, the correct platoonId info was parsed from the mobility msg (the method's input), which is broadcasted by the target platoon leader. And the corresponding variable, the "platoonId", was used throughout this method (as it should) other than these two lines.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
xuhan417: Indeed.

platooning_strategic_IHP/src/platoon_manager_ihp.cpp
xuhan417: Agreed. Set -1 as default value is the correct value.

platooning_strategic_IHP/src/platoon_manager_ihp.cpp
xuhan417: Yes, if this method is called, then this is a follower already. I think the entire case 0 is not necessary.

platooning_strategic_IHP/src/platoon_manager_ihp.cpp
```diff
@@ -270,12 +272,18 @@ namespace platoon_strategic_ihp
     int PlatoonManager::allPredecessorFollowing(){
 
         ///***** Case Zero *****///
-        // If the host vehicle is the second vehicle in this platoon,we will always follow the platoon leader in front of host vehicle
-        if(hostPosInPlatoon_ == 1) {
-            ROS_DEBUG_STREAM("As the second vehicle in the platoon, it will always follow the leader. Case Zero");
-            return 0;
+        // If the host vehicle is the second follower of a platoon, it will always follow the platoon leader in the front 
+        if(getNumberOfVehicleInFront() == 1) {
+            if (isFollower){
```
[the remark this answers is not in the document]
xuhan417: I think this case 0 is just a quick bypass for the test when there is only 2 vehicles.  Since the current  test involves more then two vehicles, I feel like the entire case 0 could be removed, as the rest of the APF works just fine without it.

platooning_strategic_IHP/include/platoon_manager_ihp.h
```diff
@@ -264,6 +264,21 @@ namespace platoon_strategic_ihp
          */
         double getPlatoonFrontDowntrackDistance();
 
+        /**
+         * \brief UCLA: Return the time headway summation of all predecessors, in m.
```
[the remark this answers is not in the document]
xuhan417: Indeed, this should be in sec.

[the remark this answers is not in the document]
saina-ramyar: fixed

platooning_strategic_IHP/src/platoon_manager_ihp.cpp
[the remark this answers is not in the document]
saina-ramyar: removed the isFollower condition.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
saina-ramyar: removed

TonysCousin (approved):

xuhan417 (approved):

TonysCousin (approved):

PR #1727, opened 2022-04-19
Feature/control plugin gap regulation

xuhan417:
# PR Details
## Description

Add IHP gap regulation in platoon control plugin. Additional information are send via platoon_info_msg. 
Apply changes to gap creation procedure: The gap creation member will use a temporary desired gap that is larger than the pre-defined configured value. The control plug-in will adjust speed to maintain the new desired gap.

## Types of changes

## Checklist:


TonysCousin (changes requested):
It's important that we have consistency in naming convention.  In keeping with the strategic plugin named PlatoonStrategicIHPPlugin, we need to stick with the baseline code name of PlatoonControlIHPPlugin (and similar application in the snake-case names, where "ihp" is prior to "plugin" in any names).  Can you please make some global edits to use the names in the baseline code.  Once that is done, it will remove a LOT of diffs in the PR and allow us to focus on the actual logic changes.  Thanks.

platooning_control_ihp/include/platoon_control_ihp_config.h
```diff
@@ -36,19 +36,37 @@ struct PlatoonControlIHPPluginConfig
...
+                                    // But in normal operating conditions it should be set to false
+  //------------------------------------------------------------------------------------------------
+
   
   
-  friend std::ostream& operator<<(std::ostream& output, const PlatoonControlIHPPluginConfig& c)
+  friend std::ostream& operator<<(std::ostream& output, const PlatooningControlIHPPluginConfig& c)
   {
-    output << "PlatoonControlIHPPluginConfig { " << std::endl
+    output << "InLaneCruisingPluginConfig { " << std::endl
```
TonysCousin, line 69: Name is totally changed!  Please be careful that you are working with the latest code base.  It's a good idea to scan over a PR page after you create it to be sure everything is in order.

platooning_control_ihp/include/platoon_control_ihp.h
```diff
@@ -79,16 +73,14 @@ namespace platoon_control_ihp
         	std::shared_ptr<ros::CARMANodeHandle> nh_, pnh_;
 
 			// platoon control worker object
-        	PlatoonControlIHPWorker pcw_;
+        	PlatoonControlWorkerIHP pcw_;
```
TonysCousin, line 76: Please revert to original name.

platooning_control_ihp/src/platoon_control_ihp.cpp
```diff
@@ -21,7 +21,7 @@ namespace platoon_control_ihp
 // @SONAR_STOP@
     PlatoonControlIHPPlugin::PlatoonControlIHPPlugin()
     {
-        pcw_ = PlatoonControlIHPWorker();
+        pcw_ = PlatoonControlWorkerIHP();
```
TonysCousin, line 24: Fix name.

platooning_control_ihp/src/platoon_control_ihp.cpp
```diff
@@ -49,25 +49,19 @@ namespace platoon_control_ihp
         pnh_->param<double>("lowpassGain", config.lowpassGain, config.lowpassGain);
         pnh_->param<double>("lookaheadRatio", config.lookaheadRatio, config.lookaheadRatio);
         pnh_->param<double>("minLookaheadDist", config.minLookaheadDist, config.minLookaheadDist);
-
-        // Global params (from vehicle config)
```
TonysCousin, line 53: Could you please restore this comment, as it helps the reader to understand that the remaining params are not from the local config file.

platooning_control_ihp/src/platoon_control_ihp.cpp
```diff
@@ -49,25 +49,19 @@ namespace platoon_control_ihp
...
         pnh_->getParam("/vehicle_id", config.vehicleID);
         pnh_->getParam("/vehicle_wheel_base", config.wheelBase);
-        pnh_->getParam("/control_plugin_shutdown_timeout", config.shutdownTimeout);
-        pnh_->getParam("/control_plugin_ignore_initial_inputs", config.ignoreInitialInputs);
 
         pcw_.updateConfigParams(config);
         config_ = config;
 
 	  	// Trajectory Plan Subscriber
-        //   TODO: update trajectory executor to call this plugin
```
TonysCousin, line 63: Are we certain this TODO is no longer relevant (@sainaramyar )?  I don't see any mods to the trajectory executor here, so what is different for IHP?

platooning_control_ihp/src/platoon_control_ihp.cpp
```diff
@@ -49,25 +49,19 @@ namespace platoon_control_ihp
...
-		trajectory_plan_sub = nh_->subscribe<cav_msgs::TrajectoryPlan>("PlatoonControlIHPPlugin/plan_trajectory", 1, &PlatoonControlIHPPlugin::trajectoryPlan_cb, this);
+		trajectory_plan_sub = nh_->subscribe<cav_msgs::TrajectoryPlan>("PlatooningControlPlugin/plan_trajectory", 1, &PlatoonControlIHPPlugin::trajectoryPlan_cb, this);
         
         // Current Twist Subscriber
         current_twist_sub_ = nh_->subscribe<geometry_msgs::TwistStamped>("current_velocity", 1, &PlatoonControlIHPPlugin::currentTwist_cb, this);
 
         // Platoon Info Subscriber
-        // TODO consider changing the topic name so it is specific to ihp plugins (not sure if it is necessary)
```
TonysCousin, line 70: Just a sanity check: is this TODO no longer needed?  If so, why?

platooning_control_ihp/src/platoon_control_ihp.cpp
```diff
@@ -49,25 +49,19 @@ namespace platoon_control_ihp
...
-        pnh_->getParam("/control_plugin_shutdown_timeout", config.shutdownTimeout);
-        pnh_->getParam("/control_plugin_ignore_initial_inputs", config.ignoreInitialInputs);
 
         pcw_.updateConfigParams(config);
         config_ = config;
 
 	  	// Trajectory Plan Subscriber
-        //   TODO: update trajectory executor to call this plugin
-		trajectory_plan_sub = nh_->subscribe<cav_msgs::TrajectoryPlan>("PlatoonControlIHPPlugin/plan_trajectory", 1, &PlatoonControlIHPPlugin::trajectoryPlan_cb, this);
+		trajectory_plan_sub = nh_->subscribe<cav_msgs::TrajectoryPlan>("PlatooningControlPlugin/plan_trajectory", 1, &PlatoonControlIHPPlugin::trajectoryPlan_cb, this);
```
TonysCousin, line 59: If we're changing the name from Platoon... to Platooning..., then this message callback reference needs to change also.  Does this actually compile?

platooning_control_ihp/src/platoon_control_ihp.cpp
```diff
@@ -108,51 +92,23 @@ namespace platoon_control_ihp
...
-        {
-            ROS_DEBUG_STREAM("returning due to first data input");
-            return false;
+    void  PlatoonControlIHPPlugin::trajectoryPlan_cb(const cav_msgs::TrajectoryPlan::ConstPtr& tp){
+        
+        if (tp->trajectory_points.size() < 2) {
+            ROS_WARN_STREAM("PlatoonControlPlugin cannot execute trajectory as only 1 point was provided");
+            return;
         }
+        cav_msgs::TrajectoryPlanPoint first_trajectory_point = tp->trajectory_points[1]; // TODO this variable appears to be misnamed. It is the second trajectory point
```
TonysCousin, line 102: I am getting more concerned that again we are not working from the same baseline code.  This and the next few lines don't match what is in the baseline (old version in this PR, several lines below on this page).  Yet it looks like you haven't attempted to modify them, as the TODO is left over from several weeks ago.

platooning_control_ihp/src/platoon_control_ihp.cpp
```diff
@@ -262,7 +226,7 @@ namespace platoon_control_ihp
 
     void PlatoonControlIHPPlugin::generateControlSignals(const cav_msgs::TrajectoryPlanPoint& first_trajectory_point, const cav_msgs::TrajectoryPlanPoint& lookahead_point){
 
-        pcw_.setCurrentSpeed(trajectory_speed_); //TODO why this and not the actual vehicle speed?  Method name suggests different use than this.
```
TonysCousin, line 265: Are we sure this TODO has been addressed and is no longer relevant?  If things are okay as is, it would be good to add some clarity for future maintainers by either a) changing method name to reflect true meaning or b) add a comment explaining why it is being used this way.

TonysCousin (changes requested):
I see several snippets of old control plugin code in the green lines here, which makes me concerned that you have applied your mods to an old version of the code.  Let's please carefully check that you have the latest feature/noetic_ucla_ihp branch pulled and that the plugin correctly builds with your changes in place.  Finally, do a personal review of all the changes visible in the PR before Saina & I look it over.

platooning_control_ihp/src/platoon_control_ihp_worker.cpp
```diff
@@ -22,30 +22,98 @@
...
 
-    void PlatoonControlIHPWorker::setCurrentPose(const geometry_msgs::PoseStamped msg)
+    void PlatoonControlWorkerIHP::setCurrentPose(const geometry_msgs::PoseStamped msg)
 	{
 		current_pose_ = msg.pose;
 	}
 
-    void PlatoonControlIHPWorker::generateSpeed(const cav_msgs::TrajectoryPlanPoint& point)
+    // UCLA: Put the IHP gap regulation at low level control plug-in
+    double PlatoonControlWorkerIHP::getIHPDesPosFollower(double leaderCurrentPosition)
```
saina-ramyar, line 49: I think I need a little more explanation of how this function calculates the desired position for the host vehicle.

platooning_control_ihp/src/platoon_control_ihp_worker.cpp
```diff
@@ -22,30 +22,98 @@
...
+        double pos_g;      // desired downtrack position calculated with time-gap, in m.
+        double pos_h;      // desired downtrack position calculated with distance headaway, in m.
+
+        // 3. IHP gap regualtion 
+        // intermediate variables 
+        double timeGapAndStepRatio = desirePlatoonGap/ctrl_config_.time_step;     // The ratio between desired platoon time gap and the current time_step.
+        double totalTimeGap = desirePlatoonGap*pos_idx;                           // The overall time gap from host vehicle to the platoon leader, in s.
+
+        // calcilate pos_gap and pos_headway
+        if ((pred_spd <= ego_spd) && (ego_spd <= ctrl_config_.ss_theta))
```
saina-ramyar, line 87: It is unclear how what this condition is checking. 
In addition, you are calculating the pose_g and pse_h which are dependent on the gap, but I don't see the speed being used anywhere, can you explain?

saina-ramyar (changes requested):

platooning_control_ihp/src/platoon_control_ihp_worker.cpp
xuhan417: This is a gap regulation algorithm that uses the leader's info, preceding vehicle's info and host info to calculate a desired position.  This was implemented in the previous stages of the project. I will find the source code and make a documentation with Xin.

TonysCousin (commented):
Closing out an apparently open previous review, so that I can see the latest in everything.

platooning_control_ihp/include/platoon_control_ihp_worker.h
```diff
@@ -99,21 +99,36 @@ namespace platoon_control_ihp
         */
         void setCurrentSpeed(double speed);
 
+        /**
+         * \brief UCLA: Put the IHP gap regulation at low level control plug-in
```
TonysCousin, line 103: Description doesn't give any hint at what this method is actually doing.  What does "DesPos" mean?  What quantity is being gotten?  Also, need description of the param and of the return value, including units of measure.

platooning_control_ihp/src/platoon_control_ihp_worker.cpp
TonysCousin: Agree, there needs to be comments explaining what is going on here. The design doc doesn't mention anything about two different conditions for this calc.

platooning_control_ihp/src/platoon_control_ihp_worker.cpp
```diff
@@ -45,6 +45,74 @@ namespace platoon_control_ihp
...
+        {
+            // pos_g 
+            pos_g = (pred_pos - ctrl_config_.vehicleLength - ctrl_config_.standstill + ego_pos*timeGapAndStepRatio) / (1 + timeGapAndStepRatio);
+            // pos_h
+            double pos_h_nom = (pred_pos - ctrl_config_.standstill + ego_pos*(totalTimeGap + tmp_time_hdw)/ctrl_config_.time_step);
```
TonysCousin, line 92: I'm concerned about use of tmp_time_hdw here, since I believe it represents all vehicles between leader and host; however, this equation is using pred_pos, which only represents the immediate predecessor.  It seems this is a mixture of gap control and headway control.  Also, it seems the sum totalTimeGap + tmp_time_hdw is sort of double-counting the time gap between leader and host.  If this is correct, then a comment explaining it would be valuable, since, again, this sum doesn't appear in the design doc.
This comment applies to code in the else block below, also.

TonysCousin (changes requested):
My final review is now complete.  Please look for the 7 new comments.

platooning_control_ihp/include/platoon_control_ihp_worker.h
```diff
@@ -100,12 +100,12 @@ namespace platoon_control_ihp
         void setCurrentSpeed(double speed);
 
         /**
-         * \brief UCLA: Put the IHP gap regulation at low level control plug-in
+         * \brief UCLA: HP gap regulation that calculate the deisred position (Dtd, in m) for the platoon members.
```
TonysCousin, line 103: Still missing description of the parameter

TonysCousin (changes requested):

platooning_control_ihp/src/platoon_control_ihp_worker.cpp
```diff
@@ -36,7 +35,8 @@ namespace platoon_control_ihp
         pp_.config_ = new_config;
     }
 
-	double PlatoonControlIHPWorker::getLastSpeedCommand() const {
+	double PlatoonControlIHPWorker::getLastSpeedCommand() const
```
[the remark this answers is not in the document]
xuhan417: Em, I change the indent to four spaces instead of tab. I think that might caused this problem, it looks OK on my machine.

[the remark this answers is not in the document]
xuhan417: Emm, it looks OK on my machine. I changed the indent to four spaces instead of tab, this might caused the issue.

saina-ramyar (approved):

Issue #1725, opened 2022-04-14
Improving the entering time (ET) estimation in lci_strategic_plugin


Saeid-Solei:
### Types of Issue

- [X] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Currently, the vehicle adds a configurable time buffer to the estimated entering time (ET) for the first time after it enters the Trajectory Smoothing (TS) area regardless of whether the initially estimated ET is within red, yellow, or green phase. To guarantee that the vehicle will not enter the intersection box at the end of a red phase and before the signal light changes from red to green, this time buffer must be added if the initially estimated ET is within a yellow\red phase (i.e., if the vehicle must enter the intersection box at the beginning of a green phase). However, if the initially estimated ET is within a green phase, there is no need to add the time buffer to the estimated ET. This causes the vehicle to slow down while it can continue moving with the speed limit and enter the intersection box in a middle of a green phase. To fix this issue, a check must be added to the code to add the time buffer to the initially estimated ET only if it is within a yellow\red phase.

### Carma version where this issue was discovered

CARMA 4.0.2

### Expected behavior

If the vehicle can continue with the speed limit and enter the intersection box during a green phase (i.e., if the initially estimated ET is already within a green phase), the time buffer shall not be added to the initially estimated ET and the vehicle shall not slow down and should continue moving toward the intersection with the speed limit.

### Actual behavior

Even if the initially estimated ET is within a green phase (i.e., the vehicle can continue with speed limit and enter the intersection box within a green phase), the time buffer is added to it which causes the vehicle to slow down when it enters the TS area.

### Steps to reproduce the actual behavior

See the "Descriptive summary" section

MishkaMN:
Potential fix includes on this branch
https://github.com/usdot-fhwa-stol/carma-platform/tree/fix/green-buffer

MishkaMN:
Switched from enhancement to anomaly as this behavior was not the intended design.

PR #1748, opened 2022-05-02
fix green buffer for TSMO UC2 algorithim

MishkaMN:
# PR Details
## Description

Fixed the issue where the green buffer is applied even when the vehicle is able to make the green signal with certainty. 

New implementation will first check if the vehicle can make early arrival (-2 sec), if so, do not apply green buffer, and cache the normal arrival time without buffer. The reason we do not consider the late arrival here is because this part is only for getting ET in green phase (+2 might make ET to yellow), and there is extra logic later that checks if vehicle can make it with certainty (+2/-2). 

Resulting implementation is that buffer will only be added if ET is within buffer range from yellow/red.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1725

## Motivation and Context
TSMO UC 2

## How Has This Been Tested?
Tested on Black Pacifica

## Types of changes

## Checklist:


lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -602,9 +602,45 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
...
+    bool can_make_early_arrival  = (early_arrival_state_green_et_optional.get().second == lanelet::CarmaTrafficSignalState::PROTECTED_MOVEMENT_ALLOWED);
+   
+    // nearest_green_entry_time is by definition on green, so only check early_arrival
+    if (can_make_early_arrival)  // Green light with Certainty
+    {
+      nearest_green_entry_time_cached_ = nearest_green_entry_time;  //don't apply buffer if ET is in green
+    }  
+    else
+    {
+      ros::Time early_arrive_signal_end_time = ros::Time(lanelet::time::toSec(early_arrival_state_green_et_optional.get().first));
```
Saeid-Solei, line 631: Is this parameter always set to the beginning of the last green phase before the estimated ET? If yes, everything is ok and we can merge it.

Saeid-Solei (approved):

lci_strategic_plugin/src/lci_strategic_plugin.cpp
MishkaMN: addressed it

PR #1761, opened 2022-05-10
Feature/uc3

aswath1:
# PR Details
## Description
For TSMO Use Case 3, CARMA Streets will be providing a schedule message to CARMA Platform containing the entry time for the signalized intersection the vehicle plans to traverse. This task is to formalize the design for what those code changes will entail. The expectation is that the existing lci_strategic_plugin can be updated so that it uses the schedule from the intersection for setting its ET algorithm parameters instead of computing them based on spat if the message is available. If the message is not available, the existing logic should be executed to compute these values.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1762

## How Has This Been Tested?
Unit Tested

## Types of changes

## Checklist:


lci_strategic_plugin/config/parameters.yaml
```diff
@@ -35,6 +35,15 @@ desired_distance_to_stop_buffer : 15.0
 # Double: The minimum period in seconds which a maneuver plan must cover if the plugin wishes to control the whole plan
 min_maneuver_planning_period : 15.1
 
+# Double: Update time interval of carma streets. Units in s
+delta_t : 1.0
```
msmcconnell, line 39: This variable should have a more descriptive name. Like carma_streets_update_interval or something. delta_t is particularly confusing in a node containing a bunch of kinematic calculations

lci_strategic_plugin/config/parameters.yaml
```diff
@@ -35,6 +35,15 @@ desired_distance_to_stop_buffer : 15.0
...
+
+# Double: Minimum vehicle reaction time. Units in s
+reaction_time : 5.0
+
+#Int: Switch case for usecase 2 and usecase 3; 0 for UC2 and 1 for UC3
+case_switch : 0
```
msmcconnell, line 45: Can you describe the usage of this flag? Since the only values listed are 0 and 1 it implies that choosing UC2 or UC3 is mutually exclusive. This is not what we want. The vehicle should dynamically determine if the upcoming intersection supports Map/Spat only (UC2) or if it is equipped with carma-streets based on the broadcast from the intersection.

For example if the west intersection did not have carma-streets but the east intersection did, we chould be able to go through both intersections in a single run without changing the configuration. 

If you want a flag to disable UC2 or UC3 behavior that is fine, but the dynamic choice needs to be supported.

lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
```diff
@@ -42,6 +45,13 @@
 
 namespace lci_strategic_plugin
 {
+
+  enum TurnDirection {
```
msmcconnell, line 49: use ```enum class TurnDirection``` as C++ best practice

lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
```diff
@@ -125,7 +135,78 @@ class LCIStrategicPlugin
...
+  unsigned long long scheduled_depart_time_ = 0;
+  // scheduled latest depart position
+  uint32_t scheduled_departure_position_ = std::numeric_limits<uint32_t>::max();
+  // flag to show if the vehicle is allowed in intersection
+  bool is_allowed_int_ = false;
+  
+  //BSM
+  std::string bsm_id_ = "default_bsm_id";
+  uint8_t bsm_msg_count_ = 0;
+  uint16_t bsm_sec_mark_ = 0;
```
msmcconnell, line 207: Any particular reason for making these variables public when the others are private?

lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin_config.h
```diff
@@ -65,6 +65,21 @@ struct LCIStrategicPluginConfig
   //! The minimum period in seconds which a maneuver plan must cover if the plugin wishes to control the whole plan
   double min_maneuver_planning_period = 15.1;
 
+  //! Double: Approximate update time interval of carma streets
+  double delta_t = 1.0;
+
+  //! Double: Vehicle reaction time to a received schedule in seconds (approximate value, only used for communication with the schedule)
+  double reaction_time = 5.0;
```
msmcconnell, line 72: This is a pretty high default value. I think 2 would be more appropriate. Also change in yaml

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -851,6 +860,139 @@ void LCIStrategicPlugin::planWhenDEPARTING(const cav_srvs::PlanManeuversRequest&
...
+
+void LCIStrategicPlugin::BSMCb(const cav_msgs::BSMConstPtr& msg)
+{
+  std::vector<uint8_t> bsm_id_vec = msg->core_data.id;
+  bsm_id_ = BSMHelper::BSMHelper::bsmIDtoString(bsm_id_vec);
+  bsm_msg_count_ = msg->core_data.msg_count;
+  bsm_sec_mark_ = msg->core_data.sec_mark;
+}
+
+void LCIStrategicPlugin::parseStrategyParams(const std::string& strategy_params)
```
msmcconnell, line 892: Its possible that we might want to use this functionality while running other operations like platooning. So not every mobility operation message can be assumed to be for this plugin. Therefore, there needs to be a check on the strategy before this is called to ensure its intended for this plugin. If its not, the message should be ignored

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -851,6 +860,139 @@ void LCIStrategicPlugin::planWhenDEPARTING(const cav_srvs::PlanManeuversRequest&
...
+    approaching_light_controlled_interction_ = true;
+    ROS_DEBUG_STREAM("Approaching light Controlled Intersection: " << approaching_light_controlled_interction_);
+
+    if (msg->m_header.recipient_id == config_.vehicle_id)
+      {
+        street_msg_timestamp_ = msg->m_header.timestamp;
+        ROS_DEBUG_STREAM("street_msg_timestamp_: " << street_msg_timestamp_);
+        parseStrategyParams(msg->strategy_params);
+      }
+    previous_strategy_params_ = msg->strategy_params;
```
msmcconnell, line 879: Why are we caching this even when the message is not meant for this vehicle?

lci_strategic_plugin/src/main.cpp
```diff
@@ -89,6 +102,8 @@ int main(int argc, char** argv)
         tf_distance_pub.publish(tf_distance);
         earliest_et_pub.publish(earliest_et);
         et_pub.publish(scheduled_et);
+
+        lcip.onSpin();
```
msmcconnell, line 106: It would be better to create another timer to call this function at a configurable rate such that you can adjust the frequency of mobilityOperation publication without impacting the plugin discovery frequency or needing to rebuild. 

Also onSpin() could be renamed to something more meaningful since it only relates to mobilityOperation publication.

msmcconnell (changes requested):

lci_strategic_plugin/src/lci_strategic_plugin.cpp
aswath1: Please check 867 its already there.

lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin_config.h
```diff
@@ -65,6 +65,24 @@ struct LCIStrategicPluginConfig
...
+  //! Double: Approximate update time interval of carma streets
+  double carma_streets_update_interval = 1.0;
+  
+  //! Double: Vehicle reaction time to a received schedule in seconds (approximate value, only used for communication with the schedule)
+  double reaction_time = 2.0;
+
+  //! Double: Minimum inter-vehicle gap
+  double min_gap = 10.0;
+
+  //! Bool: Minimum inter-vehicle gap
```
msmcconnell, line 77: Incorrect comment

lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin_config.h
```diff
@@ -65,6 +65,24 @@ struct LCIStrategicPluginConfig
...
+  bool case_switch = false;
```
msmcconnell, line 78: Can we rename this to enable_carma_streets_connection

lci_strategic_plugin/config/parameters.yaml
```diff
@@ -35,6 +35,20 @@ desired_distance_to_stop_buffer : 15.0
...
 min_maneuver_planning_period : 15.1
 
+# Double: Update time interval of carma streets. Units in s
+carma_streets_update_interval : 1.0
+
+# Double: Minimum vehicle reaction time. Units in s
+reaction_time : 2.0
+
+# Bool: Switch case for usecase 2 and usecase 3; 0 for UC2 and 1 for UC3
+case_switch : 0
```
msmcconnell, line 45: See my naming comment below for config.h. 
Also the comment here needs to be updated based on that.

msmcconnell (changes requested):

lci_strategic_plugin/config/parameters.yaml
```diff
@@ -41,8 +41,8 @@ carma_streets_update_interval : 1.0
 # Double: Minimum vehicle reaction time. Units in s
 reaction_time : 2.0
 
-# Bool: Switch case for usecase 2 and usecase 3; 0 for UC2 and 1 for UC3
-case_switch : 0
+# Bool: If enable_carma_streets_connection is true when we want to allow carma streets functionality (UC3) and if its false that means we don't want to allow carma streets behaviour and will only use UC2 behaviour.
+ enable_carma_streets_connection : 0
```
msmcconnell, line 45: I think this may fail to load at runtime if its an int, please make **false**

msmcconnell (changes requested):

msmcconnell (approved):

PR #1770, opened 2022-05-16
Initial implementation of incremental build support

kjrush:
# PR Details
## Description

Adds --ros-1-packages and --ros-2-packages options to build-image.sh to support incremental docker builds.

## Related Issue

N/a

## Motivation and Context

This change enables usage of previous builds as a baseline for an incremental build. By using the newly implemented options, builds may be limited to only the set of packages with changes, minimizing the total time required to build a CARMA Platform image for development.

## How Has This Been Tested?

Local build successfully generates an image, have not yet been able to test the correctness of that image.

## Types of changes

## Checklist:


docker/install.sh
```diff
@@ -23,31 +23,61 @@ set -ex
 # ROS1 installation
 ###
 # Source the autoware installation
-source /opt/autoware.ai/ros/install/setup.bash
+
+if [[ ! -z "$ROS1_PACKAGES$ROS2_PACKAGES" ]]; then
```
msmcconnell, line 27: Is having no space between these two variables ok?
I'm thinking its probably fine if checking for not empty but want to confirm.

docker/install.sh
```diff
@@ -23,31 +23,61 @@ set -ex
...
 
 cd ~/carma_ws
 
 echo "Building ROS2 CARMA Components"
 
-colcon build --install-base /opt/carma/install_ros2 --build-base build_ros2 --cmake-args -DCMAKE_BUILD_TYPE=Release
+if [[ ! -z "$ROS1_PACKAGES$ROS2_PACKAGES" ]]; then
+    if [[ ! -z "$ROS2_PACKAGES" ]]; then
+        echo "Incrementally building ROS2 packages: $ROS2_PACKAGES"
+        colcon build --install-base /opt/carma/install_ros2 --cmake-args -DCMAKE_BUILD_TYPE=Release --packages-above $ROS2_PACKAGES --allow-overriding $ROS2_PACKAGES
```
msmcconnell, line 74: Probably good to specify the --build-base here as well

docker/build-image.sh
```diff
@@ -44,9 +44,61 @@ while [[ $# -gt 0 ]]; do
...
             ;;
+        --ros-1-packages)
+            ROS1_PACKAGES=""
+            shift
+            ;;
+        --ros-2-packages)
+            ROS2_PACKAGES=""
+            shift
+            ;;
+        *)
```
msmcconnell, line 55: I don't suppose there's anyway to notify the user if these options were provided out of order?

kjrush: I'll look into it, the problem mostly comes in with how I'm accumulating the lists below, once it sees --ros-2-packages it assumes means that --ros-1-packages has already been processed and that it can consume all further inputs. It might be hard to differentiate between the case where --ros-1-packages is not set, and the case where it is set second. Maybe I can check if one of the options has been specified, but no packages were provided and throw an error there?

docker/install.sh
kjrush: Yeah, it's just concatenating them and checking if the result is empty, I'm using this conditional a lot to see if the incremental build option was selected without need for a separate flag.

docker/install.sh
kjrush: Good catch, I missed that option!

docker/install.sh
```diff
@@ -23,31 +23,61 @@ set -ex
...
 sudo chgrp carma /opt/carma # Set group to expose permissions for build
 
-echo "Building ROS1 CARMA Components"
-
-colcon build --install-base /opt/carma/install --cmake-args -DCMAKE_BUILD_TYPE=Release
-
+if [[ ! -z "$ROS1_PACKAGES$ROS2_PACKAGES" ]]; then
+    if [[ ! -z "$ROS1_PACKAGES" ]]; then
+        echo "Incrementally building ROS1 packages: $ROS1_PACKAGES"
+        colcon build --install-base /opt/carma/install --cmake-args -DCMAKE_BUILD_TYPE=Release --packages-above $ROS1_PACKAGES --allow-overriding $ROS1_PACKAGES
```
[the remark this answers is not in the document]
msmcconnell: We can try it like this and remove it if we see issues

msmcconnell (approved):

PR #1777, opened 2022-05-20
Updates to IHP develop merge branch

saina-ramyar:
# PR Details
## Types of changes

## Checklist:

### DIFF
--- a/docker/checkout.bash
+++ b/docker/checkout.bash
@@ -45,9 +45,9 @@ if [[ "$BRANCH" = "develop" ]]; then
       git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-utils.git --branch $BRANCH
       git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-messenger.git --branch $BRANCH
 else
-      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch develop
+      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch feature/ihp_develop_merge
       git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-utils.git --branch develop
-      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-messenger.git --branch develop
+      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-messenger.git --branch feature/ihp_develop_merge
 fi
 
 # add astuff messages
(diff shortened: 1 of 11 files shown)


docker/checkout.bash
```diff
@@ -45,9 +45,9 @@ if [[ "$BRANCH" = "develop" ]]; then
       git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-utils.git --branch $BRANCH
       git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-messenger.git --branch $BRANCH
 else
-      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch feature/ihp_develop_merge
+      git clone --depth=1 https://github.com/usdot-fhwa-stol/carma-msgs.git --branch develop
```
TonysCousin, line 48: I've never understood the logic in this section. The if block above guarantees that $BRANCH == "DEVELOP", yet uses the variable name in its commands instead of just hard-coding the word "develop".  But in the else block, where $BRANCH could be anything, we hard-code the word "develop" so that $BRANCH is simply ignored.  Seems totally backwards.  However, this isn't the place to argue the legacy logic, so I'll let it go.  But if you have some insight, I'd love to hear it.

platooning_control/src/platoon_control.cpp
```diff
@@ -49,10 +49,13 @@ namespace platoon_control
         pnh_->param<double>("lowpassGain", config.lowpassGain, config.lowpassGain);
         pnh_->param<double>("lookaheadRatio", config.lookaheadRatio, config.lookaheadRatio);
         pnh_->param<double>("minLookaheadDist", config.minLookaheadDist, config.minLookaheadDist);
+        pnh_->param<double>("wheelBase", config.wheelBase, config.wheelBase);
```
TonysCousin, line 52: @saina-ramyar I'm curious as to why this param is moved from pnh_->getParam() to this? How are the two access methods different and why only this one changed?

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -494,11 +494,12 @@ namespace platoon_strategic_ihp
                 status_msg.host_platoon_position = pm_.getNumberOfVehicleInFront();
                 ROS_DEBUG_STREAM("pm platoonsize: " << pm_.getTotalPlatooningSize() << ", platoon_leader " << platoon_leader.staticId);
 
+                // TODO: To uncomment the following lines, platooninfo msg must be updated
```
TonysCousin, line 497: Is this message fix just not a priority right now?

TonysCousin (approved):

platooning_control/src/platoon_control.cpp
saina-ramyar: The way it was before, it was using the vehicle's wheelbase value. Now, it is reading the wheelbase value from the plugin's config. I did it so I can see the impact of wheelbase value on the lane-keeping performance.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
saina-ramyar: It should be updated soon, before trying to test the IHP control plugin.

PR #1779, opened 2022-05-25
Update IHP platoon gap calculation with multiple vehciles in between

saina-ramyar:
# PR Details
## Description

IHP strategic plugin is updated to calculate the gap with dynamic leader, in case there are multiple vehcles are in between.
## Types of changes

## Checklist:

### DIFF
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -340,7 +340,7 @@ namespace platoon_strategic_ihp
         // UCLA: add indicator of lane change 
         bool safeToLaneChange = false;
 
-        int dynamic_leader_index_ = 0;
+        size_t dynamic_leader_index_ = 0;
         
     private:
 
--- a/platooning_strategic_IHP/include/platoon_manager_ihp.h
+++ b/platooning_strategic_IHP/include/platoon_manager_ihp.h
@@ -339,6 +339,8 @@ namespace platoon_strategic_ihp
         bool isCreateGap = false;
         // UCLA: add indicator of lane change 
         bool safeToLaneChange = false;
+
+        int dynamic_leader_index_ = 0;
         
     private:
 
--- a/platooning_strategic_IHP/src/platoon_manager_ihp.cpp
+++ b/platooning_strategic_IHP/src/platoon_manager_ihp.cpp
@@ -192,8 +192,9 @@ namespace platoon_strategic_ihp
             for (size_t i = 0;  i < platoon.size();  ++i)
             {
                 std::string hostFlag = " ";
-                if (i == hostPosInPlatoon_)
+                if (platoon[i].staticId == getHostStaticID)
                 {
+                    hostPosInPlatoon_ = i;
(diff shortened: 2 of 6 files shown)


platooning_strategic_IHP/src/platoon_manager_ihp.cpp
```diff
@@ -343,11 +345,11 @@ namespace platoon_strategic_ihp
         // If the distance headway between the subject vehicle and its predecessor is an issue
         // according to the "min_gap" and "max_gap" thresholds, then it should follow its predecessor
         // The following line will not throw exception because the length of downtrack array is larger than two in this case
-        double distHeadwayWithPredecessor = downtrackDistance[downtrackDistance.size() - 2] - downtrackDistance[downtrackDistance.size() - 1];
+        double distHeadwayWithPredecessor = downtrackDistance[hostPosInPlatoon_ - 2] - downtrackDistance[hostPosInPlatoon_ - 1];
```
TonysCousin, line 348: I think this needs to be 
downtrackDistance[hostPosInPlatoon_ - 1] - downtrackDistance[hostPosInPlatoon]
If host is 2nd car in platoon then hostPostInPlatoon_ - 2 is -1, which will blow up the vector.

platooning_strategic_IHP/include/platoon_manager_ihp.h
```diff
@@ -339,6 +339,8 @@ namespace platoon_strategic_ihp
         bool isCreateGap = false;
         // UCLA: add indicator of lane change 
         bool safeToLaneChange = false;
+
+        int dynamic_leader_index_ = 0;
```
TonysCousin, line 343: Since this variable is an index into a vector, it would be preferred to type it as size_t for consistency.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -494,6 +494,14 @@ namespace platoon_strategic_ihp
                 status_msg.host_platoon_position = pm_.getNumberOfVehicleInFront();
                 ROS_DEBUG_STREAM("pm platoonsize: " << pm_.getTotalPlatooningSize() << ", platoon_leader " << platoon_leader.staticId);
 
+                int numOfVehiclesGaps = pm_.getNumberOfVehicleInFront() - pm_.dynamic_leader_index_;
```
TonysCousin, line 497: The naming seems pretty clear, but I'm not sure if this is doing what it appears to be.  If host is 3rd car, then getNumberOfVehiclesInFront() should return 2, yes?  And dynamic_leader_index_ would normally be 1 (the immediately preceding vehicle), yes?
Maybe worth a discussion.

TonysCousin (changes requested):

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
saina-ramyar: The hostPosInPlatoon_ is a private member. Aslo, what you are suggesting is resulting in negative value?

TonysCousin (approved):

PR #1782, opened 2022-05-26
Feature/plugin base library

msmcconnell:
# PR Details
## Description
This PR introduces an implementation of the plugin base library for outlined in this design
https://usdot-carma.atlassian.net/wiki/spaces/CRMPLT/pages/2182545409/Detailed+Design+-+Plugin+Library

I've gone ahead and implemented it to facilitate the creation of a plugin controller in ROS2 to replace the health_monitor. 

The core change (or really addition) to the design is the integration of the lifecycle state machine for plugins. Previously plugins had one core state which was ACTIVE. Now they follow the standard ROS2 lifecycle where they are active when the when their lifecycle state is active. To decouple the plugin states from the underlying state machine implementation the template pattern is used to hide the existing lifecycle callbacks. Plugins also have a concept of Availability which has not been fully defined in CARMA3. For this reason, the get_availability() function is provided but availability is not integrated into the full state machine. If we wish to implement it, that could be accomplished by adding it as an additional state on top of active. 

To make usage of the plugin base class easier the template_package has also been updated to allow for the user to specify a plugin type "strategic", "tactical", "control" and generate the appropriate extending classes. If the type is not provided, the package will default to a standard node as before. There is a fair bit of template code duplication in the template_package now which I spent a good deal of time trying to avoid but ultimately decided it was the cleanest solution. 

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1801

## Motivation and Context
Improved system usability and consistency

## How Has This Been Tested?
Unit tests and command line tests

## Types of changes

## Checklist:


carma_guidance_plugins/include/carma_guidance_plugins/plugin_base_node.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+  private:
+    // Publishers
+    //! Publisher for plugin discovery. This publisher will begin publishing 
+    //immediately on node construction regardless of lifecycle state. This is meant to allow for fast plugin activation on startup.
+    rclcpp::Publisher<carma_planning_msgs::msg::Plugin>::SharedPtr plugin_discovery_pub_;
+
+    // Timers
+    //! Timer to trigger publication of the plugin discovery message at a fixed frequency 
+    rclcpp::TimerBase::SharedPtr discovert_timer_;
```
kjrush, line 49: typo: discovert -> discovery

carma_guidance_plugins/include/carma_guidance_plugins/plugin_base_node.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+    // WorldModel listener
+    carma_wm::WMListener wm_listener_;
```
kjrush, line 52: Is there overhead in having this in the base? What's the likelihood that a plugin doesn't need the WM? Doesn't seem super likely outside of really simple cases right?

carma_guidance_plugins/include/carma_guidance_plugins/plugin_base_node.hpp
```diff
@@ -0,0 +1,189 @@
...
+    //! Virtual destructor for safe deletion
+    virtual ~PluginBaseNode() = default;
+
+    /**
+     * \brief Returns the activation status of this plugin.
+     *        The plugins API callbacks will only be triggered when this method returns true. 
+     * 
+     * \return True if plugin is in the ACTIVE state. False otherwise.
+     */ 
+    virtual bool get_activation_status() final;
```
kjrush, line 77: Not sure I understand the usage of `virtual ... final` here, they seem at odds. Maybe the implementation will clarify.

carma_guidance_plugins/include/carma_guidance_plugins/plugin_base_node.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+  /**
+   * \brief PluginBaseNode provides default functionality for all carma guidance plugins.
+   *        This includes basic state machine management (largely delegated to ROS2 lifecycle behavior), required interfaces, and plugin discovery
+   * 
+   * Extending classes must implement the on_configure_plugin method to load parameters, and my override the other state transitions methods on_<state>_plugin if desired.
+   * Additionally, extending classes must implement the methods such as get_name() which are used to populate the plugin discovery message.
+   * 
+   */
+  class PluginBaseNode : public carma_ros2_utils::CarmaLifecycleNode
```
kjrush, line 38: Is a system_alert handler built into the base class being extended here?

carma_guidance_plugins/src/control_plugin.cpp
```diff
@@ -0,0 +1,107 @@
...
+      std::bind(&ControlPlugin::current_twist_callback, this, std_ph::_1));
+
+    trajectory_plan_sub_ = create_subscription<carma_planning_msgs::msg::TrajectoryPlan>(get_name() + "/plan_trajectory", 1,
+      std::bind(&ControlPlugin::current_trajectory_callback, this, std_ph::_1));
+
+    vehicle_cmd_pub_ = create_publisher<autoware_msgs::msg::ControlCommandStamped>("ctrl_raw", 1);
+
+    command_timer_ = create_timer(
+        get_clock(),
+        std::chrono::milliseconds(33), // Spin at 30 Hz per plugin API
```
kjrush, line 70: Should this be configurable in some way? Maybe a virtual method (get_spin_freq()) or something? Or are we definitive that the API is 30hz, no higher?

carma_guidance_plugins/include/carma_guidance_plugins/control_plugin.hpp
```diff
@@ -0,0 +1,109 @@
...
+    // These callbacks do direct assignment into their respective member variables
+    void current_pose_callback(geometry_msgs::msg::PoseStamped::UniquePtr msg);
+    void current_twist_callback(geometry_msgs::msg::TwistStamped::UniquePtr msg);
+    void current_trajectory_callback(carma_planning_msgs::msg::TrajectoryPlan::UniquePtr msg);
+
+
+  protected:
+
+    //! The most recent pose message received by this node
+    boost::optional<geometry_msgs::msg::PoseStamped> current_pose_;
```
kjrush, line 62: Is it worth documenting somewhere in here that the subscriptions/etc. are all setup to support only single-threaded plugins? We don't have any multi-threaded plugins but, in the event we needed to create one, I don't think these structures here are synchronized, so we'd need to start from a different base class.

template_package/carma_package
```diff
@@ -50,6 +51,25 @@ if [[ ! -d ${package_parent_path} ]]; then
     exit 2
 fi
 
+plugin_type=$4
+
+if [ "$plugin_type" = "strategic" ] || [ "$plugin_type" = "tactical" ] || [ "$plugin_type" = "control" ]; then
```
kjrush, line 56: Huh, that's legal syntax? I had to look it up but I learned something today!

template_package/carma_package
```diff
@@ -64,9 +84,56 @@ if [[ $PWD != $(realpath ${target_path}) ]]; then
     exit 2
 fi
 
+# Cleanup alternative node files based on user specified type befoe substitutions
+if [ "$plugin_type" = "strategic" ]; then
```
kjrush, line 88: You could streamline this block a bit by inverting the conditionals:

```
if [ "$plugin_type"  = "none" ]; then
   delete the plugin stuff that won't be caught by the next branches
if [ ! "$plugin_type" = "strategic" ]; then
   delete the strategic template
if [ ! "$plugin_type" = "tactical" ]; then
   delete tactical template
```

etc. Little cleaner and less duplication that way.

template_package/carma_package
```diff
@@ -64,9 +84,56 @@ if [[ $PWD != $(realpath ${target_path}) ]]; then
...
+    rm src/template_tactical_plugin_node.cpp
+    rm src/template_control_plugin_node.cpp
+
+fi
+
 # Perform substitutions
 find . -type f -exec sed -i "s/<SUB><package_name>/${package_name}/g" {} \;
 find . -type f -exec sed -i "s/<SUB><year>/${current_year}/g" {} \;
+find . -type f -exec sed -i "s/<SUB><base_node_dep>/${base_node_dep}/g" {} \;
+#find . -type f -exec sed -i "s/<SUB><world_model_dep>/${world_model_dep}/g" {} \;
```
kjrush, line 134: Do we need these lines still?

carma_guidance_plugins/include/carma_guidance_plugins/plugin_base_node.hpp
```diff
@@ -0,0 +1,189 @@
...
+     * \return Name of this plugin
+     */ 
+    virtual std::string get_name() = 0;
+
+    /**
+     * \brief Returns the version id of this plugin.
+     * 
+     * \return The version id represented as a string
+     */ 
+    virtual std::string get_version_id() = 0;
```
kjrush, line 119: I can't recall, do we have specific format in mind for this version string? I know we use semver by convention, is it worth enforcing? I think we could do so with a non-virtual/final get_version_id() and pure virtual methods for get_major_version, get_minor_version, get_patch_version, or something like that.

template_package/carma_package
```diff
@@ -76,7 +143,14 @@ for f in ./**; do
         continue
     fi
 
+    # Here the files will be renamed.
```
kjrush, line 146: I'm a little confused by what this block is doing, but perhaps that's not relevant for this PR provided the below rename stuff works. Just seems like a lot of hoops to jump through to rename a folder?

template_package/include/template_package/template_control_plugin_node.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+    std::string get_version_id() override;
+    
+    /**
+     * \brief This method should be used to load parameters and will be called on the configure state transition.
+     */ 
+    carma_ros2_utils::CallbackReturn on_configure_plugin();
+
+    /**
+     * TODO for USER: The following lifecycle overrides are also available if needed
```
kjrush, line 102: I think this would make the most sense in the class-level doc comment right?

template_package/include/template_package/template_tactical_plugin_node.hpp
```diff
@@ -0,0 +1,115 @@
+/*
+ * Copyright (C) <SUB><year> LEIDOS.
```
kjrush, line 2: Does it make sense to make the copyright holder a parameter as well? Just thinking in case external users end up using these templates too. Though, it's a pretty easy find/replace after the fact.

carma_guidance_plugins/src/control_plugin.cpp
```diff
@@ -0,0 +1,107 @@
...
+  {
+    return "control/trajectory_control";
+  }
+
+  uint8_t ControlPlugin::get_type() 
+  {
+    return carma_planning_msgs::msg::Plugin::CONTROL; 
+  }
+
+  void ControlPlugin::current_pose_callback(geometry_msgs::msg::PoseStamped::UniquePtr msg)
```
kjrush, line 39: Does it make sense to add logging to these methods, just to have it? Should we consider adding it to the state transitions as well? Should be easy enough to wrap the user-defined callback with a higher-level one that simply logs the state transition then calls the user's callback, I think.

carma_guidance_plugins/include/carma_guidance_plugins/plugin_base_node.hpp
msmcconnell: Yeah this does not need to be virtual.

carma_guidance_plugins/include/carma_guidance_plugins/plugin_base_node.hpp
msmcconnell: CarmaLifecycleNode does not contain a system alert handler since it delegates that behavior to the lifecycle services. It does have a system alert publisher. 

Did you have a particular usage you were thinking of?

carma_guidance_plugins/src/control_plugin.cpp
msmcconnell: It needs to be 30Hz per the API. I guess there isn't much issue with it being higher though. 
I could implement a get_spin_freq() with a static_assert here to demand 30Hz or above. 

Thoughts?

template_package/carma_package
msmcconnell: Bash never ceases to confuse and confound 😄

template_package/carma_package
msmcconnell: nope i'll remove

template_package/include/template_package/template_tactical_plugin_node.hpp
msmcconnell: I have found multi-line substitutions across multiple files which include * and / extremely hard to get working reliably, so I would prefer to leave the standard default. external users could easily find and replace.

carma_guidance_plugins/src/control_plugin.cpp
msmcconnell: I believe, the state transitions already have logs from the carma base node. I can add debug logs to the callbacks. 

Can you expand on your last sentence. I'm not quite clear on what you're asking.

template_package/include/template_package/template_control_plugin_node.hpp
msmcconnell: In the doc comment of the plugin base class maybe, but I wouldn't put it in the template package. My goal with these it to use comments like this to point users to where specifically things should be written. Having it here keeps it co-located with the configure callback which is required to implement.

carma_guidance_plugins/include/carma_guidance_plugins/plugin_base_node.hpp
msmcconnell: There is definitely some overhead. I went back and forth on it, particularly for the control plugins which currently don't use it. 
I would personally be ok either way. It's only 2-3 extra lines for someone to do it manually. 
Any preference?

template_package/carma_package
msmcconnell: In order for the cmake to work the cpp file is specified. These lines ensure that all node/plugin types result in only one file name for usage in the cmake

template_package/carma_package
msmcconnell: Done

carma_guidance_plugins/include/carma_guidance_plugins/plugin_base_node.hpp
msmcconnell: At the moment we don't have an enforced format anywhere. And as far as I know nothing actually uses this field. Probably best to wait until we have time to think about enforcing plugin versions.

carma_guidance_plugins/include/carma_guidance_plugins/control_plugin.hpp
msmcconnell: Yeah, I'll add some comments. I think it would be a good idea to provide multi-threading support but I don't think its an immediate need.

carma_guidance_plugins/include/carma_guidance_plugins/plugin_base_node.hpp
kjrush: Is it possible to somehow make it so they're optionally defined in the class? Seems like the sort of thing templates should be able to support, but I'm not sure how exactly that'd work. Could definitely do it with an `#ifdef __USE_WM__ ...` macro block but that's kinda clunky too. I think if that's too much of a lift, let's just this down from the base plugin class to the strategic and tactical bases. We can always refactor it back to the base later if we find ourselves wanting the WM in control plugins.

carma_guidance_plugins/include/carma_guidance_plugins/plugin_base_node.hpp
kjrush: Do the lifecycle services then handle the shutdown when system_alert FATAL is published? I can't recall from that base class if that was included, or if that's baked into to each subsystem management node.

carma_guidance_plugins/src/control_plugin.cpp
kjrush: Should be fine to leave it like that, if that's what the API specifies. Thinking about it, it's not even a difficult refactor if we need to parameterize this later, since we could define a default virtual get_spin_freq method here to return 30 and not have to touch all of the plugins implemented before the refactor.

carma_guidance_plugins/src/control_plugin.cpp
kjrush: If the state transitions are logged then that's really what I was looking for, just to ensure we have records.

RE: the last sentence. I think I was getting a few wires crossed mentally, I was thinking for some reason these callbacks would be user defined (e.g., something like a `set_current_pose_cb(Callback cb)` kind of a pattern. That we could still ensure those were logged by wrapping them in a callback of our own (which just logs the message and then invokes the user-defined callback) before registering that callback with the subscriber. Doesn't particularly matter here, that's not the pattern we're using.

Digression aside, adding some debug level logs here wouldn't hurt.

carma_guidance_plugins/include/carma_guidance_plugins/plugin_base_node.hpp
msmcconnell: I'll move it to the strategic and tactical bases for now. Having a macro like that would require the user to modify the cmake file to set it which is not something most of the team does on the regular and would be confusing to others looking at example plugins.

carma_guidance_plugins/include/carma_guidance_plugins/plugin_base_node.hpp
msmcconnell: The subsytem controllers check for the FATAL and then send the lifecycle service calls based on their configured required nodes.

carma_guidance_plugins/include/carma_guidance_plugins/plugin_base_node.hpp
msmcconnell: Scratch that,
Per discussion on slack we will lazily initialize the world model listener as a way of only creating subscriptions if the plugin is using that variable.

msmcconnell: done

carma_guidance_plugins/src/control_plugin.cpp
msmcconnell: Added debug logs

kjrush (approved):

PR #1788, opened 2022-05-27
Port trajectory_executor to ROS2

JonSmet:
# PR Details
## Description

This PR includes updates to port the trajectory_executor node to ROS2 and to launch it from the guidance.launch.py launch file.

## Related Issue

Issue #1787 

## How Has This Been Tested?

Nodes builds and launches with carma-platform; all unit tests pass.

## Types of changes

## Checklist:


trajectory_executor/include/trajectory_executor/trajectory_executor_node.hpp
```diff
@@ -0,0 +1,111 @@
...
+    // Timers
+    rclcpp::TimerBase::SharedPtr timer_; // Timer for publishing outbound trajectories to the control plugins
+
+    // Node configuration
+    Config config_;
+
+    // Trajectory plan tracking data. Synchronized on cur_traj_mutex_
+    std::unique_ptr<carma_planning_msgs::msg::TrajectoryPlan> cur_traj_; 
+    int timesteps_since_last_traj_ {0};
+    std::mutex cur_traj_mutex_;
```
kjrush, line 58: Do we have multi-threading at play here?

trajectory_executor/src/main.cpp
```diff
@@ -14,20 +14,20 @@
...
+int main(int argc, char **argv) 
 {
-  ros::init(argc, argv, "trajectory_executor");
-  ROS_DEBUG("Entered main function for Trajectory executor, starting node...");
-  trajectory_executor::TrajectoryExecutor te;
+  rclcpp::init(argc, argv);
+
+  auto node = std::make_shared<trajectory_executor::TrajectoryExecutor>(rclcpp::NodeOptions());
+  
+  rclcpp::executors::MultiThreadedExecutor executor;
```
kjrush, line 26: It looks like we do indeed have multithreading here, is there a reason this changed from the original trajectory executor implementation?

trajectory_executor/src/trajectory_executor_node.cpp
```diff
@@ -0,0 +1,185 @@
...
+    return CallbackReturn::SUCCESS;
+  }
+
+  void TrajectoryExecutor::onTrajEmitTick()
+  {
+    std::unique_lock<std::mutex> lock(cur_traj_mutex_);
+    RCLCPP_DEBUG_STREAM(get_logger(), "TrajectoryExecutor tick start!");
+
+    if (cur_traj_ != nullptr) {
+      if (!cur_traj_->trajectory_points.empty()) {
```
kjrush, line 127: I guess I missed it during the removal of the trajectory trimming, but this conditional is no longer needed, really. Since we're not shrinking the trajectory itself as we advance along it any more, the only way for us to get an empty trajectory is for that to have been the output from the plan_delegator, which I don't think is possible (might be worth double checking if plan_delegator has safeguards for that).

trajectory_executor/src/trajectory_executor_node.cpp
```diff
@@ -0,0 +1,185 @@
...
+    RCLCPP_DEBUG_STREAM(get_logger(), "New plan contains " << msg->trajectory_points.size() << " points");
+
+    cur_traj_ = std::unique_ptr<carma_planning_msgs::msg::TrajectoryPlan>(move(msg));
+    timesteps_since_last_traj_ = 0;
+    RCLCPP_INFO_STREAM(get_logger(), "Successfully swapped trajectories!");  
+  }
+
+  void TrajectoryExecutor::guidanceStateMonitor(carma_planning_msgs::msg::GuidanceState::UniquePtr msg)
+  {
+    std::unique_lock<std::mutex> lock(cur_traj_mutex_); // Acquire lock until end of this function scope
```
kjrush, line 172: We can move this mutex inside the conditional, that way we only acquire it when we need it. Saves us from doing an unneeded syscall - or blocking in the worst case - when it's not needed.

trajectory_executor/src/main.cpp
JonSmet: This is part of the initial ROS2 code contained our template_package main.cpp file: https://github.com/usdot-fhwa-stol/carma-platform/blob/develop/template_package/src/main.cpp

Do you think this needs to be updated to match the original trajectory executor implementation?

trajectory_executor/src/trajectory_executor_node.cpp
JonSmet: Good catch. I verified that plan_delegator has safeguards in place that restrict it from publishing any trajectory with less than 2 points. I've gone ahead and removed this conditional.

trajectory_executor/src/trajectory_executor_node.cpp
JonSmet: Fixed.

PR #1792, opened 2022-06-01
Feature/carma wm ctrl ros2

MishkaMN:
# PR Details
## Types of changes

## Checklist:


carma_wm_ctrl/test/WMBroadcasterTest.cpp
```diff
@@ -569,19 +572,20 @@ TEST(WMBroadcaster, addAndRemoveGeofence)
...
 
   wmb.addGeofence(gf_ptr);
   ASSERT_EQ(map_update_call_count, 1);
   /*Analyze prev_regems_.size()*/
-  ASSERT_EQ(gf_ptr->prev_regems_.size(), 2);
-  ASSERT_EQ(gf_ptr->prev_regems_[0].first, 10007);
-  ASSERT_EQ(gf_ptr->prev_regems_[1].second->id(), old_speed_limit->id());
+  ASSERT_EQ(gf_ptr->prev_regems_.size(), 3);
+  ASSERT_EQ(gf_ptr->prev_regems_[0].first, 10000);
+  ASSERT_EQ(gf_ptr->prev_regems_[0].second->id(), old_speed_limit->id());
```
msmcconnell, line 588: Please confirm that all these unit test fixes make conceptual sense, not just updating to what the system is outputting as such a failure could indicate a bug.

carma_wm_ctrl/test/WMBroadcasterTest.cpp
```diff
@@ -2804,12 +2826,6 @@ ASSERT_EQ(value, p1);
   ASSERT_EQ(test_map_elem->speed_limit_.value(), new_speed_limit->speed_limit_.value());
   ASSERT_EQ(test_map_elem->participants_.begin()->data(), new_speed_limit->participants_.begin()->data());
 
-  // Set the map
-  wmb.baseMapCallback(map_msg_ptr);
-
-  sample_proj_string.data = proj_string;
-  wmb.geoReferenceCallback(sample_proj_string);
```
msmcconnell, line 2811: This got deleted but doesn't seem to be moved any where. Why delete it?

carma_wm_ctrl_ros2/include/carma_wm_ctrl_ros2/Geofence.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+// Map Update Geofence Common Labels
+const std::string MAP_MSG_INTERSECTION = "MAP_MSG_INTERSECTION"; 
+const std::string MAP_MSG_TF_SIGNAL = "MAP_MSG_TF_SIGNAL";
+
+
+/**
+ * @brief An object representing a geofence use for communications with CARMA Cloud
+ *
+ * TODO: This is currently a place holder class which needs to be updated based on the final geofence specification
```
msmcconnell, line 47: We can probably remove this extremely old comment

carma/launch/environment.launch.py
```diff
@@ -69,6 +69,37 @@ def generate_launch_description():
         get_package_share_directory('motion_computation'), 'config/parameters.yaml')
 
     env_log_levels = EnvironmentVariable('CARMA_ROS_LOGGING_CONFIG', default_value='{ "default_level" : "WARN" }')
+    
+    carma_wm_ctrl_ros2_param_file = os.path.join(
+        get_package_share_directory('carma_wm_ctrl_ros2'), 'config/parameters.yaml')
+
+    carma_wm_ctrl_container = ComposableNodeContainer(
```
msmcconnell, line 76: This doesn't need it's own container. Instead, just the ComposableNode should be added to external_objects_container to take advantage of intra-process comms.

engineering_tools/mem_update_install.sh
```diff
@@ -1,53 +0,0 @@
-#! /bin/bash
```
msmcconnell, line 1: This file should not be deleted

msmcconnell (changes requested):

engineering_tools/mem_update_install.sh
MishkaMN: added back

carma_wm_ctrl/include/carma_wm_ctrl/GeofenceSchedule.hpp
```diff
@@ -67,7 +68,7 @@ class GeofenceSchedule
    *
    * @return True if time > schedule_end
    */
-  bool scheduleExpired(const ros::Time& time = ros::Time::now()) const;
+  bool scheduleExpired(const rclcpp::Time& time = rclcpp::Time(0.0,0.0,RCL_SYSTEM_TIME)) const;
```
msmcconnell, line 71: Can you confirm this is a safe change. Is this called anywhere with the default arguments?

carma_wm_ctrl/src/GeofenceSchedule.cpp
```diff
@@ -14,92 +14,111 @@
...
 {
+
+double toSec(const boost::posix_time::time_duration& duration) {
+  if (duration.is_special()) {
+    throw std::invalid_argument("Cannot convert special duration to seconds");
+  }
+  return duration.total_microseconds() / 1000000.0;
+}
+
+double toSec(const boost::posix_time::ptime& time) {
```
msmcconnell, line 29: These look like they were copied and pasted from the methods in lanelet2_extension. Why not just include the timing header from that library and call them directly?

carma_wm_ctrl/src/GeofenceSchedule.cpp
```diff
@@ -14,92 +14,111 @@
...
+// returns rclcpp::Time(0) when the schedule is expired or the next interval will be on a different day of the week
 // Argument provided as absolute time (since 1970)
-std::pair<bool, ros::Time> GeofenceSchedule::getNextInterval(const ros::Time& time) const
+std::pair<bool, rclcpp::Time> GeofenceSchedule::getNextInterval(const rclcpp::Time& time) const
 {
   if (scheduleExpired(time))
   {
-    ROS_DEBUG_STREAM("Geofence schedule expired");
-    return std::make_pair(false, ros::Time(0));  // If the schedule has expired or was never started
+    RCLCPP_DEBUG_STREAM(rclcpp::get_logger("carma_wm_ctrl::GeofenceSchedule"), "Geofence schedule expired");
```
msmcconnell, line 62: Something I realized recently is that this may not match to the values based in from the log file correctly unless only the package name is provided, so "carma_wm_ctrl" is the way to go for now even though its less descriptive.

carma_wm_ctrl/src/GeofenceSchedule.cpp
```diff
@@ -14,92 +14,111 @@
...
   }
 
   // Iterate over the day to find the next control interval
   constexpr int num_sec_in_day = 86400;
-  const ros::Duration full_day(num_sec_in_day);
+  const rclcpp::Duration full_day(num_sec_in_day* 1e9);
   bool time_in_active_period = false;  // Flag indicating if the requested time is within an active control span
 
-  while (cur_start < full_day && ros_time_of_day > ros::Time(cur_start.toSec()))
+  while (cur_start < full_day && ros_time_of_day > rclcpp::Time(cur_start.seconds() * 1e9))
```
msmcconnell, line 102: You can directly call .nanoseconds() here which will remove floating point error and the need for 1*1e9
Check your other calls which use rclcpp::Duration as well.

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -1215,8 +1217,8 @@ void WMBroadcaster::setIntersectionCoordCorrection(const std::vector<int>& inter
 
   for (auto i = 0; i != intersection_correction.size(); i+2)
   {
-    sim_.intersection_coord_correction_[intersection_ids_for_correction[i/2]].first =  intersection_correction[i]; //x
-    sim_.intersection_coord_correction_[intersection_ids_for_correction[i/2]].second = intersection_correction[i + 1]; //y
+    sim_.intersection_coord_correction_[(int)intersection_ids_for_correction[i/2]].first =  intersection_correction[i]; //x
```
msmcconnell, line 1220: is int the most accurate cast here? Or int64_t since you were using that above

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -1716,9 +1722,9 @@ cav_msgs::TrafficControlRequest WMBroadcaster::controlRequestFromRoute(const cav
 
   tcr_polygon_ = composeTCRStatus(localPoint, cB, local_projector); // TCR polygon can be visualized in UI
 
-  cB.oldest =ros::Time::now(); // TODO this needs to be set to 0 or an older value as otherwise this will filter out all controls
+  cB.oldest =rclcpp::Time(0.0,0.0,RCL_SYSTEM_TIME); // TODO this needs to be set to 0 or an older value as otherwise this will filter out all controls
```
msmcconnell, line 1725: is RCL_SYSTEM_TIME the right default? What if we are running in simulation?

carma_wm_ctrl/src/main.cpp
```diff
@@ -14,16 +14,22 @@
...
 int main(int argc, char** argv)
 {
   // Initialize node
-  ros::init(argc, argv, "wm_broadcaster");
-  carma_wm_ctrl::WMBroadcasterNode node;
+  rclcpp::init(argc, argv);
+  
+  auto node = std::make_shared<carma_wm_ctrl::WMBroadcasterNode>(rclcpp::NodeOptions());
+
+  node->initializeWorker(node);
```
msmcconnell, line 28: How would this function get called when launched from a component?
This should probably be occurring automatically in handle_on_configure instead of here.

carma_wm_ctrl/test/GeofenceScheduleTest.cpp
```diff
@@ -14,82 +14,75 @@
...
+  ASSERT_TRUE(sch.scheduleStarted(rclcpp::Time(1579886341.000)));
 }
 
 TEST(GeofenceSchedule, getNextInterval)
 {
   // Test before start
 
-  GeofenceSchedule sch(ros::Time(1), ros::Time(6), ros::Duration(2), ros::Duration(1), ros::Duration(0), ros::Duration(1),
-                       ros::Duration(2)  // This means the next schedule is a 4 (2+2)
+  GeofenceSchedule sch(rclcpp::Time(1e10 * 1), rclcpp::Time(1e10 * 6), rclcpp::Duration(1e10 * 2), rclcpp::Duration(1e10 * 1), rclcpp::Duration(1e10 * 0), rclcpp::Duration(1e10 * 1),
```
msmcconnell, line 52: Shouldn't this be 1e9?

carma_wm_ctrl/test/GeofenceSchedulerTest.cpp
```diff
@@ -14,65 +14,70 @@
...
-                       ros::Time(8),
-                       ros::Duration(2),    // Starts at 2
-                       ros::Duration(3.5),  // Ends at by 5.5
-                       ros::Duration(0),    // repetition start 0 offset, so still start at 2
-                       ros::Duration(1),    // Duration of 1 and interval of 2 so active durations are (2-3 and 4-5)
-                       ros::Duration(2)));
-  ros::Time::setNow(ros::Time(0));  // Set current time
-
-  GeofenceScheduler scheduler(std::make_unique<TestTimerFactory>());  // Create scheduler
+      GeofenceSchedule(rclcpp::Time(1e10),  // Schedule between 1 and 8
```
msmcconnell, line 63: Ditto about 1e9 vs 1e10

msmcconnell (changes requested):

carma_wm_ctrl_ros2/include/carma_wm_ctrl_ros2/Geofence.hpp
MishkaMN: deleted

carma_wm_ctrl/test/WMBroadcasterTest.cpp
MishkaMN: yeah I confirmed this is correct as this is the first lanelet of the three lanelets detected 10000, 10007, 10006

carma_wm_ctrl/test/WMBroadcasterTest.cpp
MishkaMN: I did not understand what it was trying to test as it did not have any ASSERT calls, but just realized now that if it crashes it is in a way a test as it should not crash. So I added it back with right syntax

carma_wm_ctrl/src/GeofenceSchedule.cpp
MishkaMN: added to every logger

carma_wm_ctrl/src/GeofenceSchedule.cpp
MishkaMN: using nanoseconds now

carma_wm_ctrl/src/WMBroadcaster.cpp
MishkaMN: getting clocktype from scheduler now

carma_wm_ctrl/test/GeofenceScheduleTest.cpp
MishkaMN: changed to e9

carma_wm_ctrl/test/WMBroadcasterTest.cpp
```diff
@@ -328,76 +324,77 @@ TEST(WMBroadcaster, geofenceCallback)
...
+  
   // set the points
-  cav_msgs::PathNode pt;
+  carma_v2x_msgs::msg::PathNode pt;
   // check points that are inside lanelets
-  pt.x = -8.5; pt.y = -9.5; pt.z = 0; // straight geofence line across 2 lanelets
+  pt.x = 0.5; pt.y = 0.5; pt.z = 0; // straight geofence line across 2 lanelets
   msg_v01.geometry.nodes.push_back(pt);
-  pt.x = -8.5; pt.y = -8.5; pt.z = 0;
+  pt.x = 1.5; pt.y = 1.5; pt.z = 0;
```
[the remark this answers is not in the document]
MishkaMN: reverted to original points. All unit tests were not update from when the logic expected absolute coordinates while current logic require points to be offsets. This is why many of the tests were failing

carma_wm_ctrl/src/main.cpp
MishkaMN: deleted this since it is there in handle_on_configure

carma_wm_ctrl/src/GeofenceSchedule.cpp
MishkaMN: refactored the time namespace into different file and included it here

carma_wm_ctrl/test/WMBroadcasterTest.cpp
```diff
@@ -406,58 +403,58 @@ TEST(WMBroadcaster, geofenceCallback)
...
   
   //Test throw exceptions
-  ASSERT_THROW(wmb.routeCallbackMessage(route_msg), lanelet::InvalidObjectStateError);
-  ROS_INFO_STREAM("Throw Exceptions Test Passed.");
+  ASSERT_THROW(wmb.routeCallbackMessage(std::make_unique<carma_planning_msgs::msg::Route>(route_msg)), lanelet::InvalidObjectStateError);
+  RCLCPP_INFO_STREAM(rclcpp::get_logger("carma_wm_ctrl::WMBroadcasterTest"), "Throw Exceptions Test Passed.");
   
   // Load vector map from a file start 
-  std::string file = "resource/test_vector_map1.osm";
+  std::string file = "/workspaces/carma/src/carma-platform/carma_wm_ctrl/test/resource/test_vector_map1.osm";
```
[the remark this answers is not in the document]
MishkaMN: Somehow if I don't do this on local machine it won't work. Do you know how to make it work on local machine with relative path?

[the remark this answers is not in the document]
msmcconnell: I think these paths come from whatever the working directory is set to for the test. If its not specified in the cmake maybe try specifying it.

carma_wm_ctrl/include/carma_wm_ctrl/GeofenceSchedule.hpp
msmcconnell: I think that's fine

carma_wm_ctrl/src/WMBroadcaster.cpp
MishkaMN: I think actually the right cast is `uint16_t`[ since this is the format which SPAT/MAP sends to CARMA.](https://github.com/usdot-fhwa-stol/carma-platform/blob/develop/carma_wm/include/carma_wm/SignalizedIntersectionManager.h#L145) I specified the range in yaml to avid overlow

carma_wm_ctrl/test/WMBroadcasterTest.cpp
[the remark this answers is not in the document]
MishkaMN: gotcha thanks. set the working path there and it works. all unit tests pass now

msmcconnell (approved):

Issue #1499, opened 2021-10-26
Integrate plugin management into lifecycle controllers


msmcconnell:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

In the ROS2 version of carma-platform the lifecycle of nodes will be managed by the subsystem_controllers. At the moment there is a guidance_controller but exactly how the plugins should be interacted with has not been fully defined for this lifecycle structure. A unified approach should be implemented. 

### Carma version where this issue was discovered

foxy/develop
### Expected behavior
Guidance Plugins have a fully managed lifecycle
### Actual behavior

Currently treated like all other guidance nodes
### Steps to reproduce the actual behavior

N/A

Issue #1666, opened 2022-03-12
Cannot Enable or Disable Plugins from the Web UI


GoodarzMehr:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

When running CARMA, guidance plugins cannot be enabled or disabled from the web UI. Nothing happens when clicking on a plugin's button. We want to enable the Platooning Strategic plugin in particular but we cannot currently do that. Our browser is Google Chrome.

### Carma version where this issue was discovered

develop

### Expected behavior

Guidance plugins become enabled or disabled from the web UI.

### Actual behavior

Nothing happens when clicking on a plugin's button.

### Steps to reproduce the actual behavior

1. Start CARMA.
2. Navigate to localhost on a web browser, then click START Platform.
3. Click on Advanced, then Change Plugins. Try enabling or disabling any of the plugins.

msmcconnell:
Following up on this. It seems this functionality was never fully implemented in carma3. It appears to have been an after thought since it was functional in carma2.  
The functionality is currently being added back for ROS2 plugins and will probably make it into release 4.2 or worst case 4.3.
There are not currently plans to implement this for ROS1 plugins or carma3, but users can manually disable plugins by modifying the carma/launch/guidance.launch file.

GoodarzMehr:
Thanks for the clarification.

PR #1811, opened 2022-06-10
Feature/plugin controller

msmcconnell:
# PR Details
## Description
This PR resolves https://github.com/usdot-fhwa-stol/carma-platform/issues/1499 and https://github.com/usdot-fhwa-stol/carma-platform/issues/1666 by updating the guidance subsystem controller to have a plugin manager that handles plugin discovery and state transitions. The PluginManager, Entry, and EntryManager classes are moved into the subsystem_controllers package from health_monitor. The PluginManager monitors the plugin_discovery topic and updates plugins as they appear. The user can also provide a list of required plugins and auto_activated plugins. These two lists allow users to control which plugins will be activated at startup. The remaining plugins will need to be activated via the UI or manual service calls. 

In addition, to these changes the carma_guidance_plugins package is updated to not require the user to specify the node name. Instead, the name is set based on the node name returned by rclcpp.

This is currently a draft PR as Unit tests have not yet been updated.  

PR depends on https://github.com/usdot-fhwa-stol/carma-utils/pull/123
and a yet uncreated UI PR

## Related Issue
 https://github.com/usdot-fhwa-stol/carma-platform/issues/1499
https://github.com/usdot-fhwa-stol/carma-platform/issues/1666

## Motivation and Context
Plugin management adherer's to system expectations.

## How Has This Been Tested?
Unit testing and in the silver lexus

## Types of changes

## Checklist:


carma_guidance_plugins/src/plugin_base_node.cpp
```diff
@@ -142,7 +142,7 @@ namespace carma_guidance_plugins
     }
 
     carma_planning_msgs::msg::Plugin msg;
-    msg.name = get_plugin_name();
+    msg.name = std::string(get_namespace()) + std::string(get_name());
```
kjrush, line 145: Are we using the fully qualified names here? I believe previously we had been only using the name of the node/plugin, but no namespace information as part of the plugin contract is that their service calls can be accessed by /guidance/plugins/PLUGIN_NAME/plan_maneuvers, etc. 

Not a big deal either way, but worth a quick discussion of trade-offs i think.

subsystem_controllers/src/guidance_controller/plugin_manager.cpp
```diff
@@ -0,0 +1,434 @@
...
+        {
+
+            std::vector<std::string> plugin_capability_levels; 
+            boost::split(plugin_capability_levels, plugin.capability_, boost::is_any_of("/"));
+
+            if(plugin.type_ == carma_planning_msgs::msg::Plugin::CONTROL && 
+                (req->capability.size() == 0 || (matching_capability(plugin_capability_levels, req_capability_levels) == 0 && plugin.active_ && plugin.available_)))
+            {
+                RCLCPP_DEBUG_STREAM(rclcpp::get_logger("guidance_controller"), "discovered control plugin: " << plugin.name_);
+                res->plan_service.push_back(plugin.name_ + plugin.capability_);
```
kjrush, line 377: This might be something to consider. Currently the capabilities aren't directly linked to the service names like this. E.g. even if a plugin reports "plan_tactical/plan_trajectory/lane_change" the service that is used to invoke that capability is still the plugin's planning service named "/guidance/plugins/PLUGIN_NAME/plan_trajectory". In the existing implementation the value that's pushed into the list here is `plugin.name_ + control_service_suffix_` .

I think that helps with arbitrator and plan delegator having consistent interfaces by which to query plugins during planning, but I could see benefits to making it more granular. We'd have to adjust arbitrator and plan_delegator a bit. Do you think this is worth discussing in more detail?

subsystem_controllers/include/subsystem_controllers/guidance_controller/entry.h
```diff
@@ -0,0 +1,50 @@
...
+        //! Activation flag of a plugin
+        bool active_ = false;
+        //! Fully specified node name of a plugin
+        std::string name_;
+        //! Type of the plugin from the message enum in carma_planning_msgs::Plugin
+        uint8_t type_ = 0;
+        //! The capability string of the plugin
+        std::string capability_;
+        //! Flag indicating if the user requested this plugin be activated
+        bool user_requested_activation_ = false;
```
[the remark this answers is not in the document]
msmcconnell: This flag is used to specify that the user asked for the node to be activated. This can be done in two ways either via the configuration parameter for auto activation or via the UI. Its needed as an indication to separate intent from actual state.

subsystem_controllers/include/subsystem_controllers/guidance_controller/plugin_manager.h
```diff
@@ -0,0 +1,196 @@
...
+             * 
+             * For example the capability "tactical_plan/plan_trajectory" would be described as ["tactical_plan", "plan_trajectory"]
+             * if the user wanted to match this to "tactical_plan/plan_trajectory/platooning_trajectory" then the compared_capability_levels
+             * would be ["tactical_plan", "plan_trajectory", "compared_capability_levels"]
+             * Matching those two inputs as ( ["tactical_plan", "plan_trajectory"], ["tactical_plan", "plan_trajectory", "compared_capability_levels"])
+             * would be false because the compared_capability_levels is more detailed than the base. 
+             * By contrast, matching ( ["tactical_plan", "plan_trajectory", "compared_capability_levels"], ["tactical_plan", "plan_trajectory"] ) 
+             * would be true since the base is more generic then the request
+             * 
+             * // TODO check with Kyle on this because it maybe should be the reverse since a more specific capability may require specific meta data.
```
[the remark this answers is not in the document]
msmcconnell: Ok, I'll leave it as is then

subsystem_controllers/src/guidance_controller/plugin_manager.cpp
```diff
@@ -0,0 +1,434 @@
...
+        for (size_t i=0; i < base_capability_levels.size() &&  i < compared_capability_levels.size(); i++)
+        {
+            if (compared_capability_levels[i].compare(base_capability_levels[i]) != 0)
+                return false;
+        }
+
+        return true;
+    }
+
+    void PluginManager::get_control_plugins_by_capability(SrvHeader, carma_planning_msgs::srv::GetPluginApi::Request::SharedPtr req, carma_planning_msgs::srv::GetPluginApi::Response::SharedPtr res)
```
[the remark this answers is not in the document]
msmcconnell: Yeah

subsystem_controllers/src/guidance_controller/plugin_manager.cpp
msmcconnell: I think this was a bit of an oversight on my part. I'm fine with how we had it before. If we took the approach here, then it would be difficult for the receiver to disambiguate the plugin name from the service call which could make it unclear what the original capability was.

carma_guidance_plugins/src/plugin_base_node.cpp
msmcconnell: Actually, I've changed my mind on this. While I would prefer to use the fully qualified name, it seems that a large amount of changes would be needed in the ros1 nodes to remain compatible. So instead, we'll go with the base name of the node and used the /guidance/plugins prefix

kjrush (approved):

PR #1814, opened 2022-06-13
Fix/rfp shortest path fix

### LINKED ISSUE #1813: Route Following Plugin "return to shortest path logic" triggers as a result of controller drift
### Types of Issue

### Descriptive summary

Route Following Plugin contains logic to detect when the vehicle has left the shortest path through the route and to plan lane changes as necessary to return the vehicle to the a newly-computed shortest path when control is returned to the RFP. This logic is triggering unintentionally as a result of control drift causing the vehicle's reference point to drift into an adjacent lane during RFP execution. This causes the vehicle to plan a lane change from the adjacent lane back onto a new shortest path, which immediately begins to use the adjacent lane's centerline as control input, causing undesired steering oscillation.

As a related but somewhat independent issue, this check has difficulty determining the vehicle's proper lanelet when multiple lanelet's overlap (such as at divergences or convergences) and can trigger this behavior in such a scenario as well.

### Actual behavior

CARMA System 4.1 as used in IHP testing.

kjrush:
# PR Details
## Description

To address the anomaly described in #1813 , a check has been implemented to limit the times at which RFP will evaluate the need to return to shortest path. Now it should only trigger if a plugin other than RFP was in control at the time, meaning that control drift during regular RFP operation will not be considered as leaving the shortest path. In addition, a filter has been implemented such that when evaluating the vehicle's current lanelet, only lanelets on the route are considered as possible candidates.

## Related Issue
Resolves #1813 

## Motivation and Context

This is to support IHP2 testing as resolution of a blocking issue discovered during testing.

## How Has This Been Tested?

This has not yet been tested.

## Types of changes

## Checklist:

### DIFF
--- a/route_following_plugin/include/route_following_plugin.h
+++ b/route_following_plugin/include/route_following_plugin.h
@@ -241,6 +241,7 @@ namespace route_following_plugin
         ros::Publisher plugin_discovery_pub_;
         ros::Publisher upcoming_lane_change_status_pub_;
         ros::Subscriber twist_sub_;
+        ros::Subscriber current_maneuver_plan_sub_;
         ros::Timer discovery_pub_timer_;
 
         // ROS service servers
@@ -279,6 +280,9 @@ namespace route_following_plugin
         // Current vehicle pose in map
         geometry_msgs::PoseStamped pose_msg_;
         lanelet::BasicPoint2d current_loc_;
+        
+        // Currently executing maneuver plan from Arbitrator
+        cav_msgs::ManeuverPlanConstPtr current_maneuver_plan_;
 
         //Queue of maneuver plans
         std::vector<cav_msgs::Maneuver> latest_maneuver_plan_;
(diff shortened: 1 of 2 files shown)


route_following_plugin/src/route_following_plugin.cpp
```diff
@@ -568,15 +570,30 @@ void setManeuverLaneletIds(cav_msgs::Maneuver& mvr, lanelet::Id start_id, lanele
...
+            // If another plugin may have brought us off shortest path, check our current lanelet
+            auto llts = wm_->getLaneletsFromPoint(current_loc, 10);                                          
+            // Remove any candidate lanelets not on the route
+            llts.erase(std::remove_if(llts.begin(), llts.end(),
+                [&](auto lanelet) -> bool { return !wm_->getRoute()->contains(lanelet); }));
+
+            // !!! ASSUMPTION !!!:
+            // Once non-route lanelets have been removed, it is assumed that our actual current lanelet is the only one that can remain.
+            // TODO: Verify that this assumption is true in all cases OR implement more robust logic to track current lanelet when there are overlaps
+            auto current_lanelet = llts[0];
```
msmcconnell, line 588: Adding a size check on this is good to avoid seg faults and could also provide some insight on if we ever have overlapping route lanelets while testing. At this point we are assuming it has size==1 right.

route_following_plugin/include/route_following_plugin.h
```diff
@@ -279,6 +280,9 @@ namespace route_following_plugin
         // Current vehicle pose in map
         geometry_msgs::PoseStamped pose_msg_;
         lanelet::BasicPoint2d current_loc_;
+        
+        // Currently executing maneuver plan from Arbitrator
+        cav_msgs::ManeuverPlanConstPtr current_maneuver_plan_;
```
msmcconnell, line 285: Can you do a quick find on current_maneuver_plan and make sure this doesn't cause confusion with other variables in this file?

msmcconnell (changes requested):

route_following_plugin/include/route_following_plugin.h
kjrush: Yep, it's unique to this changeset.

route_following_plugin/src/route_following_plugin.cpp
kjrush: Added size check and exception for size==0.

PR #1815, opened 2022-06-14
[DO NOT MERGE] Comparison of ihp changes with 4.1.0

msmcconnell:
This is just a way of tracking some comments of mine with visibility for others on the current IHP change set


platooning_tactical_plugin/config/parameters.yaml
```diff
@@ -1,6 +1,6 @@
 # Float: The length of the trajectory in time domain
 # Unit: s
-trajectory_time_length: 6.0 # Trajectory length in seconds
+trajectory_time_length: 10.0 # Trajectory length in seconds
```
msmcconnell, line 3: Again why extend?

PR #1831, opened 2022-06-24
Port guidance node to ROS2

JonSmet:
# PR Details
## Description

This PR includes updates to port the guidance node to ROS2 and to launch it from the guidance.launch.py launch file. 

Since eventually the health_monitor will be removed from the CARMA System, a couple minor updates were made to GuidanceWorker to support guidance state machine transitions:
- When the guidance node is activated by the lifecycle system, GuidanceWorker triggers a guidance state machine `INITIALIZE` signal. This is because node activate indicates that all of the required drivers are ready.
- When the guidance node is shutdown by the lifecycle system, GuidanceWorker triggers a guidance state machine `SHUTDOWN` signal.

## Related Issue

Issue #1830 

## How Has This Been Tested?

All unit tests pass; local integration testing conducted.

## Types of changes

## Checklist:


guidance/src/guidance_state_machine.cpp
```diff
@@ -83,34 +85,29 @@ namespace guidance
...
             // '3' indicates vehicle gearshift is currently set to PARK
-            if(msg->current_gear.gear== autoware_msgs::Gear::PARK)
+            if(msg->current_gear.gear== autoware_msgs::msg::Gear::PARK)
             {
                 onGuidanceSignal(Signal::OVERRIDE); // Required for ENTER_PARK -> INACTIVE
 
-                if(operational_drivers_){
-                    onGuidanceSignal(Signal::INITIALIZED); 
-                }
+                onGuidanceSignal(Signal::INITIALIZED); // Required for INACTIVE -> DRIVERS_READY
```
kjrush, line 98: Is there a change elsewhere that explains why this check is no longer needed?

JonSmet: I spoke with Mike about the check for all drivers being ready on /system_alert, and he mentioned that health_monitor will be removed once our system has been migrated to ROS2. Instead, the guidance subsystem will have required nodes (including required drivers), and if any of these shutdown, then I believe all guidance subsystem nodes will shutdown. Because of this, the check for the `operational_drivers_` flag is no longer needed. I did notice though that I left that boolean flag in guidance_state_machine.hpp, and will remove it now.

kjrush (approved):

PR #1861, opened 2022-07-28
Fixes for platoon lane change in parallel and merge lanes

saina-ramyar:
# PR Details
## Description

Platooning plugin is updated to generate maneuvers for an adjacent lane join, based on the type of lanelets involved.
If the starting and end lanelets have a boundary with a left/right relation, a lane change maneuver is composed.
If the starting and end lanelets form a merge lane where lanelets overlap and there is no clear boundary, a lane follow maneuver is composed.

Moreover, basic autonomy is updated to use the lanelet ids included in the maneuver msg for trajectory generation, rather than generating a list based on shortest path.

## Related Issue
#1862 

## Motivation and Context
Vehicle should be able to complete driving on the merge lane and join the platoon on the adjacent lane

## How Has This Been Tested?
During verification of release/cabin

## Types of changes

## Checklist:

### DIFF
--- a/basic_autonomy/src/basic_autonomy.cpp
+++ b/basic_autonomy/src/basic_autonomy.cpp
@@ -76,8 +76,12 @@ namespace basic_autonomy
 
             cav_msgs::LaneFollowingManeuver lane_following_maneuver = maneuver.lane_following_maneuver;
             
+            // auto lanelets = wm->getLaneletsBetween(starting_downtrack, lane_following_maneuver.end_dist + detailed_config.buffer_ending_downtrack, true, true);
+
+
             auto lanelets = wm->getLaneletsBetween(starting_downtrack, lane_following_maneuver.end_dist + detailed_config.buffer_ending_downtrack, true, true);
-            
+
+
             bool lanelets_defined = !maneuver.lane_following_maneuver.lane_ids.empty();
             ROS_DEBUG_STREAM("lanelets_defined: " << lanelets_defined);
             bool isFromPlatooning = maneuver.lane_following_maneuver.parameters.planning_strategic_plugin == "PlatooningStrategicIHPPlugin";
@@ -90,7 +94,16 @@ namespace basic_autonomy
                 ROS_DEBUG_STREAM("extracted id: " << lane_id);
                 lanelet::ConstLanelet new_lanelet = wm->getMap()->laneletLayer.get(lane_id);
                 lanelets.push_back(new_lanelet);
-            }
+
+                if (maneuver.lane_following_maneuver.lane_ids.size()>1)
+                {
+                    lane_id = stoi(maneuver.lane_following_maneuver.lane_ids[1]);
+                    ROS_DEBUG_STREAM("more extracted id: " << lane_id);
+                    new_lanelet = wm->getMap()->laneletLayer.get(lane_id);
+                    lanelets.push_back(new_lanelet);
+                }
+            } 
+
 
 
             if (lanelets.empty())
--- a/basic_autonomy/src/basic_autonomy.cpp
+++ b/basic_autonomy/src/basic_autonomy.cpp
@@ -75,9 +75,6 @@ namespace basic_autonomy
             std::vector<PointSpeedPair> points_and_target_speeds;
 
             cav_msgs::LaneFollowingManeuver lane_following_maneuver = maneuver.lane_following_maneuver;
-            
-            // auto lanelets = wm->getLaneletsBetween(starting_downtrack, lane_following_maneuver.end_dist + detailed_config.buffer_ending_downtrack, true, true);
-
 
             auto lanelets = wm->getLaneletsBetween(starting_downtrack, lane_following_maneuver.end_dist + detailed_config.buffer_ending_downtrack, true, true);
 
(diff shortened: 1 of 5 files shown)


basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -75,9 +75,34 @@ namespace basic_autonomy
             std::vector<PointSpeedPair> points_and_target_speeds;
 
             cav_msgs::LaneFollowingManeuver lane_following_maneuver = maneuver.lane_following_maneuver;
-            
+
             auto lanelets = wm->getLaneletsBetween(starting_downtrack, lane_following_maneuver.end_dist + detailed_config.buffer_ending_downtrack, true, true);
 
+
+            bool lanelets_defined = !maneuver.lane_following_maneuver.lane_ids.empty();
```
msmcconnell, line 82: These changes need to be applied to basic_autonomy_ros2 as well.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -3160,6 +3143,22 @@ namespace platoon_strategic_ihp
         // because it is a rough plan, assume vehicle can always reach to the target speed in a lanelet
         maneuver_msg.lane_following_maneuver.end_time = current_time + ros::Duration(config_.time_step);
         maneuver_msg.lane_following_maneuver.lane_ids = { std::to_string(lane_id) };
+
+        ROS_DEBUG_STREAM("in compose maneuver lane id:"<< lane_id);
+
+        lanelet::ConstLanelet current_lanelet = wm_->getMap()->laneletLayer.get(lane_id);
+        if(!wm_->getMapRoutingGraph()->following(current_lanelet, false).empty())
```
msmcconnell, line 3150: Since this is ignoring the ending downtrack distance won't this check often result in incorrect lanelet ids?

msmcconnell: Seems like this method might be more appropriate 
https://github.com/usdot-fhwa-stol/carma-platform/blob/8689cec36bf9df40ea13e53a65b92c3c70ae9238/route_following_plugin/src/route_following_plugin.cpp#L300

But maybe I'm missing the context of this function?

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
```diff
@@ -3363,12 +3343,39 @@ namespace platoon_strategic_ihp
...
 
-                    // note: Since lanelet ID is not important for arbitrary lanechange, just use first lanelet's Id to create a maneuver msg.
-                    resp.new_plan.maneuvers.push_back(composeLaneChangeManeuverMessage(current_downtrack_, lc_end_dist,  
+                    lanelet::ConstLanelet starting_lanelet = wm_->getMap()->laneletLayer.get(current_lanelet_id);
+                    lanelet::ConstLanelet ending_lanelet = wm_->getMap()->laneletLayer.get(target_lanelet_id);
+
+                    auto relation = wm_->getMapRoutingGraph()->routingRelation(starting_lanelet, ending_lanelet);
+                    bool lanechangePossible = false;
+                    
+                    if (relation == lanelet::routing::RelationType::Left || relation == lanelet::routing::RelationType::Right)
```
msmcconnell, line 3355: I think this check could fail if the lane change was planned across a lanelet boundary. Fortunately, since this is a strategic plugin, it is also enforcing its own assumptions, so it probably won't break at the moment. However, I think you should add a comment documenting this assumption and an enhancement issue on github to resolve it with a link to that issue in the code.

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -75,9 +75,34 @@ namespace basic_autonomy
...
+            ROS_DEBUG_STREAM("lanelets_defined: " << lanelets_defined);
+            bool isFromPlatooning = maneuver.lane_following_maneuver.parameters.planning_strategic_plugin == "PlatooningStrategicIHPPlugin";
+            ROS_DEBUG_STREAM("isFromPlatooning: " << isFromPlatooning);
+
+            if (lanelets_defined && isFromPlatooning)
```
msmcconnell, line 87: Per discussion on slack this plugin specific solution is not desirable. Since the general solution was already written lets try evaluating that first.

msmcconnell (changes requested):

basic_autonomy/src/basic_autonomy.cpp
msmcconnell: We can wait on this, since we'll be address the other basic_autonomy changes again next sprint.

platooning_strategic_IHP/src/platoon_strategic_ihp.cpp
msmcconnell: Can leave as is since this is limited to platooning and works for ihp2

basic_autonomy/src/basic_autonomy.cpp
msmcconnell: Per discussion with Sudhakar we will wait on this and you should add a comment/issue like in my above comment

msmcconnell (approved):

PR #1872, opened 2022-08-02
Feature/ros2 localization manager

adev4a:
# PR Details
This PR includes updates to port the localization_manager node to ROS2 and to launch it from the carma_src.launch.py launch file
## How Has This Been Tested?
Unit Tested and Integration Tested

## Types of changes

## Checklist:


localization_manager/src/LocalizationTransitionTable.cpp
```diff
@@ -14,39 +14,38 @@
...
 {
   if (new_state == state_)
   {
-    return;  // State was unchanged no need to log or trigger callbacks
+      return; //State was unchanged no need to log or trigger callbacks
   }
-  ROS_INFO_STREAM("LocalizationTransitionTable changed localization state from "
-                  << state_ << " to " << new_state << " because of signal " << source_signal << " while in mode "
-                  << mode_);
+  RCLCPP_INFO_STREAM(rclcpp::get_logger("localization_manager::LocalizationTransitionTable"), "LocalizationTransitionTable changed localization state from "<< state_ << " to "<< new_state <<" because of signal "<<source_signal << " while in mode "<<mode_);
```
msmcconnell, line 42: While a bit counter intuitive for the rclcpp::get_logger() methods to be controllable from the launch file the input should be the package name without extra namespaces.

localization_manager/src/localization_manager_node.cpp
```diff
@@ -0,0 +1,172 @@
...
+        {"ndt_frequency_fault_threshold", config_.ndt_frequency_fault_threshold},
+        {"pose_pub_rate", config_.pose_pub_rate},
+        },parameters);
+
+        auto error_2 = update_params<int>({{"auto_initialization_timeout", config_.auto_initialization_timeout},
+        {"gnss_only_operation_timeout", config_.gnss_only_operation_timeout},
+        {"sequential_timesteps_until_gps_operation", config_.sequential_timesteps_until_gps_operation},
+        {"gnss_data_timeout", config_.gnss_data_timeout},
+        {"localization_mode", config_.localization_mode}
+        }, parameters);
```
msmcconnell, line 60: Since none of these new values are actually passed to the worked this function is meaningless. For this particular node, I would be ok not having runtime configurability because of the safety risk it poses to change localization modes mid run.

localization_manager/src/localization_manager_node.cpp
```diff
@@ -0,0 +1,172 @@
...
+                                                                                std::bind(&LocalizationManager::systemAlertCallback, manager_.get(), std_ph::_1));
+        // Setup publishers
+        pose_pub_ = create_publisher<geometry_msgs::msg::PoseStamped>("selected_pose", 5);
+        state_pub_ = create_publisher<carma_localization_msgs::msg::LocalizationStatusReport>("localization_status", 5);
+        
+        // Create a publisher that will send all previously published messages to late-joining subscribers ONLY If the subscriber is transient_local too
+        rclcpp::PublisherOptions intra_proc_disabled; 
+        intra_proc_disabled.use_intra_process_comm = rclcpp::IntraProcessSetting::Disable; // Disable intra-process comms for this PublisherOptions object
+
+        auto pub_qos_transient_local = rclcpp::QoS(rclcpp::KeepAll()); // A publisher with this QoS will store all messages that it has sent on the topic
```
msmcconnell, line 124: For managed initial_pose a KeepLast(1) makes more sense than KeepAll. We wouldn't want to send multiple initial poses in a row on new connection. We only send the most recent.

msmcconnell (changes requested):

localization_manager/src/localization_manager_node.cpp
```diff
@@ -0,0 +1,145 @@
...
+    void Node::poseAndStatsCallback(const geometry_msgs::msg::PoseStamped::ConstPtr pose,
+                                     const autoware_msgs::msg::NDTStat::ConstPtr stats) const
+    {
+        try
+        {
+            manager_->poseAndStatsCallback(pose, stats);
+        }
+        catch (const std::exception& e)
+        {
+            RCLCPP_ERROR_STREAM(get_logger(), "Uncaught Exception in localization_manager. Exception: "<< e.what() );
```
msmcconnell, line 135: Why are we catching this exception and not allowing it to shutdown the system?

localization_manager/test/node_test.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+    worker_node->configure(); //Call configure state transition
+    worker_node->activate();  //Call activate state transition to get not read for runtime
+
+    std::unique_ptr<std_msgs::msg::String> msg = std::make_unique<std_msgs::msg::String>();
+    msg->data = "my string";
+
+    worker_node->example_callback(move(msg)); // Manually drive topic callbacks
+
+}
```
msmcconnell, line 40: This test should be removed. Also example_callback shouldn't compile right? Have the unit tests been run?

msmcconnell (changes requested):

localization_manager/src/localization_manager_node.cpp
msmcconnell: If the issue is that it's not caught because of message_filters then we need to elevate it into the carma lifecycle node's error handling not just log it

adev4a: This is from the ros1 implementation https://github.com/usdot-fhwa-stol/carma-platform/blob/e7fe48309097169c29d811feb2456bbfce310e44/localization_manager/src/localizer_node.cpp#L49
If you think its an issue, i can change it here or log a bug and address it

localization_manager/test/node_test.cpp
adev4a: yes you're right this doesn't compile, it wasn't included in the cmake. This file is not required and is removed now

localization_manager/src/localization_manager_node.cpp
msmcconnell: This is not equivalent because you are not passing the exception to the ros2 equivalent of handleException. That ROS1 function publishes a system_alert. You should call
handle_primary_state_exception(e); here to get the same behavior as in ROS1

msmcconnell (approved):

Issue #1908, opened 2022-08-25
Tactical plugin "plan trajectory" callbacks should be evaluated for performance optimizations


JonSmet:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

The plan_delegator node attempts to generate new trajectories for the CARMA System at a rate of 10 Hz via service calls to applicable tactical plugins. As of August 25th, 2022 on the develop branch, the plan_delegator waits up to 0.1 seconds ([line here](https://github.com/usdot-fhwa-stol/carma-platform/blob/7e28e939c6fa5b11209242055bd0d9861a4c30f7/plan_delegator/src/plan_delegator.cpp#L459)) for a tactical plugin to complete trajectory planning for provided maneuver(s). This '0.1 seconds' duration enforcement is new as of July 2022, and it is possible that tactical plugins have commonly surpassed this duration in the past for trajectory planning. 

In some cases, it has been found that tactical plugins can take longer than 0.1 seconds to plan trajectories, which cause the plan trajectory service call from plan_delegator to fail. 
- From initial testing, it looks like inlanecruising_plugin can typically generate a trajectory in under 0.1 seconds, but could fail intermittently if logging is set to DEBUG. 
- The cooperative_lanechange plugin fails more reliably, with initial testing indicating that trajectory generation can take more than 0.1 seconds with default logging levels, and up to 0.5 seconds with logging set to DEBUG.
- stop_and_wait_plugin and other use-case specific tactical plugins have not been analyzed, but it is possible that they might surpass this 0.1 second duration limit as well.

In order to ensure tactical plugins reliably meet this 0.1 second duration limit for trajectory planning, more in-depth analysis of the performance of these plugins should be conducted, and logic within both tactical plugins and the basic_autonomy library should be evaluated for possible optimization improvements.

### Carma version where this issue was discovered

Develop (August 25th, 2022)

### Expected behavior
It is expected that tactical_plugins convert maneuver(s) to detailed trajectories in at most 0.1 seconds.

### Actual behavior

Some tactical plugins can take longer than 0.1 seconds to convert maneuver(s) to detailed trajectories, causing the trajectory plan service call from plan_delegator to fail.

### Steps to reproduce the actual behavior

1. Start CARMA
[...]

PR #1930, opened 2022-09-12
optimize tactical plugin trajectory generation

saina-ramyar:
# PR Details
## Description

This PR is a fix for issue #1908 where some tactical plugins take longer to generate a trajectory, causing the service request from plan_delegator to time out without any response.
This was caused because sometimes the number of points used to calculate a spline fit is large, which contributes to the delay. Therefore, the number of points used to calculate the spline is limited to a reasonable value that is large enough to include most of the points, but small enough that the spline calculations are done in time. 

## Related Issue
#1908 

## How Has This Been Tested?

## Types of changes

## Checklist:

### DIFF
--- a/basic_autonomy_ros2/src/basic_autonomy.cpp
+++ b/basic_autonomy_ros2/src/basic_autonomy.cpp
@@ -817,7 +817,7 @@ namespace basic_autonomy
                 size_t left_points_size = basic_points.size() - resized_basic_points.size();
                 RCLCPP_DEBUG_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER), "Number of left out basic_points size: " << left_points_size);
 
-                double percent_points_lost = 100 * left_points_size/basic_points.size();
+                float percent_points_lost = 1100.0 * (float)left_points_size/basic_points.size();
 
                 if (percent_points_lost > 50.0)
                 {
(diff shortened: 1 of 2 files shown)


basic_autonomy_ros2/src/basic_autonomy.cpp
```diff
@@ -802,10 +802,32 @@ namespace basic_autonomy
...
+            // So if the basic_points vector size is large, only the first 400 points are used to compute a spline fit. 
+            if (resized_basic_points.size() > 400)
+            {
+                resized_basic_points.resize(400);
+                RCLCPP_DEBUG_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER), "Resized basic_points size: " << resized_basic_points.size());
+
+                size_t left_points_size = basic_points.size() - resized_basic_points.size();
+                RCLCPP_DEBUG_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER), "Number of left out basic_points size: " << left_points_size);
 
+                int percent_points_lost = 100 * left_points_size/basic_points.size();
```
kjrush, line 820: I think you might need a float conversion in here, I think `size_t / size_t` is going to result in something integer-typed, so as long as `left_points_size < basic_points.size()` then I think we're just going to get 0 as a result for percent_points_lost.

kjrush (approved):

Issue #1926, opened 2022-09-01
CARMA-Platform BSM Generator is not random, causing BSM ID conflicts between vehicles.


paulbourelly999:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

When starting CARMA-Platform the BSMID for vehicles seems to cycle between some set sequence of numbers starting with all zeros and then going to :
- 240
 - 58
 - 214
 - 16
 after 5 mins. Seems like it is not reading the opt/carma/vehicle/calibration/identifiers/UniqueVehicleParams.yaml configuration parameters correctly because regardless of what we set the rotation toggle to. The first bsm id is all zeros and the seconds on is the values shown here.
### Carma version where this issue was discovered

Latest
### Expected behavior
BSM Random generation for multiple vehicles should guarantee unique 
### Actual behavior

BSM Generation not random causing conflict.
### Steps to reproduce the actual behavior

1. Start CARMA Platform
2. Echo outgoing mobility operation or bsm
3. BSM ID 000000000
4. Wait 5 mins 
5. BSM ID - 2405821416

MishkaMN:
Resolved with above PR

PR #1936, opened 2022-09-20
Fix/uc1 test

MishkaMN:
# PR Details
## Description

This is a assorted fixed related to UC1 regression testing. 
- Mobility path and operation should not be published if the vehicle is not engaged  for uc1.
- bsm id will generate random id from the beginning and uses fixed id if it random_generation is toggled false
- Mobility path's timestamp should be still up to date when waiting on a stop line

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1926 BSM ID
https://github.com/usdot-fhwa-stol/carma-platform/issues/1937 Mobility Path timestamp
https://github.com/usdot-fhwa-stol/carma-platform/issues/1938 Mobility Path & Operation should not publish when not engaged

## Motivation and Context
Use case 1 regression testing fixes.

## How Has This Been Tested?
Integration tested on blue lexus and white pacifica 

## Types of changes

## Checklist:


sci_strategic_plugin/config/parameters.yaml
```diff
@@ -5,7 +5,7 @@ vehicle_decel_limit_multiplier : 0.65
 vehicle_accel_limit_multiplier : 0.65
 
 # Double: A buffer before of the stopping location which will still be considered a valid stop. Units in meters
-stop_line_buffer : 3.5
+stop_line_buffer : 10
```
msmcconnell, line 8: Is this acceptable? How does it compare with UC2?

stop_and_wait_plugin/config/parameters.yaml
```diff
@@ -26,7 +26,7 @@ centerline_sampling_spacing: 1.0
 # Double: The default buffer in meters used for where the vehicle can come to a stop during execution. 
 #         This value is overriden by the first value in maneuver.parameters.float_valued_meta_data[]
 # Units: m
-default_stopping_buffer: 5.0
+default_stopping_buffer: 15.0
```
msmcconnell, line 29: Same question on acceptability as above.

MishkaMN: This parameter is being overridden, so we can actually keep the older value. So I reversed it.

sci_strategic_plugin/config/parameters.yaml
MishkaMN: UC2 has stopping location buffer of 8 meters, which I think is comparable to this value. UC2 has a little more sophisticated guard for stopping case that allows it to stop closer (once it is decided to stop, uc2 plugin will keep it slowing, while uc1 may try to lane follow again in stopping distance). So another way to make it stop more consistently without changing the code was increasing this value.

Issue #1911, opened 2022-08-25
Inaccurate Localization


saina-ramyar:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

When the vehicle is localized, engaged and moving, the vehicle starts to drift out of the lane, but on rviz it shows that the localized position is still in the lane following the path.
### Carma version where this issue was discovered
develop

### Expected behavior
Localization is stable and the localized pose is the same as the vehicle's actual position.
### Actual behavior

 The localized position is still in the lane, when vehicle's actual position is outside the lane.
### Steps to reproduce the actual behavior

1. start carma
2. localize
3. select a route and engage
4. observe rviz for the incorrect localization

PR #1944, opened 2022-09-26
Fix/develop localization

afrye51:
# PR Details
## Description

Fixed one bug, where the timeout from state DEGRADED_NO_LIDAR_FIX would crash after completing, because in its setup it was partially written on top of the timeout from state INITIALIZING. The actual fix is just setting the timer id as a new id. 

Also fixed logging issues with the localization manager. previously, logging was different in each file and didn't work from the transition table file. After switching each call for the logger to `rclcpp::get_logger("localization.localization_manager"), "string to print");`, the logging works just fine. 

## Related Issue

https://usdot-carma.atlassian.net/browse/CAR-5631 and resolves https://github.com/usdot-fhwa-stol/carma-platform/issues/1911

## Motivation and Context

localization manager was crashing 20 seconds after setting a pose in rviz, causing the localization to no longer update properly. 

## How Has This Been Tested?

Started carma, localized with rviz, then selected a route to drive from underneath the tree to the east intersection, and the localization manager stayed up and it worked great. 

## Types of changes

## Checklist:


localization_manager/src/localization_manager_node.cpp
```diff
@@ -132,15 +134,13 @@ namespace localization_manager
         }
         catch (const std::exception& e)
         {
-            RCLCPP_ERROR_STREAM(get_logger(), "Uncaught Exception in localization_manager. Exception: "<< e.what() );
+            RCLCPP_ERROR_STREAM(rclcpp::get_logger("localization.localization_manager"), "localization_manager Uncaught Exception in localization_manager. Exception: "<< e.what() );
```
MishkaMN, line 137: On all logs, there is extra preceding string "localization_manager " on all strings to print. e.g.
`"localization_manager Uncaught Exception in localization_manager. Exception: "`. I am assuming it auto replace issue?

MishkaMN (changes requested):

localization_manager/src/localization_manager_node.cpp
afrye51: Nice catch, I had updated them when logging wasn't working properly to try and find them easier. Fixed in the latest commit.

MishkaMN (approved):

MishkaMN (approved):

Issue #1943, opened 2022-09-22
Lane following trajectory is not using correct lanelet


MishkaMN:
### Types of Issue

- [X] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

From the trajectory visualizer, it was noticed that the trajectory being generated for simple lane following looks like a weird curved trajectory. It is assumed that a different lanelet is being used other than the ones on route. Since inlanecruising_plugin uses basic_autonomy_ros2, most likely some lane_ids related specification is wrong there. Once the vehicle enters into intersection, despite a small swerve towards original direction, it gets back on track generating correct trajectory

![image](https://user-images.githubusercontent.com/20613282/191837270-f8ea7304-62f9-42d6-a0ee-b505f3d728f4.png)
![image](https://user-images.githubusercontent.com/20613282/191837371-6e27f8d6-a97c-46a1-8995-38019fe9647b.png)
![image](https://user-images.githubusercontent.com/20613282/191837416-f38583e5-5a66-4fbf-86ed-d66ca3fa98c3.png)

### Carma version where this issue was discovered

carma-system-4.2.0 develop+
### Expected behavior
Generate trajectory along route for lane following
### Actual behavior

See pics
### Steps to reproduce the actual behavior

Do normal operation

msmcconnell:
Resolved by above PRs

Issue #1939, opened 2022-09-20
Mobility Path Visualizer is crashing guidance container stack when receiving MP from another car


MishkaMN:
### Types of Issue

- [X] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Recent fix due to UC1 made platform publish mobility path with updated timestamps even if the vehicle is stopping (which means coordinates x,y in MP are 0, 0). This interestingly crashes another vehicle's platform at the instant when that MP message is received by that 2nd vehicle. It was fine without that specific change before with two vehicles. We tracked down the source to mobility path visualizer, and with that node turned off, the vehicles operate without issue.  

Fix is introduced in this PR: https://github.com/usdot-fhwa-stol/carma-platform/pull/1936
Exact lines: 
https://github.com/usdot-fhwa-stol/carma-platform/blob/6ed83e7b1bbb8fd98d2774ca936a60cdda33be98/mobilitypath_publisher/src/mobilitypath_publisher.cpp#L90-L91

### Carma version where this issue was discovered

carma-system-4.2.0+ develop
### Expected behavior
See above.
### Actual behavior

See above.
### Steps to reproduce the actual behavior

Start platform on one platform with above fixes. 
Start another platform one a different vehicle with 4.2.0+ system and this vehicle's guidance stack will crash.

MishkaMN:
fixed with above PR

MishkaMN:
Reopening this as there were more scenarios where the node could segfault, which is being addressed during UC3 integration testing.

msmcconnell:
Resolved by above PRs

PR #1958, opened 2022-10-13
Feature/uc3 full

MishkaMN:
# PR Details
## Description

All changes and major issues required to run TSMO UC3. Detailed issues are under each issue noted in this PR below.

Miscellaneous:
- Reduced minimum allowable speed from 10mph to 5mph as requested by carma-street team
- Changed trajectory_activation distance to 200meters, to allow ts early.
- Adjusted the login carma_wm_ctrl so it is less spammy.
- Turned off ray_ground_classifer because it is printing every 0.1 sec which slows down carma considerably

## Related Issue
Arbitrator: https://github.com/usdot-fhwa-stol/carma-platform/issues/1959 (fixed)
^
Another issue that is not recorded, but could be related is where the platform is generating very short trajectory (some runs alternate between short and long trajectory every second) and running out of trajectory and then get stopped by pure_pursuit. This started happening after changing arbitrator's planning from 0.5hz to 1 hz as intended. currently investigating... could be also related to 1961.
Update: Above issue mentioned does not seem to be error.

Lane selection in basic_autonomy: https://github.com/usdot-fhwa-stol/carma-platform/issues/1943 (fixed)
docker partial build failure fix for one of the two problems: https://github.com/usdot-fhwa-stol/carma-platform/issues/1960 (fixed)
Mobilitypath segfaulting: https://github.com/usdot-fhwa-stol/carma-platform/issues/1939 (fixed)
guidance failing silently... https://github.com/usdot-fhwa-stol/carma-platform/issues/1961 (potentially fixed unless it occurs again)
^ 
Current potential fix is implemented and is to separate each nodes with service calls to a separate container. It is noted in the code as well. 

Find next avaialable green... https://github.com/usdot-fhwa-stol/carma-platform/issues/1963 (not required per discussion with carma-street team)

## Motivation and Context
See above.

## Types of changes

## Checklist:


lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -274,7 +274,7 @@ void LCIStrategicPlugin::handleFailureCase(const cav_srvs::PlanManeuversRequest&
 
   resp.new_plan.maneuvers.push_back(composeIntersectionTransitMessage(
       traffic_light_down_track, intersection_end_downtrack_.get(), intersection_speed_.get(),
-      incomplete_traj_params.modified_departure_speed, current_state.stamp + ros::Duration(incomplete_traj_params.modified_remaining_time), intersection_exit_time, crossed_lanelets.back().id(), exit_lanelet_id));
+      incomplete_traj_params.modified_departure_speed, current_state.stamp + ros::Duration(incomplete_traj_params.modified_remaining_time), intersection_exit_time, crossed_lanelets.back().id(), crossed_lanelets.back().id()));
```
msmcconnell, line 277: Can I get an explanation of this change? The exit_lanelet_id appears to be the actual final lanelet. The cross_lanelets.back() is likely the next lanelet since getLaneletsBetweenWithException was called with bounds_inclusive = true. This means the ending lanelet for the maneuver would actually be wrong.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -306,7 +306,7 @@ void LCIStrategicPlugin::handleCruisingUntilStop(const cav_srvs::PlanManeuversRe
   std::vector<lanelet::ConstLanelet> lane_follow_crossed_lanelets =
       getLaneletsBetweenWithException(new_ts_params.x1_, new_ts_params.x2_, true, true);
 
-  resp.new_plan.maneuvers.push_back(composeTrajectorySmoothingManeuverMessage(current_state.downtrack, traffic_light_down_track, lane_follow_crossed_lanelets, 
+  resp.new_plan.maneuvers.push_back(composeTrajectorySmoothingManeuverMessage(current_state.downtrack, new_ts_params.x2_, lane_follow_crossed_lanelets,
```
msmcconnell, line 309: What is x2 here? And how is that different from traffic_light_down_track?

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -316,7 +316,7 @@ void LCIStrategicPlugin::handleCruisingUntilStop(const cav_srvs::PlanManeuversRe
   resp.new_plan.maneuvers.push_back(composeStopAndWaitManeuverMessage(
     new_ts_params.x2_, traffic_light_down_track, new_ts_params.v2_, case_8_crossed_lanelets.front().id(),
     case_8_crossed_lanelets.back().id(), ros::Time(new_ts_params.t2_),
-    current_state.stamp + ros::Duration(config_.min_maneuver_planning_period), decel_rate));
+    ros::Time(new_ts_params.t2_) + ros::Duration(config_.min_maneuver_planning_period), decel_rate));
```
msmcconnell, line 319: What is t2_ here?

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -362,7 +362,7 @@ void LCIStrategicPlugin::handleGreenSignalScenario(const cav_srvs::PlanManeuvers
 
     resp.new_plan.maneuvers.push_back(composeIntersectionTransitMessage(
         traffic_light_down_track, intersection_end_downtrack_.get(), intersection_speed_.get(),
-        intersection_speed_.get(), light_arrival_time_by_algo, intersection_exit_time, entry_lanelet.id(), exit_lanelet.id()));
+        intersection_speed_.get(), light_arrival_time_by_algo, intersection_exit_time, crossed_lanelets.back().id(), crossed_lanelets.back().id()));
```
msmcconnell, line 365: Same concern as above about these lanelet Ids. Needs a strong justification for this it seems like it would make the maneuver content invalid.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -607,33 +607,32 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
...
     {
       if (pair.second == lanelet::CarmaTrafficSignalState::PROTECTED_MOVEMENT_ALLOWED && 
-        lanelet::time::timeFromSec(ros::Time::now().toSec()) < pair.first ) //if not outdated
+        lanelet::time::timeFromSec(nearest_green_entry_time.toSec()) < pair.first ) 
       {
-        nearest_green_entry_time = get_nearest_green_entry_time(current_state.stamp, earliest_entry_time, traffic_light) 
-                                          + ros::Duration(0.01);
-        ROS_DEBUG_STREAM("Up to date GREEN signal exists");
+        //nearest_green_entry_time = get_nearest_green_entry_time(current_state.stamp, earliest_entry_time, traffic_light) 
+        //                                  + ros::Duration(0.01);
```
msmcconnell, line 629: Commented code. Can it be removed?

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -705,7 +704,7 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
 
       if (early_arrival_time_green_et.toSec() - nearest_green_signal_start_time.toSec() < config_.green_light_time_buffer)
       {
-        nearest_green_entry_time_cached_ = nearest_green_signal_start_time + ros::Duration(config_.green_light_time_buffer);
+        nearest_green_entry_time_cached_ = nearest_green_signal_start_time + ros::Duration(config_.green_light_time_buffer + 0.01);
```
msmcconnell, line 707: If there is already a buffer why do we need to add an extra 0.01 to it? Why not just increase the buffer parameter?

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -850,24 +852,32 @@ void LCIStrategicPlugin::planWhenWAITING(const cav_srvs::PlanManeuversRequest& r
...
+
+  ROS_DEBUG_STREAM("WAITING STATE: requested time to req.header.stamp check: " << std::to_string(req.header.stamp.toSec()));
+  ROS_DEBUG_STREAM("WAITING STATE: requested time to CURRENT STATE check: " << std::to_string(entering_time));
   
-  if (!validLightState(current_light_state_optional, current_state.stamp))
+  if (!validLightState(current_light_state_optional, ros::Time(entering_time)))
     return;
 
-  auto bool_optional_late_certainty = canArriveAtGreenWithCertainty(current_state.stamp, traffic_light, true, false);
+  auto bool_optional_late_certainty = canArriveAtGreenWithCertainty(ros::Time(entering_time), traffic_light, true, false);
```
msmcconnell, line 866: To me these changes in time read like we are evaluating when the vehicle crosses the stop bar from a stop. Previously the logic seemed to wait until the light actually changed. Does this cause the car to start moving while the light is still red? That would be unsafe in my opinion.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -906,12 +916,12 @@ void LCIStrategicPlugin::planWhenDEPARTING(const cav_srvs::PlanManeuversRequest&
...
   // Identify the lanelets which will be crossed by approach maneuvers lane follow maneuver
   std::vector<lanelet::ConstLanelet> crossed_lanelets =
-      getLaneletsBetweenWithException(current_state.downtrack, intersection_end_downtrack, true, true);
+      getLaneletsBetweenWithException(current_state.downtrack, intersection_end_downtrack, true, false);
 
   // Compose intersection transit maneuver
   resp.new_plan.maneuvers.push_back(composeIntersectionTransitMessage(
       current_state.downtrack, intersection_end_downtrack, current_state.speed, intersection_speed_limit,
-      current_state.stamp, intersection_exit_time, crossed_lanelets.front().id(), crossed_lanelets.back().id()));
+      current_state.stamp, intersection_exit_time, crossed_lanelets.back().id(), crossed_lanelets.back().id()));
```
msmcconnell, line 924: Seems like invalid maneuver setting similar to the others above.

msmcconnell (changes requested):

carma/launch/environment.launch.py
```diff
@@ -156,20 +156,20 @@ def generate_launch_description():
...
-                    {'--log-level' : GetLogLevel('ray_ground_classifier_nodes', env_log_levels) }
-                ],
-                remappings=[
-                    ("points_in", "points_in_base_link"), 
-                    ("points_nonground", "points_no_ground")
-                ],
-                parameters=[ ray_ground_classifier_param_file]
-            ),
+            #ComposableNode(
+            #    package='ray_ground_classifier_nodes',
```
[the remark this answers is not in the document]
aswath1: Created a different pr on autowareauto to turn off this error line

https://github.com/usdot-fhwa-stol/autoware.auto/pull/13

arbitrator/src/arbitrator.cpp
```diff
@@ -152,12 +152,6 @@ namespace arbitrator
 
     void Arbitrator::waiting_state()
     {
-        // Sleep in 100ms increments until our next planning cycle
-        // This ensures we spin() at least a few times
-        while (nh_->get_clock()->now() < next_planning_process_start_)
-        {
-            std::this_thread::sleep_for(std::chrono::milliseconds(100));
-        }
```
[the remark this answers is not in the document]
aswath1: In ROS1, arbitrator did not reply on timer but used internal variable next_planning_process_start_ to start next iteration. Everything was inside while function, so it slept 0.1s increments to not fully freeze other subscription functionality. In ROS2, timer is being used, so this sleep process or even next_planning_process_start_ variable is not used.

arbitrator/src/arbitrator_node.cpp
```diff
@@ -100,7 +100,7 @@ namespace arbitrator
                                 [this]() {this->arbitrator_->bumper_pose_cb();});
         
         arbitrator_run_ = create_timer(get_clock(),
-                                std::chrono::duration<double>(1/config_.planning_frequency),
+                                std::chrono::duration<double>(1/(config_.planning_frequency * 2 )), //there is waiting state between each planning state
```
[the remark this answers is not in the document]
aswath1: The intention is to keep the arbitrator planning at intended frequency - 1s. Looks like ROS2 conversion kept the WAITING state from ROS1, which occurs between PLANNING state in the state machine. So if the timer calls state callback each 1 second, the arbitrator ends up planning every 2s due to PLANNING > WAITING > PLANNING transition. That's why the frequency is doubled.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
aswath1: CASE8 means "cruise until stop", so the algorithm generates cruising from x0 to x1 to x2, and stopping maneuver from x2 to x3 (which is the stop_line or traffic_light_down_track). So TrajectorySmoothingManeuver (lci_tactical) should be only until x2, and StopAndWaitManeuever should be from x2 to x3 (traffic_light_downtrack)

lci_strategic_plugin/src/lci_strategic_plugin.cpp
aswath1: t2 is the time the vehicle should arrive at x2 which is when the stopping maneuever should start. So previously there was inconsistency in the inputs where it used correct starting time t2, but when calculating ending time, it used current_time + maneuever_planning_period instead of t2_maneuever_planning_period

lci_strategic_plugin/src/lci_strategic_plugin.cpp
aswath1: Actually including in the buffer doesn't work because it uses that same buffer to check early and late. If buffer is 2s and green starts at 10s, it will check +/-2s from 12s. If the buffer was 2.01s and green starts at 10s again, it checks +/-2.01 from 12.01, so both checks 10s. Which is why 0.01 is hardcoded

lci_strategic_plugin/src/lci_strategic_plugin.cpp
aswath1: crossed_lanelets's last element is the intersection lanelet due to inclusive bound. And this lanelet should be added to start_lanelet_id so that intersectionTransitPlugin correctly passes it to the basic_autonomy library

lci_strategic_plugin/src/lci_strategic_plugin.cpp
aswath1: One discovery that was found during testing was that the IntersectionTransitManeuvering plugin only ever uses the first starting_lanelet_id in creating lanelet_ids for lane_follow maneuever:
https://github.com/usdot-fhwa-stol/carma-platform/blob/feature/uc3_full/intersection_transit_maneuvering/src/intersection_transit_maneuvering.cpp#L252

And the function getLaneletsBetweenWithException is called from current_downtrack to traffic_light_down_track which is the 'start of the intersection', so by calling it bounds inclusive will actually include the intersection lanelet as the last element of the list, which is why crossed_lanelets.back() is picked

much of the IntersectionTransitManeuver msg composing changes are related to that. Previously it worked because lanelet_ids field requirement in basic_autonomy was introduced later on

docker/install.sh
```diff
@@ -71,7 +71,7 @@ echo "Building ROS2 CARMA Components"
 if [[ ! -z "$ROS1_PACKAGES$ROS2_PACKAGES" ]]; then
     if [[ ! -z "$ROS2_PACKAGES" ]]; then
         echo "Incrementally building ROS2 packages: $ROS2_PACKAGES"
-        colcon build --install-base /opt/carma/install_ros2 --cmake-args -DCMAKE_BUILD_TYPE=Release --packages-above $ROS2_PACKAGES --allow-overriding $ROS2_PACKAGES
+        colcon build --install-base /opt/carma/install_ros2 --cmake-args -DCMAKE_BUILD_TYPE=Release --packages-above $ROS2_PACKAGES
```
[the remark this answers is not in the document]
aswath1: I am not sure why, but the build was failing. There is an open issue what was happening here: https://github.com/usdot-fhwa-stol/carma-platform/issues/1960

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -111,9 +111,17 @@ namespace basic_autonomy
...
-                lanelets.push_back(extra_following_lanelets[0]);
+                for (size_t i = 0; i < extra_following_lanelets.size(); i++)
+                {
+                    if (llt.id() == extra_following_lanelets[i].id())
+                    {
+                        lanelets.push_back(extra_following_lanelets[i]);
+                        break;
+                    }
+                }
             }
```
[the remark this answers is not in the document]
aswath1: Previous version just blindly accepted 0th index so it did not seem to consider it either.

I added found_extra_on_shortest_path boolean as a cleaner check.

basic_autonomy_ros2/src/basic_autonomy.cpp
```diff
@@ -110,9 +110,17 @@ namespace basic_autonomy
...
             {
-                lanelets.push_back(extra_following_lanelets[0]);
+                for (size_t i = 0; i < extra_following_lanelets.size(); i++)
+                {
+                    if (llt.id() == extra_following_lanelets[i].id())
+                    {
+                        lanelets.push_back(extra_following_lanelets[i]);
+                        break;
+                    }
+                }
```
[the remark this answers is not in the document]
aswath1: Same as previous comment.

basic_autonomy/src/basic_autonomy.cpp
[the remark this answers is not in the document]
aswath1: image

basic_autonomy_ros2/src/basic_autonomy.cpp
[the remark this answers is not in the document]
aswath1: basic_autonomy is same thing, just mirror the final changes in ROS2.

carma/launch/message.launch.py
```diff
@@ -144,7 +145,8 @@ def generate_launch_description():
                     {'--log-level' : GetLogLevel('j2735_convertor', env_log_levels) }
                 ],
                 remappings=[
-                    ("outgoing_bsm", "bsm_outbound" ),
+                    ("outgoing_bsm", "bsm_outbound" )
+                    #("incoming_mobility_operation", "incoming_mobility_operation_actual" )
```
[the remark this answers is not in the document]
aswath1: For the message.launch.py, yes the commented part can be deleted, that whole line is not needed anymore.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
aswath1: Actually these changes ended up being cosmetic as they are doign exactly the same thing as uc2. Any changes inside the function planWhenWAITING should be reverted. 

You probably missed that entering_time was assigned current_state.stamp.toSec(); so it was just a renaming. 

And this condition is never satisfied
   if (config_.enable_carma_streets_connection && entering_time > current_state.stamp.toSec()) //uc3
    should_enter = false; 

so it is same as not being there and can be removed.

aswath1: So the current behavior is still just "enter when green is ON if you are waiting at the stop_line (doesn't matter the ET sent by carma-street, which is what we discussed with Misheel, Paul and Saeid)"

lci_strategic_plugin/src/lci_strategic_plugin.cpp
aswath1: Same explanation as before where crossed_lanelets's last element is the intersection lanelet due to inclusive bound. And this lanelet should be added to start_lanelet_id so that intersectionTransitPlugin correctly passes it to the basic_autonomy library.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
aswath1: Yes

lci_strategic_plugin/config/parameters.yaml
```diff
@@ -20,10 +20,10 @@ green_light_time_buffer : 1.0
 stop_light_time_buffer : 2.0
 
 # Double: Minimum allowable speed in trajectory smoothing algorithm m/s
-algo_minimum_speed : 4.4704 
+algo_minimum_speed : 2.235
 
-# Double: Minimum allowable speed of the car m/s
-absolute_minimum_speed : 4.4704
+# Double: Minimum allowable speed of the car m/s # not used
```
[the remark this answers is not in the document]
aswath1: Removed from everywhere.

aswath1 (dismissed):

arbitrator/src/arbitrator_node.cpp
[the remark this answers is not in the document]
msmcconnell: Ok @aswath1, I think what you just said should be it's own comment in the code. The bit you added is helpful but it could be clearer if you copy and paste your paragraph in.

docker/install.sh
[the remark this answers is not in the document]
msmcconnell: @aswath1 I don't want to commit this at the moment. Let's revert this change and then you can apply it locally during testing if needed. We'll pull the issue into the next sprint for someone to investigate.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
msmcconnell: Ok @aswath1 let's add that as an epsilon constant in the file with a comment describing its use.

msmcconnell (changes requested):

msmcconnell (changes requested):

msmcconnell:
This code smell also needs to be fixed
https://sonarcloud.io/project/issues?pullRequest=1958&issues=AYP2frFS77HIeUB3Zb2O&open=AYP2frFS77HIeUB3Zb2O&id=usdot-fhwa-stol_CARMAPlatform

Basically the 0 check will never be true you'll actually need to test against max unsigned int to check for rollover

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -607,33 +609,30 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
...
-
-  // Following TODO comments are tracked in this issue: https://github.com/usdot-fhwa-stol/carma-platform/issues/1947
   
-  if (config_.enable_carma_streets_connection ==false /*|| scheduled_enter_time_ == 0 */) // TODO uncomment when carma-street is capable of sending strategy params
+  if (config_.enable_carma_streets_connection ==false || scheduled_enter_time_ == 0) //UC2
   {
     nearest_green_entry_time = get_nearest_green_entry_time(current_state.stamp, earliest_entry_time, traffic_light) 
-                                          + ros::Duration(0.01); //0.01sec more buffer since green_light algorithm's timestamp picks the previous signal - Vehicle Estimation
-    is_entry_time_within_future_events = true; //UC2
+                                          + ros::Duration(0/01); //0.01sec more buffer since green_light algorithm's timestamp picks the previous signal - Vehicle Estimation
```
msmcconnell, line 616: New typo. Should be probably be EPSILON

msmcconnell (changes requested):

msmcconnell (approved):

PR #1971, opened 2022-11-08
CI: replace build-wrapper with compilation database

codygarver:


SaikrishnaBairamoni (approved):
Currently C++ tests step is disabled to check the sonar compilation database as the test failures causing 0% code coverage in sonar dashboard. Sonar cloud reports should be re-checked after fixing tests failures for accurate results.

PR #1983, opened 2022-12-07
Feature/check dynamic spat

aswath1:
# PR Details
## Description
UC3 Bug fixes

## Related Issue
Routing node crashing when spat data received.
https://github.com/usdot-fhwa-stol/carma-platform/issues/1977
The vehicle was stopping at the red SPAT but fails to move when it turns green spat. CARMA World model dynamic spat processing was not over righting the new spat received.
https://github.com/usdot-fhwa-stol/carma-platform/issues/1984

## Motivation and Context
Use Case 3 Integration Testing

## How Has This Been Tested?
Unit tested and integration tested with Blue Lexus.

## Types of changes

- [x] Defect fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that cause existing functionality to change)

## Checklist:

- [x] I have added any new packages to the sonar-scanner.properties file
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
[CARMA Contributing Guide](https://github.com/usdot-fhwa-stol/carma-platform/blob/develop/Contributing.md) 
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Issue #1982, opened 2022-12-06
Edge cases in UC3 ET logic


aswath1:
### Types of Issue

- [X] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

In lci plugin's ET logic, 2 edge cases should be considered. 

1. There is a check that verifies if the received ET is within a green phase or TBD area or not. This check is not implemented correctly. It is possible that the ET is within a red or yellow phase, but the check is satisfied and returns true (meaning ET is within green or TBD).
(Following logic should account for a case where ET can be erroneously inside the future states. In which case, the vehicle should notify the user since carma-street is not supposed to send ET that is within RED. The current behavior of "cruising until stop" can be kept the same.)

![image](https://user-images.githubusercontent.com/32398753/206007301-7257299e-3823-4399-a7d1-2cd2b29d5f71.png)

2. If ET is within green and [ET-Green_Buffer] is not within green (i.e., ET is between the start time of the green and the start time plus green buffer), then there is a logic that changes ET such that [ET-Green_Buffer] is also within green. This check is not implemented correctly either.
(Following logic's intention is to increase ET to match the green_start + green_buffer if ET is within green buffer. However it is not clear if it is exactly working that way or the comments can be better. Additionally, this logic would fail if there are more than 2 GREEN phases in the SPAT as it utilizes current time instead of ET to determine which green it is closest to.)

![image](https://user-images.githubusercontent.com/32398753/206008951-0aba4306-cc9e-472e-9acf-56d82ab8daff.png)

Here is the pseudocodes of the existing and proposed Logic:

[LCI_Plugin_ET_Logic_V2.docx](https://github.com/usdot-fhwa-stol/carma-platform/files/10196135/LCI_Plugin_ET_Logic_V2.docx)

### Carma version where this issue was discovered

carma-system-4.2.0+ (develop)

### Expected behavior
See above.
### Actual behavior

See above.

MishkaMN:
Resolved by above PR

PR #1990, opened 2022-12-15
Feature/uc3 et logic

MishkaMN:
# PR Details
## Description

During UC3 testing, there were 2 edge cases where the ET calculation can be improved. Please see the issue link for detailed description. This PR addresses those edge cases.

This PR shows extra changes in carma_wm_ctrl and carma_wm libraries. But this is already in the branch. [here](https://github.com/usdot-fhwa-stol/carma-platform/pull/1983). So those changes are tested. Not actually sure why are they showing as not changed, but those fixes are already in the release/isetta branch

Main change of this PR is only in lci_strategic_plugin.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1982

## Motivation and Context
TSMO UC3 integration testing

## How Has This Been Tested?
Integration testing in progress

## Types of changes

## Checklist:


lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -608,37 +608,66 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
...
+    }
+
   }
+
+ 
   
   ROS_DEBUG_STREAM("nearest_green_entry_time: " << std::to_string(nearest_green_entry_time.toSec()) << ", with : " << nearest_green_entry_time - current_state.stamp  << " seconds left at: " << std::to_string(current_state.stamp.toSec()));
   
-  if (!nearest_green_entry_time_cached_ && is_entry_time_within_future_events) 
+  if (nearest_green_entry_time_cached_)
```
Saeid-Solei, line 663: This check might change the nearest_green_entry_time to nearest_green_entry_time_cached_. I agree that in this case, nearest_green_entry_time_cached_ is most likely within a green. But still, I feel it is better to move this check to line 623 where we calculate nearest_green_entry_time. Then we can check the if the final nearest_green_entry_time is within green (tbd) or not.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -608,37 +608,66 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
...
   
-  if (!nearest_green_entry_time_cached_ && is_entry_time_within_future_events) 
+  if (nearest_green_entry_time_cached_) 
+  { // always pick later of buffered green entry time, or earliest entry time
+    nearest_green_entry_time = ros::Time(std::max(nearest_green_entry_time.toSec(), nearest_green_entry_time_cached_.get().toSec()));
+  }
+
+  ROS_DEBUG_STREAM("After accounting for cached - nearest_green_entry_time: " << std::to_string(nearest_green_entry_time.toSec()) << ", with : " << nearest_green_entry_time - current_state.stamp  << " seconds left at: " << std::to_string(current_state.stamp.toSec()));
+  
+  if (!nearest_green_entry_time_cached_ && is_entry_time_within_green_or_tdb)
```
Saeid-Solei, line 670: I don't understand why we need the first one? I am guessing what you want to say is if nearest_green_entry_time = nearest_green_entry_time_cached_, then none of these are required to check. Note that, it is possible that nearest_green_entry_time is not overwritten by nearest_green_entry_time_cached_. Anyway, I would say let remove the first check. If nearest_green_entry_time_cached_ exist, it will be checked earlier when calculating nearest_green_entry_time. So I don't see it necessary here. 

Also, you want to have an else condition for this check. If is_entry_time_within_green_or_tdb not true, then nearest_green_entry_time is within red or yellow, which means the vehicle shall go to handle stopping case. How are we capturing this case?

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -694,41 +723,36 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
...
   }
-  else if (nearest_green_entry_time_cached_) 
-  { // always pick later of buffered green entry time, or earliest entry time
-    nearest_green_entry_time = ros::Time(std::max(nearest_green_entry_time.toSec(), nearest_green_entry_time_cached_.get().toSec()));
-  }
-  
+
   if (nearest_green_entry_time_cached_ && nearest_green_entry_time > nearest_green_entry_time_cached_.get())
   {
     ROS_DEBUG_STREAM("Earliest entry time has gone past the cached green light. nearest_green_entry_time_cached_:" << std::to_string(nearest_green_entry_time_cached_.get().toSec()) << ", and earliest_entry_time: " << std::to_string(earliest_entry_time.toSec()));
```
Saeid-Solei, line 758: Another case is that the vehicle ET is pushed back to TBD which is pushed back in time due to a new fixed phase. Not critical, but this is another case where nearest_green_entry_time > nearest_green_entry_time_cached_ in UC3.

carma_wm_ros2/src/CARMAWorldModel.cpp
```diff
@@ -1502,26 +1516,24 @@ namespace carma_wm
...
-              curr_light->recorded_time_stamps = sim_.traffic_signal_states_[curr_intersection.id.id][current_movement_state.signal_group];
-              curr_light->recorded_start_time_stamps  = sim_.traffic_signal_start_times_[curr_intersection.id.id][current_movement_state.signal_group];
-            }
-	        }
-        } 
+            RCLCPP_DEBUG_STREAM(rclcpp::get_logger("carma_wm_ros2"), "intersection id: " << (int)curr_intersection.id.id << ", signal: " << (int)current_movement_state.signal_group
+             << ", start_time: " << std::to_string(lanelet::time::toSec(start_time_dynamic))
+             << ", end_time: " << std::to_string(lanelet::time::toSec(min_end_time_dynamic))
+             << ", state: " << received_state_dynamic);
+          }
```
Saeid-Solei, line 1529: This brace is extra. Please remove it.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
MishkaMN: - I think you concern is when last ET was in TDB (cached), but new calculated ET is closer and in green, lci might overwrite the new valid green ET with old TDB ET. If that's the case, lci resets nearest_green_entry_time_cached_ everytime it gets a new ET (it checks raw scheduled_entry_time values with old one) from the street, so this cached_ is only valid for the given ET. So it should work correctly. Let me know if you had other case in mind,

- However, the main reason I left it here is because of UC2 logic though. Taking it out to somewhere else might break UC2 or having to have similar code in two places.

MishkaMN: https://github.com/usdot-fhwa-stol/carma-platform/blob/feature/uc3_et_logic/lci_strategic_plugin/src/lci_strategic_plugin.cpp#L999

this is where it resets cached

lci_strategic_plugin/src/lci_strategic_plugin.cpp
MishkaMN: - Essentially on a higher level, this check is saying if the "logic to apply green_buffer" has been done before, avoid doing it again. Basically having nearest_green_entry_time_cached_ means buffer applying logic is run before and is good to reuse again. And this cached is only specific to one ET from street and resets if given a new one as mentioned in the previous comment.
- if not in green_or_tdb (which LCI now logs before this logic with ERROR msgs as requested), no buffer will be applied and the raw ET will be used to calculate some TS, and that TS will be failed by canArriveWithCertainty check which will make it cruiseAndStop.

carma_wm_ros2/src/CARMAWorldModel.cpp
MishkaMN: This is not extra, but a bad indentation, so fixed it

lci_strategic_plugin/src/lci_strategic_plugin.cpp
Saeid-Solei: If every time we receive a new ET that is different from the previous ET, we set the cached one to null, then I guess we will be fine. Because in this case, ET will be either equal to the previous one or the new one which is already checked over spat. The only other case I can think of is if the ET update comes later than the SPaT update. In that case, it is possible that we have an old ET with a new SPaT and the checks over spat will have different results. But I am not sure how much we need be worry about this case.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
Saeid-Solei: - I still thing the first check is not necessary. It is possible that nearest_green_entry_time_cached_ is not empty, but nearest_green_entry_time is greater than nearest_green_entry_time_cached_, and therefore, whether we need to add buffer or not needs to be checked again. With the implemented logic, nothing will be checked in this case, which is wrong.
- I see now. Ok sounds good.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
Saeid-Solei: "... has gone past the cashed entering time" is a more accurate description in my opinion.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
MishkaMN: nearest_green_entry_time_cached_ is minimum nearest_green_start_time + buffer. So if nearest_green_entry_time > nearest_green_entry_time_cached_ , automatically we don't need to apply buffer.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
MishkaMN: thanks!

lci_strategic_plugin/src/lci_strategic_plugin.cpp
MishkaMN: From logs, I can confirm that new SPAT is received every 0.1s while the ET is logged every 1s, so I am assuming that chance is very low.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
Saeid-Solei: Sort of true. I guess the only case nearest_green_start_time is greater than nearest_green_entry_time_cached_ is when EET is greater than nearest_green_entry_time_cached_. In this case, I guess EET will still be in the same green. But I am worry there might be an edge case that we are not considering now. I will leave the decision to you on whether keeping it or removing it from the check.

MishkaMN: For the sake of completeness, If EET is greater than cached, that will have already be saved into nearest_green_entry_time so that will be used as the new ET.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
Saeid-Solei: Ok. This will leave 1 sec for this possible issue. We can leave it for now, but again, let keep it in mind in case we see issues in testing. We might want to save these conversations somewhere.

MishkaMN: This is a very valid reason although unlikely. So we decided to keep it the same just to keep it same semantics as UC2 to not break anything. But will come back here if anything weird is noticed in the logs.

Issue #1985, opened 2022-12-09
UC3 LCI_Plugin HandleFailureCase Logic Bug


Saeid-Solei:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

In LCI Plugin, given an ET, the vehicle will first find a smoothed trajectory for itself (trajectory cases 1-8). Once the trajectory case is initially determined, the logic has some checks to decide its action and desired speed. One of the cases is called HandleFailureCase() which will be triggered if the vehicle was following a trajectory case previously and suddenly its ET becomes out of reach when the vehicle is close to the intersection and its distance to the stop bar is less than the safe stopping speed. In this case, since the vehicle is not able to stop at the stop bar with the max comfortable deceleration rate, HandleFailureCase() tells the vehicle to try to accelerate\decelerate to enter the intersection box with a speed close to the departure speed. 

There are two main scenarios that might trigger HandleFailureCase(). 
1. The vehicle ET is at the beginning of the green and the vehicle gets closer to the intersection that it should be based on its smoothed trajectory. In this case, the vehicle must slow down to allow the signal to change from red to green before entering the intersection box.
2. The vehicle ET is at the end of the green and as the vehicle gets closer to the intersection box, due to inaccuracy in trajectory control, the vehicle receives an ET that is pushed back to the beginning of the next green.

While the current logic might work for the second case (we can hope that if the vehicle continues with its previous plan and accelerate, it can enter the intersection box during the end of the green or the yellow change phase), it does not make much sense for the first case (if the vehicle is trying to enter the intersection box at the beginning of a green, but it is closer than it should be, it should not accelerate to enter the intersection box. Instead, it should try to slow down to allow the signal light to change from red to green). This logic is the cause of some failed test runs where the vehicle enters the intersection box before the light change from red to green, or after the end of the green and during yellow change\red clearance.

[...]

PR #1991, opened 2022-12-15
Feature/uc3 edge cases

aswath1:
# PR Details
## Description

In LCI Plugin, given an ET, the vehicle will first find a smoothed trajectory for itself (trajectory cases 1-8). Once the trajectory case is initially determined, the logic has some checks to decide its action and desired speed. One of the cases is called HandleFailureCase() which will be triggered if the vehicle was following a trajectory case previously and suddenly its ET becomes out of reach when the vehicle is close to the intersection and its distance to the stop bar is less than the safe stopping speed. In this case, since the vehicle is not able to stop at the stop bar with the max comfortable deceleration rate, HandleFailureCase() tells the vehicle to try to accelerate\decelerate to enter the intersection box with a speed close to the departure speed.

There are two main scenarios that might trigger HandleFailureCase().

The vehicle ET is at the beginning of the green and the vehicle gets closer to the intersection that it should be based on its smoothed trajectory. In this case, the vehicle must slow down to allow the signal to change from red to green before entering the intersection box.
The vehicle ET is at the end of the green and as the vehicle gets closer to the intersection box, due to inaccuracy in trajectory control, the vehicle receives an ET that is pushed back to the beginning of the next green.
## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1985

## Motivation and Context
TSMO UC3 integration testing

## How Has This Been Tested?
Integration testing in progress

## Types of changes

## Checklist:


Dockerfile
```diff
@@ -46,7 +46,7 @@ RUN ~/src/carma-platform/docker/checkout.bash
 # /////////////////////////////////////////////////////////////////////////////
 
 
-FROM base-image AS install
+FROM 4da4f7496dbb AS install
```
paulbourelly999, line 49: Is this correct way to do this?

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -369,82 +369,242 @@ void LCIStrategicPlugin::handleGreenSignalScenario(const cav_srvs::PlanManeuvers
...
-  params.case_num = CASE_1;
+  TrajectoryParams return_params;
+  TrajectoryParams traj_upper;
+  TrajectoryParams traj_lower;
 
   ROS_DEBUG_STREAM("HANDLE_LAST_RESORT_CASE: Starting...");
-  double modified_remaining_time;
+  double starting_downtrack = traffic_light_downtrack - remaining_downtrack;
+  double modified_remaining_time_upper; // upper meaning downtrack vs time trajectory is curved upwards
+  double modified_remaining_time_lower; // upper meaning downtrack vs time trajectory is curved lower
```
Saeid-Solei, line 383: // lower meaning ...

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -369,82 +369,242 @@ void LCIStrategicPlugin::handleGreenSignalScenario(const cav_srvs::PlanManeuvers
...
+  traj_upper.v0_ = starting_speed;
+  traj_upper.x0_ = starting_downtrack;
+  traj_upper.is_algorithm_successful = false;
+  traj_upper.case_num = CASE_1;
+
+  if (departure_speed > starting_speed)
+  {
+    if ((pow(departure_speed,2) - pow(starting_speed,2))/(2*max_comfort_accel_) >= remaining_downtrack)
+    {
+      ROS_DEBUG_STREAM("Failure case upper 1");
```
Saeid-Solei, line 402: It is not technically a failure case. It just means that the vehicle cannot reach to departure speed which is ok. You can say, HandleFailureCase -> Upper Trajectory -> Actual speed less than desired departure speed.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -369,82 +369,242 @@ void LCIStrategicPlugin::handleGreenSignalScenario(const cav_srvs::PlanManeuvers
...
+      traj_upper.v2_ = traj_upper.v1_;
+      traj_upper.a2_ = traj_upper.a1_;
+      traj_upper.x2_ = traj_upper.x1_;
+
+      traj_upper.modified_departure_speed = modified_departure_speed_upper;
+      traj_upper.modified_remaining_time = modified_remaining_time_upper;
+    }
+    else // NOTE: most likely will not happen as it would have happened at trajectory smoothing part
+    { 
+      ROS_DEBUG_STREAM("Failure case upper 2");
```
Saeid-Solei, line 422: It is not technically a failure case. It just means that the vehicle cannot reach to departure speed which is ok. You can say, HandleFailureCase -> Upper Trajectory -> Actual speed greater than desired departure speed.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -369,82 +369,242 @@ void LCIStrategicPlugin::handleGreenSignalScenario(const cav_srvs::PlanManeuvers
...
-  else
+
+  if (departure_speed < starting_speed) // NOTE: most cases will not run due to departure speed being equal to speed limit
   {
-    params.a1_ = max_comfort_decel_;
-    params.v1_ = sqrt(pow(starting_speed, 2) + (2 * max_comfort_decel_ * remaining_downtrack));
-  }
+    if ((pow(departure_speed,2) - pow(starting_speed,2))/(2*max_comfort_decel_) >= remaining_downtrack)
+    {
+      ROS_DEBUG_STREAM("Failure case upper 3");
```
Saeid-Solei, line 451: It is not technically a failure case. It just means that the vehicle cannot reach to departure speed which is ok. You can say, HandleFailureCase -> Upper Trajectory -> Current Speed > Desired Departure Speed,  Actual Departure Speed > Desired Departure Speed

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -369,82 +369,242 @@ void LCIStrategicPlugin::handleGreenSignalScenario(const cav_srvs::PlanManeuvers
...
+
+  // the upper ET
+  // accel case
+  traj_upper.t0_ = current_time;
+  traj_upper.v0_ = starting_speed;
+  traj_upper.x0_ = starting_downtrack;
+  traj_upper.is_algorithm_successful = false;
+  traj_upper.case_num = CASE_1;
+
+  if (departure_speed > starting_speed)
```
Saeid-Solei, line 398: Please add equality condition to one of the if statement.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -369,82 +369,242 @@ void LCIStrategicPlugin::handleGreenSignalScenario(const cav_srvs::PlanManeuvers
...
+  bool calculation_success_upper = true; // identifies places in codes where calculation can be invalid such as negative distance
+
+  // TODO change light_controlled_intersection_tactical_plugin...
+
+  // the upper ET
+  // accel case
+  traj_upper.t0_ = current_time;
+  traj_upper.v0_ = starting_speed;
+  traj_upper.x0_ = starting_downtrack;
+  traj_upper.is_algorithm_successful = false;
```
Saeid-Solei, line 395: This never turns to true. I think you missed changing this to true?

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -369,82 +369,242 @@ void LCIStrategicPlugin::handleGreenSignalScenario(const cav_srvs::PlanManeuvers
...
+  }
+
+  // Handle hard failure case such as nan or invalid states
+  if (is_return_params_found && !isnan(return_params.modified_departure_speed) && return_params.modified_departure_speed > epsilon_ &&
+      return_params.modified_departure_speed < speed_limit ) //80_mph
+  {
+    ROS_DEBUG_STREAM("Updated the speed, and using modified_departure_speed: " << return_params.modified_departure_speed);
+    print_params(return_params);
+    return return_params;
   }
```
Saeid-Solei, line 577: Let add logging to else statement to know if is_return_params_found is green and the speed is wrong.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -369,82 +369,242 @@ void LCIStrategicPlugin::handleGreenSignalScenario(const cav_srvs::PlanManeuvers
...
 
-  params.a1_ = 0;
-  params.v1_ = starting_speed;
-  params.x1_ = traffic_light_downtrack;
+  // If everything above fails, cruise but log as ERROR
```
Saeid-Solei, line 579: We are missing the emergency stop case in the last page of the pseudocode. Is there any reason?

carma/scripts/pose_offset_node.py
```diff
@@ -0,0 +1,73 @@
...
+
+  rospy.init_node('pose_offset_node')
+
+  # Load params
+  x_offset = rospy.get_param("/pose_x_offset", 1.0)
+  y_offset = rospy.get_param("/pose_y_offset", 1.0)
+  z_offset = rospy.get_param('/pose_z_offset', 0.0)
+
+  # Setup publishers
+  pub = rospy.Publisher('selected_pose', PoseStamped, queue_size=10)
```
msmcconnell, line 56: This PR contains multiple changes (such as this file) which were explicitly asked to be removed in this PR
https://github.com/usdot-fhwa-stol/carma-platform/pull/1983

Why are they back? Please clean up the commits on this branch.

msmcconnell (changes requested):

lci_strategic_plugin/src/lci_strategic_plugin.cpp
Saeid-Solei: We are still missing the emergency stop case. Is there any reason?

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -392,14 +392,14 @@ TrajectoryParams LCIStrategicPlugin::handleFailureCaseHelper(const lanelet::Carm
   traj_upper.t0_ = current_time;
   traj_upper.v0_ = starting_speed;
   traj_upper.x0_ = starting_downtrack;
-  traj_upper.is_algorithm_successful = false;
+  traj_upper.is_algorithm_successful = true;
```
Saeid-Solei, line 395: I think the better way would be setting it to false at first and changing it to true if the trajectory is created.

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -2250,7 +2250,7 @@ void WMBroadcaster::updateUpcomingSGIntersectionIds()
     }
   } //END intersections
 
-  RCLCPP_DEBUG_STREAM(rclcpp::get_logger("carma_wm_ctrl"), "MAP msg: Intersection ID = " <<  map_msg_intersection_id << ", Signal Group ID =" << cur_signal_group_id );
+  RCLCPP_ERROR_STREAM(rclcpp::get_logger("carma_wm_ctrl"), "MAP msg: Intersection ID = " <<  map_msg_intersection_id << ", Signal Group ID =" << cur_signal_group_id );
```
msmcconnell, line 2253: If you need to temporarily make this change it's fine but don't forget to revert to debug before release.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -694,41 +888,36 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
...
+
+      nearest_green_entry_time_cached_ = nearest_green_signal_start_time + ros::Duration(config_.green_light_time_buffer + EPSILON);
+      
+      // EPSILON=0.01 is there because if predictState's input exactly falls on ending_time it picks the previous state.
+      //For example, if 0 - 10s is GREEN, and 10 - 12s is YELLOW, checking exactly 10.0s will return GREEN,
+      //but 10.01s will return YELLOW. This 0.01 convention is used throughout the file, so thought it is better
+      //to keep it consistent and probably too detailed for the user to think about, which is why it is not included in the buffer.
+      //Actually including in the buffer doesn't work because it uses that same buffer to check early and late. If buffer is 2s and 
+      //green starts at 10s, it will check +/-2s from 12s. If the buffer was 2.01s and green starts at 10s again, it checks +/-2.01 
+      //from 12.01, so both checks 10s.
```
msmcconnell, line 914: These are a lot of major logic changes without unit tests? Do all the old tests still pass? 
I would think new tests would help you find issues faster if they can be written.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
aswath1: Trajectory is created in lci tactical plugin. Strategic plugin calls tactical so it needs to be true here.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
aswath1: The edge case happens 1, 2s away from the intersection. The car won't be able to react that fast enough and will go through intersection anyway, which will immediately have a different plugin plan for it to have full speed again. So essentially it is same as just cruising through.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
Saeid-Solei: I know. But it shall remain false unless we have a valid trajectory. So we should keep it false and change it to true in any condition that we successfully create a trajectory. This way, we would know if we do not have a successful trajectory.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
Saeid-Solei: This is exactly for not allowing the vehicle to go through the intersection. This case forces the vehicle to decelerate with a higher rate to stop if possible. It is not the same as cruising. I would still prefer to include this case. I should be relatively easy given the pseudocode, is that right?

aswath1: ok

lci_strategic_plugin/src/lci_strategic_plugin.cpp
aswath1: Trajectory is created in a totally different plugin (tactical). And we have to make it true here otherwise that tactical plugin will ignore all this change as it will be running last successful trajectory. So we have to make it true here.
Having it false will make the tactical plugin ignore the trajectory. So it needs to be true either way.

Saeid-Solei: I understand that if it be false, then the tactical plugin does not create any trajectory. But what I am saying is that we should not assume this is true before finding the upper trajectory. I agree that we will most likely create a successful one every time, but this parameter indicates whether we are able to fill out the trajectory info correctly or not. Anyway, not a big issue. I leave it to you.

aswath1: If its not successful trajectory then we are not using these parameters.

Saeid-Solei: I believe there is a confusion here. Please let me know if you need to talk over phone to clear this up. But you can skip this for now.

lci_strategic_plugin/CMakeLists.txt
```diff
@@ -93,10 +93,10 @@ install(DIRECTORY launch config
 
 catkin_add_gmock(${PROJECT_NAME}-test 
  test/test_main.cpp
- #test/test_fixture.h
- #test/test_strategic_plugin_helpers.cpp
+ test/test_fixture.h
+ ##test/test_strategic_plugin_helpers.cpp
```
msmcconnell, line 97: This file was modified but is commented out in the CMake why?

msmcconnell (changes requested):

msmcconnell:
CI is also showing test failures
```
]0;colcon test [108/113 done] [3 ongoing]Starting >>> lci_strategic_plugin
--- stderr: lci_strategic_plugin
Errors while running CTest
---
]0;colcon test [108/113 done] [4 ongoing]Finished <<< lci_strategic_plugin [0.45s]	[ with test failures ]
```

lci_strategic_plugin/CMakeLists.txt
aswath1: That file is edited just so that it builds, but unit tests fail so I kept it turned off because it was turned off anyways and as an indication to get back to it in the future. All test added yesterday for handle case failure passes.

msmcconnell (approved):

Issue #1996, opened 2022-12-21
Black Pacifica UC3 service call failure to tactical plugin


MishkaMN:
### Types of Issue

- [X] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

During UC3 integration testing, Black Pacifica would fail to call light_controlled_intersection_tactical_plugin service. Some runs the platform failed to call it the entire run. After turning off camera and lidar logging, the failure rate reduced to about 10% of the time it fails to call randomly. Currently it is believed to be error in ROS2 service call to components where less computing power could lead to timeouts for the service calls.

Could be related to this issue where the solution was to separate each node into different containers (it is not 100% clear it was the solution as well, but the error did not occur since the change):

https://github.com/usdot-fhwa-stol/carma-platform/issues/1961

### Carma version where this issue was discovered

carma-system:4.2.0+ release/isetta
### Expected behavior
tactical plugin service call should not fail

### Actual behavior

See above. 

### Steps to reproduce the actual behavior

Do UC3 run on black pacifica with lidar and camera logging turned on. And inspect ROS2 guidance log to see if tactical plugin's service call failed or not.

MishkaMN:
Update on the issue:
- Times when the service call fails in plan_delegator exactly matches with when it is not able to retrieve most recent current_pose and its timestamp. Currently it is not clear whether if ros_bridge is failing to relay the pose or the node itself is being locked somehow, or even the current_pose is not printing (which is ROS1).  
- ROS1 lci_strategic_plugin reliably gets its pose from tf transform. Currently not sure if current_pose is working or not.
- Sometimes when the carma-platform fails to call 100% of the time could be if the developer engaged carma-platform before lci_tactical_plugin was done configuring.
- Turning off logging (especially those with ERROR msgs) seem to improve the chance the most as opposed to turning on/off other tactical and strategic plugins (increasing plan_delegator's service call time-out to 200ms also did not have positive effect).

MishkaMN:
This issue is resolved by above downsampling PR

PR #2003, opened 2023-01-09
Improve lci_tactical_plugin service call success rate

MishkaMN:
# PR Details
## Description

This PR tries to improve the service call success rate of the lci tactical plugin by omitting heavy logic of trajectory generation when unnecessary. This plugin uses previous successful trajectory whenever possible (by its own criteria I won't get into), but still generates the new trajectory all the time even though it is not used. Therefore, this PR makes the plugin to check early on whether if new trajectory generation is required, and if not, return the old one, making the service call faster in theory.

This PR also accounts for cases when plan_delegator might lose its time when sending requests (see the issue and its updates). Since the time was based on current_pose, this PR changes to use its own internal time.  

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1996

## Motivation and Context
UC3 integration test. See the above issue

## How Has This Been Tested?
Tested on Black Pacifica with 10+ runs. 
Heavy logging seem to affect the performance as well. So only LCI tactical and strategic plugin's logging (out of the default ones) were enabled during these testing 

## Types of changes

## Checklist:


MishkaMN:
Closing and recreated here https://github.com/usdot-fhwa-stol/carma-platform/pull/2004

PR #2004, opened 2023-01-09
Feature/lci tactical sc fail

MishkaMN:
# PR Details
## Description

This PR tries to improve the service call success rate of the lci tactical plugin by omitting heavy logic of trajectory generation when unnecessary. This plugin uses previous successful trajectory whenever possible (by its own criteria I won't get into), but still generates the new trajectory all the time even though it is not used. Therefore, this PR makes the plugin to check early on whether if new trajectory generation is required, and if not, return the old one, making the service call faster in theory.

This PR also accounts for cases when plan_delegator might lose its time when sending requests (see the issue and its updates). Since the time was based on current_pose, this PR changes to use its own internal time.  

EDIT:
With downsampling increased from 12 to 30, the points_and_target_speed size of 685 reduced to 58, which significantly reduces the callback time of call services. It has been identified during testing that this plugin would take 17.8s when it should have taken 8.0s at worst worst case. This caused the plan_delegator's call service to timeout and fail almost always. It is not 100% conclusive why plan_delegator also fails to enter its location callback (which makes it not generate new trajectory again), but it is likely related as it works fine when the call service succeeds. 

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/1996

## Motivation and Context
UC3 integration test. See the above issue

## How Has This Been Tested?
Tested on Black Pacifica with 10+ runs. 
Heavy logging seem to affect the performance as well. So only LCI tactical and strategic plugin's logging (out of the default ones) were enabled during these testing 

## Types of changes

## Checklist:


localization_manager/config/parameters.yaml
```diff
@@ -40,7 +40,7 @@ sequential_timesteps_until_gps_operation: 5
 # 3 - AUTO select between NDT and GNSS without GNSS timeout
 # 4 - GNSS only with NDT initialization
 # 5 - GNSS only with fixed offset
-localization_mode: 4
+localization_mode: 5
```
msmcconnell, line 43: Probably still aren't ready to make this the default since it requires editing the map offset locally.

plan_delegator/src/plan_delegator.cpp
```diff
@@ -250,7 +250,10 @@ namespace plan_delegator
         // set current vehicle state if we have NOT planned any previous trajectories
         if(latest_trajectory_plan.trajectory_points.empty())
         {
-            plan_req->header.stamp = latest_pose_.header.stamp;
+            plan_req->header.stamp = now(); // used to be latest_pose_.header.stamp which was not reliable
```
msmcconnell, line 253: Why was that not reliable? This change creates an inconstancy between the pose in the planning request and the request time.

light_controlled_intersection_tactical_plugin/src/light_controlled_intersection_tactical_plugin.cpp
```diff
@@ -180,10 +155,51 @@ namespace light_controlled_intersection_tactical_plugin
...
         {
             resp->trajectory_plan = last_trajectory_;
-            RCLCPP_DEBUG_STREAM(rclcpp::get_logger("light_controlled_intersection_tactical_plugin"), "Last Traj's target time: " << rclcpp::Time(last_trajectory_.trajectory_points.back().target_time).seconds() << ", and stamp:" << rclcpp::Time(req->header.stamp).seconds() << ", and scheduled: " << std::to_string(last_successful_scheduled_entry_time_));
+            RCLCPP_DEBUG_STREAM(rclcpp::get_logger("light_controlled_intersection_tactical_plugin"), "Last Traj's target time: " << std::to_string(rclcpp::Time(last_trajectory_.trajectory_points.back().target_time).seconds()) << ", and stamp:" << std::to_string(rclcpp::Time(req->header.stamp).seconds()) << ", and scheduled: " << std::to_string(last_successful_scheduled_entry_time_));
             RCLCPP_DEBUG_STREAM(rclcpp::get_logger("light_controlled_intersection_tactical_plugin"), "EDGE CASE: USING LAST TRAJ: " << (int)last_case_.get());
         }  
-        else if (trajectory.trajectory_points.size () < 2)
+
+        if (!resp->trajectory_plan.trajectory_points.empty()) // if has valid trajectory saved from before return
```
msmcconnell, line 162: I'm not seeing a check to validate that this trajectory actually makes sense. 
Before just defaulting to the last trajectory there should be some form of validation that the vehicle is still following it within some error threshold.

msmcconnell (changes requested):

plan_delegator/src/plan_delegator.cpp
MishkaMN: Initially created this change because latest_pose was not being updated in the plugin due to service calls being timed out. It is actually not conclusive if the repeatedly failed service calls are causing the node to freeze or it's because something else is freezing which causing the both tasks to fail. However, decreasing the execution time of the tactical plugin helped significantly reduce the service call failures, which also removed the need for this change. So reverting it back to latest_pose's stamp which keeps the consistency between the pose and request.

light_controlled_intersection_tactical_plugin/src/light_controlled_intersection_tactical_plugin.cpp
MishkaMN: We discussed offline and we want to compare more in position and time than speed. Since that's what is relevant for crossing the stop bar.
"Am I within 3m of my expected position at the current time"
As far as the current changes go, we are tabling further implementation as enhancement and merging this version. For the record, the vehicle is preventing RED light entrance by following:

All decisions to stop the vehicle from going through red light happens on strategic level. Since it involves a different plugin, it is guaranteed to follow that stopping trajectory.
If trajectory smoothing algorithm doesnt have valid solution it decides to cruise and stop at red.
Furthermore, I think this might be irrelevant to our specific discussion, but on a strategic level, there is basically only one edge case we identified that lets the vehicle go through red but we discussed it with Ying Yang today that more implementation should be held off until we have resources to do so.
On a tactical level, there is no decisions being made using lights. Only decisions are whether if
new TScase is same as before and last traj has more than 2 points and not within its last 1 sec trajectory, then use last trajectory
if not all above, use new trajectory

MishkaMN: Created new enhancement issue: 
https://github.com/usdot-fhwa-stol/carma-platform/issues/2006

PR #2010, opened 2023-01-15
Feature/refactor lci strat and handle et in tbd case

MishkaMN:
# PR Details
## Description

- This PR refactors ET calculation logic into its own function so that the main maneuver callback function is cleaner.
- Also the lci_strategic_plugin checks if the vehicle is in desired distance to start stopping while ET is in TBD. If that's the case, the vehicle will start stopping as the ET is not determined until ET is in green and valid again. 

NOTE: Current design of the logic makes the car stop regardless of the traffic signal at that time if ET and distances are above mentioned condition. More discussion needs to occur on 
- whether the vehicle should go through the intersection if the vehicle is able to go in GREEN despite the huge ET discrepancy. 
- if it needs to go, how? (just cruise or use UC2 ET logic to trajectory smooth again)

Update on NOTE: I had this discussion with Saeid, and concluded that the case should not happen where carma-street gives different ET than the time vehicle can make it to the current GREEN. If carma-street thinks the vehicle cannot make it in current GREEN and gives different ET, then platform should also conclude the same mathematically as they are using same equations. Their respective green_buffers are only thing that could be different, but in that case the vehicle should follow carma-street's buffer anyways.

## Motivation and Context
UC3 integration testing. See above issue.

## How Has This Been Tested?
Not tested yet

## Types of changes

## Checklist:

### DIFF
--- a/lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
+++ b/lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
@@ -514,6 +514,22 @@ private:
                                                                      bool bounds_inclusive = true) const;
   
   /**
+   * \brief Get the final entry time from either vehicle's own internal ET calculation (TSMO UC2) or from carma-street (TSMO UC3)
+   *        This function also applies necessary green_buffer to adjust the ET. 
+   * \param current_state Current state of the vehicle
+   * \param earliest_entry_time Earliest entry time calculated by the vehicle
+   * \param traffic_light traffic signal the vehicle is using
+   * \param is_entry_time_within_green_or_tdb this function sets it to true if ET is in green or TDB (related to UC3, always set to true in UC2)
+   * \param in_tdb this function sets it to true if ET is in TDB (not modified if in UC2)
+   * \return final entry time the vehicle uses to enter the intersection
+  */
+  ros::Time get_final_entry_time_and_conditions(const VehicleState& current_state, 
+                                                const ros::Time& earliest_entry_time, 
+                                                lanelet::CarmaTrafficSignalPtr traffic_light, 
+                                                bool& is_entry_time_within_green_or_tdb, 
+                                                bool& in_tdb);
+
+  /**
    * \brief Provides the scheduled entry time for the vehicle in the future. This scheduled time is the earliest possible entry time that 
    *        is during green phase and after timestamps both that is kinematically possible and required times for vehicle in front to pass through first 
    *
--- a/lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
+++ b/lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
@@ -740,6 +740,8 @@ private:
   double max_comfort_accel_ = 2.0;  // acceleration rates after applying miltiplier
   double max_comfort_decel_ = -2.0; 
   double max_comfort_decel_norm_ = -1 * max_comfort_decel_;
+  double emergency_decel_norm_ = -2 * max_comfort_decel_;
+
   boost::optional<ros::Time> nearest_green_entry_time_cached_;
 
   //! World Model pointer
--- a/lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
+++ b/lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
@@ -68,6 +68,7 @@ enum TSCase {
   STOPPING=9,
   UNAVAILABLE = 10,
   EMERGENCY_STOPPING=11,
+  DEGRADED_TSCASE=12  // when not performing trajectory smoothing, but making through GREEN
 };
 
 struct TrajectoryParams
@@ -203,13 +204,12 @@ public:
 /////////////////////////////////////////////////////////////////////////////////////////////////////////////////
    /**
    * \brief Useful metrics for LCI Plugin
-   * \param case_num_ Current speed profile case generated
+   * \param last_case_num_ Current speed profile case generated
    * \param distance_remaining_to_tf_ distance_remaining_to_traffic signal meters
    * \param earliest_entry_time_ earliest_entry_time in sec
    * \param scheduled_entry_time_ scheduled_entry_time in sec
    * 
    */
-  TSCase case_num_ = TSCase::UNAVAILABLE; 
   TSCase last_case_num_ = TSCase::UNAVAILABLE; 
   double distance_remaining_to_tf_ = 0.0;
   double earliest_entry_time_ = 0.0;
--- a/lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
+++ b/lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
@@ -67,6 +67,7 @@ enum TSCase {
   CASE_8 = 8,
   STOPPING=9,
   UNAVAILABLE = 10,
+  EMERGENCY_STOPPING=11,
 };
 
 struct TrajectoryParams
--- a/lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
+++ b/lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
@@ -519,15 +519,13 @@ private:
    * \param current_state Current state of the vehicle
    * \param earliest_entry_time Earliest entry time calculated by the vehicle
    * \param traffic_light traffic signal the vehicle is using
-   * \param is_entry_time_within_green_or_tdb this function sets it to true if ET is in green or TDB (related to UC3, always set to true in UC2)
-   * \param in_tdb this function sets it to true if ET is in TDB (not modified if in UC2)
-   * \return final entry time the vehicle uses to enter the intersection
+   * \return tuple of <final entry time the vehicle uses to enter the intersection, 
+   *         is_entry_time_within_green_or_tbd: true if ET is in green or TBD (related to UC3, always set to true in UC2),
+   *         in_tbd: true if ET is in TBD (false if in UC2)>
   */
-  ros::Time get_final_entry_time_and_conditions(const VehicleState& current_state, 
+  std::tuple<ros::Time, bool, bool> get_final_entry_time_and_conditions(const VehicleState& current_state, 
                                                 const ros::Time& earliest_entry_time, 
-                                                lanelet::CarmaTrafficSignalPtr traffic_light, 
-                                                bool& is_entry_time_within_green_or_tdb, 
-                                                bool& in_tdb);
+                                                lanelet::CarmaTrafficSignalPtr traffic_light);
 
   /**
    * \brief Provides the scheduled entry time for the vehicle in the future. This scheduled time is the earliest possible entry time that 
--- a/lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
+++ b/lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
@@ -780,6 +780,8 @@ private:
   FRIEND_TEST(LCIStrategicTestFixture, composeTrajectorySmoothingManeuverMessage);
   FRIEND_TEST(LCIStrategicTestFixture, findSpeedLimit);
   FRIEND_TEST(LCIStrategicTestFixture, handleFailureCaseHelper);
+  FRIEND_TEST(LCIStrategicTestFixture, planWhenETInTBD);
+  
   // Algo Unit Tests
   FRIEND_TEST(LCIStrategicTestFixture, calc_estimated_entry_time_left);
   FRIEND_TEST(LCIStrategicTestFixture, estimate_distance_to_stop);
--- a/lci_strategic_plugin/src/lci_strategic_plugin.cpp
+++ b/lci_strategic_plugin/src/lci_strategic_plugin.cpp
@@ -738,6 +738,7 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
 
   ROS_DEBUG_STREAM("current_state_speed: " << current_state_speed);
   ROS_DEBUG_STREAM("intersection_speed_: " << intersection_speed_.get());
+  ROS_DEBUG_STREAM("distance_remaining_to_traffic_light: " << distance_remaining_to_traffic_light);
 
   intersection_end_downtrack_ =
       wm_->routeTrackPos(exit_lanelet.centerline2d().front().basicPoint2d()).downtrack;
@@ -754,7 +755,7 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
 
     return;
   }
-
+  
   /////////////  2. Start of TSMO UC2 & UC3 Algorithm : ET determination //////////////
 
   ros::Time earliest_entry_time = current_state.stamp + get_earliest_entry_time(distance_remaining_to_traffic_light, speed_limit, 
@@ -762,15 +763,11 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
 
   ROS_DEBUG_STREAM("earliest_entry_time: " << std::to_string(earliest_entry_time.toSec()) << ", with : " << earliest_entry_time - current_state.stamp  << " left at: " << std::to_string(current_state.stamp.toSec()));
 
-  //ros::Time nearest_green_entry_time; 
-  //bool is_entry_time_within_green_or_tbd = false;
-  //bool in_tbd = true;
-
   auto [nearest_green_entry_time, is_entry_time_within_green_or_tbd, in_tbd] = get_final_entry_time_and_conditions(current_state, earliest_entry_time, traffic_light);
 
   if (nearest_green_entry_time == ros::Time(0))
     return;
-    
+
   ROS_DEBUG_STREAM("Final nearest_green_entry_time: " << std::to_string(nearest_green_entry_time.toSec()));
 
   auto et_state = traffic_light->predictState(lanelet::time::timeFromSec(nearest_green_entry_time.toSec()));
@@ -792,7 +789,7 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
   print_params(ts_params);
 
   ROS_DEBUG_STREAM("SPEED PROFILE CASE:" << ts_params.case_num);
-
+   
   /////////////  4. Safety Check against traffic signals and Final Maneuver Generation //////////////
 
   double safe_distance_to_stop = pow(current_state.speed, 2)/(2 * max_comfort_decel_norm_) + config_.stopping_location_buffer / 2; //Idea is to aim the middle part of stopping buffer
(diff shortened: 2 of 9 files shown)


lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -952,22 +825,14 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
     ts_params.case_num = CASE_8;
     print_params(ts_params);
   }
+  else if (distance_remaining_to_traffic_light <= desired_distance_to_stop && in_tdb) // Given ET is in TDB, but vehicle is too close to intersection
```
Saeid-Solei, line 828: I think you must add "ts_params.is_algorithm_successful && ts_params.case_num != TSCase::CASE_8 &&" to the else if condition here as well

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -952,22 +825,14 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
...
+  {
+    ROS_DEBUG_STREAM("ET is still in TDB despite the vehicle being in desired distance to start stopping. Trying to handle this edge case gracefully...");
```
Saeid-Solei, line 830: I guess you don't need to add handlestopping again? Would it be taken care of by the following logics and checks?

lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
```diff
@@ -513,6 +513,22 @@ class LCIStrategicPlugin
                                                                      bool shortest_path_only = false,
                                                                      bool bounds_inclusive = true) const;
   
+  /**
+   * \brief Get the final entry time from either vehicle's own internal ET calculation (TSMO UC2) or from carma-street (TSMO UC3)
+   *        This function also applies necessary green_buffer to adjust the ET. 
+   * \param current_state Current state of the vehicle
+   * \param earliest_entry_time Earliest entry time calculated by the vehicle
+   * \param traffic_light traffic signal the vehicle is using
+   * \param is_entry_time_within_green_or_tdb this function sets it to true if ET is in green or TDB (related to UC3, always set to true in UC2)
```
kjrush, line 522: Minor: Should, be TBD (**T**o **B**e **D**etermined) I believe

lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
```diff
@@ -513,6 +513,22 @@ class LCIStrategicPlugin
...
+   * \param in_tdb this function sets it to true if ET is in TDB (not modified if in UC2)
```
kjrush, line 523: Not a huge fan of return via parameter reference, as a general rule. Especially since "in_tbd" reads like an input variable name because of the "in" part.

TBH, I'd recommend either defining a struct when multiple return values are necessary or simply using [`std::tuple<...>` ](https://en.cppreference.com/w/cpp/utility/tuple). std::tuple is probably more readable if we're able to use the structured binding syntax mentioned in that documentation link.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -932,10 +792,23 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
...
-  // CASE SELECTION END
+  /////////////  4. Safety Check against traffic signals and Final Maneuver Generation //////////////
+
+   double safe_distance_to_stop = pow(current_state.speed, 2)/(2 * max_comfort_decel_norm_);
+  ROS_DEBUG_STREAM("safe_distance_to_stop at max_comfort_decel:  " << safe_distance_to_stop << ", max_comfort_decel_norm_: " << max_comfort_decel_norm_);
+
+  double desired_distance_to_stop = pow(current_state.speed, 2)/(2 * max_comfort_decel_norm_ * config_.deceleration_fraction) + config_.desired_distance_to_stop_buffer;
+  ROS_DEBUG_STREAM("desired_distance_to_stop at: " << desired_distance_to_stop << ", where effective deceleration rate is: " << max_comfort_decel_norm_ * config_.deceleration_fraction);
+  
+  desired_distance_to_stop = std::max(desired_distance_to_stop, config_.stopping_location_buffer);
```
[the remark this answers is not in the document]
kjrush: It's worth clamping it either way, better to be defensive I think.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -932,10 +792,23 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
 
   ROS_DEBUG_STREAM("SPEED PROFILE CASE:" << ts_params.case_num);
 
-  // CASE SELECTION END
+  /////////////  4. Safety Check against traffic signals and Final Maneuver Generation //////////////
+
+   double safe_distance_to_stop = pow(current_state.speed, 2)/(2 * max_comfort_decel_norm_);
+  ROS_DEBUG_STREAM("safe_distance_to_stop at max_comfort_decel:  " << safe_distance_to_stop << ", max_comfort_decel_norm_: " << max_comfort_decel_norm_);
+
+  double desired_distance_to_stop = pow(current_state.speed, 2)/(2 * max_comfort_decel_norm_ * config_.deceleration_fraction) + config_.desired_distance_to_stop_buffer;
```
kjrush, line 800: Should this be absolute value'd, just in case or is max_comfort_decel_norm_ guaranteed to be always >0?

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -932,10 +792,23 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
 
   ROS_DEBUG_STREAM("SPEED PROFILE CASE:" << ts_params.case_num);
 
-  // CASE SELECTION END
+  /////////////  4. Safety Check against traffic signals and Final Maneuver Generation //////////////
+
+   double safe_distance_to_stop = pow(current_state.speed, 2)/(2 * max_comfort_decel_norm_);
```
kjrush, line 797: Same as below re absolute value

kjrush, line 797: Should this also have a buffer?

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -932,10 +792,23 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
...
+  desired_distance_to_stop = std::max(desired_distance_to_stop, config_.stopping_location_buffer);
+
+  ROS_DEBUG_STREAM("new desired_distance_to_stop: " << desired_distance_to_stop);
+
+  ROS_DEBUG_STREAM("distance_remaining_to_traffic_light:  " << distance_remaining_to_traffic_light << ", current_state.speed: " << current_state.speed);
 
   // Although algorithm determines nearest_green_time is possible, check if the vehicle can arrive with certainty (Case 1-7)
-  if (ts_params.is_algorithm_successful && ts_params.case_num != TSCase::CASE_8) 
+  if (ts_params.is_algorithm_successful && ts_params.case_num != TSCase::CASE_8 && 
+    (distance_remaining_to_traffic_light >= desired_distance_to_stop || !in_tdb)) // Given ET is in TDB, but vehicle is too close to intersection
```
kjrush, line 811: Am I misunderstanding? Comment seems inaccurate, code reads `|| !in_tdb` but comment says ET is TBD.

kjrush (changes requested):
I don't really know a lot about how this algorithm is structured, but I'm not sure I'm a huge fan of where the stop check is implemented. It's conditioned on a branch where `ts_params.is_algorithm_successful && ts_params.case_num != TSCase::CASE_8` , but something like stopping at a stop bar is fundamentally safety critical. A basic need-to-stop check really should be independent of the state of the algorithm or communications with carma-streets or anything, I think.

Aditionally, are there unit tests at all for this module? I'd really like to see a sample case that triggers this new stop-check branch and ensures that the conditions are sane.

lci_strategic_plugin/src/lci_strategic_plugin_algo.cpp
```diff
@@ -213,6 +215,158 @@ ros::Duration LCIStrategicPlugin::get_earliest_entry_time(double remaining_dista
 
 }
 
+ros::Time LCIStrategicPlugin::get_final_entry_time_and_conditions(const VehicleState& current_state, const ros::Time& earliest_entry_time, lanelet::CarmaTrafficSignalPtr traffic_light, bool& is_entry_time_within_green_or_tdb, bool& in_tdb)
```
kjrush, line 218: I'm assuming this code is essentially entirely code moved from lci_strategic_plugin.cpp's main body? That this was just lifted from the maneuver planning callback to here so that it can be referenced? Or are there more substantial changes in the body of this code?

MishkaMN: Yep! This is just refactor and doesn't have any changes in the code.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
saina-ramyar: I also have a similar question as @Saeid-Solei . won't you still face this condition if the vehicle is in the middle of the stopping process?
Also, we have "if else", does the condition have a final "else"? Because I see more "if" statements later

lci_strategic_plugin/src/lci_strategic_plugin.cpp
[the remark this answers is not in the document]
MishkaMN: Yeah, I also went ahead and noted that it should be less than 1 on the config parameter

lci_strategic_plugin/src/lci_strategic_plugin.cpp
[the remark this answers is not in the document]
MishkaMN: We should separate ET_in_TBD vs ET_in_GREEN in this case because the platform shouldn't generate handleGreenManeuver when distance is too close. While in ET_in_GREEN however, it should always try to generate it, so the 2 flags should not be lumped together.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
MishkaMN: This is just for debugging purposes, and it maybe better to cover more conditions without restricting to ts_params.is_algorithm_successful && ts_params.case_num != TSCase::CASE_8. As those conditions can be referenced from other log lines.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
MishkaMN: Yep. Next few logics directly start checking whether if the vehicle should cruise and stop, directly stop, or handleFailureCase-s when not within safety stopping distance

lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
MishkaMN: This is my bad, I keep interchanging the two. I changed all instances

lci_strategic_plugin/src/lci_strategic_plugin.cpp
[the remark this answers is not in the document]
Saeid-Solei: What I think is that by adding "is_entry_time_within_green_or_tdb == true", we can be sure that ET is not within red or yellow. It confirms that ET is within green or TBD and the "in_tbd" parameter confirms if ET is in TBD or not. Otherwise, if "in_tbd" is false, ET could be within red and we still run the logic in this if condition, right?

lci_strategic_plugin/src/lci_strategic_plugin.cpp
Saeid-Solei: I see. If the handlefailurecase and handlestoppingcase for this edge case is covered by the later logic, then this is fine.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
MishkaMN: Yeah it is guaranteed to be always >0 because the name also suggests it is the "norm"

lci_strategic_plugin/src/lci_strategic_plugin.cpp
MishkaMN: Same comment as before

lci_strategic_plugin/src/lci_strategic_plugin.cpp
saina-ramyar: sounds good. In that case add something to that log statement saying condition handled by later logic or something.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
MishkaMN: Comment was incorrect due to copy/paste and not editing. I changed it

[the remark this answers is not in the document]
MishkaMN: Ah that is true. The old logic would have checked certainty case and would have detected the RED, but I went ahead and added the check to make it more clear to readers

lci_strategic_plugin/include/lci_strategic_plugin/lci_strategic_plugin.h
MishkaMN: I went ahead and made them to tuple

MishkaMN:
> I don't really know a lot about how this algorithm is structured, but I'm not sure I'm a huge fan of where the stop check is implemented. It's conditioned on a branch where `ts_params.is_algorithm_successful && ts_params.case_num != TSCase::CASE_8` , but something like stopping at a stop bar is fundamentally safety critical. A basic need-to-stop check really should be independent of the state of the algorithm or communications with carma-streets or anything, I think.

Upon discussion with @Saeid-Solei  on Friday, he suggested that having basic stop check case without consideration of carma-streets or the algorithm in this case is significantly hard problem itself. From my understanding, having basic stop check might undermine the trajectory smoothing algorithm as it constantly changes speed to make it to the green. So it needs to be considered in the algorithm itself, which it does as best as it could. Let me know if that is what you meant @Saeid-Solei 

> Aditionally, are there unit tests at all for this module? I'd really like to see a sample case that triggers this new stop-check branch and ensures that the conditions are sane.

HandleFailureCase which is when the vehicle is already within safety stopping distance is tested recently and rigorously. I am adding few more tests to check the stopping case as well as TBD case. I know from previous discussions that the unit test for the whole plan maneuver call back was failing even before UC3 started due to changes in UC2. And since integration testing achieves pretty consistent and correct behavior, I'd suggest we do tests for entire callback later due to limited time.

Saeid-Solei:
> I don't really know a lot about how this algorithm is structured, but I'm not sure I'm a huge fan of where the stop check is implemented. It's conditioned on a branch where `ts_params.is_algorithm_successful && ts_params.case_num != TSCase::CASE_8` , but something like stopping at a stop bar is fundamentally safety critical. A basic need-to-stop check really should be independent of the state of the algorithm or communications with carma-streets or anything, I think.
> 

I totally agree that the ideal implementation would include separate logics for trajectory smoothing and safety where the recommended speed from trajectory smoothing can be overwritten by the safe\stopping speed. But this fix will take much more time and could introduce new bugs which I believe is not recommended given that we are at the verification testing stage. However, I agree that this is the ideal enhancement that we can work on in future. 

Currently, the vehicle trajectory and speed are decided purely based on the estimated entering time. Vehicle will first check if the received entering time is within an effective green or not. If it is within a fixed red or yellow phases, it will throw error with the correct implementation. But if it is within green or the TBD area (the time area where we did not fix any phases yet), then the vehicle tries to find a trajectory case. Trajectory cases 1-7 does not require the vehicle to stop. Case 8 requires the vehicle to go with min speed and then stop. There are also couple of other checks to see if the vehicle shall stop or continue with its trajectory case.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -753,191 +755,94 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
...
-  TrajectoryParams ts_params = get_ts_case(req.header.stamp.toSec(), nearest_green_entry_time.toSec(), current_state_speed, intersection_speed_.get(), speed_limit, config_.algo_minimum_speed, max_comfort_accel_, max_comfort_decel_, current_state.downtrack, traffic_light_down_track, distance_remaining_to_traffic_light, boundary_distances, boundary_traj_params);
-  print_params(ts_params);
+    if (!validLightState(stopping_arrival_state_optional, stopping_arrival_time))
+    {
+      ROS_ERROR_STREAM("Unable to resolve give signal for stopping_arrival_state_optional: " << std::to_string(stopping_arrival_time.toSec()));
+      return;
+    }
 
-  ROS_DEBUG_STREAM("SPEED PROFILE CASE:" << ts_params.case_num);
+    if (stopping_arrival_state_optional.get().second == lanelet::CarmaTrafficSignalState::STOP_AND_REMAIN)
```
Saeid-Solei, line 830: Did we decide to get rid of your earlier fix and replace it with red light safeguard? Given this new changes, we can force the vehicle to stop if the vehicle is within the stopping distance but its ET is within a green. I thought at the minimum we need your previous changes and the plan is to add the red light safeguard on top of it. Please clarify what is the plan? 
If this is still confusing, I would say we should have a quick meeting with the team about this.

Saeid-Solei: Also, here you probably want to add yellow light state as well.

kjrush: Is that the case here? It looks like it's projecting the estimated arrival time at the stop bar forward into the known signal timing data, so that if we're expected to arrive on a green we can continue to approach the stop bar even if it's currently red.

Saeid-Solei: The projection definitely helps reducing the chance of the edge case I mentioned, but it is still possible.

Another edge case to think about is when a vehicle is trying to enter the intersection box at the end of a green. Imagine a case where the vehicle ET is around the end of the green and the vehicle can technically enter during the current green. With this check alone on top of everything, the estimated stopping arrival time could be within red while the vehicle ET is within the current green and the vehicle can speed up and enter the intersection box.

Sorry if it feels like I am repeating stuff here again, but I guess I am finding new stuff as I think more about the safeguard.

lci_strategic_plugin/src/lci_strategic_plugin.cpp
```diff
@@ -753,191 +760,95 @@ void LCIStrategicPlugin::planWhenAPPROACHING(const cav_srvs::PlanManeuversReques
...
-  print_params(ts_params);
+    if (!validLightState(stopping_arrival_state_optional, stopping_arrival_time))
+    {
+      ROS_ERROR_STREAM("Unable to resolve give signal for stopping_arrival_state_optional: " << std::to_string(stopping_arrival_time.toSec()));
+      return;
+    }
 
-  ROS_DEBUG_STREAM("SPEED PROFILE CASE:" << ts_params.case_num);
+    if (stopping_arrival_state_optional.get().second == lanelet::CarmaTrafficSignalState::STOP_AND_REMAIN || 
+      stopping_arrival_state_optional.get().second == lanelet::CarmaTrafficSignalState::PROTECTED_CLEARANCE)
```
Saeid-Solei, line 836: After more discussions, I feel like maybe we can get rid of yellow change in this _if_ condition. Since we are checking the stopping arrival time, maybe we don't want the vehicle to stop if the stopping arrival time is within yellow!!!

MishkaMN: gotcha removed.

PR #2062, opened 2023-03-02
Feature/lightbar manager ros2

MishkaMN:
# PR Details
## Description

Ported lightbar_manager to ROS2
## Related Issue
TODO

## Motivation and Context
See above.

## How Has This Been Tested?
local integration tested

## Types of changes

## Checklist:


lightbar_manager/src/lightbar_manager_worker.cpp
```diff
@@ -121,7 +127,7 @@ namespace lightbar_manager
 
         if (raw_map.size() < 4)
```
adev4a, line 128: Why did the input to this function change? Wont this map always be <4, we're only assigning it one value

MishkaMN: - ROS2 doesn't have dictionary parameter loading, so I turned it into 2 vectors of 1-to-1 correlation (just like how carma_wm_ctrl loads the offset parameters).
 - Yes we are assigning only 1 element, but there has to be at least 4 elements (or key/value pairs) total in the map, since there is minimum 4 CDATypes that needs to be mapped.

lightbar_manager/include/lightbar_manager/lightbar_manager_node.hpp
```diff
@@ -82,49 +98,70 @@ class LightBarManager
...
+        * \brief Turn signal callback function helper
+        * \return 
         */
-        void setupUnitTest();
+        void processTurnSignal(const automotive_platform_msgs::msg::TurnSignalCommand& msg);
 
         // Node Data
-        std::string node_name_;
-        double spin_rate_;
+        std::string node_name_ = "lightbar_manager";
```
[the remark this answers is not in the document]
MishkaMN: I realized it was not needed in the worker class, but still needed on node level here. This is because the node uses this name to check its priority against potential other plugins

adev4a (approved):

PR #2084, opened 2023-04-12
Assorted fix for integration testing ERV use case

MishkaMN:
# PR Details
## Description

This PR includes:

1. ERV and CMV opposing direction scenario fix:

Previously the ERV and CMV's lane index is obtained from counting how many routable lanelets are to the right of the current lanelet. (if no right llts, then idx is 0, if there is 1, then idx is 1 etc)
This works if the CMV and ERV are both in the same direction. However, if the ERV is going opposite direction, it can also be 0 idx (because ERV is at right-most lane in its own direction), which the CMV detects erroneously as they are in a same lane if CMV is also on rightmost lane but opposite direction. (edited) 
So the approach here to fix this was to check if any of the ERV's shortest path lanelet is inside the CMV's all possible route lanelets. If it matches, they are on the same direction. If it doesn't match, then they are opposite direction and ignore the ERV. We only do this once to avoid heavy for loops and directions not going to change unlike lanes (edited) 

2. Ability to indicate to turn on hazard light (only notify ssc_driver not directly included here, because the driver is ROS1)
 Also correlated PR is in: https://github.com/usdot-fhwa-stol/carma-ssc-interface-wrapper/pull/149

3. Added a guard when creating route from ERV's destination points. Sometimes even if carma-cloud removes the passed destination points, for unknown reason, the passed point can still be in destination points of the ERV. Since not being able route means losing tracking of the ERV altogether, extra filter was added to only pick points ahead to make the routing stable.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/2085
https://github.com/usdot-fhwa-stol/carma-platform/issues/2086
https://github.com/usdot-fhwa-stol/carma-platform/issues/2088

## Motivation and Context
ERV Use Case

## How Has This Been Tested?
Unit tested

## Types of changes

## Checklist:

### DIFF
--- a/approaching_emergency_vehicle_plugin/config/parameters.yaml
+++ b/approaching_emergency_vehicle_plugin/config/parameters.yaml
@@ -6,7 +6,7 @@ passing_threshold : 15.0
 # Double: A threshold; if the estimated duration until an ERV passes the ego vehicle is below this, the
 #         ERV will no longer attempt to change lanes in reaction to an approaching ERV.
 # Units: Seconds
-approaching_threshold : 33.0
+approaching_threshold : 90.0
 
 # Double: A threshold; an actively-passing ERV is considered to have completed passing the ego vehicle when
 #         it is this distance in front of the ego vehicle.
@@ -26,7 +26,7 @@ bsm_processing_frequency : 1.0
 # Double: The amount that the speed limit of a lanelet will be reduced by when planning a maneuver when an
 #         ERV is actively passing the ego vehicle.
 # Units: m/s
-speed_limit_reduction_during_passing : 4.4704 # NOTE: 4.4704 m/s is 10 mph
+speed_limit_reduction_during_passing : 8.94 # NOTE: 4.4704 m/s is 10 mph
 
 # Double: The minimum speed limit that a lanelet's speed limit will be reduced to when an ERV is actively passing
 #         the ego vehicle.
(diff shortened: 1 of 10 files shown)


MishkaMN:
don't merge yet as it hasn't been integration tested

approaching_emergency_vehicle_plugin/src/approaching_emergency_vehicle_plugin_node.cpp
```diff
@@ -475,16 +510,40 @@ namespace approaching_emergency_vehicle_plugin
...
       lanelet::ConstLanelet erv_current_lanelet = erv_future_route.get().shortestPath()[0];
+
+      if (is_same_direction_.find(erv_information.vehicle_id) == is_same_direction_.end()) // no need to check again if direction is set already
+      {
+        is_same_direction_[erv_information.vehicle_id] = false;
+        for (auto llt: erv_future_route.get().shortestPath()) // checks if ERV is on the same path assuming CMV got all of its planned route when detected
+        {
+          if (wm_->getRoute()->contains(llt))
+          {
+            is_same_direction_[erv_information.vehicle_id] = true;
```
saina-ramyar, line 526: it seems the direction determination only happens once. What if there is a road closure /or route change?
Is something like that included in this use case?

approaching_emergency_vehicle_plugin/src/approaching_emergency_vehicle_plugin_node.cpp
```diff
@@ -475,16 +510,40 @@ namespace approaching_emergency_vehicle_plugin
...
+      }
+
       erv_information.lane_index = wm_->getMapRoutingGraph()->rights(erv_current_lanelet).size();
+
+     
       RCLCPP_DEBUG_STREAM(rclcpp::get_logger(logger_name), "ERV's lane index is " << erv_information.lane_index);
     }
     else{
-      RCLCPP_DEBUG_STREAM(rclcpp::get_logger(logger_name), "ERV's shortest path is empty!");
+      RCLCPP_ERROR_STREAM(rclcpp::get_logger(logger_name), "ERV's shortest path is empty!");
```
saina-ramyar, line 546: Do you want it as an error or its changed for debugging?

MishkaMN: I think this should probably be WARN message

approaching_emergency_vehicle_plugin/config/parameters.yaml
```diff
@@ -6,7 +6,7 @@ passing_threshold : 15.0
 # Double: A threshold; if the estimated duration until an ERV passes the ego vehicle is below this, the
 #         ERV will no longer attempt to change lanes in reaction to an approaching ERV.
 # Units: Seconds
-approaching_threshold : 33.0
+approaching_threshold : 90.0
```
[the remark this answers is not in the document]
MishkaMN: Config changes are not merging I can remove them

approaching_emergency_vehicle_plugin/src/approaching_emergency_vehicle_plugin_node.cpp
saina-ramyar: ok please add a note to mention this limitation.

PR #2131, opened 2023-08-01
update localization launch

adev4a:
# PR Details
This PR removes the localization launch file and updates the ros2 launch file with the required simulation mode launch argument.

## Types of changes

## Checklist:


carma/launch/carma_src.launch.py
```diff
@@ -140,6 +140,9 @@ def generate_launch_description():
     vector_map_file = LaunchConfiguration('vector_map_file')
     declare_vector_map_file = DeclareLaunchArgument(name='vector_map_file', default_value='/opt/carma/maps/vector_map.osm')
 
+    simulation_mode = LaunchConfiguration('simulation')
```
JonSmet, line 143: Can we change this arg from `simulation` to `simulation_arg` as well?

carma/launch/localization.launch.py
```diff
@@ -371,6 +373,11 @@ def generate_launch_description():
         map_param_loader_container,
         pcd_map_file_loader_container,
         ndt_matching_container,
-        ekf_localizer_container,
         subsystem_controller
     ]) 
+
+    if simulation_mode is False:
```
JonSmet, line 379: I'm not sure whether we are actually able to use LaunchArguments directly in python conditional statements like this. I think the ROS2 launch approach for this would be to use IfConditional() instead. To do this, you can add:

```
from launch.conditions import IfCondition
from launch.substitutions import LaunchConfiguration, PythonExpression

    ...

    carma_port_drayage_plugin_container = ComposableNodeContainer(
        condition=IfCondition(PythonExpression(['not ', simulation_mode])), # Nodes in this container will only be launched if 'simulation_mode' evaluates to 'False'
    
    ...
```

JonSmet (changes requested):

JonSmet (approved):

Issue #2153, opened 2023-09-21
yield_plugin to process predicted_states of external_object


MishkaMN:
### Summary

This issue is to upgrade and integration test yield plugin to yield to moving objects in an intersection. 

The developer should create a dummy node that can publish such objects in an intersection to roadway objects topic to test this.

Add the logic for processing the motion predicted states in this story!

Especially roadway objects not moving along the host vehicle’s route (perpendicular for example intersecting the route).

### Reasoning for new functionality

Currently yield_plugin assumes the roadway objects are going along the route, which is not true for pedestrians for example.

MishkaMN:
This has been addressed under VRU use case in general. 
Currently the motion_computation model assumes just "constant velocity" motion model for almost all obstacles.

PR #2154, opened 2023-09-21
Feature/yield plugin fix

MishkaMN:
# PR Details
## Description
Improved yield_plugin functionality to consider objects' predicted_states, instead of assuming the roadway_object is moving along the route. Summary of changes:

- yield_plugin doesn't rely on roadway_objects node. It has its own filtering functionality to get the objects from external_objects as some pedestrians may not be on the road, but will be in a few seconds that may result in collision. It was decided that roadway_object_node is not needed to include that functionality due to what its name indicates. 
- added collision checking algorithm that interpolates the trajectory points to a same timestamp to get a potential collision by checking a certain configurable radius. 
- changed to using nh_->get_logger() to decrease the usage of loggers with hardcoded name
-- In order for this change to work, another PR was created to fix the way logger levels are currently set in carma-platform: https://github.com/usdot-fhwa-stol/carma-utils/pull/151
- minor fixes for integration test and unit test

## Related GitHub Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/2153

## Related Jira Key

[CDAR-124](https://usdot-carma.atlassian.net/browse/CDAR-124)

## Motivation and Context
VRU use case

## How Has This Been Tested?
cdasim integration tested

## Types of changes

## Checklist:

[CDAR-124]: https://usdot-carma.atlassian.net/browse/CDAR-124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


yield_plugin/src/yield_plugin.cpp
```diff
@@ -135,17 +135,17 @@ namespace yield_plugin
     if (incoming_request.strategy == "carma/cooperative-lane-change")
     {
       if (!map_projector_) {
-        RCLCPP_ERROR_STREAM(nh_->get_logger(),"Cannot process mobility request as map projection is not yet set!");
+        RCLCPP_ERROR_STREAM(rclcpp::get_logger("yield_plugin"),"Cannot process mobility request as map projection is not yet set!");
```
adamlm, line 138: Is there a reason this was changed? Using the `rclcpp::` version tightly couples logging to the Node name. If a user were to change the Node's name, we would suddenly lose logging.

Also, we don't need the `_STREAM` logging variant here, we should use `RCLCPP_ERROR`.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -135,17 +135,17 @@ namespace yield_plugin
...
         return;
       }
       if (incoming_request.plan_type.type == carma_v2x_msgs::msg::PlanType::CHANGE_LANE_LEFT || incoming_request.plan_type.type == carma_v2x_msgs::msg::PlanType::CHANGE_LANE_RIGHT)
       {
-        RCLCPP_DEBUG_STREAM(nh_->get_logger(),"Cooperative Lane Change Request Received");
+        RCLCPP_DEBUG_STREAM(rclcpp::get_logger("yield_plugin"),"Cooperative Lane Change Request Received");
```
adamlm, line 143: Same as above.

adamlm, line 143: Same as above.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -135,17 +135,17 @@ namespace yield_plugin
...
         lc_status_msg.status = carma_planning_msgs::msg::LaneChangeStatus::REQUEST_RECEIVED;
         lc_status_msg.description = "Received lane merge request";
         if (incoming_request.m_header.recipient_id == config_.vehicle_id)
         {
-          RCLCPP_DEBUG_STREAM(nh_->get_logger(),"CLC Request correctly received");
+          RCLCPP_DEBUG_STREAM(rclcpp::get_logger("yield_plugin"),"CLC Request correctly received");
```
adamlm, line 148: Same as above

adamlm, line 148: Same as above

yield_plugin/src/yield_plugin.cpp
```diff
@@ -155,7 +155,7 @@ namespace yield_plugin
         carma_v2x_msgs::msg::Trajectory incoming_trajectory = incoming_request.trajectory;
         std::string req_strategy_params = incoming_request.strategy_params;
         clc_urgency_ = incoming_request.urgency;
-        RCLCPP_DEBUG_STREAM(nh_->get_logger(),"received urgency: " << clc_urgency_);
+        RCLCPP_DEBUG_STREAM(rclcpp::get_logger("yield_plugin"),"received urgency: " << clc_urgency_);
```
adamlm, line 158: Same as above.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -167,9 +167,9 @@ namespace yield_plugin
         int start_lanelet_id = pt.get<int>("sl");
         int end_lanelet_id = pt.get<int>("el");
         double req_traj_speed = (double)req_traj_speed_full + (double)(req_traj_fractional)/10.0;
-        RCLCPP_DEBUG_STREAM(nh_->get_logger(),"req_traj_speed" << req_traj_speed);
-        RCLCPP_DEBUG_STREAM(nh_->get_logger(),"start_lanelet_id" << start_lanelet_id);
-        RCLCPP_DEBUG_STREAM(nh_->get_logger(),"end_lanelet_id" << end_lanelet_id);
+        RCLCPP_DEBUG_STREAM(rclcpp::get_logger("yield_plugin"),"req_traj_speed" << req_traj_speed);
+        RCLCPP_DEBUG_STREAM(rclcpp::get_logger("yield_plugin"),"start_lanelet_id" << start_lanelet_id);
+        RCLCPP_DEBUG_STREAM(rclcpp::get_logger("yield_plugin"),"end_lanelet_id" << end_lanelet_id);
```
adamlm, line 172: Same as above.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -232,28 +232,30 @@ namespace yield_plugin
     if (req->initial_trajectory_plan.trajectory_points.size() < 2){
       throw std::invalid_argument("Empty Trajectory received by Yield");
     }
+    std::chrono::system_clock::time_point start_time = std::chrono::system_clock::now();  // Start timing the execution time for planning so it can be logged
```
adamlm, line 235: Is there a reason we are accessing the system clock directly instead of going through ROS? From my understanding, the ROS API provides access to system time.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -385,14 +392,19 @@ namespace yield_plugin
         double dt = (2 * original_traj_downtracks[i]) / (current_speed + prev_speed);
         jmt_tpp = original_tp.trajectory_points[i];
         jmt_tpp.target_time =  rclcpp::Time(jmt_trajectory_points[i-1].target_time) + rclcpp::Duration(dt*1e9);
+        std::cout << "non-empty x: " << jmt_tpp.x << ", y:" << jmt_tpp.y << ", t:" << std::to_string(rclcpp::Time(jmt_tpp.target_time).seconds()) << std::endl;
```
adamlm, line 395: We should be using ROS's logging features instead of directly printing to `std::cout`.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -385,14 +392,19 @@ namespace yield_plugin
...
       else
       {
-        RCLCPP_DEBUG_STREAM(nh_->get_logger(),"target speed is zero");
+        RCLCPP_DEBUG_STREAM(rclcpp::get_logger("yield_plugin"),"target speed is zero");
         // if speed is zero, the vehicle will stay in previous location.
         jmt_tpp = jmt_trajectory_points[i-1];
-        jmt_tpp.target_time =  rclcpp::Time(jmt_trajectory_points[0].target_time) + rclcpp::Duration(traj_target_time*1e9);
+        // MISH: Potential bug where if purepursuit accidentally picks a point that has previous time, then it may speed up. target_time is assumed to increase 
+        jmt_tpp.target_time =  rclcpp::Time(std::max((rclcpp::Time(jmt_trajectory_points[0].target_time) + rclcpp::Duration(traj_target_time*1e9)).seconds(), rclcpp::Time(jmt_trajectory_points[i -1 ].target_time).seconds()) * 1e9);
+        std::cout << "empty x: " << jmt_tpp.x << ", y:" << jmt_tpp.y << ", t:" << std::to_string(rclcpp::Time(jmt_tpp.target_time).seconds()) << std::endl;
```
adamlm, line 406: Same as above.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -403,106 +415,243 @@ namespace yield_plugin
...
+        if (route_llt_ids_.find(llt.id()) != route_llt_ids_.end())
+        {
+          on_route = true;
+          on_route_idx = j;
+          break;
+        }
+      }
+      if (on_route)
+        break;
+    }
```
adamlm, line 455: Do we need an indexing `for` loop here? From what I can see, a range-based `for` loop suffices, and it provides bounds safety.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -403,106 +415,243 @@ namespace yield_plugin
...
+    
+    double smallest_dist = std::numeric_limits<double>::max();
+    for (int i = 0; i < trajectory1.trajectory_points.size() - 1; ++i) 
+    {
+      auto p1a = trajectory1.trajectory_points[i];
+      auto p1b = trajectory1.trajectory_points[i + 1];
+      for (int j = on_route_idx; j < trajectory2.size() - 1; ++j) 
+      {
+        auto p2a = trajectory2[j];
+        auto p2b = trajectory2[j + 1];
```
adamlm, line 471: We should be accessing elements using `.at()` instead of `operator[]` unless performance is critical. I think the bounds checking outweighs the performance hit here.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -403,106 +415,243 @@ namespace yield_plugin
...
     return jmt_trajectory;
   }
+
+  void YieldPlugin::external_objects_callback(const carma_perception_msgs::msg::ExternalObjectList::UniquePtr msg)
+  {
+    external_objects_ = msg->objects;
+  }
   
-  carma_planning_msgs::msg::TrajectoryPlan YieldPlugin::update_traj_for_object(const carma_planning_msgs::msg::TrajectoryPlan& original_tp, double initial_velocity) 
+  boost::optional<rclcpp::Time> YieldPlugin::detect_collision_time(const carma_planning_msgs::msg::TrajectoryPlan& trajectory1, const std::vector<carma_perception_msgs::msg::PredictedState>& trajectory2, double collision_radius)
```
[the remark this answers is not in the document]
adamlm: A lot of the variables in this function can be made `const auto`.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -403,106 +415,243 @@ namespace yield_plugin
...
+
+        RCLCPP_DEBUG_STREAM(rclcpp::get_logger("yield_plugin"), "p2a.x: " << p2a.predicted_position.position.x << ", p2a.y: " << p2a.predicted_position.position.y);
+        RCLCPP_DEBUG_STREAM(rclcpp::get_logger("yield_plugin"), "p2b.x: " << p2b.predicted_position.position.x << ", p2b.y: " << p2b.predicted_position.position.y);
+
+        // Linearly interpolate positions at a common timestamp for both trajectories
+        double dt = (p2a_t - p1a_t) / (p1b_t - p1a_t);
+        double x1 = p1a.x + dt * (p1b.x - p1a.x);
+        double y1 = p1a.y + dt * (p1b.y - p1a.y);
+        double x2 = p2a.predicted_position.position.x;
+        double y2 = p2a.predicted_position.position.y;
```
adamlm, line 490: This should ideally be moved to a separate function. We can keep this as is for now, but I wanted to make a note.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -403,106 +415,243 @@ namespace yield_plugin
...
-    host_vehicle_size.y = config_.vehicle_width;
-    host_vehicle_size.z = config_.vehicle_height; 
+    boost::optional<carma_perception_msgs::msg::ExternalObject> earliest_collision_obj = boost::none;
+    std::set<int> checked_external_object_ids;
 
+    std::vector<carma_perception_msgs::msg::PredictedState> new_list;
+    RCLCPP_DEBUG_STREAM(rclcpp::get_logger("yield_plugin"), "ExternalObjects size: " << external_objects_.size());
 
-    std::vector<carma_perception_msgs::msg::RoadwayObstacle> rwol_collision;
+    if (!wm_->getRoute())
```
adamlm, line 550: Replacing this with `wm_->getRoute() != nullptr` reduces the mental overhead in my opinion. In its form now, users have to look up `getRoute` to see what it returns. Is it a boolean, `std::optional`, etc.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -403,106 +415,243 @@ namespace yield_plugin
...
-      if(current_velocity.linear.x > 0.0) {
-          RCLCPP_DEBUG_STREAM(nh_->get_logger(),"(object_down_downtrack - vehicle_downtrack)/current_velocity.linear.x");
-          RCLCPP_DEBUG_STREAM(nh_->get_logger(),dist_to_object/current_velocity.linear.x);
-
-          // Check to see if the object is in front of us and has a time-to-collision lower than our
-          // desired horizon
-          if(dist_to_object >= 0 &&
-            dist_to_object / current_velocity.linear.x < config_.collision_horizon) {
-              rwol_collision.push_back(i);
+    for (auto curr_obstacle : external_objects_)
```
adamlm, line 566: This should be using `const auto &` to avoid making copies.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -403,106 +415,243 @@ namespace yield_plugin
...
+      if (rclcpp::Time(curr_obstacle.predictions.back().header.stamp) > plan_start_time)
+      {
+        carma_perception_msgs::msg::PredictedState curr_state;
+        // artificially include current position as one of the predicted states
+        curr_state.header.stamp = curr_obstacle.header.stamp;
+        curr_state.predicted_position.position.x = curr_obstacle.pose.pose.position.x;
+        curr_state.predicted_position.position.y = curr_obstacle.pose.pose.position.y;
+        curr_state.predicted_velocity.linear.x = curr_obstacle.velocity.twist.linear.x;
+        new_list.push_back(curr_state);
+        new_list.insert(new_list.end(), curr_obstacle.predictions.begin(), curr_obstacle.predictions.end());
```
adamlm, line 580: The `curr_obstacle` iterators should be `cbegin()` and `cend()`, respectively.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -403,106 +415,243 @@ namespace yield_plugin
...
+
+        auto collision_time = detect_collision_time(original_tp, new_list, config_.intervehicle_collision_distance);
+        
+        if (collision_time)
```
adamlm, line 584: Using `collision_time != boost::none` or `collision_time != std::nullopt` is more readable.

yield_plugin/include/yield_plugin/yield_plugin.hpp
```diff
@@ -221,8 +220,22 @@ class YieldPlugin
...
    * \brief msg The proj string defining the projection.
    */ 
-  void georeferenceCallback(const std_msgs::msg::String::UniquePtr msg);
+  void georeference_callback(const std_msgs::msg::String::UniquePtr msg);
+  
+  /**
+   * \brief Callback for external objects with predictions in the environment
+   * \param msg The object list msg.
+   */ 
+  void external_objects_callback(const carma_perception_msgs::msg::ExternalObjectList::UniquePtr msg);
```
adamlm, line 229: Why are we using `std::unique_ptr` here? As far as I know, ROS does not give exclusive ownership to the received messages. If we are making a `std::unique_ptr` from a `std::shared_ptr`, I believe we are making a copy of the underlying message.

adamlm (changes requested):
Two general issues I see consistently that we should check for:

- `rclcpp::get_logger("yield_plugin")` tightly couples the Node's name to logging. I also imagine it is more computationally intensive because ROS has to search for a node with that name. I could be wrong on this point.
- There is a lot of usage of `RCLCPP_<LEVEL>_STREAM` when `RCLCPP_<LEVEL>` would suffice.
- Many variables can be made `const auto`.

yield_plugin/include/yield_plugin/yield_plugin_node.hpp
```diff
@@ -74,8 +74,8 @@ class YieldPluginNode : public carma_guidance_plugins::TacticalPlugin
     // Subscribers
     carma_ros2_utils::SubPtr<carma_v2x_msgs::msg::MobilityRequest> mob_request_sub_;
     carma_ros2_utils::SubPtr<carma_v2x_msgs::msg::BSM> bsm_sub_;
+    carma_ros2_utils::SubPtr<carma_perception_msgs::msg::ExternalObjectList> external_objects_sub_;
     carma_ros2_utils::SubPtr<std_msgs::msg::String> georeference_sub_;
```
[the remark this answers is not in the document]
MishkaMN: I think the reason we have the SubPtr is because the publisher is not `rclcpp::Publisher` but `std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<T>>;` which is `carma_ros2_utils::PubPtr<T>`. So it might better to be consistent?

yield_plugin/src/yield_plugin_node.cpp
```diff
@@ -85,7 +87,8 @@ namespace yield_plugin
     // Subscriber
     mob_request_sub_ = create_subscription<carma_v2x_msgs::msg::MobilityRequest>("incoming_mobility_request", 10, std::bind(&YieldPlugin::mobilityrequest_cb,worker_.get(),std_ph::_1));
     bsm_sub_ = create_subscription<carma_v2x_msgs::msg::BSM>("bsm_outbound", 1, std::bind(&YieldPlugin::bsm_cb,worker_.get(),std_ph::_1));
-    georeference_sub_ = create_subscription<std_msgs::msg::String>("georeference", 10, std::bind(&YieldPlugin::georeferenceCallback,worker_.get(),std_ph::_1));
+    georeference_sub_ = create_subscription<std_msgs::msg::String>("georeference", 10, std::bind(&YieldPlugin::georeference_callback,worker_.get(),std_ph::_1));
+    external_objects_sub_ = create_subscription<carma_perception_msgs::msg::ExternalObjectList>("external_object_predictions", 20, std::bind(&YieldPlugin::external_objects_callback,worker_.get(),std_ph::_1));
```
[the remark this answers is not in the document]
MishkaMN: done

yield_plugin/config/parameters.yaml
```diff
@@ -3,7 +3,8 @@
 # Value type: Desired
 # No unit
 acceleration_adjustment_factor: 4.0 
-# Time horizon for collision detection
+# Time horizon for collision detection for vehicles on the route
+# NOTE: not applied for pedestrians or other non-vehicle obstacles on the road
 # Units: s
 # Value type: Desired
 collision_horizon: 10.0
```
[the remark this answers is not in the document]
MishkaMN: done

yield_plugin/include/yield_plugin/yield_plugin_node.hpp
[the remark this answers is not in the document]
adamlm: I think we should remove `carma_ros2_utils::PubPtr<T>` as well. I'm not sure what the benefit is by aliasing the publisher type. `rclcpp::Publisher` and `rclcpp_lifecycle::LifecyclePublisher` have different interfaces and behaviors/expectations. If we want to be consistent, I suggest we use the `rclcpp` types instead of the `carma_ros2_utils` aliases. Thoughts?

inlanecruising_plugin/src/inlanecruising_plugin.cpp
```diff
@@ -107,7 +107,7 @@ void InLaneCruisingPlugin::plan_trajectory_callback(
     p.planner_plugin_name = plugin_name_;
   }
     
-  if (config_.enable_object_avoidance)
+  if (config_.enable_object_avoidance && original_trajectory.trajectory_points.size() >= 2)
```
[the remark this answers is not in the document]
MishkaMN: ILC sends trajectories to yield_plugin to be modified. However, it shouldn't send invalid trajectory such as those with 1 or 0 points.

yield_plugin/test/test_yield.cpp
```diff
@@ -39,12 +38,10 @@
...
 #include <carma_planning_msgs/msg/trajectory_plan_point.hpp>
 #include <carma_planning_msgs/msg/trajectory_plan.hpp>
-#include <carma_perception_msgs/msg/roadway_obstacle_list.hpp>
-#include <carma_perception_msgs/msg/roadway_obstacle.hpp>
 #include <carma_perception_msgs/msg/predicted_state.hpp>
 
-using namespace yield_plugin;
-
+namespace yield_plugin
+{
```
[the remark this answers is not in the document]
MishkaMN: FRIEND_TEST was not being detected with older approach, so I wrapped with namespace. However, the unit test no longer needs friend_test anymore

yield_plugin/test/test_yield.cpp
```diff
@@ -189,9 +186,7 @@ TEST(YieldPluginTest, test_update_traj)
 
   YieldPlugin plugin(nh,wm, config,[&](auto msg) {}, [&](auto msg) {});
```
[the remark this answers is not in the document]
MishkaMN: done

yield_plugin/test/test_yield.cpp
```diff
@@ -288,26 +283,241 @@ TEST(YieldPluginTest, test_update_traj)
...
 
+  rwol.objects = {rwo_1};
 
-  std::vector<carma_perception_msgs::msg::RoadwayObstacle> rw_objs;
+  std::vector<carma_perception_msgs::msg::ExternalObject> rw_objs;
 
   rw_objs.push_back(rwo_1);
 
-  wm->setRoadwayObjects(rw_objs);
+  plugin.external_objects_ = rw_objs;
```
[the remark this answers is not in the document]
MishkaMN: it was only required in 1 function, so I moved it as an input, so ended up not requiring internal member

yield_plugin/test/test_yield.cpp
```diff
@@ -288,26 +283,241 @@ TEST(YieldPluginTest, test_update_traj)
...
 
   carma_planning_msgs::msg::TrajectoryPlan tp_new = plugin.update_traj_for_object(tp, 10.0);
 
-  for (size_t i = 1; i < tp_new.trajectory_points.size(); i++) {
-   std::cout << tp_new.trajectory_points[i].x<< tp_new.trajectory_points[i].y << std::endl;
+  for (auto pt:  tp_new.trajectory_points) {
+    std::cout << "new x: " << pt.x << ", y:" << pt.y << ", t:" << std::to_string(rclcpp::Time(pt.target_time).seconds()) << std::endl;
+
```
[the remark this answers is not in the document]
MishkaMN: done

yield_plugin/test/test_yield.cpp
```diff
@@ -288,26 +283,241 @@ TEST(YieldPluginTest, test_update_traj)
...
   }
 
   EXPECT_EQ(7, tp.trajectory_points.size());
+  for (auto pt : tp.trajectory_points)
+  {
+    std::cout << "x: " << pt.x << ", y:" << pt.y << ", t:" << std::to_string(rclcpp::Time(pt.target_time).seconds()) << std::endl;
```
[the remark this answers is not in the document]
MishkaMN: done

yield_plugin/test/test_yield.cpp
```diff
@@ -288,26 +283,241 @@ TEST(YieldPluginTest, test_update_traj)
...
+  config.collision_check_radius = 90;
+  
+  // std::shared_ptr<carma_wm::CARMAWorldModel> wm = std::make_shared<carma_wm::CARMAWorldModel>();
+  auto nh = std::make_shared<yield_plugin::YieldPluginNode>(rclcpp::NodeOptions());
+
+  YieldPlugin plugin(nh,wm, config,[&](auto msg) {}, [&](auto msg) {});
+
+  
+  // save route Ids for faster access
+  for (auto llt: plugin.wm_->getRoute()->shortestPath())
```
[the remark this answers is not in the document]
MishkaMN: done

yield_plugin/include/yield_plugin/yield_plugin_node.hpp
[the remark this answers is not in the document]
MishkaMN: Yeah I think I agree. Changing back the aliases to its original syntax is not going to reduce the exception handling aspects of the carma, so I changed the ones in this package for the starters.

yield_plugin/src/yield_plugin.cpp
MishkaMN: This was because there was a bug in the way log levels set in carma-config. See `https://github.com/usdot-fhwa-stol/carma-utils/pull/151`

MishkaMN: Since it is working now, I changed it to use node's logger

yield_plugin/src/yield_plugin.cpp
MishkaMN: Yeah we need the indexing here because of the `on_route_idx` variable. This trajectory is being looped again in the next nested loop, so small performance improvement to start from the index that is already on the route.

adamlm (changes requested):

adamlm (approved):

PR #2189, opened 2023-11-14
"carma" shell script update to support dual compute system

willjohnsonk:
# PR Details
## Description

To support dual compute system design the carma shell script should be updated so that it can manage Docker services and containers across two PCs. All management of the system should be done from a single manager PC called "PC1", which will run commands similar to "carma start" and other carma commands.

## Related Jira Key

- [CF-459](https://usdot-carma.atlassian.net/browse/CF-459)
- [CF-663](https://usdot-carma.atlassian.net/browse/CF-663) 
- [CF-664](https://usdot-carma.atlassian.net/browse/CF-664)

## Motivation and Context

Improvements will allow for higher computation throughput on systems by leveraging two PCs.

## How Has This Been Tested?

Carma bash script has run successfully on dual compute hardware.

## Types of changes

## Checklist:

[CF-459]: https://usdot-carma.atlassian.net/browse/CF-459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CF-663]: https://usdot-carma.atlassian.net/browse/CF-663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CF-664]: https://usdot-carma.atlassian.net/browse/CF-664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


adamlm (changes requested):
There are several ShellCheck warnings/errors for the changed code. Please resolve those.

engineering_tools/carma
```diff
@@ -339,6 +339,174 @@ carma__stop() {
...
 }
 
+carma__swarm() {
+
+    if [ "$1" = "deploy" ]; then
+
+        echo "Checking Docker Swarm status..."
+        if [ "$(docker info | grep Swarm | sed 's/Swarm: //g')" == " inactive" ]; then
+            echo "Initalizing Docker Swarm..."
+            docker swarm init --advertise-addr 192.168.88.100 # consider making a variable
```
JonSmet, line 349: I think it could be worth having commands for setting the the following:

- Manager's IP address 
- Worker's IP address
- Worker's user name

engineering_tools/carma
```diff
@@ -143,7 +143,7 @@ carma__attach() {
     local CARMA_DOCKER_FILE="`docker run --rm --volumes-from carma-config:ro --entrypoint sh busybox:latest -c 'cat /opt/carma/vehicle/config/docker-compose.yml'`"
 
     echo "Attaching to CARMA container STDOUT..."
-    echo "$CARMA_DOCKER_FILE" | docker-compose -p carma -f - logs --follow --tail=10
+    echo "$CARMA_DOCKER_FILE" | docker-compose logs -f -t --tail=10
```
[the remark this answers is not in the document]
willjohnsonk: This was a testing change I missed rolling back because logs weren't behaving the same. good catch

engineering_tools/carma
willjohnsonk: Agreed, already had considered it. Will add.

Important note- the dual-compute PCs have the exact same hostname of `dev@[some model-specific Spectra name]`. This is bad and should be changed but wanted to check for a go-ahead. This may be a pain but I have not looked into it yet.

engineering_tools/carma
```diff
@@ -339,6 +339,174 @@ carma__stop() {
...
+            docker stack deploy --compose-file - carma-pc2
+            
+        
+        elif [ ! -z "$2" ]; then
+            echo "Unrecognized argument \"swarm deploy $2\""
+        fi
+
+
+    # Updates the carma-config-data volume based on a input tag string given by the user
+    elif [ "$1" = "update-config" ]; then
```
JonSmet, line 400: For a single PC non-swarm setup, a user would run the `carma config set image:tag` command. For consistency, could we change this to `config-set`, making the full command `carma swarm config-set image:tag`?

engineering_tools/carma
```diff
@@ -339,6 +339,174 @@ carma__stop() {
...
+
+        if [ "$(docker images | grep $USERNAME/carma-config | grep $TAG )" ]; then
+            echo "Found image, applying to carma-config-volume"
+            docker run --rm --name carma-config-data -v carma-config-data:/opt/carma/vehicle/config -d $USERNAME/carma-config:$TAG
+        else
+            echo "Could not find $USERNAME/carma-config:$TAG in local images"
+        fi
+
+
+    elif [ "$1" = "edit-config" ]; then
```
JonSmet, line 430: Similar to my previous comment on `carma swarm update-config`, could we change this to `config-edit` to remain consistent with the single PC non-swarm command (`carma config edit`)?

JonSmet, line 430: I remember at one point last week discussing that the user might need to navigate to the `/var/lib/docker/volumes` on the manager PC to actually edit files in the volume. Is that still required, or can the user use this command to manually edit certain aspects in the carma-config-data volume (for example, change a launch argument in carma_docker.launch.py)?

engineering_tools/carma
kjrush: Yeah, commands or an environment variable would be most preferable here. That way we can more easily deploy a swarm setup in places that don't match our vehicle networks (AWS).

engineering_tools/carma
```diff
@@ -339,6 +339,174 @@ carma__stop() {
...
+        echo "Checking Docker Swarm status..."
+        if [ "$(docker info | grep Swarm | sed 's/Swarm: //g')" == " inactive" ]; then
+            echo "Initalizing Docker Swarm..."
+            docker swarm init --advertise-addr 192.168.88.100 # consider making a variable
+        else
+            echo "Docker Swarm active"
+        fi
+
+        # if carma-config-data volume doesn't exist exit and request update
+        if [ "$(docker volume ls | grep carma-config-data | sed 's/local     //g')" ]; then
```
kjrush, line 355: Does this work? It's checking to see if the internal command there returns an empty string? I've seen `[[ -z ...` used for that, but I'm not super familiar with `[`

engineering_tools/carma
```diff
@@ -339,6 +339,174 @@ carma__stop() {
...
+        echo "Pushing $SOURCE_IMAGE_NAME to localhost:5000"
+
+        docker tag "$SOURCE_IMAGE_NAME" "localhost:5000/$SOURCE_IMAGE_NAME"
+        docker push "localhost:5000/$SOURCE_IMAGE_NAME"
+        docker rmi "localhost:5000/$SOURCE_IMAGE_NAME"
+
+        done
+
+
+    elif [ "$1" = "down" ]; then
```
JonSmet, line 492: To maintain consistency with the single PC non-swarm setup, can we change this to `stop`?

engineering_tools/carma
kjrush, line 400: Not the end of the world, but I think it might be worth considering 'set' for this verb, as that's the same paradigm as the non-swarm usage. Could be less confusing for users, `carma config set` and `carma swarm config set` or something like that?

Breaking the pattern is fine if we're intentional about it though, if this is doing something fundamentally different from `set` in your mind.

engineering_tools/carma
```diff
@@ -339,6 +339,174 @@ carma__stop() {
...
+        # performance will likely be poor due to combined log streams
+        for TASK in $SWARM_SERVICES; do
+            docker service logs --follow --details --tail=10 "$TASK" &
+        done
+
+        sleep infinity
+
+
+    # Takes n image tags that are saved locally and pushes them to the registry
+    elif [ "$1" = "set-image" ]; then
```
JonSmet, line 465: If `carma swarm set-image image:tag` pushes the specified image to the local registry, would the user be expected to then ssh into the worker PC and pull down the image that have been pushed from the manager PC? Just trying to check my understanding.

engineering_tools/carma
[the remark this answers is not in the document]
kjrush: Looking through this, that's probably better IMO. The large if-elif-else block here for command line parsing is a bit harder to read than function definitions and a parser loop.

engineering_tools/carma
```diff
@@ -339,6 +339,174 @@ carma__stop() {
...
+        shift
+        echo "Installing Docker images..."
+
+        if [ "$(docker volume ls | grep carma-registry | sed 's/local     //g')" ]; then
+            echo "Found carma-registry volume"
+        else
+            echo "Missing carma-registry volume, adding now..."
+            docker volume create carma-registry
```
kjrush, line 473: Does this script provide a way to clean up this volume? I could see this getting messy over time, if multiple versions were deployed without clearing this registry.

engineering_tools/carma
```diff
@@ -339,6 +339,174 @@ carma__stop() {
...
+        # this seems to loop forever and cannot be escaped with ctrl+C
+        # performance will likely be poor due to combined log streams
+        for TASK in $SWARM_SERVICES; do
+            docker service logs --follow --details --tail=10 "$TASK" &
+        done
+
+        sleep infinity
+
+
+    # Takes n image tags that are saved locally and pushes them to the registry
```
JonSmet, line 464: Small typo, and maybe update to `Takes in one or more image tags that ....`

engineering_tools/carma
```diff
@@ -339,6 +339,174 @@ carma__stop() {
...
+        fi
+
+        docker service create --constraint 'node.role==manager' -d --volume "carma-registry:/var/lib/registry" -p 5000:5000 --name carma-registry-service registry:2
+        
+        IMAGE_NAMES="$*"
+
+        for SOURCE_IMAGE_NAME in $IMAGE_NAMES
+        do
+        
+        echo "Pushing $SOURCE_IMAGE_NAME to localhost:5000"
```
JonSmet, line 483: Can we add a check to verify that `SOURCE_IMAGE_NAME` exists on the manager PC before pushing to the local registry?

engineering_tools/carma
kjrush: Also, this only uploads images into the host's (I'm assuming that's the primary host) registry? Where/how does the other PC get access to that? Is it automatically configured to point to the swarm manager's registry somehow?

Otherwise, it might make more sense to treat this as an atomic operation? Push images to local registry, ssh to client node, pull images, exit ssh, cleanup registry.

kjrush (changes requested):

engineering_tools/carma
```diff
@@ -339,6 +339,174 @@ carma__stop() {
...
+
+
+    elif [ "$1" = "down" ]; then
+        echo "Shutting down swarm..."
+
+        if [ "$(docker node ls | grep worker )" ]; then
+            ssh dev@192.168.88.101 "docker swarm leave"
+        fi
+        docker swarm leave --force
+
```
JonSmet, line 499: Another command I think we'd like to support is `carma swarm config install image:tag`, which will be similar to `carma-config__install()`, and will enable a user to provide the image:tag of a carma-config image for a dual compute system. The logic would pull the carma-config image, and then all images included in its docker-compose-background-pc1.yml, docker-compose-pc1.yml, and docker-compose-pc2.yml files.

engineering_tools/carma
```diff
@@ -339,6 +339,174 @@ carma__stop() {
...
+
```
JonSmet, line 500: Another command I think we'd like to support is `carma swarm config-status` which will be similar to `carma-config__status()`, and will printout the contents of each docker-compose file currently set. This will require that carma-config-data has been set by the user with `carma swarm update-config`.

JonSmet (changes requested):

engineering_tools/carma
willjohnsonk: This fetches the actual volume name "carma-config-data" because at one point I was checking name strings. Now that I realize I forgot to revert it back to just a grep check I am also shocked it works. Will fix for clarity

engineering_tools/carma
```diff
@@ -339,6 +339,174 @@ carma__stop() {
...
+        if [ "$(docker volume ls | grep carma-config-data | sed 's/local     //g')" ]; then
+            echo "Found carma-config-data volume"
+        else
+            echo "Missing carma-config-data volume, add with \"carma swarm update-config <image tag>\""
+            exit 1
+        fi
+
+        # carma-web-ui will always run as a background process regardless of deploy argument
+        docker run --rm -v carma-config-data:/opt/carma/vehicle/config --entrypoint sh busybox:latest -c \
+        'cat /opt/carma/vehicle/config/docker-compose-background-pc1.yml' | \
```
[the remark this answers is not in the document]
willjohnsonk: Everything is/will be documented in the [dual compute design doc](https://leidoscorpus.sharepoint.us/:w:/r/sites/STR/_layouts/15/doc2.aspx?sourcedoc=%7Bcc4c7ec8-bf15-4658-9441-b56f6205e69e%7D&action=edit&wdPid=4209b88)  which will eventually be transferred to Confluence plus into the related git readme files. It matches the previous volumes-from directory, the path just has to be more explicitly defined when using a named volume.

engineering_tools/carma
willjohnsonk: The registry will run as a swarm service so any connected host can access it. Kinda like a named volume. Honestly a registry is almost entirely overkill for this application since it will only be for installing the 1 or 2 images needed on PC2. I mostly included it for scalability and proof of concept, realistically you could just scp/download the images using PC1 if you didn't have them for some reason.

I guess this opens the question of if we want this expanded upon (I more or less made a minimal registry to make images available across Swarm) or if this isn't a high priority. The registry requires image names to be altered and won't accept uppercase characters so automating it initially seemed like it'd take some looking into.

engineering_tools/carma
willjohnsonk: This seems to work for the named volume as `carma config edit` works for images in the carma script currently, yes. Unfortunately you need sudo privileges to edit because of some /var/lib/ permissions that I didn't want to mess with. It is essentially is the exact same behavior as /var/lib/ edits, this just matches Docker intent better

engineering_tools/carma
willjohnsonk: Yes- I mentioned this in a previous registry comment to Kyle above and I @'d you for your thoughts

engineering_tools/carma
willjohnsonk: I actually meant _n_ images as in "any number of images, n" since it takes each successive argument as a new image lol

I injected algebra into my comment and for this I am sorry, will clarify

engineering_tools/carma_script_extensions/swarm
```diff
@@ -0,0 +1,352 @@
...
+
+    docker tag "$SOURCE_IMAGE_NAME" "localhost:5000/$SOURCE_IMAGE_NAME"
+    docker push "localhost:5000/$SOURCE_IMAGE_NAME"
+    docker rmi "localhost:5000/$SOURCE_IMAGE_NAME"
+
+    done
+}
+
+
+swarm__stop() {
```
adamlm, line 258: Is there a difference between stopping a swarm and tearing it down? If so, should we have commands to let users do that?

engineering_tools/carma_script_extensions/swarm
```diff
@@ -0,0 +1,352 @@
...
+}
+
+
+swarm__register() {
+    echo "Installing Docker images..."
+
+    if [ "$(docker volume ls | grep carma-registry | sed 's/local     //g')" ]; then
+        echo "Found carma-registry volume"
+    else
+        echo "Missing carma-registry volume, adding now..."
```
adamlm, line 220: `carma-registry` should be quoted since they are the names. It's a small visual aid that could help users.
```shell
echo "Found 'carma-registry' volume"
```

engineering_tools/carma_script_extensions/swarm
```diff
@@ -0,0 +1,352 @@
...
+
+    if [ "$1" = "-d" ]; then
+        local USERNAME=usdotfhwastoldev
+        local TAG=$2
+    elif [ "$1" = "-c" ]; then
+        local USERNAME=usdotfhwastolcandidate
+        local TAG=$2
+    else
+        local USERNAME=usdotfhwastol
+        local TAG=$1
```
adamlm, line 129: Should these be declared read only?
```shell
local readonly USERNAME
local readonly TAG
```

engineering_tools/carma_script_extensions/swarm
```diff
@@ -0,0 +1,352 @@
...
+    fi
+
+    if [[ -z $TAG ]]; then
+        echo "Please specify a tag string to update carma-config-data volume."
```
adamlm, line 31: This should print out to stderr, not stdout. This comment applies to several locations.

engineering_tools/carma_script_extensions/swarm
```diff
@@ -0,0 +1,352 @@
...
+        exit 1
+    fi
+
+    if [ ! -z "$(echo $TAG | grep :)" ]; then
+        local IMAGE_NAME="$TAG"
+    else
+        local IMAGE_NAME="$USERNAME/carma-config:$TAG"
```
adamlm, line 38: Should this be declared read only?
```shell
local readonly IMAGE_NAME
```

engineering_tools/carma_script_extensions/swarm
```diff
@@ -0,0 +1,352 @@
...
+    fi
+
+    echo "Downloading $IMAGE_NAME..."
+    docker pull $IMAGE_NAME
+
+    docker run --rm --name carma-config-data -v carma-config-data:/opt/carma/vehicle/config -d usdotfhwastol/carma-config:carma-system-4.4.3-47-g40f710d-SNAPSHOT-development
```
adamlm, line 44: Is this container going to be generalized? Why is the specific snapshot hardcoded?

engineering_tools/carma_script_extensions/swarm
```diff
@@ -0,0 +1,352 @@
...
+    echo "$WORKER_COMPOSE" | docker-compose -f - pull
+
+}
+
+
+swarm__start() {
+    echo "Checking Docker Swarm status..."
+    if [ "$(docker info | grep Swarm | sed 's/Swarm: //g')" == " inactive" ]; then
+        echo "Initalizing Docker Swarm..."
+        docker swarm init --advertise-addr 192.168.88.100 # consider making a variable
```
adamlm, line 65: This should be a variable.

engineering_tools/carma_script_extensions/swarm
```diff
@@ -0,0 +1,352 @@
...
+        docker stack deploy --compose-file - carma-manager
+
+
+    # For use with two connected PCs with known IPs
+    # Saves the generated swarm join token and ssh's into the worker to execute and deploy
+    elif [ "$1" = "dual" ]; then
+        echo "Deploying CARMA to dual compute Swarm stack..."
+
+        local TKN=$(docker swarm join-token -q worker)
+        ssh dev@192.168.88.101 "docker swarm join --token $TKN 192.168.88.100:2377"
```
adamlm, line 101: These should be variables.

adamlm (changes requested):

engineering_tools/carma_script_extensions/swarm
willjohnsonk: We discussed that stderr is probably best defined as a function in the main carma script to use in extensions instead of defining it in the extension itself. I'll compile a list of edits/enhancements that the primary carma script needs such as this.

engineering_tools/carma_script_extensions/swarm
willjohnsonk: For our purposes I don't think so. Technically Swarm can remain initialized but without a stack deployed to it there's no functionality/need for it to be running in the background.

engineering_tools/carma
willjohnsonk: This is semi-resolved. Will follow the registry functionality used in pushmi-pullyu with a separate install function for downloading images.

engineering_tools/carma_script_extensions/swarm
```diff
@@ -60,6 +59,14 @@ swarm__install() {
 
 swarm__start() {
     echo "Checking Docker Swarm status..."
+    # if carma-config-data volume doesn't exist exit and request update
+    if [ ! -z "$(docker volume ls | grep carma-config-data)" ]; then
+        echo "Found 'carma-config-data' volume"
+    else
+        echo "Missing 'carma-config-data' volume, add with \"carma swarm update-config <image tag>\""
```
adamlm, line 66: Should this output to the stderr instead of stdout?

engineering_tools/carma_script_extensions/swarm
```diff
@@ -243,11 +245,19 @@ swarm__register() {
         exit 1
     fi
 
-    local REGISTRY_PORT=5000
-    local REGISTRY_HOST="localhost"
+    local carma_base=$(__get_image_from_config carma-base:)
```
adamlm, line 248: Should this be marked read-only?

engineering_tools/carma_script_extensions/swarm
```diff
@@ -261,23 +271,22 @@ swarm__register() {
...
+          && docker rmi "$REGISTRY_HOST:$REGISTRY_PORT/$TARGET_IMAGE_NAME"
     done
-
-    # May need to find the service and remove it after use since Swarm will likely keep it alive
 }
 
 
 swarm__stop() {
     echo "Shutting down swarm..."
+    local carma_base=$(__get_image_from_config carma-base:)
```
adamlm, line 284: Should this be marked read-only? Also, the naming convention deviates from the rest of the script. Other variables use `SCREAMING_SNAKE_CASE`.

adamlm (approved):
General comment: ShellCheck raises several issues. Please resolve those.

engineering_tools/carma
```diff
@@ -143,7 +143,7 @@ carma__attach() {
     local CARMA_DOCKER_FILE="`docker run --rm --volumes-from carma-config:ro --entrypoint sh busybox:latest -c 'cat /opt/carma/vehicle/config/docker-compose.yml'`"
 
     echo "Attaching to CARMA container STDOUT..."
-    echo "$CARMA_DOCKER_FILE" | docker-compose -p carma -f - logs --follow --tail=10
+    echo "$CARMA_DOCKER_FILE" | docker-compose -p carma logs -f - logs --follow --tail=10
```
JonSmet, line 146: Is this change required?

JonSmet (changes requested):

engineering_tools/carma_script_extensions/swarm
willjohnsonk: Will first need to add an error function in the top level carma script. Not in scope but will track

engineering_tools/carma_script_extensions/swarm
```diff
@@ -60,31 +60,31 @@ swarm__install() {
...
     fi
 
-    MANAGER_INFO=$(docker run -it --rm -v carma-config-data:/opt/carma/vehicle/config $carma_base /bin/sh -c '{ echo "$MANAGER_USER"; echo "@"; echo "$MANAGER_IP"; } | tr -d "\n"')
-    WORKER_INFO=$(docker run -it --rm -v carma-config-data:/opt/carma/vehicle/config $carma_base /bin/sh -c '{ echo "$WORKER_USER"; echo "@"; echo "$WORKER_IP"; } | tr -d "\n"')
+    MANAGER_INFO=$(docker run -it --rm -v carma-config-data:/opt/carma/vehicle/config "$CARMA_BASE" /bin/sh -c '{ echo "$MANAGER_USER"; echo "@"; echo "$MANAGER_IP"; } | tr -d "\n"')
+    WORKER_INFO=$(docker run -it --rm -v carma-config-data:/opt/carma/vehicle/config "$CARMA_BASE" /bin/sh -c '{ echo "$WORKER_USER"; echo "@"; echo "$WORKER_IP"; } | tr -d "\n"')
 
     MANAGER_IP=${MANAGER_INFO#*@}
-    MANAGER_USER=${MANAGER_INFO%%@*}
+    # MANAGER_USER=${MANAGER_INFO%%@*}
```
JonSmet, line 80: Can we remove this commented-out line?

engineering_tools/carma_script_extensions/swarm
```diff
@@ -60,31 +60,31 @@ swarm__install() {
...
 
-    WORKER_IP=${WORKER_INFO#*@}
-    WORKER_USER=${WORKER_INFO%%@*}
+    # WORKER_IP=${WORKER_INFO#*@}
+    # WORKER_USER=${WORKER_INFO%%@*}
```
JonSmet, line 83: Can we remove these commented-out lines?

JonSmet (changes requested):

adamlm (changes requested):
Resolve yellow (warning) ShellCheck issues. Prefferrably green ones too but not required.

carma/launch/carma_src.launch.py
```diff
@@ -144,8 +160,10 @@ def generate_launch_description():
     simulation_mode = LaunchConfiguration('simulation_mode')
     declare_simulation_mode = DeclareLaunchArgument(name='simulation_mode', default_value = 'False', description = 'True if CARMA Platform is launched with CARLA Simulator')
 
+
     # Launch ROS2 rosbag logging
     ros2_rosbag_launch = GroupAction(
+        condition=IfCondition(PythonExpression(["'", host_placement, "' == 'manager'"])),
```
adev4a, line 166: If all the nodes here need to be launched under this condition, can we not have this launch file itself called under the condition instead of defining it for each one?

system_controller/config/config.yaml
```diff
@@ -6,7 +6,7 @@
 
     # Long: Timeout for each service to be detected as available in milliseconds
     # Units: milliseconds
-    service_timeout_ms : 200
+    service_timeout_ms : 18000
```
adev4a, line 9: Was this change meant to be committed to develop? 18 seconds seems too large for the timeout

carma/launch/carma_src.launch.py
```diff
@@ -116,6 +116,22 @@ def generate_launch_description():
         default_value= 'False',
         description='Flag to enable opening http tunnesl to CARMA Cloud'
     )
+
+    # Declare system_architecture
+    system_architecture = LaunchConfiguration('system_architecture')
+    declare_system_architecture = DeclareLaunchArgument(
+        name = 'system_architecture',
+        default_value = 'single',
+        description = 'Flag to define whether a single compute system or a dual compute system is being used'
```
adev4a, line 125: Can you also specify the accepted values for this config in the description?

carma/launch/carma_src.launch.py
```diff
@@ -116,6 +116,22 @@ def generate_launch_description():
...
+    )
+
+    # Declare host_placement
+    host_placement = LaunchConfiguration('host_placement')
+    declare_host_placement = DeclareLaunchArgument(
+        name = 'host_placement',
+        default_value = 'manager',
+        description = 'Flag to define whether the current active host is a manager or worker for ROS node allocation'
```
adev4a, line 133: Same as above, can you define the accepted values for the config in the description?

carma/launch/carma_src.launch.py
willjohnsonk: They aren't- guidance control is defined as launching on the worker. In the final implementation more nodes may need to be specified for drivers as well.

system_controller/config/config.yaml
willjohnsonk: 18000 is a very arbitrary value assigned during testing. This relates to specific Swarm behavior of how it deploys services differently (and a bit slower) than our normal Docker deployment. I can do some testing to see what this value actually needs to be, it should be lower but unsure by how much. I can show you a demo sometime as an example.

carma/launch/carma_src.launch.py
willjohnsonk: Technically the accepted values are shown in both of these but I can see how it's not clear. It'd be better to be bolded but can't have that, do you think putting it as 'single' 'dual' / 'manager' 'worker' with single quotes in the text be clear enough?

adev4a: Yes, i think having them in single quotes would make it clear that those are the values we can use here

MishkaMN:
@willjohnsonk Should we close this PR if it is not getting merged?
We can keep the branch alive or move this to https://github.com/usdot-fhwa-stol/carma-developer-tools/

JonSmet:
Closing this PR due to inactivity, and since this effort was paused in December 2023.

PR #2216, opened 2023-12-15
Feature/time sync

MishkaMN:
# PR Details
## Description

This PR supports https://github.com/usdot-fhwa-stol/carma-ns3-adapter/pull/11
During integration testing of VRU, it was discovered that the CARLA's time which we call "game time" is not synced with the "simulation time" which is used and synced between sumo and mosaic. 
CARLA's game time goes up until the sumo is hit start, which is when the all timelins are synced correctly, but the game time still has the initial offset from when CARLA started and up until sumo was hit started.

Upon trying to fix this issue, the ns3-adapter (which should be renamed to cdasim-adapter due to its new functionality) was identified as the potential source to broadcast the /clock for ROS network to provide the correct simulation time. However, doing so makes the carma-platform not behave the way it should and debugging could be non-trivial:
[TODO: github link](https://github.com/usdot-fhwa-stol/carma-platform/issues/2217)

Therefore, the only place the carma-platform has a time difference with the cdasim is when it receives SPAT message. Therefore, this specific logic is modified to adjust for that time difference.

## Related Jira Key
[CDAR-602](https://usdot-carma.atlassian.net/browse/CDAR-602)

## Motivation and Context
Integration testing of VRU

## How Has This Been Tested?
cdasim integration tested on the simulation PC1 

## Types of changes

## Checklist:

[CDAR-602]: https://usdot-carma.atlassian.net/browse/CDAR-602?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


carma_wm/src/WMListener.cpp
```diff
@@ -85,25 +87,37 @@ WMListener::WMListener(
...
-  route_sub_ = rclcpp::create_subscription<carma_planning_msgs::msg::Route>(node_topics_, "route", 1, 
+  route_sub_ = rclcpp::create_subscription<carma_planning_msgs::msg::Route>(node_topics_, "route", 1,
                                   std::bind(&WMListenerWorker::routeCallback, worker_.get(), std::placeholders::_1), route_options);
 
+  ros1_clock_sub_ = rclcpp::create_subscription<rosgraph_msgs::msg::Clock>(node_topics_, "/clock", 1,
+                                  std::bind(&WMListenerWorker::ros1ClockCallback, worker_.get(), std::placeholders::_1), ros1_clock_options);
+
+  sim_clock_sub_ = rclcpp::create_subscription<rosgraph_msgs::msg::Clock>(node_topics_, "/sim_clock", 1,
+                                  std::bind(&WMListenerWorker::simClockCallback, worker_.get(), std::placeholders::_1), sim_clock_options);
+
```
adamlm, line 115: These should be lambdas instead of `std::bind`. Lambdas can completely replace `std::bind` and is the recommended implementation pattern. Consider:
```cpp
ros1_clock_sub_ = rclcpp::create_subscription<rosgraph_msgs::msg::Clock>(node_topics_, "/clock", 1,
[this](const rosgraph_msgs::msg::Clock::SharedPtr msg) {
  this->worker_.simClockCallback(msg);
},
ros1_clock_options);
```

carma_wm/include/carma_wm/CARMAWorldModel.hpp
```diff
@@ -159,7 +159,15 @@ class CARMAWorldModel : public WorldModel
...
   void setRouteName(const std::string& route_name);
- 
+
+  /*! \brief Set current Ros1 clock (only used in simulation runs)
+   */
+  void setRos1Clock(rclcpp::Time time_now);
+
+  /*! \brief Set simulation clock clock (only used in simulation runs)
+   */
+  void setSimulationClock(rclcpp::Time time_now);
```
adamlm, line 169: These can be made const references: `const rclcpp::Time &`.

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -613,6 +613,16 @@ namespace carma_wm
     return semantic_map_;
   }
 
+  void CARMAWorldModel::setRos1Clock(rclcpp::Time time_now)
+  {
+    ros1_clock_ = time_now;
+  }
+
+  void CARMAWorldModel::setSimulationClock(rclcpp::Time time_now)
```
adamlm, line 621: These can be made const references: `const rclcpp::Time &`

carma_wm/src/CARMAWorldModel.cpp
```diff
@@ -1419,6 +1429,18 @@ namespace carma_wm
 
   boost::posix_time::ptime CARMAWorldModel::min_end_time_converter_minute_of_year(boost::posix_time::ptime min_end_time,bool moy_exists,uint32_t moy, bool is_simulation)
   {
+    double simulation_time_difference_in_seconds = 0.0;
+
+    // NOTE: In simulation, ROS1 clock (often coming from CARLA) can have a large time ahead.
+    // the timing calculated here is in Simulation time, which is behind. Therefore, the world model adds the offset to make it meaningful to carma-platform:
+    // https://github.com/usdot-fhwa-stol/carma-platform/issues/2217
+    if (is_simulation && ros1_clock_.has_value() && simulation_clock_.has_value())
```
adamlm, line 1437: `std::optional` is implicitly convertible to `bool`, so we don't need the `.has_value()` calls. Being explicit in this context also seems to go against the ISO C++ code guidelines on verbosity (see [here](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es87-dont-add-redundant--or--to-conditions)). Consider replacing with
```cpp
if (is_simulation && ros1_clock_ && simulation_clock_)
```

carma_wm/src/WMListener.cpp
```diff
@@ -85,25 +87,37 @@ WMListener::WMListener(
...
     auto traffic_spat_group = node_base_->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive);
-    traffic_spat_options.callback_group = traffic_spat_group;  
+    traffic_spat_options.callback_group = traffic_spat_group;
 
-  }  
+    auto ros1_clock_group = node_base_->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive);
+    ros1_clock_options.callback_group = ros1_clock_group;
+
+    auto sim_clock_group = node_base_->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive);
+    sim_clock_options.callback_group = sim_clock_group;
```
adamlm, line 102: Consider assigning the `create_callback_group` to avoid having to copy the `std::shared_ptr` and an extra variable:
```cpp
ros1_clock_options.callback_group = node_base_->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive);
```

carma_wm/src/WMListener.cpp
adamlm: This is an optional suggestion, but it would be good to remove `std::bind` usage throughout this file.

adamlm (changes requested):

PR #2231, opened 2023-12-20
Arbitrator should retry the failed service calls

MishkaMN:
# PR Details
## Description

This PR caches the strategic_plugin clients previously created instead of recreating it every 1hz as before.
This PR adds a guard for when none of the strategic plugins return any valid maneuvers (due to ROS failing the service calls or somehow the node froze or shut down), arbitrator should shut down with error. Currently there is no guard for empty responses, which I am guessing is the potential segfault place. Tested it by running platform and forcing shut down the only available strategic plugin and got the following error:

> [environment_perception_controller-8] 1703074934.381615035 | INFO | on_system_alert:59 | Received SystemAlert message of type: 6 with message: SHUTDOWN /guidance subsytem has failed with error: Uncaught Exception from node: arbitrator exception: None of the strategic plugins generated any valid plans! Please check if any is turned and returning valid maneuvers, shutting down arbitrator... while in ACTIVE state.
[v2x_controller-19] 1703074934.381624890 | INFO | on_system_alert:59 | Received SystemAlert message of type: 6 with message: SHUTDOWN /guidance subsytem has failed with error: Uncaught Exception from node: arbitrator exception: None of the strategic plugins generated any valid plans! Please check if any is turned and returning valid maneuvers, shutting down arbitrator... while in ACTIVE state.
[drivers_controller-2] 1703074934.381628844 | INFO | on_system_alert:59 | Received SystemAlert message of type: 6 with message: SHUTDOWN /guidance subsytem has failed with error: Uncaught Exception from node: arbitrator exception: None of the strategic plugins generated any valid plans! Please check if any is turned and returning valid maneuvers, shutting down arbitrator... while in ACTIVE state.
[guidance_controller-32] 1703074934.381633900 | INFO | on_system_alert:59 | Received SystemAlert message of type: 6 with message: SHUTDOWN /guidance subsytem has failed with error: Uncaught Exception from node: arbitrator exception: None of the strategic plugins generated any valid plans! Please check if any is turned and returning valid maneuvers, shutting down arbitrator... while in ACTIVE state.
[localization_controller-17] 1703074934.381671092 | INFO | on_system_alert:59 | Received SystemAlert message of type: 6 with message: SHUTDOWN /guidance subsytem has failed with error: Uncaught Exception from node: arbitrator exception: None of the strategic plugins generated any valid plans! Please check if any is turned and returning valid maneuvers, shutting down arbitrator... while in ACTIVE state.
[localization_controller-17] 1703074934.381698191 | INFO | on_system_alert:79 | Alert not from a sensor node. Alerting node: /system_controller
[INFO] [system_controller-35]: process has finished cleanly [pid 388]

- this PR also added functionality to retry some of the strategic_plugin service calls because sometimes they could fail due to ROS unknown DDS error. Since strategic plugins are only ran at 1Hz, it is relatively important that we don't miss this planning. So added retry attempt, and the most calls are successful only in its 2nd or 3rd try on Simulation PC1 with fastdds rmw. These failed service calls gave a clue that arbitrator maybe segfaulting if by chance none of the strategic plugins' service call succeeded (or had non-empty maneuver). 

## Related GitHub Issue
Potentially fixes this issue: https://github.com/usdot-fhwa-stol/carma-platform/issues/2245

## Related Jira Key
https://usdot-carma.atlassian.net/browse/CDAR-631
CDAR-631

## Motivation and Context
integration testing VRU use case in simulation

## How Has This Been Tested?
Simulation PC 1

## Types of changes

## Checklist:

### DIFF
--- a/arbitrator/include/capabilities_interface.tpp
+++ b/arbitrator/include/capabilities_interface.tpp
@@ -73,7 +73,7 @@ namespace arbitrator
                 try {
                     using std::literals::chrono_literals::operator""ms;
 
-                    if (registered_strategic_plugins_.find(topic) == registered_strategic_plugins_.end())
+                    if (registered_strategic_plugins_.count(topic) == 0)
                         registered_strategic_plugins_[topic] = nh_->create_client<carma_planning_msgs::srv::PlanManeuvers>(topic);
 
                     auto client = registered_strategic_plugins_[topic];
@@ -92,6 +92,8 @@ namespace arbitrator
                         case std::future_status::ready:
                             responses.emplace(topic, response.get());
                             break;
+                        case std::future_status::deferred:
+                        case std::future_status::timeout:
                         default:
                             RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"), "failed...: " << topic);
                     }
--- a/arbitrator/include/capabilities_interface.tpp
+++ b/arbitrator/include/capabilities_interface.tpp
@@ -73,8 +73,18 @@ namespace arbitrator
                 try {
                     using std::literals::chrono_literals::operator""ms;
 
-                    const auto client = nh_->create_client<carma_planning_msgs::srv::PlanManeuvers>(topic);
-                    RCLCPP_DEBUG_STREAM(rclcpp::get_logger("arbitrator"), "found client: " << topic);
+                    if (registered_strategic_plugins_.find(topic) == registered_strategic_plugins_.end())
+                        registered_strategic_plugins_[topic] = nh_->create_client<carma_planning_msgs::srv::PlanManeuvers>(topic);
+
+                    auto client = registered_strategic_plugins_[topic];
+
+                    if (client->wait_for_service(500ms))
+                        RCLCPP_DEBUG_STREAM(rclcpp::get_logger("arbitrator"), "found client: " << topic);
+                    else
+                    {
+                        topics_to_retry.push_back(topic);
+                        RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"), "Following client timed out: " << topic << ", retrying, attempt no: " << retry_attempt);
+                    }
 
                     const auto response = client->async_send_request(msg);
 
--- a/arbitrator/include/capabilities_interface.tpp
+++ b/arbitrator/include/capabilities_interface.tpp
@@ -94,9 +94,14 @@ namespace arbitrator
                             responses.emplace(topic, response.get());
                             break;
                         case std::future_status::deferred:
+                            RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"), "service call to " << topic << " is deferred... Please check if the plugin is active");
+                            break;
                         case std::future_status::timeout:
+                            RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"), "service call to " << topic << " is timed out... Please check if the plugin is active");
+                            break;
                         default:
-                            RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"), "failed...: " << topic);
+                            RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"), "service call to " << topic << " is failed... Please check if the plugin is active");
+                            break;
                     }
                     if (retry_attempt > 0)
                     {
--- a/arbitrator/include/capabilities_interface.tpp
+++ b/arbitrator/include/capabilities_interface.tpp
@@ -35,13 +35,12 @@ namespace arbitrator
     {
         std::map<std::string, std::shared_ptr<MSrvRes>> responses;
         bool topics_call_successful = false;
-        std::vector<std::string> topics_to_retry;
-        std::vector<std::string> current_topics_to_check;
         int retry_attempt = 0;
+        std::vector<std::string> detected_topics;
         while (!topics_call_successful && retry_attempt < MAX_RETRY_ATTEMPTS)
         {
             try {
-            topics_to_retry = get_topics_for_capability(query_string);
+            detected_topics = get_topics_for_capability(query_string);
             topics_call_successful = true;
             if (retry_attempt > 0)
             {
@@ -62,6 +61,8 @@ namespace arbitrator
             return responses;
         }
 
+        std::vector<std::string> topics_to_retry = detected_topics;
+        std::vector<std::string> current_topics_to_check;
         retry_attempt = 0;
         while (!topics_to_retry.empty() && retry_attempt < MAX_RETRY_ATTEMPTS)
         {
@@ -81,8 +82,7 @@ namespace arbitrator
                         case std::future_status::ready:
                             responses.emplace(topic, response.get());
                             break;
-                        case std::future_status::deferred:
-                        case std::future_status::timeout:
+                        default:
                             RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"), "failed...: " << topic);
                     }
                     if (retry_attempt > 0)
@@ -94,13 +94,12 @@ namespace arbitrator
                     RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"),
                         "Cannot make service request for service '" << topic << "': " << error.what() <<". So retrying, attempt no: " << retry_attempt);
                     topics_to_retry.push_back(topic);
-                    continue;
                 }
             }
             retry_attempt ++;
         }
 
-        if (retry_attempt == MAX_RETRY_ATTEMPTS)
+        if (retry_attempt >= MAX_RETRY_ATTEMPTS)
         {
             RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"),
                             "Failed to get a valid response from one or all of the strategic plugins...");
--- a/arbitrator/include/capabilities_interface.tpp
+++ b/arbitrator/include/capabilities_interface.tpp
@@ -26,35 +26,83 @@
 
 namespace arbitrator
 {
+    constexpr auto MAX_RETRY_ATTEMPTS {10};
+
     template<typename MSrvReq, typename MSrvRes>
     std::map<std::string, std::shared_ptr<MSrvRes>>
         CapabilitiesInterface::multiplex_service_call_for_capability(const std::string& query_string, std::shared_ptr<MSrvReq> msg)
     {
         std::map<std::string, std::shared_ptr<MSrvRes>> responses;
-
-        for (const auto & topic : get_topics_for_capability(query_string))
+        bool topics_call_successful = false;
+        std::vector<std::string> topics_to_retry;
+        std::vector<std::string> current_topics_to_check;
+        int retry_attempt = 0;
+        while (!topics_call_successful && retry_attempt < MAX_RETRY_ATTEMPTS)
         {
             try {
-                using std::literals::chrono_literals::operator""ms;
-
-                const auto client = nh_->create_client<carma_planning_msgs::srv::PlanManeuvers>(topic);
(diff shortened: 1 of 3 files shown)


arbitrator/include/capabilities_interface.tpp
```diff
@@ -23,38 +23,91 @@
...
         {
             try {
-                using std::literals::chrono_literals::operator""ms;
+            topics_to_retry = get_topics_for_capability(query_string);
+            topics_call_successful = true;
+            if (retry_attempt > 0)
+            {
+                RCLCPP_INFO_STREAM(rclcpp::get_logger("arbitrator")," Service call to get available strategic plugins successfully finished after retrying: " << retry_attempt);
+            }
+            } catch (const std::exception& error) {
```
adamlm, line 50: If there is a specific ROS error we expect to get thrown, we should catch that instead of `std::exception`. I'm assuming it's `rclcpp::exceptions::RCLError` that gets thrown.

arbitrator/include/capabilities_interface.tpp
```diff
@@ -23,38 +23,91 @@
...
+        int retry_attempt = 0;
+        while (!topics_call_successful && retry_attempt < MAX_RETRY_ATTEMPTS)
         {
             try {
-                using std::literals::chrono_literals::operator""ms;
+            topics_to_retry = get_topics_for_capability(query_string);
+            topics_call_successful = true;
+            if (retry_attempt > 0)
+            {
+                RCLCPP_INFO_STREAM(rclcpp::get_logger("arbitrator")," Service call to get available strategic plugins successfully finished after retrying: " << retry_attempt);
```
adamlm, line 48: This should probably be at `DEBUG` level.

arbitrator/include/capabilities_interface.tpp
```diff
@@ -23,38 +23,91 @@
...
+            }
+            } catch (const std::exception& error) {
+                RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"),
```
adamlm, line 51: This should probably be at `DEBUG` level.

arbitrator/include/capabilities_interface.tpp
```diff
@@ -23,38 +23,91 @@
...
+                            "Could not make a service call to get available strategic service topic names, with detected error: " << error.what() <<". So retrying, attempt no: " << retry_attempt);
+                retry_attempt ++;
+            }
+
+        }
+
+        if (!topics_call_successful)
+        {
+            RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"),
```
adamlm, line 60: This should probably be at `ERROR` level.

arbitrator/include/capabilities_interface.tpp
```diff
@@ -23,38 +23,91 @@
...
+                            "Failed to retrieve available strategic plugin service topics! Returning empty maneuver responses");
+            return responses;
+        }
+
+        retry_attempt = 0;
+        while (!topics_to_retry.empty() && retry_attempt < MAX_RETRY_ATTEMPTS)
+        {
+            current_topics_to_check = topics_to_retry;
+            topics_to_retry = {};
```
adamlm, line 69: This should be changed to `.clear()`. I assume clearing the `std::vector` will be faster than reassignment.

arbitrator/include/capabilities_interface.tpp
```diff
@@ -23,38 +23,91 @@
...
+                    const auto response = client->async_send_request(msg);
 
-                switch (const auto status{response.wait_for(500ms)}; status) {
-                    case std::future_status::ready:
-                        responses.emplace(topic, response.get());
-                        break;
-                    case std::future_status::deferred:
-                    case std::future_status::timeout:
-                        RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"), "failed...: " << topic);
+                    switch (const auto status{response.wait_for(500ms)}; status) {
```
adamlm, line 80: I think in this context we don't need the `; status` part. The declaration should be implicitly convertible to `bool` here.

arbitrator/include/capabilities_interface.tpp
```diff
@@ -23,38 +23,91 @@
...
+                        RCLCPP_INFO_STREAM(rclcpp::get_logger("arbitrator"),"Service call to: " << topic << ", successfully finished after retrying: " << retry_attempt);
+                    }
+
+                } catch(const rclcpp::exceptions::RCLError& error) {
+                    RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"),
+                        "Cannot make service request for service '" << topic << "': " << error.what() <<". So retrying, attempt no: " << retry_attempt);
+                    // Fast repeated calls may fail and on average in one of our environment was failing at 70ms repeated calls
+                    // https://github.com/ros2/rmw_fastrtps/issues/46
+                    std::chrono::milliseconds duration(100);
+                    std::this_thread::sleep_for(duration);
```
adamlm, line 99: These two lines can be combined:
```cpp
std::this_thread::sleep_for(std::chrono::milliseconds{100});
```
or 
```cpp
using std::chrono::literals::operator""ms;  // May not be needed if declared elsewhere
std::this_thread::sleep_for(100ms);
```

arbitrator/src/tree_planner.cpp
```diff
@@ -87,20 +87,26 @@ namespace arbitrator
...
             }
-            
+
             open_list_to_evaluate = temp_open_list;
         }
 
+        if (final_open_list.empty())
+        {
+            RCLCPP_ERROR_STREAM(rclcpp::get_logger("arbitrator"), "None of the strategic plugins generated any valid plans! Please check if any is turned and returning valid maneuvers, shutting down arbitrator...");
+            throw std::runtime_error("None of the strategic plugins generated any valid plans! Please check if any is turned and returning valid maneuvers, shutting down arbitrator...");
```
adamlm, line 107: This error message is slightly misleading. The `TreePlanner` is not shutting down the arbitrator node here. It only reports that there is an error.

adamlm (changes requested):

arbitrator/include/capabilities_interface.tpp
MishkaMN: I think it is better to be INFO because the earlier failure catch prints WARN saying "retrying", so the user knows the issue was resolved

arbitrator/include/capabilities_interface.tpp
MishkaMN: I think it is better to be INFO because the earlier failure catch prints WARN saying "retrying", so the user knows the issue was resolved

arbitrator/include/capabilities_interface.tpp
MishkaMN: Made it to ERROR, but did not throw because the caller is throwing in this PR.

arbitrator/include/capabilities_interface.tpp
MishkaMN: Removed this section because it did not fix or improve anything

arbitrator/src/tree_planner.cpp
MishkaMN: Removed the log

arbitrator/include/capabilities_interface.tpp
MishkaMN: catching the correct one now

arbitrator/include/capabilities_interface.tpp
```diff
@@ -23,38 +23,87 @@
...
-                    case std::future_status::ready:
-                        responses.emplace(topic, response.get());
-                        break;
-                    case std::future_status::deferred:
-                    case std::future_status::timeout:
-                        RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"), "failed...: " << topic);
+                    switch (const auto status{response.wait_for(500ms)}) {
+                        case std::future_status::ready:
+                            responses.emplace(topic, response.get());
+                            break;
```
paulbourelly999, line 83: don't we need to include break statements for each case here?

arbitrator/include/capabilities_interface.tpp
```diff
@@ -23,38 +23,87 @@
...
                 }
-            } catch(const rclcpp::exceptions::RCLError& error) {
-                RCLCPP_ERROR_STREAM(rclcpp::get_logger("arbitrator"),
-                    "Cannot make service request for service '" << topic << "': " << error.what());
-                continue;
             }
+            retry_attempt ++;
+        }
+
+        if (retry_attempt == MAX_RETRY_ATTEMPTS)
```
paulbourelly999, line 103: Just for safety we should check for retry_attempt >= MAX_RETRY_ATTEMPT incase we have a logically error some were

arbitrator/include/capabilities_interface.tpp
```diff
@@ -23,38 +23,87 @@
...
+
     template<typename MSrvReq, typename MSrvRes>
     std::map<std::string, std::shared_ptr<MSrvRes>>
         CapabilitiesInterface::multiplex_service_call_for_capability(const std::string& query_string, std::shared_ptr<MSrvReq> msg)
     {
         std::map<std::string, std::shared_ptr<MSrvRes>> responses;
-
-        for (const auto & topic : get_topics_for_capability(query_string))
+        bool topics_call_successful = false;
+        std::vector<std::string> topics_to_retry;
```
paulbourelly999, line 38: Make different variables for topics to call and topics to retry

paulbourelly999 (changes requested):

MishkaMN:
After reviews from Paul and Adam, I had a little epiphany to make the node cache previously created clients instead of creating it all the time every 1Hz. The client still waits to see if the service is available and is only called if the topic was received every 1Hz, so it is not different from before in terms of functionality. This completely removed all the failures I have been getting consistently at every run.

arbitrator/include/capabilities_interface.tpp
```diff
@@ -23,38 +23,96 @@
...
+        while (!topics_to_retry.empty() && retry_attempt < MAX_RETRY_ATTEMPTS)
+        {
+            current_topics_to_check = topics_to_retry;
+            topics_to_retry.clear();
+            for (const auto & topic : current_topics_to_check)
+            {
+                try {
+                    using std::literals::chrono_literals::operator""ms;
+
+                    if (registered_strategic_plugins_.find(topic) == registered_strategic_plugins_.end())
```
adamlm, line 76: I think using `count()` would be better here since we don't need an iterator in this context.
```cpp
if (registered_strategic_plugins_.count(topic) == 0)
```

arbitrator/include/capabilities_interface.tpp
```diff
@@ -23,38 +23,96 @@
...
-                        responses.emplace(topic, response.get());
-                        break;
-                    case std::future_status::deferred:
-                    case std::future_status::timeout:
-                        RCLCPP_WARN_STREAM(rclcpp::get_logger("arbitrator"), "failed...: " << topic);
+                    switch (const auto status{response.wait_for(500ms)}) {
+                        case std::future_status::ready:
+                            responses.emplace(topic, response.get());
+                            break;
+                        default:
```
adamlm, line 95: I think having the explicit cases for `deferred` and `timeout` make reading this a little clearer.

adamlm (changes requested):

arbitrator/include/capabilities_interface.tpp
MishkaMN: added both back

arbitrator/include/capabilities_interface.tpp
MishkaMN: changed

paulbourelly999 (approved):
Looks good.

paulbourelly999 (approved):

PR #2269, opened 2024-01-10
[cooperative perception] Drop unassociated duplicate detections

### LINKED ISSUE #2266: [cooperative perception] Multiple object tracker should drop unassociated duplicate detections
### Summary

The `MultipleObjectTrackerComponent` converts all unassociated detections to tracks, but this will lead to duplicated tracks since we multiple detections about the same object. We should add a filtering in the pipeline before creating tracks. The filter would remove all unassociated detections that are "close" enough to an existing track.

> [!NOTE]
> Our current association/update method supports only one-to-one associations between tracks and detections, so these unassociated detections would have been associated if they were useful.

### Reasoning for new functionality

_No response_

adamlm:
# PR Details
## Description

This PR adds a step to our tracking pipeline to drop any unassociated detections that are "close" enough to an existing track. Multiple actors publish detection lists that overlap with others. Our association algorithm currently supports only one-to-one associations, so some detections will go unassociated even though they represent an existing track. To reduce the number of duplicate tracks, we drop any unassociated detection that we deem close enough.

## Related GitHub Issue

Closes #2266 

## Related Jira Key

Closes [CDAR-711](https://usdot-carma.atlassian.net/browse/CDAR-711)

## Motivation and Context

Performance improvement

## How Has This Been Tested?

## Types of changes

## Checklist:

[CDAR-711]: https://usdot-carma.atlassian.net/browse/CDAR-711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/carma_cooperative_perception/src/multiple_object_tracker_component.cpp
+++ b/carma_cooperative_perception/src/multiple_object_tracker_component.cpp
@@ -630,7 +630,9 @@ auto MultipleObjectTrackerNode::execute_pipeline() -> void
     }
   }
 
-  // This is the erase-remove idiom (https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom)
+  // We want to remove unassociated tracks that are close enough to existing tracks
+  // to avoid creating duplicates. Duplicate tracks will cause association inconsistencies
+  // (flip flopping associations between the two tracks).
   auto remove_start{std::remove_if(
     std::begin(unassociated_detections), std::end(unassociated_detections),
     [&scores](const auto & detection) {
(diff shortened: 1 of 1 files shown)


carma_cooperative_perception/src/multiple_object_tracker_component.cpp
```diff
@@ -630,6 +630,24 @@ auto MultipleObjectTrackerNode::execute_pipeline() -> void
     }
   }
 
+  auto remove_start{std::remove_if(
```
MishkaMN, line 633: Let's add comment what are we doing here and why we need it

MishkaMN (changes requested):

MishkaMN (approved):

PR #2272, opened 2024-01-10
Handle throw from negative speed conversion in trajectory

MishkaMN:
# PR Details
## Description
This PR supports https://github.com/usdot-fhwa-stol/carma-utils/pull/210
Due to this PR change where negative speed is being detected and thrown, the only other plugin that generates negative speed is in stop_and_wait_plugin. 
This is because when stopping, using the simple kinematic equation: dx/dv = dt can blow up to be inf if dv = 0 (when stopping). 

This PR is to detect that negative speed and log it. The suggested change here still just assigns 0 values as a safety feature to the negative speed, which is the intention of the stop_and_wait_plugin (and all other stopping behaviors where this issue arises). 

However, more appropriate fix would be to fix it at the origin plugin, for example stop_and_wait_plugin, which will be in a future issue (https://github.com/usdot-fhwa-stol/carma-utils/issues/213). 

## Related GitHub Issue
https://github.com/usdot-fhwa-stol/carma-utils/issues/211

## Related Jira Key
https://usdot-carma.atlassian.net/browse/CDAR-717
CDAR-717

## How Has This Been Tested?
integration tested with cdasim pc1 under vru use case 1

## Types of changes

## Checklist:


stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -375,9 +375,9 @@ std::vector<carma_planning_msgs::msg::TrajectoryPlanPoint> StopandWait::compose_
   {
     if (times[i] != 0 && !std::isnormal(times[i]) && i != 0)
     {  // If the time
-      RCLCPP_WARN_STREAM(rclcpp::get_logger("stop_and_wait_plugin"),"Detected non-normal (nan, inf, etc.) time. Making it same as before: " << times[i-1]);
-      // NOTE: overriding the timestamps in assumption that pure_pursuit_wrapper will detect it as stopping case
-      times[i] = times[i - 1];
+      RCLCPP_WARN_STREAM(rclcpp::get_logger("stop_and_wait_plugin"),"Detected non-normal (nan, inf, etc.) time."
+        "This happens due to  dv = dx/0.0 = inf speed. Incrementing time by arbitrary large number to achieve stopping behavior");
+      times[i] = times[i - 1] + 3600.0; //assigning 10hr difference between what is normally 0.1-0.2 sec
```
adamlm, line 380: Why are we unable to set the time to `inf`?

adamlm: I'm not sure this correction should be here. I think it should move to the `trajectory_from_points_times_orientations` function. That function generates a trajectory, so it should be the one to take corrective actions if the inputs are collectively invalid.

adamlm (changes requested):

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
MishkaMN: So I inserted this function into the function you specified. I made the `times` value pass by reference because we are modifying it. I know we don't technically use it after the function, so we don't need to pass it by reference. However, I find the log statement that has `downtrack, t, and v` really helpful:
`RCLCPP_DEBUG_STREAM(rclcpp::get_logger("stop_and_wait_plugin"),"1d: " << downtracks[i] << " t: " << times[i] << " v: " << filtered_speeds[i]);`

MishkaMN: I also ended up adding more logic since you last reviewed that was throwing as well. So I fixed it.

stop_and_wait_plugin/include/stop_and_wait_plugin.hpp
```diff
@@ -104,9 +104,17 @@ class StopandWait
...
+   *  NOTE: If times has non-normal value (nan, inf etc), which can happen when speed is 0, it increments current time by 10hr from previous
+   * \param points The list of points that define the trajectory
+   * \param times The list if times associated to points. The input is modified if times has inf or nan values
+   * \param yaws The list if yaws associated to points
+   * \param starTime start time of the trajectory
+   * \return A trajectory plan
+   */
   std::vector<carma_planning_msgs::msg::TrajectoryPlanPoint> trajectory_from_points_times_orientations(
-      const std::vector<lanelet::BasicPoint2d>& points, const std::vector<double>& times,
+      const std::vector<lanelet::BasicPoint2d>& points, std::vector<double>& times,
```
paulbourelly999, line 117: Are we modifying the times passed here? Why did we remove the const keyword

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -216,6 +216,14 @@ std::vector<carma_planning_msgs::msg::TrajectoryPlanPoint> StopandWait::trajecto
   for (size_t i = 0; i < points.size(); i++)
   {
     carma_planning_msgs::msg::TrajectoryPlanPoint tpp;
+
+    if (times[i] != 0 && !std::isnormal(times[i]) && i != 0)
+    {  // If the time
+      RCLCPP_WARN_STREAM(rclcpp::get_logger("stop_and_wait_plugin"),"Detected non-normal (nan, inf, etc.) time."
```
paulbourelly999, line 222: Are these times or speeds?

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -345,7 +353,7 @@ std::vector<carma_planning_msgs::msg::TrajectoryPlanPoint> StopandWait::compose_
 
   std::vector<double> filtered_speeds = basic_autonomy::smoothing::moving_average_filter(speeds, config_.moving_average_window_size);
 
-  for (size_t i = 0; i < filtered_speeds.size(); i++)
+  for (size_t i = 1; i < filtered_speeds.size(); i++)
```
paulbourelly999, line 356: Why has this index changed

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -216,6 +216,14 @@ std::vector<carma_planning_msgs::msg::TrajectoryPlanPoint> StopandWait::trajecto
   for (size_t i = 0; i < points.size(); i++)
   {
     carma_planning_msgs::msg::TrajectoryPlanPoint tpp;
+
+    if (times[i] != 0 && !std::isnormal(times[i]) && i != 0)
+    {  // If the time
+      RCLCPP_WARN_STREAM(rclcpp::get_logger("stop_and_wait_plugin"),"Detected non-normal (nan, inf, etc.) time."
+        "This happens due to  dv = dx/0.0 = inf speed. Incrementing time by arbitrary large number to achieve stopping behavior");
+      times[i] = times[i - 1] + 3600.0; //assigning 10hr difference between what is normally 0.1-0.2 sec
```
paulbourelly999, line 224: It sounds to me that the times should never be modified. If anything is modified it should be the points. The issue of infinite acceleration arises when two points are the same but the initial point has a non zero velocity. Am I missing something here. The times should always be separated by 0.1 right?

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -202,7 +202,7 @@ std::vector<PointSpeedPair> StopandWait::maneuvers_to_points(const std::vector<c
 }
 
 std::vector<carma_planning_msgs::msg::TrajectoryPlanPoint> StopandWait::trajectory_from_points_times_orientations(
-    const std::vector<lanelet::BasicPoint2d>& points, const std::vector<double>& times, const std::vector<double>& yaws,
+    const std::vector<lanelet::BasicPoint2d>& points, std::vector<double>& times, const std::vector<double>& yaws,
```
paulbourelly999, line 205: Better than having these methods modify their inputs on invalid conditions, throw an exception and have the caller correct the inputs right?

paulbourelly999 (changes requested):
I might be misunderstanding this PR but some of this looks off from what our discussion was.

stop_and_wait_plugin/include/stop_and_wait_plugin.hpp
MishkaMN: yeah the logic is modifying the `times` solely to print it later outside the function for debugging purpose. It is much more meaningful to see the `times` with `downtracks` in the debug message outside, so I purposefully letting it pass by reference. I am also open to other suggestions.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
MishkaMN: Times. The time is going to `inf` because speed is zero and dt = dx / dv = dx / 0 = `inf`

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
MishkaMN: This is to let the keep the first velocity, which is the vehicle's current velocity. We shouldn't force the vehicle's first velocity to be 0 because plugins supposed to plan starting from the last available speed to be continuous.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
MishkaMN: Time should still be modified because the list would end up having `inf` values, which is not usable down the line. However, we can have it separated by 0.1 sec, I think that's no problem.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
MishkaMN: So the correct steps after keeping the first velocity is to calculate the dx it takes to slow down to 0 and keep that distance in the trajectory as well. This is same behavior we discussed. Otherwise due to the 0 velocity, dx will be 0, and again we will have the original problem of zigzag. Because despite using 0 velocity here to calculate the dx, the later components still uses the first non-zero first velocity to convert back, which conflicts. So this dx calculation is handled in `is_first_extrapolation_point` now. And then after that dx, the speed is 0, so the dx can be safely 0.

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
MishkaMN: This node is the main source of the trajectory. So this is the source place to modify and correct the inputs. I originally had it outside, but @adamlm suggested that from the name `trajectory_from_points_times_orientations`, this is the correct place to handle the wrong time and create correct trajectory

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
MishkaMN: 1. `speed_to_time` is still setting it to `nan` that part is not changed.
2. you are right, I changed it so that it is incrementing by config_.stop_timestep now
3. I also applied this change and reverted the pass-by-reference change

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
MishkaMN: After Adam's comment, I also reverted this change as we don't need to

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
MishkaMN: EDIT: Correction, they are `nan` not `inf` because dx and dv both 0 where 0/0 = `nan` is happening

stop_and_wait_plugin/include/stop_and_wait_plugin.hpp
```diff
@@ -104,7 +104,14 @@ class StopandWait
    */
   void splitPointSpeedPairs(const std::vector<PointSpeedPair>& points, std::vector<lanelet::BasicPoint2d>* basic_points,
                             std::vector<double>* speeds) const;
-
+  /**
+   * \brief Combines the points, times, and yaws to create a trajectory that starts from startTime
+   * \param points The list of points that define the trajectory
+   * \param times The list if times associated to points.
+   * \param yaws The list if yaws associated to points
```
adamlm, line 111: Should be `of` instead of `if`?

stop_and_wait_plugin/include/stop_and_wait_plugin.hpp
```diff
@@ -104,7 +104,14 @@ class StopandWait
...
+   * \param starTime start time of the trajectory
```
adamlm, line 112: `startTime`?

stop_and_wait_plugin/src/stop_and_wait_plugin.cpp
```diff
@@ -217,6 +217,14 @@ std::vector<carma_planning_msgs::msg::TrajectoryPlanPoint> StopandWait::trajecto
   {
     carma_planning_msgs::msg::TrajectoryPlanPoint tpp;
     rclcpp::Duration relative_time(times[i] * 1e9);
+
+    if (times[i] != 0 && !std::isnormal(times[i]) && i != 0)
+    {  // If the time
```
adamlm, line 222: Do we need the `// If the time`?

adamlm (changes requested):

MishkaMN:
I am still integration testing the changes, hold on.

docker/install.sh
```diff
@@ -33,6 +33,9 @@ fi
 
 cd ~/carma_ws
 
+echo "Installing multiple object tracking dependencies"
+sudo ./src/multiple_object_tracking/scripts/install_dependencies.sh
+
```
MishkaMN, line 38: not sure why it is showing as a change, this change already exists in develop.

adamlm (approved):

paulbourelly999 (approved):

PR #2279, opened 2024-01-12
Add string to int ID conversion

adamlm:
# PR Details
## Description

This PR fixes an issue when converting Track IDs (from the CP stack) to ExternalObject ID. The Track ID to ExternalObject ID conversion (string to int) now handles cases where the mapping is not one-to-one.

## Related GitHub Issue

Closes #2278 

## Related Jira Key

Closes [CDAR-720](https://usdot-carma.atlassian.net/browse/CDAR-720)

## Motivation and Context

Bug fix

## How Has This Been Tested?

Manually in integration testing

## Types of changes

- [x] Defect fix (non-breaking change that fixes an issue)

## Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/usdot-fhwa-stol/carma-platform/blob/develop/Contributing.md) document.

[CDAR-720]: https://usdot-carma.atlassian.net/browse/CDAR-720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2288, opened 2024-01-18
Prepend BSM ID to detections

### LINKED ISSUE #2286: [cooperative perception] Change SDSM to detection conversion to include BSM ID in Detection ID
### Summary

In the SDSM to detection converter, the detection ID includes only the associated SDSM object's ID, which leads to conflicting detections. The object ID is not unique across SDSMs. Concatenating BSM ID and Object ID will make the resulting detection's ID unique across all detections.

Related Jira story: [CDAR-538](https://usdot-carma.atlassian.net/browse/CDAR-538)

### Reasoning for new functionality

_No response_

[CDAR-538]: https://usdot-carma.atlassian.net/browse/CDAR-538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

adamlm:
# PR Details
## Description

This PR changes the way detection IDs are generated from incoming SDSMs. Previously, the detection IDs were simply the object IDs from the SDSM's contained objects. Now the detection IDs will have the SDSM's originator source ID prepended to it. This will detection IDs unique across all detections.

## Related GitHub Issue

Closes #2286

## Related Jira Key

Progresses [CDAR-538](https://usdot-carma.atlassian.net/browse/CDAR-538)

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:

[CDAR-538]: https://usdot-carma.atlassian.net/browse/CDAR-538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/carma_cooperative_perception/src/msg_conversion.cpp
+++ b/carma_cooperative_perception/src/msg_conversion.cpp
@@ -244,7 +244,21 @@ auto to_detection_list_msg(
 
     detection.header.stamp = to_time_msg(detection_time);
 
-    detection.id = std::to_string(common_data.detected_id.object_id);
+    static constexpr auto to_string = [](const std::vector<std::uint8_t> bsm_id) {
+      std::string bsm_string;
+
+      std::array<char, 2> buffer;
+      for (const auto & element : bsm_id) {
+        std::to_chars(std::begin(buffer), std::end(buffer), element, 16);
+        bsm_string.push_back(std::toupper(std::get<0>(buffer)));
+        bsm_string.push_back(std::toupper(std::get<1>(buffer)));
+      }
+
+      return bsm_string;
+    };
+
+    detection.id =
+      to_string(sdsm.source_id.id) + "-" + std::to_string(common_data.detected_id.object_id);
 
     const auto pos_offset{PositionOffsetXYZ::from_msg(common_data.pos)};
     detection.pose.pose.position = to_position_msg(MapCoordinate{
--- a/carma_cooperative_perception/src/msg_conversion.cpp
+++ b/carma_cooperative_perception/src/msg_conversion.cpp
@@ -244,17 +244,19 @@ auto to_detection_list_msg(
 
     detection.header.stamp = to_time_msg(detection_time);
 
-    static constexpr auto to_string = [](const std::vector<std::uint8_t> bsm_id) {
-      std::string bsm_string;
+    // TemporaryID and octet string terms come from the SAE J2735 message definitions
+    static constexpr auto to_string = [](const std::vector<std::uint8_t> & temporary_id) {
+      std::string str;
+      str.reserve(2 * std::size(temporary_id));  // Two hex characters per octet string
 
       std::array<char, 2> buffer;
-      for (const auto & element : bsm_id) {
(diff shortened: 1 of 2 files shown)


carma_cooperative_perception/src/msg_conversion.cpp
```diff
@@ -244,7 +244,21 @@ auto to_detection_list_msg(
 
     detection.header.stamp = to_time_msg(detection_time);
 
-    detection.id = std::to_string(common_data.detected_id.object_id);
+    static constexpr auto to_string = [](const std::vector<std::uint8_t> bsm_id) {
```
MishkaMN, line 247: it sounds like this is `uuid` instead of bsm_id, so let's rename it?

MishkaMN (changes requested):

carma_cooperative_perception/test/test_msg_conversion.cpp
```diff
@@ -118,7 +119,7 @@ TEST(ToDetectionMsg, Simple)
   EXPECT_DOUBLE_EQ(detection.accel.accel.linear.y, 1.0);
   EXPECT_NEAR(detection.accel.accel.linear.z, 2.4 * 9.80665, 1e-4);
 
-  EXPECT_EQ(detection.id, std::to_string(0xBEEF));
+  EXPECT_EQ(detection.id, "BADDCAFE-1");
```
MishkaMN, line 122: So this goes back to the conversation we had for https://github.com/usdot-fhwa-stol/carma-platform/pull/2279. Would it be possible by chance to have identical track names that are supposed to be representing different objects (I know we are making detections here, but I believe tracks just directly use the detection id when first created). 
Source1: {0xB1, 0xDD, 0xCA, 0xFE} first object -> `B1DDCAFE-1` -----> (when turned into external object, we only leave digits I believe, so) `11`?
Source2: {0xC1, 0xDD, 0xCA, 0xFE} first object -> `C1DDCAFE-1` ------> also `11`?

MishkaMN (changes requested):

carma_cooperative_perception/src/msg_conversion.cpp
adamlm: I took a look at the J2735 document and saw the type for that message was TemporaryID. I've updated the terminology to reflect that.

carma_cooperative_perception/test/test_msg_conversion.cpp
adamlm: Yes, that's a limitation on the track->external object conversion. We can open a new issue for that. I think we can implement a monotonically-increasing number generator so that all new tracks (regardless of their originating detection's ID) will have a unique value. Thoughts?

MishkaMN: Yeah I think that works too. Let's do that

MishkaMN (approved):

PR #2298, opened 2024-02-05
Add DetectionList message visualization

adamlm:
# PR Details
## Description

This PR adds a new node to convert `DetectionList.msg` messages to `MarkerArray.msg` messages that can be used in RViz or Foxglove Studio to visualize detections. This will help with troubleshooting cooperative perception pipeline issues.

## Related Jira Key

Closes [CDAR-758](https://usdot-carma.atlassian.net/browse/CDAR-758)

## Motivation and Context

Helps with troubleshooting cooperative perception stack issues.

## How Has This Been Tested?

## Types of changes

## Checklist:

[CDAR-758]: https://usdot-carma.atlassian.net/browse/CDAR-758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/carma_cooperative_perception/src/detection_list_viz_component.cpp
+++ b/carma_cooperative_perception/src/detection_list_viz_component.cpp
@@ -1,0 +1,56 @@
+// Copyright 2024 Leidos
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "carma_cooperative_perception/detection_list_viz_component.hpp"
+
+#include <rclcpp_components/register_node_macro.hpp>
+
+namespace carma_cooperative_perception
+{
+
+DetectionListVizNode::DetectionListVizNode(const rclcpp::NodeOptions & options)
+: Node("detection_list_viz_node", options)
+{
+  detection_list_sub_ =
+    create_subscription<carma_cooperative_perception_interfaces::msg::DetectionList>(
+      "input/detections_lists", 1,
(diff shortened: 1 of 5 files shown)


carma_cooperative_perception/src/detection_list_viz_component.cpp
```diff
@@ -0,0 +1,56 @@
...
+          marker.scale.z = 2.0;
+        }
+
+        marker_pub_->publish(markers);
+      });
+
+  marker_pub_ = create_publisher<visualization_msgs::msg::MarkerArray>("output/markers", 1);
+}
+
+}  // namespace carma_cooperative_perception
```
MishkaMN, line 50: Shouldn't we include DELETEALL marker otherwise some markers from previous publication may still get visualized?

```
visualization_msgs::msg::MarkerArray viz_msg;
    //delete all markers before adding new ones
    visualization_msgs::msg::Marker marker;
    viz_msg.markers.reserve(msg_ptr->detections.size() + 1); //+1 to account for delete all marker
    marker.id = 0;
    marker.action = visualization_msgs::msg::Marker::DELETEALL;
    viz_msg.markers.push_back(marker);
```

MishkaMN (changes requested):

carma_cooperative_perception/src/detection_list_viz_component.cpp
MishkaMN: I am worried that the troubleshooting tool giving wrong impression may confuse the user even more. I am assuming the marker is flashing because it is receiving empty sdsms? If that's the case, can we only publish markers if the list has objects?

MishkaMN (changes requested):
I am worried that the troubleshooting tool giving wrong impression may confuse the user even more. I am assuming the marker is flashing because it is receiving empty sdsms? If that's the case, can we only publish markers if the list has objects? If that still doesn't work, then I am okay merging it (maybe just add a comment about the situation). Let me know.

carma_cooperative_perception/src/detection_list_viz_component.cpp
adamlm: It is flashing because it receives a list of detections from CARMA in one time step then a list of detections from an SDSM in the next.

MishkaMN: Okay that makes sense

MishkaMN (approved):

PR #2311, opened 2024-02-14
[cooperative perception] Change UUID when creating tracks

adamlm:
# PR Details
## Description

This PR changes how UUIDs are generated for tracks. Previously, the multiple object tracking node would reuse the UUID from a detection when converting it to a track. This made the track identifiers misleading because tracks' states are fused from several sources, not only detections with the same identifier. It also caused issues downstream when converting the UUID to an integer due to data size limits.

The node now maintains a monotonically increasing count of lifetime-generated tracks. It uses this number when creating new tracks.

## Related Jira Key

Closes [CDAR-792](https://usdot-carma.atlassian.net/browse/CDAR-792)

## How Has This Been Tested?

## Types of changes

## Checklist:

[CDAR-792]: https://usdot-carma.atlassian.net/browse/CDAR-792?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/carma_cooperative_perception/src/multiple_object_tracker_component.cpp
+++ b/carma_cooperative_perception/src/multiple_object_tracker_component.cpp
@@ -582,6 +582,8 @@ auto MultipleObjectTrackerNode::execute_pipeline() -> void
     for (const auto & cluster : clusters) {
       const auto detection{std::cbegin(cluster.get_detections())->second};
       const auto uuid_str{mot::get_uuid(detection).value()};
+      // CARLA uses three-digit actor identifiers. We want to UUID scheme to be
+      // <track_number>-<carla_actor_id> for easier visual analysis by users.
       const mot::Uuid new_uuid{
         std::to_string(lifetime_generated_track_count_++) +
         uuid_str.substr(std::size(uuid_str) - 4, 4)};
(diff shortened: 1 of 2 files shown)


carma_cooperative_perception/src/multiple_object_tracker_component.cpp
```diff
@@ -577,7 +581,12 @@ auto MultipleObjectTrackerNode::execute_pipeline() -> void
     const auto clusters{mot::cluster_detections(detections_, 0.75)};
     for (const auto & cluster : clusters) {
       const auto detection{std::cbegin(cluster.get_detections())->second};
-      track_manager_.add_tentative_track(std::visit(make_track_visitor, detection));
+      const auto uuid_str{mot::get_uuid(detection).value()};
+      const mot::Uuid new_uuid{
+        std::to_string(lifetime_generated_track_count_++) +
+        uuid_str.substr(std::size(uuid_str) - 4, 4)};
```
MishkaMN, line 587: Let's add a comment here why it is 4. I believe it is solely because CARLA's unique ids are in 3 digits and we need 1 more for the "-" character right? Same in the place below.

MishkaMN (changes requested):

carma_cooperative_perception/src/multiple_object_tracker_component.cpp
adamlm: Done

MishkaMN (approved):

PR #2314, opened 2024-02-16
[cooperative perception] Fix distance calculation for host vehicle filter

adamlm:
# PR Details
## Description

This PR fixes an implementation issue with the host vehicle filter's distance calculation. The `euclidean_distance_squared()` function included angular distance as part of the calculation, which is different from Euclidean distance. This caused filter issues do to noisy rotations.

This PR also makes a small improvement to the `is_within_distance` lambda by adding a semantic class check. If a detection's semantic class is known to not be a vehicle, the lambda will return false because the object cannot be the host vehicle.

## Related Jira Key

Closes [CDAR-805](https://usdot-carma.atlassian.net/browse/CDAR-805)

## Motivation and Context

Bug fix

## How Has This Been Tested?

## Types of changes

## Checklist:

### DIFF
--- a/carma_cooperative_perception/src/host_vehicle_filter_component.cpp
+++ b/carma_cooperative_perception/src/host_vehicle_filter_component.cpp
@@ -167,6 +167,12 @@ auto HostVehicleFilterNode::attempt_filter_and_republish(
     return;
   }
 
+  if (
+    detection.semantic_class != detection.SEMANTIC_CLASS_UNKNOWN &&
+    detection.semantic_class != detection.SEMANTIC_CLASS_SMALL_VEHICLE) {
+    return false;
+  }
+
   const auto is_within_distance = [this](const auto & detection) {
     return euclidean_distance_squared(host_vehicle_pose_.value().pose, detection.pose.pose) <=
            this->squared_distance_threshold_meters_;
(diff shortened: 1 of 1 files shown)


carma_cooperative_perception/src/host_vehicle_filter_component.cpp
```diff
@@ -184,10 +190,7 @@ auto euclidean_distance_squared(
   const geometry_msgs::msg::Pose & a, const geometry_msgs::msg::Pose & b) -> double
 {
   return std::pow(a.position.x - b.position.x, 2) + std::pow(a.position.y - b.position.y, 2) +
-         std::pow(a.position.z - b.position.z, 2) + std::pow(a.orientation.x - b.orientation.x, 2) +
-         std::pow(a.orientation.y - b.orientation.y, 2) +
-         std::pow(a.orientation.z - b.orientation.z, 2) +
-         std::pow(a.orientation.w - b.orientation.w, 2);
+         std::pow(a.position.z - b.position.z, 2);
```
MishkaMN, line 193: Should we just take off z component as well since carma doesn't use this component?

MishkaMN (changes requested):

carma_cooperative_perception/src/host_vehicle_filter_component.cpp
adamlm: I don't think so because Euclidean (without any other qualifiers) is typically 3D. If we reduce it to 2D, we should rename the function to `pythagorean_distance_squared()` or something similar.

Since CARMA does not use the z-axis, this term will be zero anyways.

MishkaMN (approved):

Issue #2323, opened 2024-02-28
Yield plugin can sometimes detect vehicle is passed but can detect collision again


MishkaMN:
### Summary

Yield_plugin is ignoring some collision because it detected that the vehicle would be passed the object. 
However, it also detects collisions again on the next iteration, so it can have false positives for collision in between the iterations.

MishkaMN:
Some cases have been handled in above 2325 PR.
However, I am fairly confident rest of the issues happen because of issue: https://github.com/usdot-fhwa-stol/carma-platform/issues/2326

Another reason it can happen is because we are downsampling the trajectory when checking the collision. Therefore, we may end up missing the collision if the collision were to happen duing the downsampled place. Although this case is rare and can be resolved if the collision detection algorithm was much better and faster. So we can close this speific issue if 2326 is addressed

PR #2325, opened 2024-02-28
Fix platform speeding up briefly while speeding up

MishkaMN:
# PR Details
## Description
This PR adds consecutive_clearance_counter feature where platform doesn't accidentally detect that the vehicle would be passing the pedestrian. The functionality to detect whether if pedestrian was passed was first introduced so that the vehicle doesn't stop indefinitely if a pedestrian was within collision radius, but behind the vehicle. However, it is also making yield_plugin to ignore close calls and collisions if vehicle is predicted to barely miss the pedestrian. So the counter now should check consecutive 20 (which amounts to 2 sec approximately) to confirm if the vehicle is passed the pedestrian, otherwise err on the safer side and yield.
NOTE: it doesn't completely fix all occasions please see this issue. https://github.com/usdot-fhwa-stol/carma-platform/issues/2326
Following shows a log statement where the counter went up as high as 11. 
image

Following shows a log statements where the vehicle continued detecting collision way past the first time it falsely detected that the vehicle is passed and continued safely stopping
image

image

## Related GitHub Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/2323

## Related Jira Key
https://usdot-carma.atlassian.net/browse/CDAR-811

## Motivation and Context
vru use case 1 verification testing prep

## How Has This Been Tested?
sim pc 1

## Types of changes

## Checklist:

### DIFF
--- a/yield_plugin/include/yield_plugin/yield_plugin.hpp
+++ b/yield_plugin/include/yield_plugin/yield_plugin.hpp
@@ -63,11 +63,11 @@ struct PointSpeedPair
 /**
  * \brief Convenience class for saving collision results
  */
-struct CollisionData
+struct GetCollisionResult
 {
   rclcpp::Time collision_time;
-  lanelet::BasicPoint2d predicted_vehicle_collision_point;
-  lanelet::BasicPoint2d predicted_obstacle_collision_point;
+  lanelet::BasicPoint2d point1;
+  lanelet::BasicPoint2d point2;
 };
 
 /**
@@ -245,15 +245,14 @@ public:
 
   /**
    * \brief Return naive collision time and locations based on collision radius given two trajectories with one being obstacle's predicted steps
-   * \param object_id object id corresponding to the trajectory2
    * \param trajectory1 trajectory of the ego vehicle
    * \param trajectory2 trajectory of predicted steps
    * \param collision_radius a distance to check between two trajectory points at a same timestamp that is considered a collision
    * NOTE: Currently Traj2 is assumed to be a simple cv model to save computational performance
-   * NOTE: Collisions are based on only collision radius at the same predicted time even if ego vehicle maybe past the obstacle. To filter these cases, see `is_object_behind_the_vehicle()`
+   * NOTE: Collisions are based on only collision radius at the same predicted time even if ego vehicle maybe past the obstacle. To filter these cases, see `is_object_behind_vehicle()`
    * \return data of time of collision if detected, otherwise, std::nullopt
    */
-  std::optional<CollisionData> get_collision_data(uint32_t object_id, const carma_planning_msgs::msg::TrajectoryPlan& trajectory1, const std::vector<carma_perception_msgs::msg::PredictedState>& trajectory2, double collision_radius);
+  std::optional<GetCollisionResult> get_collision(const carma_planning_msgs::msg::TrajectoryPlan& trajectory1, const std::vector<carma_perception_msgs::msg::PredictedState>& trajectory2, double collision_radius);
 
   /**
    * \brief Return collision time given two trajectories with one being external object with predicted steps
@@ -267,12 +266,14 @@ public:
   /**
    * \brief Check if object location is behind the vehicle using estimates of the vehicle's length and route downtracks
    * \param object_id object id to use for the consecutive_clearance_count_for_obstacles_
-   * \param collision_data info about collision such as collision_time, vehicle and obstacle positions at the time of collision
+   * \param collision_time predicted time of collision
+   * \param vehicle_downtrack at the time of collision
+   * \param object_downtrack at the time of collision
    * NOTE: Uses internal counter low pass filter to confirm the object is behind only if it counted continuously above
            config_.consecutive_clearance_count_for_obstacles_threshold
    * \return return true if object is behind the vehicle
    */
-  bool is_object_behind_the_vehicle(uint32_t object_id, const CollisionData& collision_data);
+  bool is_object_behind_vehicle(uint32_t object_id, const rclcpp::Time& collision_time, double vehicle_point, double object_downtrack);
 
   /**
    * \brief Return the earliest collision object and time of collision pair from the given trajectory and list of external objects with predicted states.
--- a/yield_plugin/include/yield_plugin/yield_plugin.hpp
+++ b/yield_plugin/include/yield_plugin/yield_plugin.hpp
@@ -61,6 +61,16 @@ struct PointSpeedPair
 };
 
 /**
+ * \brief Convenience class for saving collision results
+ */
+struct CollisionData
+{
+  rclcpp::Time collision_time;  // possibly unneeded
+  lanelet::BasicPoint2d predicted_vehicle_collision_point;
+  lanelet::BasicPoint2d predicted_obstacle_collision_point;
+};
+
+/**
  * \brief Class containing primary business logic for the In-Lane Cruising Plugin
  *
  */
@@ -234,15 +244,16 @@ public:
   void set_external_objects(const std::vector<carma_perception_msgs::msg::ExternalObject>& object_list);
 
   /**
-   * \brief Return collision time given two trajectories with one being predicted steps
+   * \brief Return naive collision time and locations based on collision radius given two trajectories with one being obstacle's predicted steps
    * \param object_id object id corresponding to the trajectory2
    * \param trajectory1 trajectory of the ego vehicle
    * \param trajectory2 trajectory of predicted steps
    * \param collision_radius a distance to check between two trajectory points at a same timestamp that is considered a collision
    * NOTE: Currently Traj2 is assumed to be a simple cv model to save computational performance
-   * \return time_of_collision if collision detected, otherwise, std::nullopt
+   * NOTE: Collisions are based on only collision radius at the same predicted time even if ego vehicle maybe past the obstacle. To filter these cases, see `is_object_behind_the_vehicle()`
+   * \return data of time of collision if detected, otherwise, std::nullopt
    */
-  std::optional<rclcpp::Time> get_collision_time(uint32_t object_id, const carma_planning_msgs::msg::TrajectoryPlan& trajectory1, const std::vector<carma_perception_msgs::msg::PredictedState>& trajectory2, double collision_radius);
+  std::optional<CollisionData> get_collision_data(uint32_t object_id, const carma_planning_msgs::msg::TrajectoryPlan& trajectory1, const std::vector<carma_perception_msgs::msg::PredictedState>& trajectory2, double collision_radius);
 
   /**
    * \brief Return collision time given two trajectories with one being external object with predicted steps
@@ -254,15 +265,14 @@ public:
   std::optional<rclcpp::Time> get_collision_time(const carma_planning_msgs::msg::TrajectoryPlan& original_tp, const carma_perception_msgs::msg::ExternalObject& curr_obstacle);
 
   /**
-   * \brief Check if object location is behind the vehicle using estimates of the vehicle's length
+   * \brief Check if object location is behind the vehicle using estimates of the vehicle's length and route downtracks
    * \param object_id object id to use for the consecutive_clearance_count_for_obstacles_
-   * \param vehicle_downtrack vehicle downtrack of its back axle along the route
-   * \param object_downtrack object downtrack along the route
+   * \param collision_data info about collision such as collision_time, vehicle and obstacle positions at the time of collision
    * NOTE: Uses internal counter low pass filter to confirm the object is behind only if it counted continuously above
            config_.consecutive_clearance_count_for_obstacles_threshold
    * \return return true if object is behind the vehicle
    */
-  bool is_object_behind_the_vehicle(uint32_t object_id, double vehicle_downtrack, double object_downtrack);
+  bool is_object_behind_the_vehicle(uint32_t object_id, const CollisionData& collision_data);
 
   /**
    * \brief Return the earliest collision object and time of collision pair from the given trajectory and list of external objects with predicted states.
--- a/yield_plugin/include/yield_plugin/yield_plugin.hpp
+++ b/yield_plugin/include/yield_plugin/yield_plugin.hpp
@@ -235,13 +235,14 @@ public:
 
   /**
    * \brief Return collision time given two trajectories with one being predicted steps
+   * \param object_id object id corresponding to the trajectory2
    * \param trajectory1 trajectory of the ego vehicle
    * \param trajectory2 trajectory of predicted steps
    * \param collision_radius a distance to check between two trajectory points at a same timestamp that is considered a collision
    * NOTE: Currently Traj2 is assumed to be a simple cv model to save computational performance
    * \return time_of_collision if collision detected, otherwise, std::nullopt
    */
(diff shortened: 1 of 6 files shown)


adamlm (changes requested):
1. This PR proposes a bug fix and a new feature. We should split this up into two separate PRs. Can we do that?

2. I'm not sure I understand the `consecutive_clearance_count` solution. It appears we are keeping track of how often a specific obstacle is beyond the collision threshold and assuming this count somehow translates to an obstacle being behind the vehicle. Is this correct?

    Perhaps we could do an alternative implementation based on a line passing through the vehicle's rear axis. If the obstacle's position is on one side of the line (I believe this would be the _epigraph_) then it is not behind the vehicle. Otherwise, it is behind the vehicle.

MishkaMN (commented):
I separated the PR. New one here for the goal position calculation: https://github.com/usdot-fhwa-stol/carma-platform/pull/2331/files

yield_plugin/src/yield_plugin.cpp
```diff
@@ -659,22 +662,35 @@ namespace yield_plugin
         double vehicle_downtrack = wm_->routeTrackPos(vehicle_point).downtrack;
         double object_downtrack = wm_->routeTrackPos(object_point).downtrack;
 
-        if (vehicle_downtrack > object_downtrack + config_.vehicle_length / 2)  // if half a length of the vehicle past, it is considered behind
+        if (vehicle_downtrack > object_downtrack + config_.vehicle_length / 2) // if half a length of the vehicle past, it maybe considered behind
```
MishkaMN, line 665: > I'm not sure I understand the consecutive_clearance_count solution. It appears we are keeping track of how often a specific obstacle is beyond the collision threshold and assuming this count somehow translates to an obstacle being behind the vehicle. Is this correct?

It is actually checking a case where the obstacle is within collision threshold. It would be detected a collision all the time without this check even in cases where vehicle is passed the pedestrian

> Perhaps we could do an alternative implementation based on a line passing through the vehicle's rear axis. If the obstacle's position is on one side of the line (I believe this would be the epigraph) then it is not behind the vehicle. Otherwise, it is behind the vehicle.

Current implementation is exactly doing that. vehicle_downtrack and object_downtrack are distances from the vehicle's engagement position projected from their x,y position to along the road. So in this case, the line passing through vehicle's rear axis would be vehicle_downtrack. And if object downtrack is more than half of the vehicle_length (to be conservative) away from the vehicle_downtrack, it means it is passed.

adamlm, line 665: The vehicle's down track is based on the position of its rear axle, right? If so, why are we moving the value forward by half a vehicle length? This seems like it would cause issues when an obstacle is near the vehicle's rear.

adamlm (commented):
After reading your description and looking through the code some more, it seems that we are essentially low-passing the "is behind vehicle" check based on the down track difference. If the vehicle's down track is consistently ahead of the obstacle's, we consider the obstacle _behind_ the vehicle. Is this a correct interpretation?

If it is correct, why do we need this kind of filtering?

yield_plugin/src/yield_plugin.cpp
MishkaMN: 2. Yes, it is based on the rear-axle. We can manually check it. Consider such scenarios (veh_length is 4 meters): 
- vehicle_downtrack = 80m and object_downtrack 82m.  It means ped is 2 meters in front of the back-axle. `if statement` goes: 80 > 82 + 2 is false, so vehicle is NOT passed. Correct
- vehicle_downtrack = 82m and object_downtrack 82m.  It means ped is at the back-axle. `if statement` goes 82 > 82 + 2 is false, so vehicle is NOT passed. Correct
- vehicle_downtrack = 84m and object_downtrack 82m.  It means ped is at the 1/2 veh length passed the rear-axle (1/2 veh len is more than distance between rear-axle and rear bumper). `if statement` goes 84 > 82 + 2 is false, so vehicle is NOT passed. Incorrect-> I guess I could change the expression to `>=`

adamlm: > * vehicle_downtrack = 84m and object_downtrack 82m.  It means ped is at the 1/2 veh length passed the rear-axle (1/2 veh len is more than distance between rear-axle and rear bumper for all cars except trucks maybe). `if statement` goes 84 > 82 + 2 is `false`, so vehicle is NOT passed. Incorrect-> I guess I could change the expression to `>=` . Since the `if statement` here should say `true`

It's unclear (to me at least) from the code alone that we modify the down track to accommodate for the vehicle overhang. I suggest changing

```cpp
if (vehicle_downtrack > object_downtrack + config_.vehicle_length / 2)
```

to 

```cpp
if (object_downtrack < vehicle_downtrack - config_.vehicle_length / 2)
```

and adding a clarifying comment about how we need to compensate for the vehicle's body overhang and that half the vehicle's total length is a conservative value.

adamlm: > I am hesitant to move the obstacle_behind_filter out of the function. It is mainly to avoid another 2 for loops to get data we already have.

We can avoid doing another two loops by refactoring the `get_collision_time()` function. Consider the following alternative:

```cpp
struct get_collision_result {
  rclcpp::Time collision_time;  // possibly unneeded
  lanelet::BasicPoint2d point1;
  lanelet::BasicPoint2d point2;
};

std::optional<get_collision_result> get_collision(/* traj1 */, /* traj2 */);

if (const auto result{get_collision(obstacle_trajectory, vehicle_trajectory)}) {
  const auto obstacle_downtrack{wm_->routeTrackPos(result.point1).downtrack};
  const auto vehicle_downtrack{wm->routeTrack(result.point2).downtrack};

  if (obstacle_behind_vehicle(obstacle_id, obstacle_downtrack, vehicle_downtrack) {
    // ...
  }
}
```

This is a similar architecture to the [`std::ranges::mismatch()`](https://en.cppreference.com/w/cpp/algorithm/ranges/mismatch) function.

> And having obstacle_behind_filter inside the get_collision_time makes sense because without it, get_collision_time returns wrong or incomplete result.

The `get_collision_time()` function's only job is to determine if there is a collision between two trajectories. Noisy trajectories mean we will get some false positives, but `get_collision_time()` doesn't need to know or care about that context. That's the caller's responsibility. From the function's perspective, the two provided trajectories resulted in a collision.

> The user would have to know to call this extra edge case handler function to get correct result.

The users in this case are downstream or service-calling ROS nodes. The code we are discussing is an implementation detail, so it is our responsibility to properly handle the edge case.

MishkaMN: Okay I incorporated your suggestions and added unit test for the new function.

yield_plugin/include/yield_plugin/yield_plugin.hpp
```diff
@@ -234,14 +244,16 @@ class YieldPlugin
...
    * \param trajectory1 trajectory of the ego vehicle
    * \param trajectory2 trajectory of predicted steps
    * \param collision_radius a distance to check between two trajectory points at a same timestamp that is considered a collision
    * NOTE: Currently Traj2 is assumed to be a simple cv model to save computational performance
-   * \return time_of_collision if collision detected, otherwise, std::nullopt
+   * NOTE: Collisions are based on only collision radius at the same predicted time even if ego vehicle maybe past the obstacle. To filter these cases, see `is_object_behind_the_vehicle()`
+   * \return data of time of collision if detected, otherwise, std::nullopt
    */
-  std::optional<rclcpp::Time> get_collision_time(const carma_planning_msgs::msg::TrajectoryPlan& trajectory1, const std::vector<carma_perception_msgs::msg::PredictedState>& trajectory2, double collision_radius);
+  std::optional<CollisionData> get_collision_data(uint32_t object_id, const carma_planning_msgs::msg::TrajectoryPlan& trajectory1, const std::vector<carma_perception_msgs::msg::PredictedState>& trajectory2, double collision_radius);
```
adamlm, line 256: 1. It looks like this function no longer needs the `object_id` parameter.
2. Consider dropping the `_data` suffix. We are looking for a collision, not a collision data. The results, if there is a collision, are stored in a `GetCollisionData` object. This matches closer to the `std::mismatch` / `std::mismatch_result` naming style.

yield_plugin/include/yield_plugin/yield_plugin.hpp
```diff
@@ -60,6 +60,16 @@ struct PointSpeedPair
   double speed = 0;
 };
 
+/**
+ * \brief Convenience class for saving collision results
+ */
+struct CollisionData
```
adamlm, line 66: Consider renaming this to `GetCollisionResult` to better align with the Standard Library's `*_result` types. This kind of struct, coupled with its usage, is intended to facilitate multiple return values from `get_collision()`. It's not really something to pass around to other functions.

yield_plugin/include/yield_plugin/yield_plugin.hpp
```diff
@@ -60,6 +60,16 @@ struct PointSpeedPair
...
+{
+  rclcpp::Time collision_time;
+  lanelet::BasicPoint2d predicted_vehicle_collision_point;
+  lanelet::BasicPoint2d predicted_obstacle_collision_point;
```
adamlm, line 70: The `get_collision()` or (`get_collision_data()`) function is agnostic to the trajectory types considering the `trajectory1` and `trajectory2` parameter names. Consider renaming to `point1` and `point2`, respectively. This may seem antithetical to clearly-named variables, but I think this is a good exception case. The `get_collision()` function is generic, so it's return type should be as well.

yield_plugin/include/yield_plugin/yield_plugin.hpp
```diff
@@ -252,6 +264,16 @@ class YieldPlugin
...
 
+  /**
+   * \brief Check if object location is behind the vehicle using estimates of the vehicle's length and route downtracks
+   * \param object_id object id to use for the consecutive_clearance_count_for_obstacles_
+   * \param collision_data info about collision such as collision_time, vehicle and obstacle positions at the time of collision
+   * NOTE: Uses internal counter low pass filter to confirm the object is behind only if it counted continuously above
+           config_.consecutive_clearance_count_for_obstacles_threshold
+   * \return return true if object is behind the vehicle
+   */
+  bool is_object_behind_the_vehicle(uint32_t object_id, const CollisionData& collision_data);
```
adamlm, line 275: Consider changing this function's signature to use `lanelet::BasicPoint2d` values directly. This function does not need to be coupled to `get_collision_data()` calls.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -653,31 +652,58 @@ namespace yield_plugin
...
+
+    const auto distance{std::hypot(
+      collision_data.predicted_vehicle_collision_point.x() - collision_data.predicted_obstacle_collision_point.x(),
+      collision_data.predicted_vehicle_collision_point.y() - collision_data.predicted_obstacle_collision_point.y()
+    )}; //for debug
+
+    RCLCPP_WARN_STREAM(nh_->get_logger(), "Collision detected for object: " << object_id << ", at timestamp " << std::to_string(collision_data.collision_time.seconds()) <<
+      ", x: " << collision_data.predicted_vehicle_collision_point.x() << ", y: " << collision_data.predicted_vehicle_collision_point.y() <<
+      ", within actual downtrack distance: " << object_downtrack - vehicle_downtrack <<
+      ", and collision distance: " << distance);
```
adamlm, line 702: Comment follows from my other comment about changing the function's sigature.

This should be moved to the calling code. It could look something like:

```cpp
const auto is_behind{is_object_behind_the_vehicle(/* args */);
if (is_behind) {
  // Print logging ...
}
```

yield_plugin/src/yield_plugin.cpp
```diff
@@ -653,31 +652,58 @@ namespace yield_plugin
...
+    if (object_downtrack < vehicle_downtrack - config_.vehicle_length / 2)
+    {
+      consecutive_clearance_count_for_obstacles_[object_id] = std::min(consecutive_clearance_count_for_obstacles_[object_id] + 1, config_.consecutive_clearance_count_for_obstacles_threshold);
+      RCLCPP_INFO_STREAM(nh_->get_logger(), "Detected an object nearby might be behind the vehicle at timestamp: " << std::to_string(collision_data.collision_time.seconds()) <<
+        ", and consecutive_clearance_count_for obstacle: " <<  object_id << ", is: " << consecutive_clearance_count_for_obstacles_[object_id]);
+    }
+    // confirmed false positive for a collision
+    if (consecutive_clearance_count_for_obstacles_[object_id] >= config_.consecutive_clearance_count_for_obstacles_threshold)
+    {
+      RCLCPP_INFO_STREAM(nh_->get_logger(), "Confirmed that the object: " << object_id << " is behind the vehicle at timestamp " << std::to_string(collision_data.collision_time.seconds()));
```
adamlm, line 684: Same comment about moving this to calling code.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -653,31 +652,58 @@ namespace yield_plugin
...
+        collision_data.collision_time = rclcpp::Time(p2a.header.stamp);
+        return collision_data;
       }
     }
 
     // No collision detected
     return std::nullopt;
   }
 
+  bool YieldPlugin::is_object_behind_the_vehicle(uint32_t object_id, const CollisionData& collision_data)
```
adamlm, line 667: (optional) Consider removing `the` to save some horizontal space.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -653,31 +652,58 @@ namespace yield_plugin
...
+  {
+    const double vehicle_downtrack = wm_->routeTrackPos(collision_data.predicted_vehicle_collision_point).downtrack;
+    const double object_downtrack = wm_->routeTrackPos(collision_data.predicted_obstacle_collision_point).downtrack;
+
+    auto previous_clearance_count = consecutive_clearance_count_for_obstacles_[object_id];
```
adamlm, line 672: Can be made `const`.

adamlm (changes requested):

yield_plugin/include/yield_plugin/yield_plugin.hpp
MishkaMN: 1. Removed
2. replaced

yield_plugin/include/yield_plugin/yield_plugin.hpp
MishkaMN: Renamed

yield_plugin/include/yield_plugin/yield_plugin.hpp
MishkaMN: Renamed

yield_plugin/include/yield_plugin/yield_plugin.hpp
MishkaMN: Extracted to use the individual members of collision result

yield_plugin/src/yield_plugin.cpp
MishkaMN: made it const

yield_plugin/src/yield_plugin.cpp
MishkaMN: refactored to follow this

yield_plugin/src/yield_plugin.cpp
MishkaMN: moved out the logging

adamlm (approved):

Issue #2329, opened 2024-02-29
A steering jerk can occasionally occur at the beginning of a lane change trajectory 


JonSmet:
### Summary

During Freight Emergency Response Phase 1 Testing (April/May 2024), it was found that a steering jerk could occasionally occur at the beginning of a lane change trajectory. It seems that the first point of the lane change trajectory becomes set to a random location on the map (perhaps an overflow error of some kind?); since the trajectory point is not on the road path, this results in a steering jerk when that portion of the vehicle attempts to follow that part of the trajectory.

### Version

4.3.0 (Current)

### Expected Behavior

Trajectory plan points follow the lane centerline during lane following operations or along a reasonable path on the road during other maneuvers (lane changes, for example).

### Actual Behavior

It seems one trajectory plan point at the beginning of lane changes is set to a random point on the map, leading to an erratic trajectory.

### Steps to Reproduce the Actual Behavior

1. Engage CARMA Platform on a route with a lane change.
2. Examine the trajectory at the beginning of the lane change. 
(This occurs intermittently, so the test might need to be conducted multiple times)

JonSmet:
Closed by #2330

PR #2330, opened 2024-02-29
Fix unintended size_t wrap-around behavior in basic_autonomy package

JonSmet:
# PR Details
## Description

This PR fixes unintended size_t wrap-around behavior in the `basic_autonomy` package's `get_nearest_index_by_downtrack()` function. This function receives [1] a vector of points within increasing downtracks and [2] a target downtrack, and it finds the index of the point with the greatest downtrack that is less than the target downtrack (returning index 0 if no points have a downtrack less than the target downtrack). 

Prior to this fix, the line (slightly altered here) `size_t best_index = std::max((size_t)0, i - 1);` led to unintended behavior if `i` was set to 0, since size_t(-1) results in a very large positive number.

NOTE: Two unit tests currently fail in the `basic_autonomy` package. These will be worked on as part of a separate effort.

## Related GitHub Issue

#2329 

## Related Jira Key

CF-785

## Motivation and Context

Fixes the above linked GitHub issue. This is a safety improvement to the system and also improves ride comfort.

## How Has This Been Tested?

A unit test was created for the `get_nearest_index_by_downtrack()` function (no unit tests existed previously for this function) to test its results for several cases, including the case that failed in the original implementation.

## Types of changes

## Checklist:

### DIFF
--- a/basic_autonomy/src/helper_functions.cpp
+++ b/basic_autonomy/src/helper_functions.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021-2022 LEIDOS.
+ * Copyright (C) 2021-2024 LEIDOS.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -14,6 +14,7 @@
  * the License.
  */
 
+#include <algorithm>
 #include <basic_autonomy/helper_functions.hpp>
 
 
@@ -65,18 +66,21 @@ namespace waypoint_generation
     int get_nearest_index_by_downtrack(const std::vector<lanelet::BasicPoint2d>& points, const carma_wm::WorldModelConstPtr& wm, double target_downtrack)
     {
         size_t best_index = points.size() - 1;
-        for(size_t i = 0;i < points.size(); i++){
-            double downtrack = wm->routeTrackPos(points[i]).downtrack;
-            if(downtrack > target_downtrack){
-                //If value is negative, best index should be index 0
-                best_index = std::max((size_t)0, i - 1);
 
-                RCLCPP_DEBUG_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER), "get_nearest_index_by_downtrack>> Found best_idx: " << best_index<<", points[i].x(): " << points[best_index].x() << ", points[i].y(): " << points[best_index].y() << ", downtrack: "<< downtrack);
-                break;
-            }
+        // Find first point with a downtrack greater than target_downtrack
+        const auto itr = std::find_if(std::cbegin(points), std::cend(points), 
+            [&](const auto & point) { return wm->routeTrackPos(point).downtrack > target_downtrack; });
+
+        // Set best_index to the last point with a downtrack less than target_downtrack without going below index 0
+        if(itr != points.begin()){
+            best_index = std::distance(points.begin(), itr) - 1;
         }
-        RCLCPP_DEBUG_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER), "get_nearest_index_by_downtrack>> Found best_idx: " << best_index<<", points[i].x(): " << points[best_index].x() << ", points[i].y(): " << points[best_index].y());
+        else{
+            best_index = 0;
+        }
 
+        RCLCPP_DEBUG_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER), "get_nearest_index_by_downtrack>> Found best_index: " << best_index<<", points[i].x(): " << points[best_index].x() << ", points[i].y(): " << points[best_index].y());
+        
         return static_cast<int>(best_index);
     }
 
--- a/basic_autonomy/test/test_waypoint_generation.cpp
+++ b/basic_autonomy/test/test_waypoint_generation.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019-2022 LEIDOS.
+ * Copyright (C) 2019-2024 LEIDOS.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -113,7 +113,7 @@ namespace basic_autonomy
 
     }
 
-     TEST(BasicAutonomyTest, constrain_to_time_boundary)
+    TEST(BasicAutonomyTest, constrain_to_time_boundary)
     {
 
         std::vector<waypoint_generation::PointSpeedPair> points;
@@ -201,7 +201,7 @@ namespace basic_autonomy
         ASSERT_EQ(3, waypoint_generation::get_nearest_point_index(points, state));
     }
 
-TEST(BasicAutonomyTest, get_nearest_basic_point_index)
+    TEST(BasicAutonomyTest, get_nearest_basic_point_index)
     {
         std::vector<waypoint_generation::PointSpeedPair> points;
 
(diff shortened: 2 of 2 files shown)


basic_autonomy/src/helper_functions.cpp
```diff
@@ -65,18 +66,21 @@ namespace waypoint_generation
...
-            if(downtrack > target_downtrack){
-                //If value is negative, best index should be index 0
-                best_index = std::max((size_t)0, i - 1);
 
-                RCLCPP_DEBUG_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER), "get_nearest_index_by_downtrack>> Found best_idx: " << best_index<<", points[i].x(): " << points[best_index].x() << ", points[i].y(): " << points[best_index].y() << ", downtrack: "<< downtrack);
-                break;
-            }
+        // Find first point with a downtrack greater than target_downtrack
+        const auto itr = std::find_if(std::cbegin(points), std::cend(points), 
+            [&](const auto & point) { return wm->routeTrackPos(point).downtrack > target_downtrack; });
```
adamlm, line 72: Prefer explicit captures over default ones:
```cpp
[&wm = std::as_const(wm), target_downtrack] (const auto & point) { /* ... */ }
```

basic_autonomy/src/helper_functions.cpp
```diff
@@ -65,18 +66,21 @@ namespace waypoint_generation
...
+
+        // Set best_index to the last point with a downtrack less than target_downtrack without going below index 0
+        if(itr != points.begin()){
+            best_index = std::distance(points.begin(), itr) - 1;
```
adamlm, line 76: Prefer `const` iterators over mutable ones wherever possible. Also prefer using free functions over member functions:
```cpp
if (itr != std::cbegin(points)) {
  best_index = std::distance(std::cbegin(points), itr) - 1;
} else {
  best_index = 0;
}
```

basic_autonomy/src/helper_functions.cpp
```diff
@@ -65,18 +66,21 @@ namespace waypoint_generation
     int get_nearest_index_by_downtrack(const std::vector<lanelet::BasicPoint2d>& points, const carma_wm::WorldModelConstPtr& wm, double target_downtrack)
     {
         size_t best_index = points.size() - 1;
```
adamlm, line 68: `best_index` should be an `int` type to avoid rollover. If `points` is empty, the function should return `-1` to indicate that there is no nearest index.

adamlm (changes requested):

basic_autonomy/src/helper_functions.cpp
JonSmet: Addressed this.

basic_autonomy/src/helper_functions.cpp
JonSmet: Updated to use `const` iterators. And I might be missing something, is there an instance where I'm using a member function in this PR where a free function should be used?

basic_autonomy/src/helper_functions.cpp
JonSmet: Good idea. Updated `best_index` to be an `int` and had an additional slight update to handle an empty `points` vector.

basic_autonomy/src/helper_functions.cpp
```diff
@@ -65,23 +65,28 @@ namespace waypoint_generation
 
     int get_nearest_index_by_downtrack(const std::vector<lanelet::BasicPoint2d>& points, const carma_wm::WorldModelConstPtr& wm, double target_downtrack)
     {
-        size_t best_index = points.size() - 1;
+        int best_index = points.size() - 1;
```
adamlm, line 68: Let's move this down further in the function closer to where it's used.

basic_autonomy/src/helper_functions.cpp
```diff
@@ -65,23 +65,28 @@ namespace waypoint_generation
...
         const auto itr = std::find_if(std::cbegin(points), std::cend(points), 
-            [&](const auto & point) { return wm->routeTrackPos(point).downtrack > target_downtrack; });
+            [&wm = std::as_const(wm), target_downtrack](const auto & point) { return wm->routeTrackPos(point).downtrack > target_downtrack; });
 
-        // Set best_index to the last point with a downtrack less than target_downtrack without going below index 0
-        if(itr != points.begin()){
-            best_index = std::distance(points.begin(), itr) - 1;
+        // Set best_index to the last point with a downtrack less than target_downtrack
+        if(itr != std::cbegin(points)){
+            best_index = std::distance(std::cbegin(points), itr) - 1;
```
adamlm, line 81: Let's change this to use `std::prev`:

```cpp
best_index = std::distance(std::cbegin(points), std::prev(itr));
```

basic_autonomy/src/helper_functions.cpp
```diff
@@ -65,23 +65,28 @@ namespace waypoint_generation
...
         }
         else{
             best_index = 0;
         }
 
         RCLCPP_DEBUG_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER), "get_nearest_index_by_downtrack>> Found best_index: " << best_index<<", points[i].x(): " << points[best_index].x() << ", points[i].y(): " << points[best_index].y());
```
adamlm, line 87: Consider using the `at()` member function instead of `operater[]()`. The former provides bounds checking while the latter does not.

basic_autonomy/test/test_waypoint_generation.cpp
```diff
@@ -231,7 +231,7 @@ namespace basic_autonomy
         ASSERT_EQ(3, waypoint_generation::get_nearest_point_index(points, state));
     }
 
-    TEST(BasicAutonomyTest, get_nearest_index_by_downtrack)
+    TEST(BasicAutonomyTest, get_nearest_index_by_downtrack_with_nonempty_points)
```
adamlm, line 234: Optional note: Google Test's documentation discourages underscores in test names. I'm not sure if there's a technical justification or if it's a style guide decision. Either way, we should refactor all test names to use `PascalCase`.

basic_autonomy/test/test_waypoint_generation.cpp
```diff
@@ -267,7 +267,23 @@ namespace basic_autonomy
...
+        std::shared_ptr<carma_wm::CARMAWorldModel> wm = std::make_shared<carma_wm::CARMAWorldModel>();
+        auto map = carma_wm::test::buildGuidanceTestMap(5.0, 10.0); // Lanelet width 5.0 meters; Lanelet length 10.0 meters
+        wm->setMap(map);
+        carma_wm::test::setRouteByIds({1200, 1201, 1202, 1203}, wm);
+
+        // Create vector of points; this will remain empty
+        std::vector<lanelet::BasicPoint2d> points;
+
+        // Test with arbitrary target_downtrack to verify proper return value due to empty points vector
+        double target_downtrack = 4.0;
```
adamlm, line 283: This can be declared `static constexpr auto`:

```cpp
static constexpr auto target_downtrack{4.0};
```

basic_autonomy/test/test_waypoint_generation.cpp
```diff
@@ -267,7 +267,23 @@ namespace basic_autonomy
...
+    TEST(BasicAutonomyTest, get_nearest_index_by_downtrack_with_empty_points)
+    {
+        // Create CARMA World Model with custom Guidance Test Map
+        std::shared_ptr<carma_wm::CARMAWorldModel> wm = std::make_shared<carma_wm::CARMAWorldModel>();
+        auto map = carma_wm::test::buildGuidanceTestMap(5.0, 10.0); // Lanelet width 5.0 meters; Lanelet length 10.0 meters
+        wm->setMap(map);
+        carma_wm::test::setRouteByIds({1200, 1201, 1202, 1203}, wm);
+
+        // Create vector of points; this will remain empty
+        std::vector<lanelet::BasicPoint2d> points;
```
adamlm, line 280: This can be declared `const`.

adamlm (changes requested):

basic_autonomy/src/helper_functions.cpp
```diff
@@ -65,23 +65,28 @@ namespace waypoint_generation
 
     int get_nearest_index_by_downtrack(const std::vector<lanelet::BasicPoint2d>& points, const carma_wm::WorldModelConstPtr& wm, double target_downtrack)
     {
-        size_t best_index = points.size() - 1;
+        int best_index = points.size() - 1;
+
+        if(points.empty()){
+            RCLCPP_WARN_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER), "Empty points vector received, returning best_index of " << best_index);
+            return best_index;
```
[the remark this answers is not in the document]
JonSmet: Agreed, and addressed

basic_autonomy/src/helper_functions.cpp
JonSmet: Good idea, addressed

basic_autonomy/src/helper_functions.cpp
JonSmet: Good catch, addressed this

basic_autonomy/test/test_waypoint_generation.cpp
JonSmet: Addressed

basic_autonomy/test/test_waypoint_generation.cpp
JonSmet: Addressed

basic_autonomy/test/test_waypoint_generation.cpp
JonSmet: As we discussed already, this is interesting and I wasn't aware of this before. I will make a GitHub issue to allow us to track this fix across our repo.

JonSmet: Created GitHub issue #2332 for this

basic_autonomy/src/helper_functions.cpp
adamlm: In your original implementation, you were using `points.begin()`, which is a `std::vector` member function. The free function alternative is `std::begin()`.

It's a minor detail here since we are specifically using `std::vector`, but if we had a more general function that worked on `std::vector`, `std::array`, or raw arrays, we would have to use `std::begin()` because raw arrays have no `.begin()` member function.

basic_autonomy/src/helper_functions.cpp
```diff
@@ -64,20 +65,28 @@ namespace waypoint_generation
...
+        }
 
-                RCLCPP_DEBUG_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER), "get_nearest_index_by_downtrack>> Found best_idx: " << best_index<<", points[i].x(): " << points[best_index].x() << ", points[i].y(): " << points[best_index].y() << ", downtrack: "<< downtrack);
-                break;
-            }
+        // Find first point with a downtrack greater than target_downtrack
+        const auto itr = std::find_if(std::cbegin(points), std::cend(points), 
+            [&wm = std::as_const(wm), target_downtrack](const auto & point) { return wm->routeTrackPos(point).downtrack > target_downtrack; });
+
+        int best_index = points.size() - 1;
```
adamlm, line 77: Optional: Consider using `std::size(points) - 1` instead of the member function. This aligns with the free function use for `std::cbegin()` and `std::cend()`.

adamlm (approved):

basic_autonomy/src/helper_functions.cpp
JonSmet: Good catch, just updated

adamlm (approved):

Issue #2346, opened 2024-04-02
Yield_plugin is ignoring static objects on its route and don't yield


MishkaMN:
### Summary

In this PR, https://github.com/usdot-fhwa-stol/carma-platform/pull/2313/
we included more efficient way to handle obstacles on the road. One of which was a special treatment for static objects. 
The idea behind of that improvement was that if an object is stopping, there is no need to look at every one of its predicted states which can be costly. 

However, the current algorithm for collision check is ignoring the time aspect of a stopped object. Because the static objects' future states are ignored, it is always only checking 1 position at 1 timestamp of the run, which is the obstacles' position/time at the beginning of the run. In other words, since the collision check "time aligns" vehicle's trajectory to the obstacles' each timestamps in its trajectory to check collision, it was always aligning the time to the beginning of the run and NOT detect a collision while ignoring rest of the timestamps which could generate collision. 

There are two places it ignores future states:
- https://github.com/usdot-fhwa-stol/carma-platform/blob/release/lavida/yield_plugin/src/yield_plugin.cpp#L578
Here, we shouldn't ignore even if it is 0 speed.
- https://github.com/usdot-fhwa-stol/carma-platform/blob/release/lavida/yield_plugin/src/yield_plugin.cpp#L555
Here, iteration_stride calculation also becomes huge (10000+) and only runs 1 loop due to zero speed of traj2. Therefore iteration_stride also needs to be capped or need different way to derive it.

The solution to above problem should keep in mind not to increase the processing time of the collision check dramatically as its intended fix was to reduce the processing time.

### Commit

5a0244eb48f0e65d28300d302955ef09cabe5371

### Expected Behavior

Yield plugin should stop to static objects

### Actual Behavior

Yield is ignoring static objects

### Steps to Reproduce the Actual Behavior

Put a static object on the road and have the vehicle run through it

PR #2347, opened 2024-04-03
Fix yield_plugin ignoring static objects

MishkaMN:
# PR Details
## Description
This PR resolves this issue: https://github.com/usdot-fhwa-stol/carma-platform/issues/2346
Please see the issue link for the problem. 
This PR fixes the issue by modifying the code in two ways:
- Nested loop for collision check checks rest of the possible combinations even if traj2 has 0m/s speed.
- Iteration_stride previously become 10000+ huge number due to traj2 0m/s speed, so we needed different way to calculate it. It is not derived from accounting both trajectories' speeds and `intervehicle_collision_check_distance_m`

These changes will not affect original issue of low performance where the yield_plugin was taking too long to process obstacles because the main improvement came from parallel processing. And if the vehicle can process efficiently a moving object, then it should take similar time for static object and not exceed the max trajectory generation time. 

## Related GitHub Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/2346

## Related Jira Key
https://usdot-carma.atlassian.net/browse/CDAR-881

## Motivation and Context
VRU verification testing
VOICES demo 

## How Has This Been Tested?
Sim PC 1

## Types of changes

## Checklist:

### DIFF
--- a/yield_plugin/src/yield_plugin.cpp
+++ b/yield_plugin/src/yield_plugin.cpp
@@ -65,6 +65,11 @@ namespace yield_plugin
     return fabs(get_trajectory_end_time(trajectory) - get_trajectory_start_time(trajectory));
   }
 
+  double get_predicted_steps_duration(const std::vector<carma_perception_msgs::msg::PredictedState>& trajectory)
+  {
+    return (rclcpp::Time(trajectory.back().header.stamp) - rclcpp::Time(trajectory.front().header.stamp)).seconds();
+  }
+
   std::vector<std::pair<int, lanelet::BasicPoint2d>> YieldPlugin::detect_trajectories_intersection(std::vector<lanelet::BasicPoint2d> self_trajectory, std::vector<lanelet::BasicPoint2d> incoming_trajectory) const
   {
     std::vector<std::pair<int, lanelet::BasicPoint2d>> intersection_points;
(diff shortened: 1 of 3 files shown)


yield_plugin/src/yield_plugin.cpp
```diff
@@ -258,26 +263,35 @@ namespace yield_plugin
     carma_planning_msgs::msg::TrajectoryPlan original_trajectory = req->initial_trajectory_plan;
     carma_planning_msgs::msg::TrajectoryPlan yield_trajectory;
 
+
+    double initial_velocity = req->vehicle_state.longitudinal_vel;
+    // If vehicle_state is stopped, non-zero velocity from the trajectory
+    // should be used. Otherwise, vehicle will not move.
+    if (initial_velocity < EPSILON)
+    {
+      initial_velocity = original_trajectory.initial_longitudinal_velocity;
```
MishkaMN, line 272: This is a small unrelated bug I found if obstacle is placed near the starting point of the vehicle, the vehicle would not move although it has plenty runway until stopping before the obstacle. This is because `vehicle_state.longitudinal_vel` is 0 when stopped and trajectory generated from yield starts from that speed. Inlanecruising_plugin also had similar problem, which we resolved back in the day, by forcing the first commanded speed to be non-zero, which is where the `initial_longitudinal_velocity` partially came from.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -65,6 +65,11 @@ namespace yield_plugin
     return fabs(get_trajectory_end_time(trajectory) - get_trajectory_start_time(trajectory));
   }
 
+  double get_predicted_steps_duration(const std::vector<carma_perception_msgs::msg::PredictedState>& trajectory)
```
adamlm, line 68: Optional: Consider renaming to `get_trajectory_duration()` to be more general and more self-explanatory. `PredictedStates` (or `_predicted_steps_`) seems to be an implementation detail.

adamlm (approved):

PR #2390, opened 2024-05-30
add architecture docs for CP

MishkaMN:
# PR Details
## Description
Adding new docs generated from the final report to the repo itself so it doesn't get lost

## Related GitHub Issue
https://github.com/usdot-fhwa-stol/multiple_object_tracking/pull/152

## Related Jira Key
https://usdot-carma.atlassian.net/browse/CDAR-887

## Motivation and Context
CDAR VRU use case final report first draft originally included this document, but it is not documented anywhere.

## How Has This Been Tested?
NA

## Types of changes

## Checklist:


carma_cooperative_perception/docs/stack_design.md
paulbourelly999, line 1: Your changes look fine, but the final page makes little since to me. It starts with assumptions that are hard to understand without very specific context and goes directly into diagrams. How is documentation different from what Adam had in the multiple object tracking library. Also why do we not put this there ( in the multiple object tracking library)?

paulbourelly999 (changes requested):
Your changes look fine, but the final page makes little since to me. It starts with assumptions that are hard to understand without very specific context and goes directly into diagrams. How is documentation different from what Adam had in the multiple object tracking library. Also why do we not put this there ( in the multiple object tracking library)?

carma_cooperative_perception/docs/stack_design.md
MishkaMN: I can try giving more context. 
The current diagram I uploaded specifically shows what is implemented in this stack. MOT is designed as a library of sub functionalities and can be used differently than drawn in this diagram, so I decided to include it here.

paulbourelly999: Ok It that is the case I think we should have more context in this README.
" For example (this is documentation of our cooperative perception stack. It uses internal detection and external detection (SDSM) to perceive objects/obstacles in its surroundings. The first diagram show X ... explaining X "

I know that is probably not a great example but as it sits without any context text this is very hard to decipher.

paulbourelly999 (changes requested):
Please add more context in the README

paulbourelly999 (approved):

Issue #2398, opened 2024-06-05
Motion Computation CTRV prediction is wrong


MishkaMN:
### Summary

During testing of VOICES with live vehicle driving before carma vehicle, carma vehicle yields based on the CTRV model of the observed vehicle. However, it was noticed that sometimes the prediction would go backwards:

![image](https://github.com/usdot-fhwa-stol/carma-platform/assets/20613282/bc4abd27-5622-49d4-8035-987c46c7c6a4)

Where a live vehicle is driving north on the right lane and carma is on the left driving north as well. Here, the red line indicates the predicted trajectory of vehicle which incorrectly shows south. This prediction switches back and forth a lot, which significantly decreases yielding capability of carma.

**Background on the CTRV Model**
The CTRV model assumes that the object moves with a constant velocity and a constant yaw rate (turn rate). The equations of motion for the object's position and orientation are derived from these assumptions.

**Original (Incorrect) Equations**

```
next_state.x = v_w * sin(wT + state.yaw) - v_w * sin_yaw + state.x;
next_state.y = -v_w * cos(wT + state.yaw) + v_w * sin_yaw + state.y;
```
These equations improperly handle the trigonometric relationships for the new position, resulting in incorrect predictions.

**Corrected Equations**
```
next_state.x = state.x + v_w * (sin(state.yaw + wT) - sin_yaw);
next_state.y = state.y + v_w * (cos_yaw - cos(state.yaw + wT));
```

Actual equation derivations:
https://www.mathworks.com/help/fusion/ug/motion-model-state-and-process-noise.html

### Version

4.5.0 (Current)

### Expected Behavior

See above

### Actual Behavior

See above

### Steps to Reproduce the Actual Behavior

it was discovered in simulation.
1. started carma in TFHRC CARLA
2. nearby spawn a live vehicle driving in CARLA
3. manually drive the vehicle and see motion prediction red lines on rviz

MishkaMN:
Resolved with above PR

PR #2399, opened 2024-06-05
Make motion models in motion comp configurable

MishkaMN:
# PR Details
## Description
Utility function to make motion models in motion computation configurable to easily test

## Related GitHub Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/2398

## Related Jira Key
https://usdot-carma.atlassian.net/browse/CAR-6057

## Motivation and Context
See above link

## How Has This Been Tested?
sim pc tested

## Types of changes

## Checklist:

### DIFF
--- a/motion_computation/config/parameters.yaml
+++ b/motion_computation/config/parameters.yaml
@@ -33,8 +33,8 @@ enable_mobility_path_processing: false
 enable_sensor_processing: false
 
 # Boolean: True if CTRV motion model should be used for the object type. False, if CV should be used:
-enable_ctrv_for_unknown_obj: false
-enable_ctrv_for_motorcycle_obj: false
-enable_ctrv_for_small_vehicle_obj: false
-enable_ctrv_for_large_vehicle_obj: false
-enable_ctrv_for_pedestrian_obj: true
+enable_ctrv_for_unknown_obj: true
+enable_ctrv_for_motorcycle_obj: true
+enable_ctrv_for_small_vehicle_obj: true
+enable_ctrv_for_large_vehicle_obj: true
+enable_ctrv_for_pedestrian_obj: false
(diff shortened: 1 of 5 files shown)


motion_computation/src/motion_computation_worker.cpp
```diff
@@ -35,6 +35,10 @@ void MotionComputationWorker::predictionLogic(
   carma_perception_msgs::msg::ExternalObjectList sensor_list;
   sensor_list.header = obj_list->header;
 
+  if (!obj_list || obj_list->objects.empty())
+  {
+    return;
+  }
```
JonSmet, line 41: Nit: Could we add a line of whitespace after this code block before the for-loop just to improve readability at a quick glance?

motion_computation/config/parameters.yaml
```diff
@@ -31,3 +31,10 @@ enable_mobility_path_processing: false
 # Boolean: If true then ExternalObjects generated from sensor data will be processed.
 #          If other object sources are enabled, they will be synchronized but no fusion will occur (objects may be duplicated)
 enable_sensor_processing: false
+
+# Boolean: True if CTRV motion model should be used for the object type. False, if CV should be used:
+enable_ctrv_for_unknown_obj: true
```
JonSmet, line 36: In `MotionComputationWorker::predictionLogic()`, we used to use the CV model for unknown objects. Is there a reason we are changing it to CTRV by default? Want to understand if there is a reason, and if it may change any previous Vulnerable Road User use case behavior.

JonSmet (changes requested):

motion_computation/config/parameters.yaml
MishkaMN: Originally, it was actually a little confusing because there were two different places unknown could be used. 1) If it was obj.UNKNOWN, it used CTRV. 2) but when enum didn't match anything such as UNKNOWN or PEDESTRIAN VEHICLE etc, it forced the type to UNKNOWN and used CV. So I just picked CTRV because that looked like what most instances would have executed.

motion_computation/src/motion_computation_worker.cpp
MishkaMN: Added

JonSmet (approved):

PR #2408, opened 2024-06-14
Add logic to carma script to account for env var in carma-config

MishkaMN:
# PR Details
## Description

This PR fixes the carma script to be able to run from any directory without relying on .env file to exist in the same directory. 
Depends on this PR: https://github.com/usdot-fhwa-stol/carma-config/pull/352

The proposed change on a high level:
- Bake the `.env` file into the `carma-config` container. (previously it didn't exist inside `carma config edit` for example).
- Whenever `carma *` script is run, **parse** all variables inside the `.env` file from the carma-config image set through volume
- use that parsed variables to **sed change** all instances found in the `docker-compose.yml` to run the individual services. This should work for any environment variables specified, not just `DOCKER_TAG` or `DOCKER_ORG` since some services require different org/tagging such as `OPS` or `v2xhub` etc.

Resulting behavior:
- Since the change is made by sed command to get hardcoded image names as before, the local `.env` file in host dir where `carma` is called will no longer work unless the `.env` file in carma-config was NOT able to replace. Then the local .env file will takeover using docker's native functionality and the variables in the docker-compose.yml wil be replaced. This is desirable for simulation runs for example where multiple variables such as ROUTE_FILE are specified in .env file 
- If the user wants to change the image name/tags, they can simply modify the .env file and rebuild or `carma config edit` and modify the `/opt/carma/vehicle/config/.env` file and restart. The change should persist as how `carma config edit` normally works. Or even hardcoding as before in the docker-compose.yml work as well.

## Related GitHub Issue
NA

## Related Jira Key
https://usdot-carma.atlassian.net/browse/ARC-82

## Motivation and Context
During 4.5.0 release, we modifed the way DOCKER_ORG and DOCKER_TAG are specified by centralizing them into a single .env file in each carma-config folder for docker-compose.yml to load. However, currnently, when a carma-config image is built by our GitHub Actions CI processes, the resulting image still contains the {DOCKER_ORG} and {DOCKER_TAG} environment variable placeholders in the docker-compose.yml and docker-compose-background.yml files. This breaks the carma config install <carma-config image> functionality of the carma shell script. The carma start all functionality breaks with this as well. 

Workaround: If .env file exists in the same directory from where `carma` is called, then everything works fine. However, this is not a desired functionality as we want to call carma from anywhere.

## How Has This Been Tested?
locally in VM. Results of each commands working:

file: 
[carma command results.txt](https://github.com/user-attachments/files/15842992/carma.command.results.txt)
in colored:
image
image

## Types of changes

## Checklist:

### DIFF
--- a/engineering_tools/carma
+++ b/engineering_tools/carma
@@ -360,7 +360,7 @@ carma__start() {
     fi
 
     # Extract environment variables from the carma-config container
-    echo "Extracting environment variables from carma-config..." > /dev/tty  # Make sure to print to terminal instead of returning
+    echo "Extracting environment variables from carma-config..."
     local env_vars=$(docker run --rm --volumes-from carma-config:ro --entrypoint sh busybox:latest -c \
         'cat /opt/carma/vehicle/config/.env' | xargs)
 
@@ -393,7 +393,7 @@ carma__stop() {
     fi
 
     # Extract environment variables from the carma-config container
-    echo "Extracting environment variables from carma-config..." > /dev/tty  # Make sure to print to terminal instead of returning
+    echo "Extracting environment variables from carma-config..."
     local env_vars=$(docker run --rm --volumes-from carma-config:ro --entrypoint sh busybox:latest -c \
         'cat /opt/carma/vehicle/config/.env' | xargs)
 
--- a/engineering_tools/carma
+++ b/engineering_tools/carma
@@ -94,6 +94,8 @@ __get_image_from_config() {
             local TARGET_IMAGE=$IMAGE_BACKUP
             echoerr "Could not identify usable image from config so defaulting to image: $TARGET_IMAGE"
         fi
+
+        rm /tmp/.env
     else
         echoerr "No config detected or target image specified so using default: $TARGET_IMAGE"
     fi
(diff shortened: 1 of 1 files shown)


engineering_tools/carma
```diff
@@ -80,11 +80,14 @@ __get_image_from_config() {
...
-            "sed -n -e '/image:\s*\($SUPPORTED_ORANIZATIONS\)\/\($SUPPORTED_IMAGES\)/ s/.*\image: *//p' /opt/carma/vehicle/config/docker-compose.yml | grep -m 1 '.*'")
-            
+        # Extract the .env file content from the carma-config volume
+        docker run --rm --volumes-from carma-config:ro busybox:latest cat /opt/carma/vehicle/config/.env > /tmp/.env
+
+        # Extract DOCKER_ORG and DOCKER_TAG from the .env file
+        source /tmp/.env
+
+        # Construct the image name using the values from the .env file
+        TARGET_IMAGE="${DOCKER_ORG}/carma-base:${DOCKER_TAG}"
```
MishkaMN, line 90: TODO: for @MishkaMN this hardcode needs to change

MishkaMN:
After writing a whole lot of helper functions and sed operations. I realized I could have just done the same thing with a simple method... 🤦‍♂️ Here: https://github.com/usdot-fhwa-stol/carma-platform/pull/2409
So probably we don't need this PR...

PR #2409, opened 2024-06-14
Add logic to carma script to account for env var in carma-config

MishkaMN:
# PR Details
## Description

This PR fixes the carma script to be able to run from any directory without relying on .env file to exist in the same directory. 
Depends on this PR: https://github.com/usdot-fhwa-stol/carma-config/pull/352

**The proposed change on a high level:**
- Bake the `.env` file into the `carma-config` container. Since `.env` file specifies carma version now, it makes sense to include it in the config, which is intended to track carma version.
- Use COMPOSE_ENV_FILES functionality under the hood from docker which allows the user to specify multiple .env files (NOTE: needs docker-compose 2.23.0+ version). `carma` script now modifies that variable as below (highest to lowest precedence):
    - 1 user defined shell COMPOSE_ENV_FILES values
    - 2 local .env file where `carma` is called
    - 3 .env file specified in `carma config` currently set
This way, `carma` script always has default values from carma-config wherever it's called. Additionally, user can still use the variable as docker intended. 
NOTE: docker actually uses local `.env` file only as the last resort such as if no `.env` file is specified, so it doesn't always use. Currently I implemented such that it uses .env as 2nd to last precedence as long as it exists, as it makes sense in my head. For anyone who is interested, here is the docker [env file precedence](https://docs.docker.com/compose/environment-variables/envvars-precedence/). 
- Fixed previously broken feature of using `extra_args` in `carma start all extra_args` as input to `docker-compose up` command. Now it works as intended where the user can input extra flags `--env-file` or `--progress` etc. 

**Resulting behavior:**
- **NEW Behavior**: docker-compose needed to be updated to latest (or more than 2.23.0) which is now 2.27.1
- **NEW Behavior**: If multiple `.env` files are specified, such as local `.env`, `.env` files in COMPOSE_ENV_FILES etc, all of them are loaded including `.env` from `carma-config`. If overlapping values exist, above precedence is used to overwrite.
NOTE: if `--env-file` argument is used in `carma start all --env-file /path/to/env`, docker ignores COMPOSE_ENV_FILES. It is still same behavior as that.
- **OLD Behavior**: `carma config edit` uses `carma-base` that tries to use same version as specified in its `.env` inside. If fails, just uses most recent `carma-base`. In the end, it shouldn't matter which version of `carma-base` is used as it is only for editing `volume`
- **Old behavior**: `carma config install <image>` tries to pull down images specified in `.env` of `<config-image>`. It doesn't make sense to use any other `.env` here as the user is trying to download specific version in `<config-image>`
- **OLD Behavior + fixes**: `carma exec` 
     - 1 first tries to open `carma-platform` or `carma-messenger` using version specified inside current `.env` in `carma-config`. 
     - 2 If it fails, try to use most recent `carma-platform` or `carma-messenger` images. 
     NOTE: Here, previously the script tried to use `usdotfhwastol/carma-base:latest` instead which often doesn't exist and script failed. This was undesirable such as when custom image like `local/carma-platform:latest` is used, then the script used to fail.
     - 3 if that also fails, try to use most recent `carma-base` image. 
     - 4 if that fails, return by logging error
     NOTE: notice here that `carma exec` is uses only `.env` file from carma-config to open the container, which how it worked before. Using any other `.env` file is overengineering and not needed. `carma exec` has extra argument of `-i` to specify specific image anyways. 
     
- **OLD Behavior**: In most cases, if the user wants to change the image name/tags (or any other variable), they can simply modify the .env file and rebuild `<config-image>` or `carma config edit` and modify the `/opt/carma/vehicle/config/.env` file and restart. Or even hardcoding as before in the docker-compose.yml work as well without worrying about any `.env` files. However, they can now use different `.env` files now as mentioned before, which has higher precedence than the one in `<carma-config>` 
-  **OLD Behavior**: `carma` script is still backwards compatible to `<config-image>` before (and including) carma-system-4.5.0, which didn't have `.env` files. I accomplished it by keeping old relevant logics in `__get_image_from_config` (most new ones are backwards compatible anyways). 
NOTE: I suppressed some potential error messages that would have occured due to older versions as they are not errors (such as not finding `.env` file in `<config-image>` and there will be other actual error messages if there was an actual issue). 

## Related GitHub Issue
NA

## Related Jira Key
https://usdot-carma.atlassian.net/browse/ARC-82

## Motivation and Context
During 4.5.0 release, we modifed the way DOCKER_ORG and DOCKER_TAG are specified by centralizing them into a single .env file in each carma-config folder for docker-compose.yml to load. However, currently, when a carma-config image is built by our GitHub Actions CI processes, the resulting image still contains the {DOCKER_ORG} and {DOCKER_TAG} environment variable placeholders in the docker-compose.yml and docker-compose-background.yml files. This breaks all of the `carma *` functionality of the carma shell script. 

Workaround: If .env file exists in the same directory from where `carma` is called, then everything works fine. However, this is not a desired functionality as we want to call carma from anywhere.

## How Has This Been Tested?
locally in VM. 

## Types of changes

## Checklist:

I also updated [carma installation guide here](https://usdot-carma.atlassian.net/wiki/spaces/CRMPLT/pages/486178841/Setup+CARMA+Platform+Prerequisites) to use the most recent version of docker-compose instead of hardcoding it

### DIFF
--- a/engineering_tools/carma
+++ b/engineering_tools/carma
@@ -55,7 +55,7 @@ echoerr() { echo "$@" 1>&2; }
 # If the config is not set or an image cannot be found the method will return a carma-base image as the default.
 # The user can specify optional organizations or images as sed compatable regex lists.
 # For example calling
-#  $(__get_image_from_config 'carma-platform:\|carma-messenger:' 'usdotfhwastol\|usdotfhwastoldev\|usdotfhwastolcandidate')
+#  $(__get_image_from_config 'carma-platform\|carma-messenger' 'usdotfhwastol\|usdotfhwastoldev\|usdotfhwastolcandidate')
 #
 #  If no parameters are provided the defaults will be used
 ##
@@ -63,7 +63,7 @@ __get_image_from_config() {
     # Default image is the most recent carma-base image
     local DEFAULT_IMAGE=$(docker images --format '{{.Repository}}:{{.Tag}} {{.CreatedAt}}' | grep -E 'carma-base' | grep -v '<none>' | sort -k 2 -r | head -n 1 | awk '{print $1}')
     local SUPPORTED_ORGANIZATIONS="usdotfhwastol\|usdotfhwastoldev\|usdotfhwastolcandidate"
-    local SUPPORTED_IMAGES="carma-platform:\|carma-messenger:"
+    local SUPPORTED_IMAGES="carma-platform\|carma-messenger"
 
     if [ -n "$1" ]; then
         SUPPORTED_IMAGES="$1"
@@ -81,8 +81,21 @@ __get_image_from_config() {
         # Extract DOCKER_ORG and DOCKER_TAG from the .env file
         source /tmp/.env
 
-        # Construct the image name using the values from the .env file
-        local TARGET_IMAGE="${DOCKER_ORG}/carma-base:${DOCKER_TAG}"
+        # Construct the image name using the values from the .env file and given SUPPORTED_IMAGES options
+
+        # Split SUPPORTED_IMAGES into an array
+        local IFS='\|'
+        read -ra IMAGE_ARRAY <<< "$SUPPORTED_IMAGES"
+
+        for IMAGE in "${IMAGE_ARRAY[@]}"; do
+            TARGET_IMAGE="${DOCKER_ORG}/${IMAGE}:${DOCKER_TAG}"
+
+            # Verify target image exists
+            if [ -n "$(docker images -q $TARGET_IMAGE 2> /dev/null)" ]; then
+                IMAGE_FOUND=1
+                break
+            fi
+        done
 
         # Check if TARGET_IMAGE exists locally
         if ! docker images --format '{{.Repository}}:{{.Tag}}' | grep -q "^${TARGET_IMAGE}$"; then
@@ -159,7 +172,7 @@ carma-config__edit() {
 
     echo "Opening shell inside carma-config container with read/write privileges..."
 
-    local carma_base=$(__get_image_from_config carma-base:)
+    local carma_base=$(__get_image_from_config carma-base)
     if [[ -z $carma_base ]]; then
         __pull_newest_carma_base
         carma_base=$(__get_most_recent_carma_base)
@@ -449,7 +462,7 @@ carma__exec() {
         shift
     # If the image was not specified try to use what was in the carma config
     else
-        local TARGET_IMAGE=$(__get_image_from_config 'carma-platform:\|carma-messenger:')
+        local TARGET_IMAGE=$(__get_image_from_config 'carma-platform\|carma-messenger')
     fi
 
     # Verify target image exists
--- a/engineering_tools/carma
+++ b/engineering_tools/carma
@@ -190,10 +190,18 @@ carma-config__set() {
 }
 
 carma__attach() {
+    # Extract the .env file content from the carma-config volume
+    docker run --rm --volumes-from carma-config:ro busybox:latest cat /opt/carma/vehicle/config/.env > /tmp/.env 2>/dev/null  # suppress misleading error message that will occur for versions before 4.5.0
+
+    # Extract DOCKER_ORG and DOCKER_TAG (and others) from the .env file
+    set -o allexport
+    source /tmp/.env 2>/dev/null # suppress misleading error message that will occur for versions before 4.5.0
+    set +o allexport
+
     local CARMA_DOCKER_FILE="`docker run --rm --volumes-from carma-config:ro --entrypoint sh busybox:latest -c 'cat /opt/carma/vehicle/config/docker-compose.yml'`"
 
     echo "Attaching to CARMA container STDOUT..."
-    echo "$CARMA_DOCKER_FILE" | docker-compose -p carma -f - logs --follow --tail=10
+    echo "$CARMA_DOCKER_FILE" | docker-compose -p carma logs -f - logs --follow --tail=10
 }
 
 carma-config__edit() {
--- a/engineering_tools/carma
+++ b/engineering_tools/carma
@@ -99,8 +99,22 @@ __get_image_from_config() {
         # Check if TARGET_IMAGE exists locally
         if [ $IMAGE_FOUND -eq 0 ]; then
             echoerr "TARGET_IMAGE with DOCKER_ORG=${DOCKER_ORG} DOCKER_TAG=${DOCKER_TAG} does not exist locally for these images: ${SUPPORTED_IMAGES},
-                defaulting to most recent carma-base image: $DEFAULT_IMAGE"
-            TARGET_IMAGE=$DEFAULT_IMAGE
+                trying to get most recent images from these: ${SUPPORTED_IMAGES}"
+
+            for IMAGE in "${IMAGE_ARRAY[@]}"; do
+                TARGET_IMAGE=$(docker images --format '{{.Repository}}:{{.Tag}} {{.CreatedAt}}' | grep -E "$IMAGE" | grep -v '<none>' | sort -k 2 -r | head -n 1 | awk '{print $1}')
+
+                # Verify target image exists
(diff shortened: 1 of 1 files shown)


MishkaMN:
TODO:
- If the user made a typo accidentally in the `.env` file. They cannot go back into fix it since this new `carma` script don't have backup feature for that
- We should probably support IP configs while we are at it because during simulation, all platforms run in different IPs and not on localhost. Therefore the `carma exec` end up not being able to communicate.

MishkaMN:
- If no valid image is found using the `.env` use most recent carma-base image available (which includes carma-base:latest previously set as default)
- if no carma-base image is available, then log error and return which will rarely happen
- Image below shows series of commands 1) if image is not available, use most recent image 2) use different image names such as dev and stol 3) make sure it is working

image

willjohnsonk:
@MishkaMN https://github.com/usdot-fhwa-stol/carma-platform/blob/623f636dd6b2ccbacc6b9cd97546de9f1f92d7e9/engineering_tools/carma
OBE'd but updated carma script for dual compute. Should be mostly isolated from dual compute functionality

MishkaMN:
Discussed offline with @paulbourelly999 and @adev4a to include an option for the user to specify the env file directory for the docker-compose

engineering_tools/carma
```diff
@@ -50,54 +50,85 @@ __pull_newest_carma_base() {
...
-        local TARGET_IMAGE=$(docker run --rm --volumes-from carma-config:ro --entrypoint sh busybox:latest -c \
-            "sed -n -e '/image:\s*\($SUPPORTED_ORANIZATIONS\)\/\($SUPPORTED_IMAGES\)/ s/.*\image: *//p' /opt/carma/vehicle/config/docker-compose.yml | grep -m 1 '.*'")
-            
-
-        # If the sed resulted in an empty string then reset the tag
-        if [ -z "$TARGET_IMAGE" ]; then
-            local TARGET_IMAGE=$IMAGE_BACKUP
-            echoerr "Could not identify usable image from config so defaulting to image: $TARGET_IMAGE"
+        local IMAGE_FOUND=0
+        # Extract the .env file content from the carma-config volume
```
adev4a, line 77: Already mentioned in your comment, but just adding a note, this will use the local path to the .env file if specified

engineering_tools/carma
```diff
@@ -249,6 +280,14 @@ carma-config__install() {
     fi
     docker run --name carma-config-tmp $IMAGE_NAME
 
+    # Extract the .env file content from the carma-config volume
```
adev4a, line 283: Also here, should allow passing the .env file as well

engineering_tools/carma
```diff
@@ -249,6 +280,14 @@ carma-config__install() {
...
+    docker run --rm --volumes-from carma-config:ro busybox:latest cat /opt/carma/vehicle/config/.env > /tmp/.env
+
+    # Extract DOCKER_ORG and DOCKER_TAG (and others) from the .env file
```
adev4a, line 286: What happens if an .env file doesn't exist at this step?

engineering_tools/carma
```diff
@@ -292,12 +331,20 @@ carma__start() {
         exit 1
     fi
 
+    # Extract the .env file content from the carma-config volume
```
adev4a, line 334: Similar note, should allow passing path to .env file

engineering_tools/carma
```diff
@@ -292,12 +331,20 @@ carma__start() {
...
+    docker run --rm --volumes-from carma-config:ro busybox:latest cat /opt/carma/vehicle/config/.env > /tmp/.env
+
+    # Extract DOCKER_ORG and DOCKER_TAG (and others) from the .env file
```
adev4a, line 337: What happens if there is no .env file found at this step?

engineering_tools/carma
MishkaMN: Although we discussed this offline, I had a second thought where I realized that we don't need to use local .env for this function. This function `__get_image_from_config` is used in only 3 places:
- carma_exec: which can reliably use carma-config's `.env` most of the time since that would be same behavior as when we didnt use `.env`. if they want to use specific image, `-i` option is available. Making it use local `.env` seemed to be was not straightforward and over-engineering. 
- carma config edit: doesn't have to be any specific image version anyways.
- carma install: we shouldn't use local `.env` since the function is intending to install version in the config. 

So I decided to only keep config's .env here

engineering_tools/carma
MishkaMN: here we actually shouldn't use passing env since the carma install function is intending to install version in the config. If they need to install different version, they have to download different config or download images individually.

engineering_tools/carma
MishkaMN: This should allow extra passing of arguments now. extra arguments are same arguments as `docker-compose up` where not only `--env-file` but also others

engineering_tools/carma
```diff
@@ -285,27 +343,60 @@ carma__fix_bag() {
...
 }
+
+__get_compose_env_files() {
+    # If variable is not set, modify to get default values from carma-config
+    RETURN_COMPOSE_ENV_FILES="/tmp/.env"
+
+    # If .env file exists in current directory, it takes higher precedence over carma-config .env file
+    if [ -f ".env" ]; then
+        echoerr "NOTE: .env file detected in current directory, considering its values for current carma session."
+        RETURN_COMPOSE_ENV_FILES="$RETURN_COMPOSE_ENV_FILES,.env"
```
MishkaMN, line 354: @adev4a I currently coded to consider current directory `.env` as long as it exists. This is the only  change from how docker works where it uses it if neither of COMPOSE_ENV_FILES nor `--env-file` is set. Since my finctionality uses COMPOSE_ENV_FILES, it is always technically set. But we still want the local `.env` to overwrite it, so I had to include it this way. lmk if it's confusing, but I think this is really minor and benign side-effect since carma users don't usually use COMPOSE_ENV_FILES files anyways.

adev4a: Yup, it makes sense to me to have the local .env file take precedence over the carma-config .env file. Thanks for including the comments in the script as well

engineering_tools/carma
```diff
@@ -50,54 +50,101 @@ __pull_newest_carma_base() {
...
+    if [ -n "$1" ]; then
+        SUPPORTED_IMAGES="$1"
     fi
 
     # Check if carma-config is set. If so, use what was in the carma config
     if docker container inspect carma-config > /dev/null 2>&1; then
 
-        local IMAGE_BACKUP=$TARGET_IMAGE
-
+        # Before and including carma-system-4.5.0, image can be directly grabbed from config docker-compose using sed
```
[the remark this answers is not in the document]
MishkaMN: added

engineering_tools/carma
```diff
@@ -50,54 +50,104 @@ __pull_newest_carma_base() {
...
+# If none is possible from all above, returns error message.
+# The user can specify optional images as sed compatable regex lists.
+# For example calling
+#  $(__get_image_from_config 'carma-platform\|carma-messenger')
+#
 #  If no parameters are provided the defaults will be used
 ##
 __get_image_from_config() {
-    local TARGET_IMAGE="usdotfhwastol/carma-base:latest"
+    # Default image is the most recent carma-base image
```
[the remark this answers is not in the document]
MishkaMN: time of creation, not the latest flag

engineering_tools/carma
```diff
@@ -50,54 +50,104 @@ __pull_newest_carma_base() {
...
+# combinations from SUPPORTED_IMAGES that exist locally.
+# If the config is not set or an image cannot be found the method will return most recent
+# SUPPORTED_IMAGES image as the default.
+# If that fails, returns the most recent carma-base image.
+# If none is possible from all above, returns error message.
+# The user can specify optional images as sed compatable regex lists.
+# For example calling
+#  $(__get_image_from_config 'carma-platform\|carma-messenger')
+#
 #  If no parameters are provided the defaults will be used
```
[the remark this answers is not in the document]
MishkaMN: removed

adev4a (approved):

Issue #2385, opened 2024-05-16
Arbitrator is getting delayed due to failed calls 


MishkaMN:
### Summary

During VRU validation testing, carma vehicle stopped generating trajectory right in the middle of the intersection. Upon investigation, the location of the vehicle was not the main issue and just happened to shut down there. 

As you can see from below, 4.5.0 included new functionality in arbitrator to retry maneuver call back if failed as sometimes it can randomly fail due to ROS2 middleware issue. The retry however has timeout of 500ms and tries 10 times, which is 5 seconds. This results in arbitrator not being able to generate maneuver for entire 5 seconds despite all but 1 (which is optional) plugin successfully generated the high priority maneuver.  Please observe the timestamps below:

```
guidance/plugins/approaching_emergency_vehicle_plugin/plan_maneuvers, retrying, attempt no: 0
1713294495.8355138 [carma_component_container_mt-25] 1713294495.835187321 | WARN | multiplex_service_call_for_capability:86 | Following client timed out: /guidance/plugins/approaching_emergency_vehicle_plugin/plan_maneuvers, retrying, attempt no: 1
1713294496.3357017 [carma_component_container_mt-25] 1713294496.335434161 | WARN | multiplex_service_call_for_capability:86 | Following client timed out: /guidance/plugins/approaching_emergency_vehicle_plugin/plan_maneuvers, retrying, attempt no: 2
1713294496.8359416 [carma_component_container_mt-25] 1713294496.835693847 | WARN | multiplex_service_call_for_capability:86 | Following client timed out: /guidance/plugins/approaching_emergency_vehicle_plugin/plan_maneuvers, retrying, attempt no: 3
1713294497.3361895 [carma_component_container_mt-25] 1713294497.335941837 | WARN | multiplex_service_call_for_capability:86 | Following client timed out: /guidance/plugins/approaching_emergency_vehicle_plugin/plan_maneuvers, retrying, attempt no: 4
1713294497.8365588 [carma_component_container_mt-25] 1713294497.836221242 | WARN | multiplex_service_call_for_capability:86 | Following client timed out: /guidance/plugins/approaching_emergency_vehicle_plugin/plan_maneuvers, retrying, attempt no: 5
[...]

Issue #2245, opened 2023-12-23
arbitrator shutting down without any log


MishkaMN:
### Summary

Arbitrator is getting RCLError a lot, which was getting caught by the try catch.
However, at some point, right after catching the error the arbitrator shuts down without any log.

> 1703287546.6583834 [carma_component_container_mt-24] [37;1m2023-12-22 23:25:46.658 [31;1m[[37;1mPARTICIPANT[31;1m Error] [37mType : carma_planning_msgs::srv::dds_::PlanManeuvers_Response_ Not Registered[34;1m -> Function [36mcreateSubscriber[m
1703287546.6586072 [carma_component_container_mt-24] 1703287546.658097221 | WARN | arbitrator | multiplex_service_call_for_capability:94 | Cannot make service request for service '/guidance/plugins/stop_and_dwell_strategic_plugin/plan_maneuvers': could not create client: failed to create client response subscriber, at /tmp/binarydeb/ros-foxy-rmw-fastrtps-cpp-1.3.2/src/rmw_client.cpp:214, at /tmp/binarydeb/ros-foxy-rcl-1.1.14/src/rcl/client.c:175. So retrying, attempt no: 0

This is an example of the error being caught. However, at some point it reports that the RCLError has occured but doesn't print the carma's error log before shutting down the arbitrator node, like this:

> 1703287546.6583834 [carma_component_container_mt-24] [37;1m2023-12-22 23:25:46.658 [31;1m[[37;1mPARTICIPANT[31;1m Error] [37mType : carma_planning_msgs::srv::dds_::PlanManeuvers_Response_ Not Registered[34;1m -> Function [36mcreateSubscriber[m

The error is happening at least once during a run.

### Version

4.3.0 (Current)

### Expected Behavior

Arbitrator should not throw exception

### Actual Behavior

See above

### Steps to Reproduce the Actual Behavior

start carma in cdasim on simulation pc 1 with rmw fast dds

MishkaMN:
Above PR highly likely fixes the issue; however it could be too early to tell. Better to confirm on a different PC

MishkaMN:
Did not see this issue anymore, so closing.

PR #2415, opened 2024-07-08
Update capabilities_interface.tpp

MishkaMN:
# PR Details
## Description
Arbitrator is getting frozen due to too many repeated failed calls. 
With 500ms service call timeout, the arbitrator can only afford maximum 2 calls to satisfy 1hz operation. 
Arbitrator should at least retry 2nd time again because sometimes ROS service call can fail due to unknown error. 
Without retrying, the next planning will be after 1second, which is too late for some planning such as checking red light in lci_strategic_plugin.

## Related GitHub Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/2385

## Related Jira Key
https://usdot-carma.atlassian.net/browse/CAR-6039

## Motivation and Context
Demos after 4.5.0 frequently encounter this issue by default because some nodes in ROS2 are not integration tested well but converted. This makes some of the nodes fail to activate or just due to host machine's performance, and repeated calls to such failed nodes freeze the arbitrator for up to 5 sec with 10 retry attempts.

## How Has This Been Tested?
integration tested locally

## Types of changes

## Checklist:

### DIFF
--- a/arbitrator/include/capabilities_interface.tpp
+++ b/arbitrator/include/capabilities_interface.tpp
@@ -27,7 +27,7 @@
 #include <thread>
 namespace arbitrator
 {
-    constexpr auto MAX_RETRY_ATTEMPTS {10};
+    constexpr auto MAX_RETRY_ATTEMPTS {2};
 
     template<typename MSrvReq, typename MSrvRes>
     std::map<std::string, std::shared_ptr<MSrvRes>>


paulbourelly999:
@MishkaMN do we have any information about the nature of these ROS call failures. Maybe there are node configurations we can set to reduce the amount of failed calls to mitigate the impact of retry attempts. I have had similar issues with SNMP calls that would fail to frequently. I was able to reduce failed by increasing the SNMP Client timeout for responses since almost all the failures were related to client timeouts. Just curious if we have looked into ROS node configurations relate to these calls to see if we can reduce the frequency of failing calls.

MishkaMN:
> @MishkaMN do we have any information about the nature of these ROS call failures. Maybe there are node configurations we can set to reduce the amount of failed calls to mitigate the impact of retry attempts. I have had similar issues with SNMP calls that would fail to frequently. I was able to reduce failed by increasing the SNMP Client timeout for responses since almost all the failures were related to client timeouts. Just curious if we have looked into ROS node configurations relate to these calls to see if we can reduce the frequency of failing calls.

if we have time, we can probably investigate more about node configurations, specifically Quality of Service settings. 
https://github.com/usdot-fhwa-stol/carma-platform/issues/2245 is the original issue that triggered this code implementation during VRU. And although I believe that specific issue was resolved, I have seen over the years that ROS service call can just fail without much logs. I don't have documented logs unfortunately but it looks somewhat similar like the link. 

Moreover, the current failure in arbitrator is mostly just symptom rather than an issue itself at this point (rarely the mysterious ROS error that this retry logic attempts to resolve). This PR is merely reducing the occurrence of the symptom so carma doesn't fail as often. 

There are other issues that can cause this arbitrator failure such as some nodes not correctly ported to ROS2 but enabled anyways, or not having enough time to initialize carma so some nodes are not activated, or deactuvated nodes publishing services as available etc. They are open issues that will be resolved in time.

paulbourelly999 (approved):
Arbitrarily modifying this value does not seem like a long term solution for failing ROS2 system calls. The reason this seems arbitrary to me is that we initially set the value to 10. Not sure what setting it to 2 does other than fail earlier. We may want to determine the root cause of this failure and try to mitigate it.

MishkaMN:
> Arbitrarily modifying this value does not seem like a long term solution for failing ROS2 system calls. The reason this seems arbitrary to me is that we initially set the value to 10. Not sure what setting it to 2 does other than fail earlier. We may want to determine the root cause of this failure and try to mitigate it.

Yes we should plan to figure out the root cause by getting some data to work with. Currently this safeguard was put out of personal experience over the years. 10 was arbitrary number retry number that is plain out wrong since it exceeds the 1hz operation we intend (10 comes out to 0.2Hz) and doing significantly more harm than good. This PR is only to fix that at the moment, and this 2nd attempt with 500ms timeout is not arbitrary because it is aiming to keep strategic planning close to the 1hz that is anticipated.

PR #2417, opened 2024-07-18
Fix platooning_control

adev4a:
# PR Details
This PR adds a fix to the platooning_control node which was recently ported to ros2.

Fixes a bug in delta time calculation for trajectory speed and updates the duration to be in nanoseconds.
Fixes the control topic in the trajectory_executor node to reference updated name of platooning_control package.
 

## Related Jira Key

https://usdot-carma.atlassian.net/browse/CAR-6012

## Types of changes

## Checklist:

### DIFF
--- a/carma_guidance_plugins/src/control_plugin.cpp
+++ b/carma_guidance_plugins/src/control_plugin.cpp
@@ -76,7 +76,7 @@ namespace carma_guidance_plugins
           {
             auto vehicle_cmd = this->generate_command();
             //If vehicle cmd is outdated - dont publish
-            if ((this->now() - rclcpp::Time(vehicle_cmd.header.stamp)).seconds() < 5.0){
+            if ((this->now() - rclcpp::Time(vehicle_cmd.header.stamp)).seconds() < cmd_timeout_in_s_){
               this->vehicle_cmd_pub_->publish(vehicle_cmd);
             }
 
--- a/carma_guidance_plugins/src/control_plugin.cpp
+++ b/carma_guidance_plugins/src/control_plugin.cpp
@@ -31,9 +31,9 @@ namespace carma_guidance_plugins
     return "control/trajectory_control";
   }
 
-  uint8_t ControlPlugin::get_type() 
+  uint8_t ControlPlugin::get_type()
   {
-    return carma_planning_msgs::msg::Plugin::CONTROL; 
+    return carma_planning_msgs::msg::Plugin::CONTROL;
   }
 
   void ControlPlugin::current_pose_callback(geometry_msgs::msg::PoseStamped::UniquePtr msg)
@@ -74,10 +74,15 @@ namespace carma_guidance_plugins
         [this]() {
           if (this->get_activation_status()) // Only trigger when activated
           {
-            this->vehicle_cmd_pub_->publish(this->generate_command());
+            auto vehicle_cmd = this->generate_command();
+            //If vehicle cmd is outdated - dont publish
+            if ((this->now() - rclcpp::Time(vehicle_cmd.header.stamp)).seconds() < 5.0){
+              this->vehicle_cmd_pub_->publish(vehicle_cmd);
+            }
+
           }
         });
-    
+
     return PluginBaseNode::handle_on_configure(prev_state);
   }
 
@@ -107,4 +112,3 @@ namespace carma_guidance_plugins
   }
 
 } // carma_guidance_plugins
-
--- a/carma_guidance_plugins/src/control_plugin.cpp
+++ b/carma_guidance_plugins/src/control_plugin.cpp
@@ -74,11 +74,7 @@ namespace carma_guidance_plugins
         [this]() {
           if (this->get_activation_status()) // Only trigger when activated
           {
-            auto vehicle_cmd = this->generate_command();
-            //If vehicle cmd is outdated - dont publish
-            if ((this->now() - rclcpp::Time(vehicle_cmd.header.stamp)).seconds() < cmd_timeout_in_s_){
-              this->vehicle_cmd_pub_->publish(vehicle_cmd);
-            }
+            this->vehicle_cmd_pub_->publish(this->generate_command());
 
           }
         });
--- a/platooning_control/src/platooning_control.cpp
+++ b/platooning_control/src/platooning_control.cpp
@@ -45,7 +45,6 @@ namespace platooning_control
     config_.emergency_stop_distance = declare_parameter<double>("emergency_stop_distance", config_.emergency_stop_distance);
     config_.speed_thres_traveling_direction = declare_parameter<double>("speed_thres_traveling_direction", config_.speed_thres_traveling_direction);
     config_.dist_front_rear_wheels = declare_parameter<double>("dist_front_rear_wheels", config_.dist_front_rear_wheels);
-    cmd_timeout_in_s_ = declare_parameter<double>("cmd_timeout_in_s", cmd_timeout_in_s_);
 
     config_.dt = declare_parameter<double>("dt", config_.dt);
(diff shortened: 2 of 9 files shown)


platooning_control/src/platooning_control.cpp
```diff
@@ -359,7 +359,7 @@ namespace platooning_control
     double dx1 = trajectory_points[trajectory_points.size()-1].x - trajectory_points[0].x;
     double dy1 = trajectory_points[trajectory_points.size()-1].y - trajectory_points[0].y;
     double d1 = sqrt(dx1*dx1 + dy1*dy1);
-    double t1 = rclcpp::Time((trajectory_points[trajectory_points.size()-1].target_time)).seconds() - rclcpp::Time(trajectory_points[0].target_time).nanoseconds();
+    double t1 = rclcpp::Time((trajectory_points[trajectory_points.size()-1].target_time)).seconds() - rclcpp::Time(trajectory_points[0].target_time).seconds();
```
saina-ramyar, line 362: changing from nanosec to sec, has any impact somewhere else?
I dont remember if it does

adev4a: I kept this in seconds since were using seconds precision in the earlier implementation and the speed is being calculcated as m/s 
https://github.com/usdot-fhwa-stol/carma-platform/blob/11047f2bde3dc2f37b7d2744cd41a3d9772bbc42/platooning_control/src/platoon_control.cpp#L298

But using nanoseconds instead will not have any impact, probably better to do since rounding of mps can have greater impact on mph speed.

adev4a:
This PR also needs https://github.com/usdot-fhwa-stol/carma-vehicle-calibration/pull/127, which updates parameters for pure pursuit

saina-ramyar (dismissed):

carma_guidance_plugins/src/control_plugin.cpp
```diff
@@ -74,10 +74,15 @@ namespace carma_guidance_plugins
         [this]() {
           if (this->get_activation_status()) // Only trigger when activated
           {
-            this->vehicle_cmd_pub_->publish(this->generate_command());
+            auto vehicle_cmd = this->generate_command();
+            //If vehicle cmd is outdated - dont publish
+            if ((this->now() - rclcpp::Time(vehicle_cmd.header.stamp)).seconds() < 5.0){
```
JonSmet, line 79: Let's make this 5.0 value a named member variable (likely `cmd_timeout_in_sec` or similar) in the `ControlPlugin` class to avoid the use of a magic number here

JonSmet, line 79: Separately, I'm trying to think through issues that could arise with this approach. If we start a normal single vehicle test, CARMA Platform will use Pure Pursuit as the controller, but if we then begin platooning with another vehicle, then the system will switch to Platooning Control as the controller. Does this mean that for five seconds (or the value that we have configured here), we'll have overlap where both control nodes will publish commands?

platooning_control/src/platooning_control.cpp
```diff
@@ -45,6 +45,7 @@ namespace platooning_control
     config_.emergency_stop_distance = declare_parameter<double>("emergency_stop_distance", config_.emergency_stop_distance);
     config_.speed_thres_traveling_direction = declare_parameter<double>("speed_thres_traveling_direction", config_.speed_thres_traveling_direction);
     config_.dist_front_rear_wheels = declare_parameter<double>("dist_front_rear_wheels", config_.dist_front_rear_wheels);
+    cmd_timeout_in_s_ = declare_parameter<double>("cmd_timeout_in_s", cmd_timeout_in_s_);
```
JonSmet, line 48: Might be missing a commit here, but `cmd_timeout_in_s_` will need to be dined in the `ControlPlugin` class for this to build

carma_guidance_plugins/src/control_plugin.cpp
JonSmet: After taking another look, in this example I think the overlap between both Pure Pursuit and Platooning Control both generating and publishing valid control commands would happen indefinitely since both control nodes would be able to generate commands in line with the current timestamp. 

Referencing [PurePursuitWrapperNode::generate_command()](https://github.com/usdot-fhwa-stol/carma-platform/blob/a92c44e692e7f36d0ace654dfc0c793769693ecd/pure_pursuit_wrapper/src/pure_pursuit_wrapper.cpp#L132), Pure Pursuit would just continue using the latest received Pure Pursuit trajectory (could be even from a minute ago) and would generate a command for it.

JonSmet (changes requested):

platooning_control/src/platooning_control.cpp
adev4a: Yup, looks like that didn't get pushed earlier. Just updated it

carma_guidance_plugins/src/control_plugin.cpp
adev4a: The trajectory executor will publish the trajectory to the control plugin that is supposed to generate the control commands https://github.com/usdot-fhwa-stol/carma-platform/blob/a92c44e692e7f36d0ace654dfc0c793769693ecd/trajectory_executor/src/trajectory_executor_node.cpp#L133
So there should be no overlap between any control plugin generating valid commands

JonSmet: Resolving this comment since we are removing and updates related to the vehicle command timeout from this PR.

JonSmet: Resolving this comment since we are removing and updates related to the vehicle command timeout from this PR.

platooning_control/src/platooning_control.cpp
JonSmet: Resolving this comment since we are removing and updates related to the vehicle command timeout from this PR.

trajectory_executor/src/trajectory_executor_node.cpp
```diff
@@ -55,8 +55,8 @@ namespace trajectory_executor
       out[config_.default_control_plugin] = config_.default_control_plugin_topic;
 
       //Hardcoding platoon control plugins
-      std::string control_plugin2 = "platoon_control";
-      std::string control_plugin_topic2 = "/guidance/plugins/platoon_control/plan_trajectory";
+      std::string control_plugin2 = "platooning_control";
+      std::string control_plugin_topic2 = "/guidance/plugins/platooning_control/plan_trajectory";
```
JonSmet, line 59: From the `CI: Run Tests` CI step, it looks like a unit test failure was introduced here:
```
2024-07-22T20:35:33.9337470Z 1: /opt/carma/src/carma-platform/trajectory_executor/test/test_trajectory_executor.cpp:95: Failure
2024-07-22T20:35:33.9339501Z 1: Expected: (test_suite_node->msg_count) > (0), actual: 0 vs 0
2024-07-22T20:35:33.9340622Z 1: Failed to receive whole trajectory from TrajectoryExecutor node.
2024-07-22T20:35:34.1153518Z 1: [  FAILED  ] TrajectoryExecutorTestSuite.test_emit_traj (3661 ms)
```

I think we will need to update the control topic used for the unit test here: https://github.com/usdot-fhwa-stol/carma-platform/blob/a92c44e692e7f36d0ace654dfc0c793769693ecd/trajectory_executor/test/trajectory_executor_test_suite.cpp#L68

JonSmet (changes requested):

JonSmet:
Now that the scope of changes has been adjusted for this PR (i.e. not introducing a command timeout), can we please update the PR description, and also mention the two specific fixes that are included in this changeset (changing the control topic for platooning_control and fixing the seconds vs. nanoseconds bug)

JonSmet (approved):

PR #2445, opened 2024-09-24
remove ros1 mock drivers and add ros2 mock controller driver

saina-ramyar:
# PR Details
## Description

Removed mock_rosbag_drivers and ported mock_controller_driver to ROS2 to be used for basic autonomy local test.

## Related Jira Key

[CAR-6069](https://usdot-carma.atlassian.net/browse/CAR-6069)

## Motivation and Context

Transition of CARMA Platform to ROS2

## How Has This Been Tested?

Locally tested

## Types of changes

## Checklist:

[CAR-6069]: https://usdot-carma.atlassian.net/browse/CAR-6069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


carma/launch/drivers.launch.py
```diff
@@ -47,6 +48,13 @@ def generate_launch_description():
         description = "True if simulation mode is on"
     )
 
+    use_mock_controller = LaunchConfiguration('use_mock_controller')
```
MishkaMN, line 51: I think if we are calling this node, we expect to use the node, so could it be unnecessary to have this flag?

mock_controller_driver/README.md
```diff
@@ -0,0 +1,3 @@
+# mock_controller_driver
+
+Mock controller driver replicates the controller driver behavior in for local deployment.
```
MishkaMN, line 3: typo:
"Mock controller driver replicates the controller driver behavior (for example returning engaged status) in local testing deployment."

mock_controller_driver/include/mock_controller_driver/mock_controller_driver_node.hpp
```diff
@@ -0,0 +1,121 @@
...
+#include <carma_driver_msgs/srv/set_enable_robotic.hpp>
+#include <carma_driver_msgs/msg/robot_enabled.hpp>
+#include <autoware_msgs/msg/vehicle_cmd.hpp>
+#include <carma_ros2_utils/carma_lifecycle_node.hpp>
+
+namespace mock_controller_driver
+{
+
+  /**
+   * \brief TODO for USER: Add class description
```
MishkaMN, line 33: Let's resolve all TODO.

MishkaMN: by either doing the TODO or removing the comments altogether. If it's really something we should track, we could add Jira Ticket and link it in the comment.

carma/launch/drivers.launch.py
```diff
@@ -102,10 +110,23 @@ def generate_launch_description():
...
+        package='mock_controller_driver',
+        name='mock_controller_driver',
+        executable='mock_controller_driver_node_exec',
+        parameters=[
+            {"use_mock_controller" : use_mock_controller}],
+        arguments=['--ros-args', '--log-level', GetLogLevel('mock_controller_driver', env_log_levels)],
+        condition=IfCondition(LaunchConfiguration('use_mock_controller'))
+    )
+
     return LaunchDescription([
```
MishkaMN, line 124: Actually maybe we should not even add a call here because we normally call it form a different drivers.launch.py in carma-config/development.

mock_controller_driver/src/mock_controller_driver_node.cpp
```diff
@@ -0,0 +1,95 @@
...
+    // Return success if everything initialized successfully
+    return CallbackReturn::SUCCESS;
+  }
+
+  carma_ros2_utils::CallbackReturn MockControllerDriver::handle_on_activate(const rclcpp_lifecycle::State &prev_state)
+  {
+    // Setup timers
+    spin_timer_ = create_timer(
+        get_clock(),
+        std::chrono::milliseconds(50), // 20 Hz
```
MishkaMN, line 53: should be 30Hz for controller I think

mock_controller_driver/src/mock_controller_driver_node.cpp
```diff
@@ -0,0 +1,95 @@
...
+    {
+      robot_active_ = false;
+    }
+
+    return true;
+  }
+
+  void MockControllerDriver::timer_callback()
+  {
+    RCLCPP_DEBUG(get_logger(), "Example timer callback");
```
MishkaMN, line 83: lets either remove this debug line or add more informational fields such as the message being generated here

mock_controller_driver/include/mock_controller_driver/mock_controller_driver_node.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+    // Sub
+    const std::string vehicle_cmd_topic_ = "vehicle_cmd";
+
+    // Service
+    const std::string enable_robotic_srv_ = "controller/enable_robotic";
+
+    // Robot Status flags
+    bool robot_active_ = false;
+    bool robot_enabled_ = false;
```
MishkaMN, line 67: What is the difference between the two?
for example, which one is for when the vehicle is engaged? I am struggling to understand the [comment on the msg itself too](https://github.com/usdot-fhwa-stol/carma-msgs/blob/develop/carma_driver_msgs/msg/RobotEnabled.msg#L7C1-L11C19)

Could we give more descriptive or commonly used term by the team?

mock_controller_driver/src/mock_controller_driver_node.cpp
```diff
@@ -0,0 +1,95 @@
...
+    }
+    else
+    {
+      robot_active_ = false;
+    }
+
+    return true;
+  }
+
+  void MockControllerDriver::timer_callback()
```
MishkaMN, line 81: Let's give more descriptive name for this since it is only doing one function
`publish_robot_status_callback()` or sth like that

mock_controller_driver/src/mock_controller_driver_node.cpp
```diff
@@ -0,0 +1,95 @@
...
+    return CallbackReturn::SUCCESS;
+  }
+
+
+  void MockControllerDriver::vehicle_cmd_callback(const autoware_msgs::msg::VehicleCmd::UniquePtr msg)
+  {
+    robot_enabled_ = true;  // If a command was received set the robot enabled status to true
+  }
+
+  bool MockControllerDriver::enable_robotic_srv(std::shared_ptr<rmw_request_id_t>,
```
MishkaMN, line 65: Did we switch the variables? The function "enables" the robot, but it is modifying the variable related to "active" or not and expects the robot_enabled variable to be set already.

MishkaMN (changes requested):

carma/launch/drivers.launch.py
saina-ramyar: I agree, in my story it was mentioned call mock driver if an argument is set. You let me know if I should remove it.

carma/launch/drivers.launch.py
saina-ramyar: I agree, removing the parameter and launch description.

mock_controller_driver/src/mock_controller_driver_node.cpp
saina-ramyar: changed. But thats the control plugins and the original mock controller was 20hz.

mock_controller_driver/src/mock_controller_driver_node.cpp
saina-ramyar: No this is the way mock driver works too, the robot_enable is set to true when a command is generated.

mock_controller_driver/include/mock_controller_driver/mock_controller_driver_node.hpp
saina-ramyar: From what I understood from this driver and guidance plugin, the setRobotEnable service call is when the vehicle is first engaged, and robot_active_ is when vehicle is engaged and receiving commands.
I couldn't find any documentation to explain the sequence. Even the guidance plugin description are brief and kind of vague.

mock_controller_driver/src/mock_controller_driver_node.cpp
saina-ramyar: updated

mock_controller_driver/include/mock_controller_driver/mock_controller_driver_node.hpp
saina-ramyar: removed all todos.

mock_controller_driver/README.md
saina-ramyar: fixed

carma/launch/drivers.launch.py
saina-ramyar: removed

mock_controller_driver/include/mock_controller_driver/mock_controller_driver_node.hpp
saina-ramyar: added a sentence to the robot enable topic.

mock_controller_driver/src/mock_controller_driver_node.cpp
MishkaMN: Okay this looks about right. I can integration test it with basic-driving-simulator later if there is any issue

mock_controller_driver/include/mock_controller_driver/mock_controller_driver_node.hpp
MishkaMN: yeah that's my understanding as well. I am just getting confused by the order of the service call and vehicle_cmd subscription order. I would think vehicle_cmd topic subscription will only receive values (unless it normally receives speed=0 commands constantly) after the vehicle is engaged (the enable service call). But I could be wrong, so I am approving it and we can integration test it later.

MishkaMN (approved):

PR #2449, opened 2024-10-10
Update setup opt-carma script according to install guide

MishkaMN:
# PR Details
## Description
During creation of CDASim comprehensive tutorial guide, I needed to include a step to setup CARMA. That writeup prompted opportunity to clean up our script that is referenced in our CARMA installation guide here: https://usdot-carma.atlassian.net/wiki/spaces/CRMPLT/pages/488472599/Setup+CARMA+Platform+Runtime

Changes include:
- No longer downloads autoware.ai just to copy its yolo contents. CARMA doesn't use camera at the moment so it is not needed. That repo is almost 400 MB+ so it is not ideal.
- Added support for setting up common static folders needed for CDASim because it is not documented anywhere, and my current write-up and existing installation guide will include them. 

## Related GitHub Issue
NA

## Related Jira Key
https://usdot-carma.atlassian.net/browse/CXC-94

## Motivation and Context
I am writing CDASim tutorial guide. And its configuration step required some changes to our installation guides, which I discovered these issues. 

## How Has This Been Tested?
local integration tested by calling scripts in https://github.com/usdot-fhwa-stol/carma-platform/pull/2449
and building images locally

## Types of changes

## Checklist:

### DIFF
--- a/engineering_tools/opt_carma_setup.bash
+++ b/engineering_tools/opt_carma_setup.bash
@@ -68,7 +68,7 @@ sudo groupadd --gid $GRP_ID carma # create the carma group if it does not alread
 USERNAME=$(whoami)
 sudo usermod -a -G $GRP_ID $USERNAME
 
-mkdir -p /opt/carma/logs /opt/carma/.ros /opt/carma/maps /opt/carma/routes
+mkdir -p /opt/carma/logs /opt/carma/.ros /opt/carma/maps /opt/carma/routes /opt/carma/yolo /opt/carma/data
 
 # Check if downloading maps is required
 if [ "$DOWNLOAD_MAPS" == "true" ]; then
--- a/engineering_tools/opt_carma_setup.bash
+++ b/engineering_tools/opt_carma_setup.bash
@@ -1,13 +1,13 @@
 #!/bin/bash
 
-#  Copyright (C) 2018-2021 LEIDOS.
-# 
+#  Copyright (C) 2018-2024 LEIDOS.
+#
 #  Licensed under the Apache License, Version 2.0 (the "License"); you may not
 #  use this file except in compliance with the License. You may obtain a copy of
 #  the License at
-# 
+#
 #  http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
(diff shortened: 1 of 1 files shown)


engineering_tools/opt_carma_setup.bash
```diff
@@ -27,29 +68,44 @@ sudo groupadd --gid $GRP_ID carma # create the carma group if it does not alread
...
+# Link the provided vehicle calibration folder
+if [ -L /opt/carma/vehicle ]; then
+    sudo rm /opt/carma/vehicle
+fi
+
+ln -s "$CALIBRATION_FOLDER" /opt/carma/vehicle
+
+if [ "$ADD_SIM_FOLDER" == "true" ]; then
+    echo "Adding folders used for CDASim to be able to save logs"
+    mkdir -p /opt/carma/logs/carma_1 /opt/carma/logs/carma_2 /opt/carma-simulation/logs /opt/carma-simulation/scripts /opt/carma-streets/scripts
```
adev4a, line 104: nit: Manually adding log directories for each component seems like this will change pretty frequently. Since the permissions are already elevated, can we not let these directories be created on their own?

MishkaMN: These folders do not exist yet. Therefore, currently when the carma runs, it creates with permissions of root root. This is why carma keeps failing to write logs and won't start. If the folder exists with correct permissions, files can be written. 

it must be possible to make our docker configuration to write with correct permissions. But I just don't have enough bandwidth to investigate and fix.

engineering_tools/opt_carma_setup.bash
```diff
@@ -27,29 +68,44 @@ sudo groupadd --gid $GRP_ID carma # create the carma group if it does not alread
...
+    sudo chgrp -R $GRP_ID /opt/carma/
```
adev4a, line 105: These can be moved to a single line. See
https://stackoverflow.com/questions/19093613/how-to-give-777-permission-to-multiple-directories-at-a-time-in-unix

adev4a (approved):

PR #2457, opened 2024-11-05
carma exec to accept network

MishkaMN:
# PR Details
## Description
For CDASim, carma exec doesn't work because network configuration is hardcoded to host, but cdasim uses different IPs. I could use `docker exec` but this won't last in between test and have to do the sourcing exporting again, which is not user-friendly. This will be useful only for Sim PCs as vehicles don't need it right away. And it can be updated whenever the script is updated next. 

This PR lets the user specify network to connect to. Usage:
`carma exec -n network_name` 

## Related GitHub Issue
NA

## Related Jira Key

NA
## Motivation and Context
This happens a lot when I try to test something in CDASim, I need to change the network myself in the /usr/bin/carma script. 

## How Has This Been Tested?
CDASIM
image

## Types of changes

## Checklist:

### DIFF
--- a/engineering_tools/carma
+++ b/engineering_tools/carma
@@ -502,42 +502,58 @@ carma__exec() {
     local TAG=latest
 
     local TARGET_IMAGE="$USERNAME/carma-platform:$TAG"
+    local NETWORK="host"  # Default network mode
+    local RUNTIME=""      # Default runtime is empty
+    local COMMAND=""
+    local INIT_FILE="/home/carma/.base-image/init-env.sh"
 
-    # Check if the -i argument was provided
-    # If the image was specified then use that
+    # Process all arguments in any order
+    while [[ $# -gt 0 ]]; do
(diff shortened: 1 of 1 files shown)


engineering_tools/carma
```diff
@@ -587,12 +600,14 @@ carma__exec() {
             --volume=/opt/carma/routes:/opt/carma/routes \
             --env="XAUTHORITY=${XAUTH}" \
             --device=/dev/dri:/dev/dri \
-            --network=host \
+            --network=$NETWORK \
             --entrypoint="$entrypoint" \
+            -p 8765:8765 \
```
adev4a, line 605: Is this port specification required?

MishkaMN: Not exactly. I remember it being helpful when I was trying out foxglove live. However, we can add that when we actually need it. So I removed it

adev4a (approved):

PR #2469, opened 2024-11-20
more processes

MishkaMN:
# PR Details
## Types of changes

## Checklist:

### DIFF
--- a/engineering_tools/monitor-ros-cpu.py
+++ b/engineering_tools/monitor-ros-cpu.py
@@ -7,6 +7,46 @@ import os
 import argparse
 from datetime import datetime
 
+# Define ROS-related keywords to filter processes
+ROS_KEYWORDS = {
+    'ros',
+    'node',
+    'python3',
+    'rviz',
+    'rqt',
+    'launch',
+    'carma',  # CARMA-specific processes
+    'guidance',
+    'planning',
+    'control',
+    'perception',
+    'localization',
+    'vehicle',
+    'environment',
+    '/opt/ros/',  # ROS installation path
+    'roscore',
+    'rosmaster',
+    'roslaunch',
+    'rostopic',
+    'rosnode',
+    'rosbag',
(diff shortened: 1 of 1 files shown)


engineering_tools/monitor-ros-cpu.py
```diff
@@ -7,6 +7,37 @@
 import argparse
 from datetime import datetime
 
+# Define ROS-related keywords to filter processes
+ROS_KEYWORDS = {
+    "ros",
+    "node",
+    "rviz",
+    "rqt",
+    "ros2 launch",
```
john-chrosniak, line 16: Nit: Shouldn't need `ros2 launch` since `ros2` is already a keyword

engineering_tools/monitor-ros-cpu.py
```diff
@@ -15,17 +46,61 @@ def parse_args():
...
+def is_ros_related_process(proc_info, cmdline):
+    """
+    Check if a process is ROS-related based on name and command line
+    """
+    # Convert process information to lowercase for case-insensitive matching
+    name_lower = proc_info["name"].lower()
+    cmdline_lower = cmdline.lower()
+
+    # Check exclusions first
+    if any(excl in name_lower or excl in cmdline_lower for excl in EXCLUDE_KEYWORDS):
```
john-chrosniak, line 74: If there is an excluded keyword in the process and also a ROS keyword this would still return false, is that the desired behavior?

MishkaMN: Yeah that is the behavior. Current keywords are picked just like that, but I included a note on the comment to also emphasize that.

engineering_tools/monitor-ros-cpu.py
MishkaMN: Makes sense removed

john-chrosniak (approved):

Issue #2501, opened 2025-02-26
Yield_plugin is shutting down due to rclcpp::Time negative


MishkaMN:
### Are you Saxton Lab/Leidos employee?

Yes

### Summary

During testing of yield_plugin to other vehicles on the road, sometimes the vehicle shuts down due to error time cannot be negative. Not sure where is this coming from

### Version

4.5.0 (Current)

### Expected Behavior

yield_plugin shouldn't shut down carma

### Actual Behavior

yield_plugin shuts down carma

### Steps to Reproduce the Actual Behavior

run carma and yielding in simulation

PR #2502, opened 2025-02-26
Improve yield_plugin to remove exceptions and shut downs

MishkaMN:
# PR Details
## Description
During testing of TM/TSP with background vehicles, sometimes the vehicle is shutting down due to 2 exceptions. 
This PR makes the service call safer so that it doesn't let it shut down unnecessarily for the time being. 
This is because currently there are 2 open issues (see below) that is blocking any normal testing of carma-platform in Town05 simulation. If the exception is safely ignored for now, vehicle resumes normal operation and finishes the route. 

## Related GitHub Issue
These are the 2 issues notices:
Not sure why but yield is generating negative time sometimes: https://github.com/usdot-fhwa-stol/carma-platform/issues/2501
Due to not safely handling osm map issue, sometimes trajectory generation can fail due to erroneous spline: https://github.com/usdot-fhwa-stol/carma-platform/issues/2503

Currently it is just applying workaround so it doesn't block other people. 
With these exceptions being ignored, the vehicle is able to finish its route. 

## Related Jira Key
NA

## Motivation and Context
TM/TSP Town05 integration testing

## How Has This Been Tested?
Block Sim PC 3

## Types of changes

## Checklist:

### DIFF
--- a/yield_plugin/src/yield_plugin.cpp
+++ b/yield_plugin/src/yield_plugin.cpp
@@ -1054,6 +1054,8 @@ namespace yield_plugin
       auto llts = wm_->getLaneletsFromPoint(veh_pos, 1);
       if (llts.empty())
       {
+        // This should technically never happen
+        // However, trajectory generation currently may fail due to osm map issue https://github.com/usdot-fhwa-stol/carma-platform/issues/2503
         RCLCPP_WARN_STREAM(nh_->get_logger(),"Trajectory point: x= " << original_tp.trajectory_points.at(i).x << "y="<< original_tp.trajectory_points.at(i).y);
         RCLCPP_WARN_STREAM(nh_->get_logger(),"Trajectory is not on the road, so was unable to get the digital minimum gap. Returning default minimum_safety_gap_in_meters: " << config_.minimum_safety_gap_in_meters);
         return desired_gap;
--- a/yield_plugin/src/yield_plugin.cpp
+++ b/yield_plugin/src/yield_plugin.cpp
@@ -312,7 +312,7 @@ namespace yield_plugin
         resp->trajectory_plan = yield_trajectory;
       }
     }
-    catch(const std::exception& e)
+    catch(const std::runtime_error& e)
     {
       RCLCPP_WARN_STREAM(nh_->get_logger(), "Yield Plugin failed to plan trajectory: " << e.what());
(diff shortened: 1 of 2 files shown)


yield_plugin/src/yield_plugin.cpp
```diff
@@ -263,55 +263,66 @@ namespace yield_plugin
...
+      else
+      {
+        yield_trajectory.header.frame_id = "map";
+        yield_trajectory.header.stamp = nh_->now();
+        yield_trajectory.trajectory_id = original_trajectory.trajectory_id;
+        resp->trajectory_plan = yield_trajectory;
+      }
     }
-    else
+    catch(const std::runtime_error& e)
```
paulbourelly999, line 315: Is there a more specific error we can catch here. Sonar scanner discourages catching generic exceptions for error handling and encourages catching more specific ones.

paulbourelly999 (changes requested):
Small change

yield_plugin/src/yield_plugin.cpp
paulbourelly999: https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=2502&id=usdot-fhwa-stol_CARMAPlatform&open=AZVEDtGbGJGytsS7SHKc

MishkaMN: There is no specific error we can catch. Please look at my comment above. the error thrown is itself runtime_error

MishkaMN: Okay I figured out a way to catch the specific error based on the error message not the error type.

MishkaMN:
Also because we already started the merging of develop-humble, the docker build and CI should fail until everything is merged.

yield_plugin/src/yield_plugin.cpp
paulbourelly999: If that is the case I actually prefer you change it back to what you had originally. The idea of throwing specific exceptions and catching is to allow error handling to respond differently based on the exception thrown. It sounds like we should throw a more specific exception and then catch that specific one here but for the sake of this PR I would prefer if you revert the change looking for a specific error message. That is not very maintainable.

paulbourelly999 (changes requested):
Left a comment. I think you should revert to what you had initially for error handling.

paulbourelly999 (approved):

PR #2525, opened 2025-03-27
Add branch to v2x ros conversion deps

adev4a:
# PR Details
## Types of changes

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,13 +35,13 @@
 ARG DOCKER_ORG="usdotfhwastoldev"
 ARG DOCKER_TAG="develop"
 FROM ${DOCKER_ORG}/autoware.ai:${DOCKER_TAG} as base-image
-
+ARG GIT_BRANCH="develop"
 FROM base-image AS source-code
 
 RUN mkdir ~/src
 COPY --chown=carma ./docker /home/carma/src/carma-platform/docker
 
-ARG GIT_BRANCH="develop"
+
 RUN ~/src/carma-platform/docker/checkout.bash -b ${GIT_BRANCH}
 
 COPY --chown=carma . /home/carma/src/carma-platform
@@ -57,7 +57,7 @@ ENV PACKAGES=${PACKAGES}
 RUN mkdir ~/carma_ws
 COPY --from=source-code --chown=carma /home/carma/src /home/carma/carma_ws/src
 
-RUN ~/carma_ws/src/carma-platform/docker/install.sh
+RUN ~/carma_ws/src/carma-platform/docker/install.sh -b ${GIT_BRANCH}
 
 # /////////////////////////////////////////////////////////////////////////////
 # Stage 3 - Finalize deployment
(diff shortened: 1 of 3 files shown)


docker/checkout.bash
```diff
@@ -74,3 +78,10 @@ cd ${dir}/src/ros2_tracing
 git sparse-checkout init --cone
 git sparse-checkout set tracetools
 cd ${dir}/src
+
+cd ${dir}
+
+# Install dependencies
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+sudo chmod +x ${SCRIPT_DIR}/install_dependencies.sh
+${SCRIPT_DIR}/install_dependencies.sh -b $BRANCH -r $dir
```
MishkaMN, line 87: Can we wrap this with `if [[ -z "$PACKAGES" ]]; then` like before so that we don't install this again if we are building only select packages. The assumption there is that these dependencies would already be there.

Dockerfile
```diff
@@ -45,19 +47,9 @@ ARG GIT_BRANCH="develop"
...
-# Copy the source files from the previous stage and build/install
-RUN mkdir ~/carma_ws
-COPY --from=source-code --chown=carma /home/carma/src /home/carma/carma_ws/src
 
-RUN ~/carma_ws/src/carma-platform/docker/install.sh
+## Build and install the software
+RUN ~/src/carma-platform/docker/install.sh
 
 # /////////////////////////////////////////////////////////////////////////////
 # Stage 3 - Finalize deployment
```
MishkaMN, line 55: Stage 2 is removed it looks like. Can we still keep that `install` stage?

docker/checkout.bash
MishkaMN, line 87: NIt: Can we wrap this with `if [[ -z "$PACKAGES" ]]; then` so that it is a little bit more obvious if we are running this or not.  Although I know that if clause is still inside.

docker/install_dependencies.sh
```diff
@@ -0,0 +1,46 @@
...
+if [[ -z "$PACKAGES" ]]; then
+    echo "Installing multiple object tracking dependencies"
+    sudo ./src/multiple_object_tracking/scripts/install_dependencies.sh
+fi
+
+# Install dependencies for v2x-ros-conversion - needs branch to pull correct debian version
+if [[ -z "$PACKAGES" ]]; then
+    echo "Installing cpp_message dependencies"
+    sudo ./src/v2x-ros-conversion/docker/install_dependencies.sh -b $BRANCH
+fi
```
MishkaMN, line 46: Nit: We can remove the if clauses here if we wrap the script entirely with if clause before calling like in my previous comment

Dockerfile
adev4a: Based on in person conversation, keeping the current changes with the checkout and install called on same step - since this avoids having to copy the source code twice.

MishkaMN (approved):

Issue #2345, opened 2024-03-28
yield_plugin is NOT stopping with comfortable decel when CP is on


MishkaMN:
### Summary

During verification testing data analysis, it was discovered that the yield_plugin is generating trajectory with emergency decel rate even with CP turned on. Following are from graphs from two different runs, but describe same behavior:

![example-deceleration-and-speed](https://github.com/usdot-fhwa-stol/carma-platform/assets/20613282/44d487db-4f9e-4805-b8a5-a1013f40abb7)
![decel](https://github.com/usdot-fhwa-stol/carma-platform/assets/20613282/95890baf-0f9d-453f-89ea-dd9c29673aeb)

### Commit

951ca9a2bdd5212af032aedde0891bbe4b22f9a8

### Expected Behavior

platform should decelerate with lower decel rate

### Actual Behavior

see above

### Steps to Reproduce the Actual Behavior

run cdasim with CP enabled and yield_plugin's minimum_safety_distance set to small number such as 1.0 meter

MishkaMN:
One remaining issue here that I could identify was about the initial inputs that we are giving. Initial inputs specify:
initial_access: 0. (https://github.com/usdot-fhwa-stol/carma-platform/blob/95f33725259f87a4da1268e35c67ed54dc997a93/yield_plugin/src/yield_plugin.cpp#L420)
This condition arises from trying to achieve minimum jerk, which is 0 initial and end accel among other conditions. 

However, the yield plugin is running at 10hz and everytime it is generating new trajectory assuming 0 acceleration EVEN if it is already decelerating with certain non-zero accel. Therefore, the calculation is correct only on the first iteration, but gradually becomes more inaccurate. This is probably being displayed by the sudden cliff at the end where the accel suddenly drops to make the math work.

MishkaMN:
Modification done under VRU use case was heavily influenced by stopping for potential obstacles whereas the original algorithm comes from more of "platooning" behavior, which is why the minimum jerk algorithm is picked rather than just assigning comfortable deceleration and stop. Therefore, any more fixes should keep both cases in mind.

PR #2530, opened 2025-04-02
Fx initial conditions in JMT algorithm in yield plugin

MishkaMN:
# PR Details
## Description
Previously we ignored the initial condition especially acceleration in JMT algorithm - it always assumed 0 m/s^2 acceleration initial condition. 

This will make the trajectory try to plan a brand new minimum jerk trajectory from the beginning which would eventually run out of time to have a perfect speed profile shape. By adjusting the initial condition and accounting for the fact that the JMT algorithm is already on the way, we should successfully achieve the desired speed profile shape theoretically.

This is a speed profile pic from before (doesn't slow down as quite as fast compared to after pic):
yield-original
After (more consistent deceleration, instead of "continuously under-decelerate then decelerate fast toward the end):
yield-after-solving
JMT algorithm should show aggressive deceleration in the middle and small decel/accel at the beginning/end.
So it seems to me that the trajectory got better, but not sure if it can be even better. Although I didn't repeat tests many times and thoroughly. At least it is not breaking anything I think.

NOTE: The speed "falling off" near the end around 2s is probably not related to yield_plugin but `vehicle_response_lag` 
As it does look a bit better without it (but it will make the vehicle stop too close to the target - probably CARLA acceleration is messing up the speed profile https://usdot-carma.atlassian.net/browse/CDAD-96)
image

## Related GitHub Issue
https://github.com/usdot-fhwa-stol/carma-platform/issues/2345

## Related Jira Key
NA

## Motivation and Context
Yield improvement for demo

## How Has This Been Tested?
Sim PC, see the comparison speed profile

## Types of changes

## Checklist:

### DIFF
--- a/yield_plugin/config/parameters.yaml
+++ b/yield_plugin/config/parameters.yaml
@@ -19,11 +19,11 @@ min_obj_avoidance_plan_time_in_s: 2.0
 # Deceleration value used for yielding trajectories
 # Units: m/s^2
 # Value type: Desired
-yield_max_deceleration_in_ms2: 2.0
+yield_max_deceleration_in_ms2: 3.0
 # Minimum safety gap with an Object/Obstacle
 # Units: meters
 # Value type: Desired
-minimum_safety_gap_in_meters: 30.0
+minimum_safety_gap_in_meters: 4.0
 # Maximum speed value to consider the ego vehicle stopped
 # Units: m/s
 # Value type: Desired
--- a/yield_plugin/config/parameters.yaml
+++ b/yield_plugin/config/parameters.yaml
@@ -23,7 +23,7 @@ yield_max_deceleration_in_ms2: 3.0
 # Minimum safety gap with an Object/Obstacle
 # Units: meters
 # Value type: Desired
-minimum_safety_gap_in_meters: 4.0
+minimum_safety_gap_in_meters: 10.0
 # Maximum speed value to consider the ego vehicle stopped
 # Units: m/s
 # Value type: Desired
--- a/yield_plugin/config/parameters.yaml
+++ b/yield_plugin/config/parameters.yaml
@@ -19,15 +19,15 @@ min_obj_avoidance_plan_time_in_s: 2.0
 # Deceleration value used for yielding trajectories
 # Units: m/s^2
 # Value type: Desired
-yield_max_deceleration_in_ms2: 3.0
+yield_max_deceleration_in_ms2: 2.0
 # Minimum safety gap with an Object/Obstacle
 # Units: meters
 # Value type: Desired
-minimum_safety_gap_in_meters: 10.0
+minimum_safety_gap_in_meters: 30.0
 # Maximum speed value to consider the ego vehicle stopped
 # Units: m/s
 # Value type: Desired
-max_stop_speed_in_ms: 1.0
+max_stop_speed_in_ms: 0.15
 # parameter to enable cooperative behavior
 # Value type: Desired
 enable_cooperative_behavior: false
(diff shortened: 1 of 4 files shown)


yield_plugin/config/parameters.yaml
```diff
@@ -27,7 +27,7 @@ minimum_safety_gap_in_meters: 10.0
 # Maximum speed value to consider the ego vehicle stopped
 # Units: m/s
 # Value type: Desired
-max_stop_speed_in_ms: 1.0
+max_stop_speed_in_ms: 0.15
```
MishkaMN, line 30: Below this value, the vehicle forces the traj speed to 0m/s. So to command 0 from 1.0 m/s seemed pretty high. This is almost a pedestrian walking speed. Since it was in simulation the effect may not have felt as much. I got the new values from looking at the data and how it is printing towards the end.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -427,11 +427,27 @@ namespace yield_plugin
     // Up until goal_pos (which also can be until end of the entire original trajectory), generate new speeds at
     // or near original trajectory points by generating them at a fixed time interval using the JMT polynomial equation
     const double initial_time = 0;
-    const double initial_accel = 0;
+    double initial_accel = 0;
+    if (last_speed_)
+    {
+      initial_accel = (initial_velocity - last_speed_.value()) /
```
MishkaMN, line 433: I am not sure at the moment whether if instantaneous acceleration I am using here is a good variable to use. Maybe low pass filter or some other way to approximate the ongoing acceleration is better?

MishkaMN:
This may not be the ultimate fix for https://github.com/usdot-fhwa-stol/carma-platform/issues/2345
Other approach I can think of is to use previously generated trajectory if detected collision conditions are around the same. But that would require more drastic changes, so I am holding off on that before the potential demo we have soon.

yield_plugin/src/yield_plugin.cpp
saina-ramyar: maybe we can get actual vehicle acceleration from vehicle status or something, but instantaneous is also reasonable

yield_plugin/src/yield_plugin.cpp
```diff
@@ -427,11 +427,27 @@ namespace yield_plugin
     // Up until goal_pos (which also can be until end of the entire original trajectory), generate new speeds at
     // or near original trajectory points by generating them at a fixed time interval using the JMT polynomial equation
     const double initial_time = 0;
-    const double initial_accel = 0;
+    double initial_accel = 0;
+    if (last_speed_)
```
saina-ramyar, line 431: can you check if last_speed_ and last_speed_time_ both have values here?

yield_plugin/src/yield_plugin.cpp
```diff
@@ -450,15 +466,21 @@ namespace yield_plugin
...
                                                                                                 planning_time);
-    RCLCPP_DEBUG_STREAM(nh_->get_logger(),"Used original_max_speed: " << original_max_speed);
+    RCLCPP_DEBUG_STREAM(rclcpp::get_logger("yield_plugin"),"Used original_max_speed: " << original_max_speed);
+    for (size_t i = 0; i < polynomial_coefficients.size(); i++) {
+      RCLCPP_DEBUG_STREAM(rclcpp::get_logger("yield_plugin"),"Coefficient " << i << ": " << polynomial_coefficients[i]);
+    }
     const auto smallest_time_step = get_smallest_time_step_of_traj(original_tp);
+    int new_traj_idx = 1;
+    int original_traj_idx = 1;
+    bool is_accelerating = (initial_velocity < goal_velocity) ? true : false;
```
saina-ramyar, line 476: I dont see where is_accelerating is used?

yield_plugin/src/yield_plugin.cpp
MishkaMN: added both

yield_plugin/src/yield_plugin.cpp
MishkaMN: Oh good catch this was an artifact from a different method I was trying

saina-ramyar (approved):

Issue #2395, opened 2024-05-31
Unsuccessful calls to Cooperative Lanechange Plugin 


MishkaMN:
### Summary

During preparation of 06/11/2024 [unsignalized workzone demo ](https://usdot-carma.atlassian.net/wiki/spaces/CRMPLT/pages/2877980673/TFHRC+Non-signalized+Work+Zone+Closed+Lane+Demo+with+CARMA+Platform), BP and Fusion both showed frequent trajectory loss observed from rviz. It looks like the trajectory is blinking, which results in vehicle sometimes yanking a little bit out of the lanechange. 

Inspecting the logs shows 81+ instances of `unsuccessful calls to cooperative_lanechange`.

Increasing [tactical_plugin_timeout parameter in carma-config](https://github.com/usdot-fhwa-stol/carma-config/blob/carma-system-4.5.0/chrysler_pacifica_ehybrid_s_2019/VehicleConfigParams.yaml#L48) from 100ms to 300ms largely resolved the issue.

Therefore, it is assumed that the plugin is inefficient or needs more tuning to perform better.   

### Version

4.5.0 (Current)

### Expected Behavior

See above

### Actual Behavior

See above

### Steps to Reproduce the Actual Behavior

perform any kind of lanechange

PR #2531, opened 2025-04-03
Cooperative Lanechange is not using the resample_step_size distance

MishkaMN:
# PR Details
## Description
This PR fixes lanechange trajectory generation in basic_autonomy to actually use the resampled points from resampling with `curve_resample_step_size` meters. It was previously ignoring this, so that the cooperative_lanechange's trajectory has very tiny distance between them which is hard to see on rviz.

With this fix, you can actually see the trajectory and its color (meaning the speed). See below
image

## Related GitHub Issue
Partially resolves https://github.com/usdot-fhwa-stol/carma-platform/issues/2395
Because if it has that many points, it might be taking long to respond.

## Related Jira Key
CAR-6115
https://usdot-carma.atlassian.net/browse/CAR-6115

## Motivation and Context
Personally noticed first in 4.5.0 while doing workzone demo. Another demo came around and thought it is better to address it as workzone demo is presented a lot.

## How Has This Been Tested?
Tested on workzone in Black Pacifica with the map that works.

## Types of changes

## Checklist:


MishkaMN:
Uni  tests are passing. I think the CI is just failing due to resource constraints? Some tests are just timing out.
image
image

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -1151,53 +1151,94 @@ namespace basic_autonomy
...
-            double scaled_steps_along_curve = 0.0; //from 0 (start) to 1 (end) for the whole trajectory
+            std::vector<lanelet::BasicPoint2d> resampled_points;
+            resampled_points.reserve(total_step_along_curve + 1);
 
-            for(int steps_along_curve = 0; steps_along_curve < total_step_along_curve; steps_along_curve++){
-                lanelet::BasicPoint2d p = (*fit_curve)(scaled_steps_along_curve);
-
-                all_sampling_points.push_back(p);
+            // Add the current vehicle position as first point
+            resampled_points.push_back(current_vehicle_point);
```
adev4a, line 1174: Current vehicle position was already included in future_geom_points on line 1157. If the intention here is to make sure that the resampled points have the vehicle position it is probably better to replace the first point with the vehicle position at this point.

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -1151,53 +1151,94 @@ namespace basic_autonomy
...
-            std::vector<double> downtracks = carma_wm::geometry::compute_arc_lengths(future_geom_points);
+            // Compute downtracks for original points (for interpolation reference)
+            std::vector<double> original_downtracks = carma_wm::geometry::compute_arc_lengths(future_geom_points);
 
-            auto total_step_along_curve = static_cast<int>(downtracks.back() /detailed_config.curve_resample_step_size);
+            // Now create resampled points using the spline
+            auto total_step_along_curve = static_cast<int>(original_downtracks.back() / detailed_config.curve_resample_step_size);
+            if (total_step_along_curve <= 0) {
+                RCLCPP_WARN_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER), "Total steps along curve is 0 or negative. Using original points.");
+                total_step_along_curve = 1; // Avoid division by zero
```
adev4a, line 1167: This makes sense - total steps along curve will atleast include the vehicle current location. 
nit: Can we update the log statement to "Using original points. Total steps along curve is 0 or negative"

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -1151,53 +1151,94 @@ namespace basic_autonomy
...
 
-            for(int steps_along_curve = 0; steps_along_curve < total_step_along_curve; steps_along_curve++){
-                lanelet::BasicPoint2d p = (*fit_curve)(scaled_steps_along_curve);
-
-                all_sampling_points.push_back(p);
+            // Add the current vehicle position as first point
+            resampled_points.push_back(current_vehicle_point);
 
+            double scaled_steps_along_curve = 0.0; // from 0 (start) to 1 (end) for the whole trajectory
+            for(int steps_along_curve = 0; steps_along_curve < total_step_along_curve; steps_along_curve++){
```
adev4a, line 1177: nit: can we update the iterator name to "step"

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: Ah I think I missed the future_points having the state. I removed this line

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: renamed

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: rewrote the log

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -1151,53 +1151,94 @@ namespace basic_autonomy
...
+                total_step_along_curve = 1; // This also avoids division by zero
+            }
 
-                all_sampling_points.push_back(p);
+            std::vector<lanelet::BasicPoint2d> resampled_points;
+            resampled_points.reserve(total_step_along_curve);
 
+            double scaled_steps_along_curve = 0.0; // from 0 (start) to 1 (end) for the whole trajectory
+            for(int step = 0; step < total_step_along_curve; step++){
                 scaled_steps_along_curve += 1.0 / total_step_along_curve;
```
adev4a, line 1178: This should probably be incremented by `(step) / (total_step_along_curve - 1)`

**With current method** 
for step = 0 -> 10:
scaled_steps range from (0.1 -> 0.9)

**With suggested update**
for step = 0->10:
scaled_steps range from (0 ->1)

This is probably why the vehicle current_pose was being added back to the resampled point?

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -1151,53 +1151,94 @@ namespace basic_autonomy
...
+            RCLCPP_DEBUG_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER), "Got resampled points with size:" << resampled_points.size());
+
+            // Compute downtracks for the resampled points
+            std::vector<double> resampled_downtracks = carma_wm::geometry::compute_arc_lengths(resampled_points);
+
+            // Interpolate speeds based on downtracks
+            std::vector<double> resampled_speeds;
+            resampled_speeds.reserve(resampled_points.size());
+
+            for (const auto& downtrack : resampled_downtracks) {
```
adev4a, line 1191: Do we need this manual speed interpolation? Seems like moving average filter should smooth this out?

MishkaMN: Sorry I forgot we discussed this offline. I remove the interpolation.

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: It was actually missign the last point i think (until 0.9). So I included the last point

adev4a (dismissed):

PR #2534, opened 2025-04-09
Feature/add carma autoupdate

EricChen-Lei:
# PR Details
## Description

added daily auto update feature to carma script
To start using type "carma autoupdate_on hh mm"
hh is the hour in a day, from 0 to 23
mm is the minute, from 00 to 59
To stop the auto update type "carma autoupdate_stop" 
## Related Jira Key
OMDO-38

## Types of changes

## Checklist:

### DIFF
--- a/engineering_tools/carma
+++ b/engineering_tools/carma
@@ -497,6 +497,33 @@ carma-config__status() {
     fi
 }
 
+carma__autoupdate_stop() {
+  crontab -l | grep -v "docker pull" | crontab -
+  echo "Docker image auto-updates stopped."
+}
+
+carma__autoupdate_on() {
+  local hour="$1"
+  local minute="$2"
+
+  if ! [[ "$hour" =~ ^([0-1]?[0-9]|2[0-3])$ ]] || ! [[ "$minute" =~ ^([0-5]?[0-9])$ ]]; then
+    echo "Invalid hour or minute. Use hour (0-23) and minute (0-59)."
+    return 1
+  fi
+
+  # local images=$(__get_image_from_config 'carma-platform\|carma-messenger')
+
+  #if [ -z "$images" ]; then
+  #  echo "Could not retrieve images from carma-config. Please ensure carma-config is set."
+  #  return 1
+  #fi
+
+  # Schedule the Docker image updates using cron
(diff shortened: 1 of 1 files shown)


paulbourelly999 (approved):
Looks good to me.

engineering_tools/carma
```diff
@@ -497,6 +497,27 @@ carma-config__status() {
...
+  local minute="$2"
+
+  if ! [[ "$hour" =~ ^([0-1]?[0-9]|2[0-3])$ ]] || ! [[ "$minute" =~ ^([0-5]?[0-9])$ ]]; then
+    echo "Invalid hour or minute. Use hour (0-23) and minute (0-59)."
+    return 1
+  fi
+  COMPOSE_ENV_FILES_WITH_DEFAULT="$(pwd)/../.env"
+
+  # Schedule the Docker image updates using cron
+  (crontab -l 2>/dev/null; echo "${minute} ${hour} * * * docker run --rm --volumes-from carma-config:ro--ntrypoint cat busybox:latest /opt/carma/vehicle/config/docker-compose.yml | COMPOSE_ENV_FILES="$COMPOSE_ENV_FILES_WITH_DEFAULT", docker compose -f - pull --ignore-pull-failures") | crontab -
```
kjrush, line 516: ```carma-config:ro--ntrypoint``` Should that be carma-config:ro --entrypoint?

Also, I recall from the past I think that if docker mostly checks the tags to see if an image is equivalent, not the actual hashes. E.g., if you have `.../carma-platform:develop` locally but there's a newer one on the remote repo, when you run `docker pull .../carma-platform:develop` it'll think you already have one tagged develop and not pull the newer one from remote.

Have we confirmed that behavior isn't an issue here? If it is, we may need to add a flag to force the pull, or possibly delete or temporarily re-tag the local images before pulling?

kjrush (changes requested):
See comment about typo, and would be good to get confirmation on other possible concern in that comment.

engineering_tools/carma
EricChen-Lei: Yes, thank you for catching the typo.
As for docker compose pull, I think the behavior described would be closer to docker compose up, I think pull checks if there's a newer version remote every time we run it.

kjrush (approved):

PR #2540, opened 2025-04-17
Overwrite pedestrian speed and orientation

MishkaMN:
# PR Details
## Description
1. This PR overwrites the speed and the orientation of the PSM and External Object (which means SDSM) message to have hardcoded configurable values. 
NOTE1: Only tested PSM and not SDSM.
NOTE2: Default orientation is just kind of north. One can use Rviz2 to use "2D select pose" to get desired quaternion printed in the rviz2's terminal and use that.
2. This actually has a fix for PSM prediction, which was using transformations wrong.
3. This needed the PSM to be broadcasted at 20Hz OR CP pipeline to broadcast at 10Hz otherwise the object is blinking (because current logic needs at least empty external_objects to be published with a frequency that matches PSM)

Tested using basic_travel_simulator. Used command to publish PSM (on TFHRC_lanechange_pocx.osm map):
```
ros2 topic pub /message/incoming_j2735_psm j2735_v2x_msgs/msg/PSM "presence_vector: 15
basic_type:
  type: 1
sec_mark:
  millisecond: 36000
msg_cnt:
  msg_cnt: 1
id:
  id: [1, 2, 3, 4]
position:
  latitude: 389548206
  longitude: -771483733
  elevation: 1000
  elevation_exists: true
accuracy:
  semi_major: 100
  semi_minor: 100
  orientation: 0
speed:
  velocity: 1000
heading:
  heading: 9000
path_prediction:
  radius_of_curvature: 10000
  confidence: 100
propulsion:
  choice: 1
  human:
    type: 1" --rate 10

```
Interesting thing is that the rviz is updating its location. Maybe it is motion_computation pruning and updating, not sure but it works nicely.

## Types of changes

## Checklist:

### DIFF
--- a/carma/launch/carma_src.launch.py
+++ b/carma/launch/carma_src.launch.py
@@ -197,7 +197,7 @@ def generate_launch_description():
     is_cp_mot_enabled = LaunchConfiguration('is_cp_mot_enabled')
     declare_is_cp_mot_enabled = DeclareLaunchArgument(
         name='is_cp_mot_enabled',
-        default_value = 'False',
+        default_value = 'True',
         description = 'True if user wants Cooperative Perception capability using Multiple Object Tracking to be enabled'
     )
 
--- a/carma/launch/environment.launch.py
+++ b/carma/launch/environment.launch.py
@@ -525,7 +525,7 @@ def generate_launch_description():
                 ],
                 remappings=[
                     ("output/track_list", "cooperative_perception_track_list"),
-                    ("input/detection_list", "filtered_detection_list"), #TODO effectivly bypass
+                    ("input/detection_list", "filtered_detection_list"),
                 ],
                 parameters=[
                     cp_multiple_object_tracker_node_file,
--- a/carma/launch/environment.launch.py
+++ b/carma/launch/environment.launch.py
@@ -508,7 +508,7 @@ def generate_launch_description():
                     {'--log-level' : GetLogLevel('cp_track_list_to_external_object_list_node', env_log_levels) },
                 ],
                 remappings=[
-                    ("input/track_list", "filtered_detection_list"),
+                    ("input/detection_list", "filtered_detection_list"),
                     ("output/external_object_list", "fused_external_objects"),
                 ],
                 parameters=[
--- a/carma/launch/environment.launch.py
+++ b/carma/launch/environment.launch.py
@@ -508,7 +508,7 @@ def generate_launch_description():
                     {'--log-level' : GetLogLevel('cp_track_list_to_external_object_list_node', env_log_levels) },
                 ],
                 remappings=[
-                    ("input/detection_list", "filtered_detection_list"),
+                    ("input/track_list", "cooperative_perception_track_list"),
                     ("output/external_object_list", "fused_external_objects"),
                 ],
                 parameters=[
@@ -525,7 +525,7 @@ def generate_launch_description():
                 ],
                 remappings=[
                     ("output/track_list", "cooperative_perception_track_list"),
-                    #("input/detection_list", "filtered_detection_list"), #TODO effectivly bypass
+                    ("input/detection_list", "filtered_detection_list"), #TODO effectivly bypass
                 ],
                 parameters=[
                     cp_multiple_object_tracker_node_file,
--- a/carma/launch/environment.launch.py
+++ b/carma/launch/environment.launch.py
@@ -251,7 +251,7 @@ def generate_launch_description():
         ]
     )
 
-                                                          
+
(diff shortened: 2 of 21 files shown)


motion_computation/src/psm_to_external_object_convertor.cpp
```diff
@@ -306,16 +318,14 @@ std::vector<geometry_msgs::msg::Pose> sample_2d_linear_motion(
   double total_dt = 0;
 
   while (total_dt < period) {
-    // Compute the 2d position and orientation in the Pose frame
+    // Increment time
     total_dt += step_size;
-    double dx_from_start = velocity * total_dt;  // Assuming linear motion in pose frame
+    double distance = velocity * total_dt;  // Total distance traveled
 
-    double x = pose.position.x + dx_from_start;
```
adev4a, line 313: Still trying to wrap my head around this but putting down my though - the pose.position.x in the map frame should be a fairly large number - having an error of this size should be noticeable.

lci_strategic_plugin/config/parameters.yaml
adev4a, line 1: Are these changes required?

docker/checkout.bash
adev4a, line 1: Lets revert these changes since they will break CI  build

docker/install_dependencies.sh
adev4a, line 1: Same note here about reverting changes here since they will break CI

carma/launch/environment.launch.py
```diff
@@ -525,7 +525,7 @@ def generate_launch_description():
                 ],
                 remappings=[
                     ("output/track_list", "cooperative_perception_track_list"),
-                    ("input/detection_list", "filtered_detection_list"),
+                    ("input/detection_list", "filtered_detection_list"), #TODO effectivly bypass
```
adev4a, line 528: I'm guessing there is no change required here since this has already been tested to work well - can we update this comment with more description of the suggested future change or remove the comment?

MishkaMN: yes this is not needed

docker/install_dependencies.sh
MishkaMN: reverted

docker/checkout.bash
MishkaMN: reverted

motion_computation/src/psm_to_external_object_convertor.cpp
MishkaMN: Correct. It was wrong because it is arbitrarily adding the x value of the original pose, which is already captured by the first transform.

carma/launch/carma_src.launch.py
```diff
@@ -197,7 +197,7 @@ def generate_launch_description():
     is_cp_mot_enabled = LaunchConfiguration('is_cp_mot_enabled')
     declare_is_cp_mot_enabled = DeclareLaunchArgument(
         name='is_cp_mot_enabled',
-        default_value = 'False',
+        default_value = 'True',
```
JonSmet, line 200: Is MOT being enabled so that the CP pipeline will publish, which the PR description states is required?

JonSmet (changes requested):
PR is looking good. Made a request for a "nice to have" change, but it is alright if there isn't time to implement.

Will wait to approve until this is tested on a vehicle. We will only test this on a vehicle if the team is ready to test with yielding to the real pedestrian dummy PSMs.

carma/launch/carma_src.launch.py
MishkaMN: This is largely due to previous implementation design where we expected external objects to be always published with other msgs like PSM. And having one subscriber's intention is to synchronize all timestamps into one.

motion_computation/config/parameters.yaml
```diff
@@ -2,7 +2,13 @@
 prediction_time_step: 0.2
 
 # Period of prediction in seconds
-prediction_period: 10.0
+prediction_period: 30.0
+
```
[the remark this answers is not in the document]
JonSmet: Oh yes thanks for the reminder. I was thinking that a negative pedestrian_speed could result in the orientation still being overwritten, but that is not the case.

Can we add a comment to pedestrian_orientation as well to state that a negative pedestrian_speed will result in the orientation not being overwritten?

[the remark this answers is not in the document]
MishkaMN: Ah yeah added the comment

adev4a (approved):

JonSmet:
Closing this PR since FLIR Camera PSMs will not be used for the demonstration, so overwriting of the heading/speed values is not necessary.

PR #2549, opened 2025-04-23
Update BSM Generator to correctly pull heading from orientation in Cu…

paulbourelly999:
# PR Details
## Description
Update BSM Generator to pull heading from Pose since this should work both in SIM and in Vehicle

## Related Jira Key
TT-177

## Motivation and Context
Fix BSM heading in SIM

## How Has This Been Tested?
Unit testing 

## Types of changes

## Checklist:

### DIFF
--- a/bsm_generator/src/bsm_generator_worker.cpp
+++ b/bsm_generator/src/bsm_generator_worker.cpp
@@ -97,4 +97,29 @@ namespace bsm_generator
     {
         return std::max(std::min(heading, 359.9875f), 0.0f);
     }
+
+    float BSMGeneratorWorker::getHeading(const geometry_msgs::msg::Quaternion & quaternion)
+    {
+        tf2::Quaternion orientation;
+        orientation.setX(quaternion.x);
+        orientation.setY(quaternion.y);
+        orientation.setZ(quaternion.z);
+        orientation.setW(quaternion.w);
+      
+        double roll;
+        double pitch;
+        double yaw;
+        tf2::Matrix3x3(orientation).getRPY(roll, pitch, yaw);
+        // Convert yaw radians to degrees
+        yaw = yaw * 180 / M_PI;
+        // Convert to NED frame
+        yaw = -yaw+90;
+        // Convert to range [0, 360]
+        yaw = std::fmod(yaw, 360.0);
+        if ( yaw < 0  )
+        {
+          yaw = 360 + yaw;
+        }
+        return static_cast<float>(yaw);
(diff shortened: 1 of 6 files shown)


bsm_generator/src/bsm_generator_worker.cpp
```diff
@@ -97,4 +97,29 @@ namespace bsm_generator
...
+        orientation.setW(quaternion.w);
+      
+        double roll;
+        double pitch;
+        double yaw;
+        tf2::Matrix3x3(orientation).getRPY(roll, pitch, yaw);
+        // Convert yaw radians to degrees
+        yaw = yaw * 180 / M_PI;
+        // Convert to NED frame
+        yaw = -yaw+90;
```
willjohnsonk, line 116: I would suggest we do this conversion in quaternion world so that we do not run into bad elevation values down the line that comes from doing calculations after removing a dimension. Here's a link that provides a good (albeit long) explanation with examples: https://stackoverflow.com/questions/49790453/enu-ned-frame-conversion-using-quaternions

bsm_generator/test/test_bsm_generator_worker.cpp
```diff
@@ -107,6 +107,21 @@ TEST(BSMWorkerTest, testHeading)
     EXPECT_NEAR(300.001f, worker.getHeadingInRange(300.001f), 0.0001);
 }
 
+TEST(BSMWorkerTest, testHeading) {
```
willjohnsonk, line 110: I'd add a couple of more test cases with various orientations (looked like CARLA provided rpy and associated quat values?) as a sanity check if we have verified values across the board

willjohnsonk (changes requested):

bsm_generator/src/bsm_generator_worker.cpp
afrye51: Are we confident that the PoseStamped topic's frame is the same as the BSM frame? (save ENU->NED frame conversion)

paulbourelly999: This conversion is based on this https://github.com/Unity-Technologies/ROS-TCP-Connector/issues/163. I have also confirmed that on CDASim testing data that this value is ENU since driving east gives 0.0, 0.0, 0.0, 0.0 quaterian. I have also tested this conversion for the many different quaterians. See unit tests.

willjohnsonk (approved):
quaternion wizard 🔮

PR #2573, opened 2025-06-04
install needed items for public ami deploy

EricChen-Lei:
install nice xdcv and packages the public ami is missing

# PR Details
## Description
used new public ami from https://aws.amazon.com/marketplace/pp/prodview-ch24i42e34ezm?applicationId=AWS-EC2-Console&ref_=beagle&sr=0-5
added installation of missing packages inside the public ami

## Related Jira Key
STRT-11

## Types of changes

## Checklist:

### DIFF
--- a/engineering_tools/carma_script_extensions/cloud-sim
+++ b/engineering_tools/carma_script_extensions/cloud-sim
@@ -173,6 +173,10 @@ cloud-sim__start() {
     local CONFIG_IMAGE=$(cat ${CARMA_CLOUD_SIM_CONFIG_PATH} | grep "CONFIG_IMAGE=" | sed "s|CONFIG_IMAGE=||")
     local AMI_ID=$(cat ${CARMA_CLOUD_SIM_CONFIG_PATH} | grep "AMI_ID=" | sed "s|AMI_ID=||")
     local REGION=$(cat ${CARMA_CLOUD_SIM_CONFIG_PATH} | grep "REGION=" | sed "s|REGION=||")
+    if [ -z "$REGION" ]; then
+        REGION="us-east-1"
+        echo "No region specified in config. Defaulting to us-east-1."
+    fi
 
     if [ -z "$USER_ARN" ]; then
         echo "Configuration does not include AWS User ARN. Please run: carma cloud-sim config -a <User ARN>"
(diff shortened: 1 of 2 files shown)


engineering_tools/carma_script_extensions/cloud-sim
```diff
@@ -173,6 +173,10 @@ cloud-sim__start() {
     local CONFIG_IMAGE=$(cat ${CARMA_CLOUD_SIM_CONFIG_PATH} | grep "CONFIG_IMAGE=" | sed "s|CONFIG_IMAGE=||")
     local AMI_ID=$(cat ${CARMA_CLOUD_SIM_CONFIG_PATH} | grep "AMI_ID=" | sed "s|AMI_ID=||")
     local REGION=$(cat ${CARMA_CLOUD_SIM_CONFIG_PATH} | grep "REGION=" | sed "s|REGION=||")
+    if [ -z "$REGION" ]; then
```
kjrush, line 176: Did you mention before we needed some sort of mapping table between region and AMI ID? I don't see that in this PR, are we only supporting us-east-1?

EricChen-Lei: I think this is kind of the tricky part, cause the region here will have to tied to the ami that they chose to use when using the public ami, the process is that you will need to choose a region when you first subscribe to the public ami, so it kind of depends on the user's choice

engineering_tools/xil-cloud-formation-v0.yaml
```diff
@@ -466,6 +502,48 @@ Resources:
...
+
+          # Enable and start dcvserver on boot
+          sudo systemctl enable dcvserver
+          sudo systemctl start dcvserver
+
+          # (Optional) Create a virtual DCV session
+          sudo -u ubuntu dcv create-session ubuntu --owner ubuntu --type virtual
+          
+          # Create systemd service to recreate DCV session after reboot
+          cat <<EOF | sudo tee /etc/systemd/system/create-dcv-session.service
```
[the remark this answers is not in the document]
EricChen-Lei: yes, this does start up the session after reboot, I think bash doesn't treat sudo tee as a part of the heredoc, it's outside the quoted input

PR #2583, opened 2025-06-19
CDAD-105 Handle cases obj disappearing

MishkaMN:
# PR Details
## Types of changes

## Checklist:

### DIFF
--- a/yield_plugin/config/parameters.yaml
+++ b/yield_plugin/config/parameters.yaml
@@ -1,8 +1,6 @@
-
-# Adjustment factor for safe and comfortable acceleration/deceleration
-# Value type: Desired
-# No unit
-acceleration_adjustment_factor: 1.0
+###################################################
+# COLLISION CHECK RELATED PARAMETERS
+###################################################
 # Time horizon for collision detection for vehicles on the route
 # NOTE: not applied for pedestrians or other non-vehicle obstacles on the road
 # Units: s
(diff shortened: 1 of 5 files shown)


yield_plugin/config/parameters.yaml
```diff
@@ -12,11 +10,38 @@ on_route_vehicle_collision_horizon_in_s: 10.0
 # Units: m/s
 # Value type: Desired
 obstacle_zero_speed_threshold_in_ms: 0.25
+# Time horizon until collision to commit to last stopping trajectory
+# NOTE: Only applies to situations if the ego was expected to stop completely due to the collision
+# not just yield and follow with a lower speed
+# Units: s
+time_horizon_until_collision_to_commit_to_stop_in_s: 3.0
+# Number of consecutive times the vehicle detects that the obstacle is behind
```
paulbourelly999, line 18: What does behind in this context?

yield_plugin/src/yield_plugin.cpp
```diff
@@ -299,10 +342,16 @@ namespace yield_plugin
         yield_trajectory = update_traj_for_object(original_trajectory, external_objects_, initial_velocity); // Compute the trajectory
       }
 
+
       // return original trajectory if no difference in trajectory points a.k.a no collision
       if (fabs(get_trajectory_end_time(original_trajectory) - get_trajectory_end_time(yield_trajectory)) < EPSILON)
       {
+        
         resp->trajectory_plan = original_trajectory;
+        // Reset the collision prevention variables
```
paulbourelly999, line 351: Is this intended to correct logic to account for multiple stops of the vehicle.

yield_plugin/config/parameters.yaml
paulbourelly999: remove this configuration parameter as currently it is unreachable since when there is no collision/object is behind the method returns nullptr anyway and resets the consecutive counter prevent it from incrementing to the limit (see)
https://github.com/usdot-fhwa-stol/carma-platform/blob/f4ec70ad51c2691eb66d595589c0ac2152a68186/yield_plugin/src/yield_plugin.cpp#L838

adev4a: Added a TODO comment to investigate instead of removing the change in this PR.

paulbourelly999 (dismissed):

PR #2586, opened 2025-07-01
CDAD-137 Fix Traffic Incident Parser so that it doesn’t ignore some messages and speed limit change

saina-ramyar:
# PR Details
## Description

Two issues are addressed in this PR:
CDAD-137: TIM Use case: traffic incident parser is ignoring some messages
CDAD-138: TIM Use case: carma-platform is ignoring speed limit change from ERV

It seems like if the move-over MOM is published before the route is selected, it is being ignored. The reason is that a geofence is created from the first received MOM, but since route is not loaded yet, it is not applied by world model.
However, the Traffic Incident Parser will save the message and will ignore all the future messages if they are the same as the first one. So when the route is selected, new geofences wont be created anymore.

There was also a mismatch between speed units. The MOM has an advisory speed value in mph, which was not being converted to m/s

## Related Jira Key

https://usdot-carma.atlassian.net/browse/CDAD-137
https://usdot-carma.atlassian.net/browse/CDAD-138

## Motivation and Context

To have a functioning TIM use case

## Types of changes

## Checklist:

### DIFF
--- a/traffic_incident_parser/include/traffic_incident_parser/traffic_incident_parser_worker.hpp
+++ b/traffic_incident_parser/include/traffic_incident_parser/traffic_incident_parser_worker.hpp
@@ -46,7 +46,7 @@ class TrafficIncidentParserWorker
  public:
 
   using PublishTrafficControlCallback = std::function<void(const carma_v2x_msgs::msg::TrafficControlMessage&)>;
-  constexpr double MphToMetersPerSec = 0.44704; //Convert miles per hour to meters per second
+  const double MphToMetersPerSec = 0.44704; //Convert miles per hour to meters per second
 
 
   /*!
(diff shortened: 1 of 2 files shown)


john-chrosniak (dismissed):

traffic_incident_parser/include/traffic_incident_parser/traffic_incident_parser_worker.hpp
```diff
@@ -46,6 +46,8 @@ class TrafficIncidentParserWorker
  public:
 
   using PublishTrafficControlCallback = std::function<void(const carma_v2x_msgs::msg::TrafficControlMessage&)>;
+  constexpr double MphToMetersPerSec = 0.44704; //Convert miles per hour to meters per second
```
john-chrosniak: Looks like CI doesn't like this being declare a `constexpr`, maybe just need `const`?

saina-ramyar: yeah I had fixed it locally but didnt commit.

adev4a (approved):

PR #2587, opened 2025-07-02
Fix is object behind logic in Yield

adev4a:
# PR Details

This PR updates the implementation in yield plugin for checking if an object is behind the vehicle. Currently collision detection is happening over a defined radius, so if an object is approaching the vehicle from behind the vehicle moving path it will still be considered under collision. 
Current implementation is checking over a configured duration if the object is detected as behind, before rejecting collision. The map storing the count is never incremented, this change set adds an increment to the count on successive iterations. 
Also the count was being reset at every iteration, which invalidates the configuration parameter and is removed. 

## Types of changes

## Checklist:

### DIFF
--- a/yield_plugin/src/yield_plugin.cpp
+++ b/yield_plugin/src/yield_plugin.cpp
@@ -746,12 +746,6 @@ namespace yield_plugin
     {
       return true;
     }
-    // if the clearance counter didn't increase by this point, true collision was detected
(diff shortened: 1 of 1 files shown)


MishkaMN:
Discussed over the call that the count is in fact increasing in previous logic. So this PR should be OBE

PR #2589, opened 2025-07-09
Catch sdsm detection error

adev4a:
# PR Details
During testing at Summit Point it was seen that when SDSM's are received before the georefence has completely loaded into the system, it causes carma_cooperative_perception to crash and in turn carma-platform. 
On investigation it appears that the node in trying to project the incoming SDSM detections in the map frame tries to use the existing georeference to make this projection, however it ignores cases where the georeference has not yet been loaded.

This fix adds a try catch to the sdsm to detection object logic, to ignore SDSM's causing the system to crash with an error log.
Alternative would be to add a georeference check explicitly, but invalid messages should not be allowed to crash the system instead should be logged.

## Types of changes

## Checklist:

### DIFF
--- a/carma_cooperative_perception/src/msg_conversion.cpp
+++ b/carma_cooperative_perception/src/msg_conversion.cpp
@@ -569,107 +569,113 @@ auto to_detection_list_msg(
   -> carma_cooperative_perception_interfaces::msg::DetectionList
 {
   carma_cooperative_perception_interfaces::msg::DetectionList detection_list;
-  const auto ref_pos_3d{Position3D::from_msg(sdsm.ref_pos)};
+  try{
+  
+    const auto ref_pos_3d{Position3D::from_msg(sdsm.ref_pos)};
 
-  units::length::meter_t elevation(0.0);
-  if(ref_pos_3d.elevation){
-    elevation = ref_pos_3d.elevation.value();
-  }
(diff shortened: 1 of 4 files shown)


carma_cooperative_perception/src/msg_conversion.cpp
```diff
@@ -569,107 +569,113 @@ auto to_detection_list_msg(
...
+      convert_object_type(detection, common_data.obj_type);
 
-    detection_list.detections.push_back(std::move(detection));
+      detection_list.detections.push_back(std::move(detection));
+    }
   }
-
+  catch (...) {
+    RCLCPP_ERROR_STREAM(rclcpp::get_logger("sdsm_to_detection_list_node"), "Error converting SDSM to object, ignoring sdsm message.")
+  }
```
MishkaMN: If the issue is really only due to missing georeference, we should just add that guard and not catch all error like this.
The georeference guard would be on top of the sdsm_callback https://github.com/usdot-fhwa-stol/carma-platform/blob/fix-cp-sdsm-missing-georeference/carma_cooperative_perception/include/carma_cooperative_perception/sdsm_to_detection_list_component.hpp#L110

MishkaMN (changes requested):

MishkaMN (dismissed):

saina-ramyar (approved):

PR #2596, opened 2025-07-21
Improve plan_delegator to reuse last successful traj if new one failed

MishkaMN:
# PR Details
## Description

I applied a fix to the plan_delegator regarding the flickering trajectory. I confirmed on all the 10 runs that the flickering issue was happening 1-3 times (mostly due to yield_plugin timeout and our basic_autonomy not being able to generate traj near the end of lanelet), but the platform was able to overcome them. 

This happens because plan_delegator publishes a trajectory even if it is not usable because it is considered valid if it has 2 points. So sometimes out of A B C plugins (in that order), B may not generate valid trajectory due to timeout (yielding).  Then if A gave only 5 points, which is not useful as the vehicle is almost passing those points but still valid, plan_delegator uses that trajectory.

We could further discuss what should be valid trajectory.

But for now, I coded it to use last successful trajectory if plan_delegator was not able to generate traj fully from all plugins instructed.

We uploaded the runs in: https://drive.google.com/drive/folders/1-CKYiyqI_TKpzsBhKUxzKMm35RvLhvo1?usp=sharing

## Related GitHub Issue
NA

## Related Jira Key

CDAD-152
CDAD-159
## Motivation and Context
Tempest release testing and ITSWC 2025 Demo

## How Has This Been Tested?
Summit Point 10/10 successful runs
image

In each runs, the issue would have happened following times:
```
July 21 Testing
1. Happened 3 times - highest 0.34s
2. Happened 3 times - highest ILC ExecutionTime: 0.277474 ( ExecutionTime Yield: 0.264261)
3. Happened 2 times - highest ExecutionTime Yield: 0.398444
4. Happened 2 times - highest ILC ExecutionTime: 0.275721 ( ExecutionTime Yield: 0.241054)
5. Happened 1 time - ILC ExecutionTime: 0.268358 ( ExecutionTime Yield: 0.260101), but happened because of 'is at or past the planned end distance. Couldn't generate trajectory"
6. Happened 1 time - ILC ExecutionTime: 0.237326 ( ExecutionTime Yield: 0.225053),  but happened because of 'is at or past the planned end distance. Couldn't generate trajectory"
7. 0 happened
8 Happened 2 times - ILC ExecutionTime: 0.311489 (ExecutionTime Yield: 0.501016), 1 time for " 'is at or past the planned end distance. Couldn't generate trajectory""
9. Happened 1 time -ILC ExecutionTime: 0.317286 (ExecutionTime Yield: 0.288118)
10. Happened 2 times - ILC ExecutionTime: 0.305264 ( ExecutionTime Yield: 0.343301)
```

## Types of changes

## Checklist:

### DIFF
--- a/cooperative_lanechange/src/cooperative_lanechange_node.cpp
+++ b/cooperative_lanechange/src/cooperative_lanechange_node.cpp
@@ -292,7 +292,17 @@ namespace cooperative_lanechange
     RCLCPP_DEBUG_STREAM(get_logger(), "Starting CLC downtrack: " << maneuver_plan[0].lane_change_maneuver.start_dist);
 
     if(current_downtrack < maneuver_plan[0].lane_change_maneuver.start_dist - config_.starting_downtrack_range){
+<<<<<<< Updated upstream
       RCLCPP_DEBUG_STREAM(get_logger(), "Lane change trajectory will not be planned. current_downtrack is more than " << config_.starting_downtrack_range << " meters before starting CLC downtrack");
+=======
+      RCLCPP_ERROR_STREAM(get_logger(), "Lane change trajectory will not be planned. current_downtrack is more than " << config_.starting_downtrack_range << " meters before starting CLC downtrack");
+
+      std::chrono::system_clock::time_point end_time = std::chrono::system_clock::now();  // Planning complete
+
+      auto duration = end_time - start_time;
+      RCLCPP_ERROR_STREAM(get_logger(), "CLC ExecutionTime: " << std::chrono::duration<double>(duration).count());
+      RCLCPP_WARN_STREAM(get_logger(), "CLC ExecutionTime: " << std::chrono::duration<double>(duration).count());
+>>>>>>> Stashed changes
       return;
     }
     auto current_lanelets = lanelet::geometry::findNearest(wm_->getMap()->laneletLayer, veh_pos, 10);
@@ -424,7 +434,15 @@ namespace cooperative_lanechange
       p.planner_plugin_name = get_plugin_name();
     }
 
+<<<<<<< Updated upstream
     return;
+=======
+    std::chrono::system_clock::time_point end_time = std::chrono::system_clock::now();  // Planning complete
+
+    auto duration = end_time - start_time;
+    RCLCPP_ERROR_STREAM(get_logger(), "CLC ExecutionTime: " << std::chrono::duration<double>(duration).count());
+    RCLCPP_WARN_STREAM(get_logger(), "CLC ExecutionTime: " << std::chrono::duration<double>(duration).count());
+>>>>>>> Stashed changes
   }
 
   void CooperativeLaneChangePlugin::add_trajectory_to_response(carma_planning_msgs::srv::PlanTrajectory::Request::SharedPtr req,
--- a/cooperative_lanechange/src/cooperative_lanechange_node.cpp
+++ b/cooperative_lanechange/src/cooperative_lanechange_node.cpp
@@ -292,17 +292,14 @@ namespace cooperative_lanechange
     RCLCPP_DEBUG_STREAM(get_logger(), "Starting CLC downtrack: " << maneuver_plan[0].lane_change_maneuver.start_dist);
 
     if(current_downtrack < maneuver_plan[0].lane_change_maneuver.start_dist - config_.starting_downtrack_range){
-<<<<<<< Updated upstream
-      RCLCPP_DEBUG_STREAM(get_logger(), "Lane change trajectory will not be planned. current_downtrack is more than " << config_.starting_downtrack_range << " meters before starting CLC downtrack");
-=======
       RCLCPP_ERROR_STREAM(get_logger(), "Lane change trajectory will not be planned. current_downtrack is more than " << config_.starting_downtrack_range << " meters before starting CLC downtrack");
 
       std::chrono::system_clock::time_point end_time = std::chrono::system_clock::now();  // Planning complete
 
       auto duration = end_time - start_time;
-      RCLCPP_ERROR_STREAM(get_logger(), "CLC ExecutionTime: " << std::chrono::duration<double>(duration).count());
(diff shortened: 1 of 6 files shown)


yield_plugin/src/yield_plugin.cpp
```diff
@@ -739,10 +746,10 @@ namespace yield_plugin
         const auto distance{std::hypot(x1 - x2, y1 - y2)};
 
         smallest_dist = std::min(distance, smallest_dist);
-        RCLCPP_DEBUG_STREAM(nh_->get_logger(), "Smallest_dist: " << smallest_dist << ", distance: " << distance << ", dt: " << dt
```
MishkaMN, line 742: Logging too much, I removed it, and added in the end

plan_delegator/src/plan_delegator.cpp
```diff
@@ -716,13 +728,23 @@ namespace plan_delegator
...
+            last_successful_traj_ = trajectory_plan;
             traj_pub_->publish(trajectory_plan);
         }
         else
         {
             RCLCPP_WARN_STREAM(rclcpp::get_logger("plan_delegator"),
-                "Guidance is engaged, but planned trajectory has less than 2 points. " <<
+                "Guidance is engaged, but new planned trajectory has less than 2 points. " <<
                 "It will not be published!");
+            if (last_successful_traj_.has_value())
```
adev4a: Would we expect this last successful trajectory to always be within a reasonable duration prior to a failure? It would feel a lot more comfortable to me if we were checking this against a fixed time (Maybe something specific like 1-2 iterations of planning)

MishkaMN: I made it 3 consecutive tries until failure because I saw cases it failed 2 consecutively. 
Made it a configurable parameter as it might work differently across different computers.

cooperative_lanechange/src/cooperative_lanechange_node.cpp
```diff
@@ -292,7 +292,16 @@ namespace cooperative_lanechange
...
 
     if(current_downtrack < maneuver_plan[0].lane_change_maneuver.start_dist - config_.starting_downtrack_range){
-      RCLCPP_DEBUG_STREAM(get_logger(), "Lane change trajectory will not be planned. current_downtrack is more than " << config_.starting_downtrack_range << " meters before starting CLC downtrack");
+      RCLCPP_WARN_STREAM(get_logger(),
+      "Lane change trajectory will not be planned. current_downtrack is more than "
+      << config_.starting_downtrack_range << " meters before starting CLC downtrack");
+
+      std::chrono::system_clock::time_point end_time = std::chrono::system_clock::now();  // Planning complete
+
+      auto duration = end_time - start_time;
```
JonSmet, line 303: I don't see `start_time` defined in this scope, will this build?

cooperative_lanechange/src/cooperative_lanechange_node.cpp
```diff
@@ -424,7 +433,12 @@ namespace cooperative_lanechange
       p.planner_plugin_name = get_plugin_name();
     }
 
-    return;
+    std::chrono::system_clock::time_point end_time = std::chrono::system_clock::now();  // Planning complete
+
+    auto duration = end_time - start_time;
```
JonSmet, line 440: Similar comment to above - I don't see `start_time` defined in this scope, will this build?

cooperative_lanechange/src/cooperative_lanechange_node.cpp
```diff
@@ -292,7 +292,16 @@ namespace cooperative_lanechange
...
     if(current_downtrack < maneuver_plan[0].lane_change_maneuver.start_dist - config_.starting_downtrack_range){
-      RCLCPP_DEBUG_STREAM(get_logger(), "Lane change trajectory will not be planned. current_downtrack is more than " << config_.starting_downtrack_range << " meters before starting CLC downtrack");
+      RCLCPP_WARN_STREAM(get_logger(),
+      "Lane change trajectory will not be planned. current_downtrack is more than "
+      << config_.starting_downtrack_range << " meters before starting CLC downtrack");
+
+      std::chrono::system_clock::time_point end_time = std::chrono::system_clock::now();  // Planning complete
+
+      auto duration = end_time - start_time;
+      RCLCPP_DEBUG_STREAM(
```
JonSmet, line 304: Is this the right location for these newly added statements, since CLC planning has just started at this point?

JonSmet (changes requested):

cooperative_lanechange/src/cooperative_lanechange_node.cpp
MishkaMN: Oops, I don't know how I missed that

cooperative_lanechange/src/cooperative_lanechange_node.cpp
MishkaMN: added

cooperative_lanechange/src/cooperative_lanechange_node.cpp
MishkaMN: yeah because the function returns here. I also have on at the end outside this if scope.

plan_delegator/src/plan_delegator.cpp
```diff
@@ -730,21 +732,41 @@ namespace plan_delegator
...
                 traj_pub_->publish(last_successful_traj_.value());
             }
-
+            // case where traj generation fails from the beginning
+            else if (!last_successful_traj_.has_value() &&
+                consecutive_traj_gen_failure_num_ <= config_.max_traj_generation_reattempt)
+            {
+                RCLCPP_WARN_STREAM(rclcpp::get_logger("plan_delegator"),
+                    "Instead, tried publishing last available trajectory, but it is not available!");
+            }
```
adev4a, line 762: This should be throwing an error as well, since we don't have a valid trajectory to return. Do we need this to be different from the else condition?

adev4a: Discussed offline, not throwing an error here because we want a minimum number of replans before throwing an error.

adev4a (approved):

JonSmet (approved):

PR #2601, opened 2025-07-22
Account expired/passed maneuvers in indexing in plan_delegator

MishkaMN:
# PR Details
## Description

Currently plan_delegator could ignore a whole maneuver intermittently. 

For context, ILC can plan over multiple maneuvers using the related_maneuvers field even if only 1 maneuver was sent to be planned. Due to this, plan_delegator accounts for it, and increments its index for what maneuver to plan next using related_maneuvers size.

The problem arises when plan_delegator detects that current maneuver is passed or expired (due to vehicle traveling and planning is little delayed due to 1hz update rate of arbitrator), plan_delegator increments its index and moves on to the next one. However, it doesn't account for that early increment later when it adjusts its index for the related_maneuvers. 

For example, if it had if 2 contiguous LANE_FOLLOW maneuvers, but one is expired, its index increments to 1 (index 0 to 1). However, later when it adjusts for its related_maneuvers, it forgot it had 1 expired maneuver and shifts 2 indexes anyways, adding extra 1 than needed (index 1 + 2 = 3). So the index is 3 now, although 2 should have been planned.

This results in the next maneuver being ignored. This is intermittent because 1hz planning of strategic_plugin may or may not update in time for plan_delegator to not make that mistake.

And I currently only included LANE_FOLLOW as detection as I don't think other plugins use related_maneuvers, but wanted to be explicit.

Example log: 2025-07-22-11-44-34-332730-carma-1
Example mcap: rosbag2_2025-07-22_114434
in here: https://drive.google.com/drive/u/2/folders/18_9g4zMEtdCDIsmRTFPdJowpdmDA1GP-
Picture of CLC being ignored:
image

## Related GitHub Issue
NA

## Related Jira Key
CDAD-159
CDAD-152

## Motivation and Context
Summit Point testing tempest release

## How Has This Been Tested?
TODO
Black Pacifica Summit point

## Types of changes

## Checklist:

### DIFF
--- a/plan_delegator/src/plan_delegator.cpp
+++ b/plan_delegator/src/plan_delegator.cpp
@@ -607,7 +607,7 @@ namespace plan_delegator
 
         // Track the index of the starting maneuver in the maneuver plan that this trajectory plan service request is for
         uint16_t current_maneuver_index = 0;
-
+        int count_expired_maneuvers = 0;
         // Loop through maneuver list to make service call to applicable Tactical Plugin
         while(current_maneuver_index < latest_maneuver_plan_.maneuvers.size())
         {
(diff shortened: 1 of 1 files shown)


plan_delegator/src/plan_delegator.cpp
```diff
@@ -607,7 +607,7 @@ namespace plan_delegator
 
         // Track the index of the starting maneuver in the maneuver plan that this trajectory plan service request is for
         uint16_t current_maneuver_index = 0;
-
+        int count_expired_maneuvers = 0;
```
JonSmet, line 610: We skip manuevers that are "expired" and "passed", so maybe we can make this more general by renaming to `count_skipped_maneuvers`?

MishkaMN:
Upon testing, I may have been wrong on this. New hypothesis is here: https://github.com/usdot-fhwa-stol/carma-platform/pull/2602

PR #2631, opened 2025-09-10
Add support for a global parameter override yaml file in carma-config

MishkaMN:
# PR Details
## Description
Adding a global parameter override yaml file to carma-config. This will help version control any map or scenario specific parameters using carma-config. The GlobalParamsOverride.yaml file stored in the carma-config will include the parameters that will override the parameters in each individual package's yaml file.

This also has a benefit of not needing to create docker volumes from carma-config to multiple packages. We can add the desired parameter in carma-config's GlobalParamsOverride.yaml file to directly override package-level parameters.

Design Document: https://usdot-carma.atlassian.net/wiki/spaces/CRMPLT/pages/3759669249/High+Level+Design+-+CARMA+Config+Update+Configuration+Management

## Related Issue
CDAS-74
Supports: https://github.com/usdot-fhwa-stol/carma-config/pull/475

## Motivation and Context
- Some tests and demos require special parameters in carma-platform docker image. Those values are not released so it takes time to re-discover those again. With the new way, carma-config truly houses all the configurations needed to make the demo work again (TODO: in the near future, we are planning to include necessary route and map files to fully complete the configurations as well, so map and routes version controlled as well and works out of the box).
- CDASim had multiple params.yaml file for each plugins. This approach gives them the unified way to change the parameters.

## How Has This Been Tested?
Locally on sim PC. All nested namespaces were tested.

## Types of changes

## Checklist:

### DIFF
--- a/carma/launch/carma_src.launch.py
+++ b/carma/launch/carma_src.launch.py
@@ -82,7 +82,7 @@ def generate_launch_description():
     declare_vehicle_config_dir_arg = DeclareLaunchArgument(
         name = 'vehicle_config_dir',
         default_value = "/opt/carma/vehicle/config",
-        description = "Path to file containing vehicle config directories"
+        description = "Path to vehicle configuration directory populated by carma-config"
     )
 
     vehicle_characteristics_param_file = LaunchConfiguration('vehicle_characteristics_param_file')
--- a/carma/launch/carma_src.launch.py
+++ b/carma/launch/carma_src.launch.py
@@ -105,7 +105,7 @@ def generate_launch_description():
     global_params_override_file = LaunchConfiguration('global_params_override_file')
     declare_global_params_override_file_arg = DeclareLaunchArgument(
         name = 'global_params_override_file',
-        default_value = [vehicle_config_dir, "/global_params_overwrite.yaml"],
+        default_value = [vehicle_config_dir, "/GlobalParamsOverride.yaml"],
         description = "Path to global file containing the parameters overwrite"
     )
 
--- a/carma/launch/carma_src.launch.py
+++ b/carma/launch/carma_src.launch.py
@@ -349,7 +349,8 @@ def generate_launch_description():
         executable='system_controller',
         parameters=[
                     system_controller_param_file,
-                    {"use_sim_time" : use_sim_time}],
+                    {"use_sim_time" : use_sim_time},
+                    global_params_override_file],
         on_exit = Shutdown(), # Mark the subsystem controller as required for segfaults
         arguments=['--ros-args', '--log-level', GetLogLevel('system_controller', env_log_levels)]
     )
--- a/carma/launch/carma_src.launch.py
+++ b/carma/launch/carma_src.launch.py
@@ -100,6 +100,15 @@ def generate_launch_description():
         description = "Path to file contain vehicle configuration parameters"
     )
 
+    # Declare the global_params_override_file launch argument
+    # Parameters in this file will override any parameters loaded in their respective packages
+    global_params_override_file = LaunchConfiguration('global_params_override_file')
+    declare_global_params_override_file_arg = DeclareLaunchArgument(
+        name = 'global_params_override_file',
+        default_value = [vehicle_config_dir, "/global_params_overwrite.yaml"],
+        description = "Path to global file containing the parameters overwrite"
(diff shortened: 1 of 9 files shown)


MishkaMN:
After this PR, I am planning on 
- creating another PR to modify each yaml files, so we can add a note that those parameters may get overriden by GlobalParamsOverride.yaml in the carma-config if launching from docker. And add PR description change to include checklist whether if the override yaml is included in the launch file of the node. 
- update rest of the driver repos and their launch files

carma/launch/drivers.launch.py
```diff
@@ -61,6 +61,22 @@ def generate_launch_description():
         description = "Path to file containing override parameters for the subsystem controller"
     )
 
+    vehicle_config_dir = LaunchConfiguration('vehicle_config_dir')
+    declare_vehicle_config_dir_arg = DeclareLaunchArgument(
+        name = 'vehicle_config_dir',
+        default_value = "/opt/carma/vehicle/config",
+        description = "Path to file containing vehicle config directories"
```
JonSmet: Nit for minor typo: Since this isn't a file, consider changing to "Path to vehicle configuration directory" or "Path to vehicle configuration directory populated by carma-config."

Same comment applies to the description in a few other locations of the PR.

carma/launch/carma_src.launch.py
```diff
@@ -245,6 +254,7 @@ def generate_launch_description():
                 PythonLaunchDescriptionSource([ThisLaunchFileDir(), '/environment.launch.py']),
                 launch_arguments = {
                     'subsystem_controller_param_file' : [vehicle_config_dir, '/SubsystemControllerParams.yaml'],
+                    'global_params_override_file' : global_params_override_file,
```
JonSmet, line 257: I'm trying to remember why we add this as a launch argument here if we still end up needing to declare the `global_params_override_file` launch argument in each of the downstream launch files (environment.launch.py, drivers.launch.py, etc.). Do you know if this is needed?

carma/launch/carma_src.launch.py
```diff
@@ -100,6 +100,15 @@ def generate_launch_description():
         description = "Path to file contain vehicle configuration parameters"
     )
 
+    # Declare the global_params_override_file launch argument
```
JonSmet, line 103: Just a comment for the PR description: Can we add a link to the Confluence document for background on the design? Additionally can we consider updating the PR name to "Add support for a global parameter override yaml file in carma-config" or similar?

MishkaMN: Added both

carma/launch/carma_src.launch.py
MishkaMN: Because from my understanding, each launch files are scoped only in themselves. 
DeclareLaunchArgument → "This parameter exists with these properties" (makes the launch file accept this parameter as an input). But this don't actually fill the variable with the value at runtime if we try to use it. ONLY relevant to the launch file it is in.
LaunchConfiguration → "Give me the actual runtime value of this parameter". This one actually makes it possible to use the value, either to pass it to another launch file or a node. ONLY relevant to the launch file it is in.

So we should typically need both.

carma/launch/drivers.launch.py
MishkaMN: Edited all

carma/launch/carma_src.launch.py
JonSmet: That makes sense and I do remember using them this way as well, thanks!

carma/launch/drivers.launch.py
```diff
@@ -61,6 +61,27 @@ def generate_launch_description():
         description = "Path to file containing override parameters for the subsystem controller"
     )
 
+    vehicle_config_dir = LaunchConfiguration('vehicle_config_dir')
+    declare_vehicle_config_dir_arg = DeclareLaunchArgument(
+        name = 'vehicle_config_dir',
+        default_value = "/opt/carma/vehicle/config",
+            vehicle_config_dir = LaunchConfiguration('vehicle_config_dir')
```
JonSmet: Looks like a small copy+paste error, an additional vehicle_config_dir and declare_vehicle_confi_dir_arg were added

JonSmet (approved):

PR #2634, opened 2025-10-10
update repos and fix bug

EricChen-Lei:
# PR Details
## Description

added pull of cdasim-config and fix bug in cloud-sim script
## Types of changes

## Checklist:

### DIFF
--- a/engineering_tools/xil-cloud-formation-v0.yaml
+++ b/engineering_tools/xil-cloud-formation-v0.yaml
@@ -459,8 +459,8 @@ Resources:
           sudo usermod -aG docker ubuntu
 
           # Install carma-config
-          # carma config install ${CarmaConfigImage}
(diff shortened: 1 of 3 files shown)


engineering_tools/xil-cloud-formation-v0.yaml
```diff
@@ -459,8 +459,8 @@ Resources:
           sudo usermod -aG docker ubuntu
 
           # Install carma-config
-          carma config install ${CarmaConfigImage}
-          carma config set ${CarmaConfigImage}
+          # carma config install ${CarmaConfigImage}
```
kjrush: Why are we removing this? Is this where the bug is? I thought the point of these lines was to automatically download the images while setting up the instance so the user doesn't need to do it themselves?

PR #2639, opened 2026-03-30
Improve Rviz Visualization

saina-ramyar:
# PR Details
## Description

1- Update tcm marker shapre based on geofence's type. Orange dots for workzone and blue vehicle as emergency vehicle.
The emergency vehicle mesh is included in carma_wm_ctrl/meshes and in carma_calibration_dir/visualization_meshes.
The parameter is over-written in environment.launch.py and the mesh from calibration_dir is used in rviz.

2- Some minor fixes to correctly extract labels.

workzone:
image

Emergency Vehicle (it is straightened and sits along the road)
image

## Motivation and Context

To make demo experience more understandable by having relevant graphics.

## How Has This Been Tested?
Tested with  Black Pacifica

## Types of changes

## Checklist:


carma/launch/environment.launch.py
```diff
@@ -314,6 +314,10 @@ def generate_launch_description():
                 ],
                 parameters=[carma_wm_ctrl_param_file,
                             vehicle_config_param_file,
+                            {'tim_icon_path': [
```
JonSmet, line 317: Why is this `tim_icon_path` needed here with the file in `vehicle_calibration_dir` if the .obj file is already included in carma_wm_ctrl/meshes?

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -1865,19 +1873,87 @@ carma_v2x_msgs::msg::TrafficControlRequestPolygon WMBroadcaster::composeTCRStatu
...
+    marker.color.g = 0.0F;
+    marker.color.b = 1.0F;
+    marker.color.a = 1.0F;
+
+    marker.scale.x = tim_icon_scale_;
+    marker.scale.y = tim_icon_scale_;
+    marker.scale.z = tim_icon_scale_;
+
+    size_t mid = input.size() / 2;
+    marker.pose.position.x = input[mid].x();
```
JonSmet, line 1921: [No code change needed] The actual cop car might be on the shoulder of the road, but this will place it in RVIZ on the nearby lanelet. I think this is fine, but it might be something to keep in mind when we present this in demos. Might be worth noting this in the PR description as well?

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -1865,19 +1873,87 @@ carma_v2x_msgs::msg::TrafficControlRequestPolygon WMBroadcaster::composeTCRStatu
...
+    marker.pose.position.y = input[mid].y();
+    marker.pose.position.z = 2.0;
+    marker.pose.orientation.x = 0.0;
+    marker.pose.orientation.y = 0.0;
+    marker.pose.orientation.z = -0.7071;
```
JonSmet, line 1929: Does this orientation only work for the current CP demo? Can we make it generic to point in the travel direction of the lane?

carma_wm_ctrl/src/WMBroadcaster.cpp
```diff
@@ -1865,19 +1873,87 @@ carma_v2x_msgs::msg::TrafficControlRequestPolygon WMBroadcaster::composeTCRStatu
...
+
+  if (label == "workzone")
+  {
+    marker.type = visualization_msgs::msg::Marker::SPHERE_LIST;
+
+    marker.scale.x = 2.0;
+    marker.scale.y = 2.0;
+    marker.scale.z = 2.0;
+
+    // Orange Color
```
JonSmet, line 1889: [No code change needed] I remember in the earlier discussions on this RVIZ update that we considered displaying traffic cones to make work zones geofences more obvious when visualized. I think this is good for demos with presenters but just curious if you think that would still be possible eventually or if there were issues with it?

plan_delegator/config/plan_delegator_params.yaml
```diff
@@ -37,4 +37,4 @@ duration_to_signal_before_lane_change: 2.5
 # Int: The maximum number of times plan_delegator attempts to generate a trajectory before giving up
 # When trajectory generations fails, it reuses the last successful trajectory until tries again
 # in the next iteration.
-max_traj_generation_reattempt: 10
+max_traj_generation_reattempt: 1000
```
JonSmet, line 42: Can we add a note or a TODO to describe why the large value is selected here and that it is due to an underlying issue?

carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcaster.hpp
```diff
@@ -461,6 +470,8 @@ class WMBroadcaster
    * It should not be incremented for updates that do not require a full map publication.
    */
   size_t current_map_version_ = 0;
+  std::string tim_icon_path_ = "package://carma_wm_ctrl/meshes/cop.obj";
+  double tim_icon_scale_ = 0.1;
```
JonSmet: Very minor nit: Consider using the same value as in config/parameters.yaml

carma_wm_ctrl/include/carma_wm_ctrl/WMBroadcasterConfig.hpp
```diff
@@ -36,6 +36,10 @@ namespace carma_wm_ctrl
     double config_limit = 6.67; //config speed limit in m/s
     std::string vehicle_id = "CARMA";
     std::string participant = "vehicle:car";
+    // Path to the model file for tim visualization
+    std::string tim_icon_path = "package://carma_wm_ctrl/meshes/cop.obj";
+    // Scale factor to apply to the tim icon model
+    double tim_icon_scale = 0.002;
```
JonSmet: Very minor nit: Consider using the same value as in config/parameters.yaml

carma/launch/environment.launch.py
saina-ramyar: to be honest I just followed the same approach as 
https://github.com/usdot-fhwa-stol/carma-platform/tree/develop/object_visualizer
But now after testing I believe having the mesh in vehicle_calibration_dir helps with switching icon.

carma_wm_ctrl/src/WMBroadcaster.cpp
saina-ramyar: You are correct, but the visualization is based on the generated geofence not the raw message from vehicle that includes its location. 
plus, this symbol shows why there is a road closure. I agree, something to have in mind during demo

plan_delegator/config/plan_delegator_params.yaml
saina-ramyar: This is a known plan_delegator issue where trajectories take long to generate, so callback fails. Happens mostly with multiple trajectories, such as lane change or yield.
Added comment.

carma_wm_ctrl/src/WMBroadcaster.cpp
saina-ramyar: I tried adding cones, but the view was top-down, so it did not look like a cone at all. The larger orange dots seem like a better option.
image

carma_wm_ctrl/src/WMBroadcaster.cpp
saina-ramyar: Not sure how to make it more generic without heading information

carma/launch/environment.launch.py
JonSmet: As mentioned in the PR description, we're keeping the duplicate mesh file in carma_wm_ctrl as well for cases when a user might build and test the package invidually/locally (rather than with all of CARMA Platform)

carma_wm_ctrl/src/WMBroadcaster.cpp
JonSmet: If based on feedback there is still a request to try to get traffic cones included, we can also consider 2D icons (liked the pedestrian icon) that show a traffic cone straight on, rather than a 3D traffic cone.

carma_wm_ctrl/src/WMBroadcaster.cpp
JonSmet: Discussed offline but we'll add this to the backlog to resolve after the May demonstration. We might be able to find the closest lanelet centerline point to this location, and then orient the marker in the travel direction of the centerline?

carma_wm_ctrl/src/WMBroadcaster.cpp
saina-ramyar: Rviz supports limited formats of mesh files, and most of files available online are not free. So that is difficult to find.

JonSmet (approved):

Issue #2316, opened 2024-02-20
yield_plugin can be faster by not considering objects from past route


MishkaMN:
### Summary

Yield_plugin can get really slow if there are multiple moving objects to check collision. Therefore, the logic should be made faster as much as possible.  
Currently first filter the yield_plugin does is to check whether if the objects will be on the vehicle's route. 
This check doesn't consider the cases where vehicle has passed the number of lanelets in the route. 
Therefore if some vehicle crosses the carma's past route it will still seriously consider it, which can slow. This applies to cases such as in Voices where multiples cars can follow the carma but will never be on its active route.

MishkaMN:
This was resolved on this: https://github.com/usdot-fhwa-stol/carma-platform/pull/2643

PR #2643, opened 2026-05-28
Yield Plugin: CUDA GPU Collision Detection with CPU Fallback

MishkaMN:
# PR Details

## Description
supporting PR: https://github.com/usdot-fhwa-stol/carma-config/pull/481

This PR, on a high-level, replaces the legacy per-object serial/async CPU collision check with a batched CUDA kernel
that processes all ego–obstacle segment pairs in parallel, while preserving the original
`std::async` + `get_collision_time` path as a CPU-only fallback for non-GPU deployments.
TLDR: Black pacifica testing showed that the plugin now generates a trajectory in max 0.02s with much better accuracy compared to 0.07s with CPU only and bad collision location accuracy. GPU one should scale well with more obstacles on the road as well.

Specific updates include:
---

### JMT trajectory generation optimization

- When recreating yield trajectory, the JMT algorithm resamples in every smallest_time_step timestep. However, this can be as small as 0.02s which is too high resolution and gave no accuracy benefit but caused O(1/dt) loop blow-up. Capped at 0.1s.

---

### check_traj_for_digital_min_gap unnecessary trajectory loop

- check_traj_for_digital_min_gap function checks for digiltal_min_gap regulatory element along the trajectory. When doing so, it traversed all trajectory points, get lanelets, get the regems. However, this was proved to be costly operation around 69-90ms, which is almost all the planning period the plugin has.
- Instead, used shortestPath ids directly which is same as trajectory to get regems which is much faster by skipping trajectory points altogether.

---

### On-Route Filter Optimization (both paths old CPU and new GPU methods)

- Replaced `getLaneletsFromPoint` (~163 µs spatial tree query per prediction point) with a shortest path bounding box checks. Bounding boxes are built once when the route is set and reused across all calls.
- The only remaining `getLaneletsFromPoint` call is in `check_traj_for_digital_min_gap`, which uses minimal amount to check the start and end of the lanelets on the shortest path. 
- Measured speedup of the on-route filter: **~163 µs/call → ~4 µs/call**. 
NOTE: During the test, 163 µs was measured at some point which was a large reason this new logic was implemented. However, since then, I am unable to recreate it. It is consistently fast with 4 µs/call. However, by theory, checking bounding boxes from route against a point should be slightly faster than getLaneletsFromPoint because under the hood, getLaneletsFromPoint does that anyways but also traverses R-tree which is unnecessary.
---

### CUDA Kernel — `get_collision_times_concurrently` GPU Path

- Analytically minimizes squared distance over every (ego segment, obstacle segment) pair with
  temporal overlap using a closed-form quadratic solve — exact for piecewise-linear motion, no
  stride-based misses.
- All objects are batched into a single `cuda_check_all_collisions` kernel call. Objects are
  pre-filtered by radius and route bbox before GPU upload to minimize data transfer.
- Targets sm_75/sm_80/sm_86 SASS; sm_89 (RTX 40xx / Ada Lovelace) supported via PTX JIT
  fallback since CUDA 11.7 (what carma-base currently uses) does not natively support sm_89 (CUDA 12.4+ compiles it natively, but correct strategy to support both spectra and sim pc should be discussed).
- End-to-end measured speedup vs. legacy `std::async` + `get_collision_time`: **~200× on RTX 4090**.

| PC | GPU | Architecture | Support |
| ---|---|---|---|
| Spectra PC | RTX 2080 | sm_75 (Turing) | Native SASS ✓ |
| Extra | A100 | sm_80 (Ampere) | Native SASS ✓ |
| Extra | RTX 3080/3090 | sm_86 (Ampere) | Native SASS ✓ |
| Sim PC 1&2 | RTX 4090 | sm_89 (Ada) | PTX JIT from sm_86 ✓ |

---

### CUDA Cold-Start Warmup

The first `cuda_check_all_collisions` call in a fresh process pays a one-time ~150–200 ms CUDA
runtime initialization cost (driver context creation, cubin loading, heap allocator setup). To
avoid this latency hitting the first real collision event during a mission, `on_configure_plugin`
now performs a dummy kernel launch with minimal input at node startup. The startup log confirms it:

```
[INFO] CUDA warmup complete (153 ms) — first call latency paid at startup
```

All subsequent calls take < 1 ms for the actual kernel work.

---

### CPU Fallback

- Preserved as-is: `std::async` + `get_collision_time` per object, now also using the bbox
  on-route filter.
- Despite compiling with CUDA (which is almost always successful because the underlying carma-base has it), at runtime it can fall back to CPU only mode for dev environments or CI/CD cases.

---

### Startup Log

Node logs which collision path is active at configure time:

```
[INFO] YieldPlugin collision detection: GPU path (bbox on-route filter + CUDA kernel)
[INFO] YieldPlugin collision detection: CPU path (bbox on-route filter + std::async get_collision_time)
```

---

### Debug Logging

- Added `[GPU]` / `[CPU]` prefixed `RCLCPP_DEBUG_STREAM` logs at every key decision point:
- Removed excessive debug logs that print trajectory which was causing up to 6s delay at some point and 0.3-0.4s delay frequently
---

### Tests — `test_collision_cuda_benchmark`

Three GTest cases:

| Test | What it verifies |
|---|---|
| `CollisionDetectionAccuracy.BetweenTimestampMiss` | CUDA finds a collision that falls between two consecutive prediction timestamps (old algorithm misses it) |
| `CollisionDetectionAccuracy.StrideSkipMiss` | CUDA finds a collision at an index skipped by the old stride logic |
| `CollisionDetectionBenchmark.WorstCasePerformance` | 100 objects × 50 states: times legacy CPU path vs new GPU path; cross-checks a planted collision (t ≈ 0.7 s) is detected by both |

Example output that investigated what was the bottleneck (suspected function `getLaneletsFromPoint` and later `routeTrackPos` all of which were technically fine). But found that, just adding CUDA improved it by a lot
```
[getLaneletsFromPoint micro-benchmark]
  Per call           : 4.38971 μs
  Predicted overhead : 3.51177 ms  (100 objects × ~8 calls each)

[WorstCasePerformanceBenchmark]
  Ego points     : 100
  Objects        : 100
  Pred states    : 50 per object
  Total seg pairs: 485100
  CPU (legacy)    : 20.9863 ms  (median of 3, std::async + get_collision_time)
  GPU (new path)  : 0.386837 ms  (median of 3, bbox filter + CUDA kernel)
  Speedup gpu/cpu : 54.2509×
  CPU collision   : no
  GPU collision   : no
```

## Related GitHub Issue
NA

## Related Jira Key
https://usdot-carma.atlassian.net/browse/CAR-6147
https://usdot-carma.atlassian.net/browse/CDAD-141

## Motivation and Context
Platform doesn't use GPU at all currently.

## How Has This Been Tested?
- Unit tests and made sure docker builds.
- Tested with actual TFHRC dummy pedestrian in simulation on simpc. And got following improvements:
```
2m means 2m collision distance
1st with CUDA 2m MAX: ExecutionTime Yield: 0.019861
2nd with CUDA 2m MAX: ExecutionTime Yield: 0.013156
3rd with CUDA 2m MAX: ExecutionTime Yield: 0.012081

logic before:
without CUDA 2m MAX: ExecutionTime Yield: 0.035870s
without CUDA 2m MAX: ExecutionTime Yield: 0.020495s
```
Then also did the real life vehicle tests.

- Black Pacifica showed max 0.02s trajectory generation which is pretty fast

### Old collision location (pls ignore the raw ROS msg), the red dot (just approximate locations and inaccurate):
This happens on the cpu case because in order to run fast, the collision detection logic extrapolated every trajectory segments into the future on a straight line to check if they collide, which can result in inaccurate locations albeit fast if objects are relatively moving straight. This helped speed, but created inaccurate collision locations (detecting collision way earlier etc).
image

### New collision location (pls ignore the raw ROS msg), the red dot (correct collision location):
image

## Types of changes

## Checklist

- [X] I have read the CONTRIBUTING document.


MishkaMN:
Tested on black Pacifica. Accuracy improved - which helps NOT to miss any collision intermittently which would have degraded lower control performance vehicles such as Black Pacifica. And runtime is max 0.02s for a collision, which is good because it is well under 0.1s planning period.

MishkaMN:
CI is only failing at the moment because of carma-msgs 2016 to 2024 update, which broke v2x-ros-conversion. yield_plugin still builds and tests pass.

yield_plugin/config/parameters.yaml
```diff
@@ -21,9 +21,10 @@ time_horizon_until_collision_to_commit_to_stop_in_s: 3.0
 # TODO: Further investigation required for the logic calling this parameter.
 consecutive_clearance_count_for_passed_obstacles_threshold: 20
 # Intervehicle distance that is considered a collision
+# From rear axle of the vehicle to the front of the obstacle
 # Unit: m
 # Value type: Desired
-intervehicle_collision_distance_in_m: 6.0
+intervehicle_collision_distance_in_m: 5.0
```
willjohnsonk: Is there context for this change?

yield_plugin/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 # the License.
 #
 
-cmake_minimum_required(VERSION 3.8)
+cmake_minimum_required(VERSION 3.18)
```
willjohnsonk, line 17: This is a pretty significant version change, do we know if there are any implications of this? I assume the CUDA dependencies require it and it is probably needed, just not sure if we have any legacy packages that could have issues.

yield_plugin/include/yield_plugin/yield_plugin.hpp
```diff
@@ -314,6 +330,8 @@ class YieldPlugin
   LaneChangeStatusCB lc_status_publisher_;
   std::shared_ptr<carma_ros2_utils::CarmaLifecycleNode> nh_;
   std::set<lanelet::Id> route_llt_ids_;
+  // Store lanelet polygons for precise geometric containment checking instead of just bounding boxes
```
willjohnsonk, line 363: Do we have any other Platform logic that uses the old bounding box method which could benefit from this enhancement?

yield_plugin/src/yield_plugin.cpp
```diff
@@ -529,9 +538,6 @@ namespace yield_plugin
       const double downtrack_at_target_time = polynomial_calc(polynomial_coefficients, target_time);
       double velocity_at_target_time = polynomial_calc_d(polynomial_coefficients, target_time);
 
-      RCLCPP_DEBUG_STREAM(rclcpp::get_logger("yield_plugin"), "Calculated speed velocity_at_target_time: " << velocity_at_target_time
```
willjohnsonk, line 532: Seems like we removed the excess debug streams to help performance but would like @JonSmet or someone to verify that the removed logs aren't valuable.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -871,51 +858,351 @@ namespace yield_plugin
...
-    const std::vector<carma_perception_msgs::msg::ExternalObject>& external_objects, double original_tp_max_speed)
+  static lanelet::BasicPoint2d interp_trajectory_pt_at_time(
+    double t,
+    const std::vector<CudaPoint>& pts,
+    int n,
+    const carma_planning_msgs::msg::TrajectoryPlan& tp)
   {
+    lanelet::BasicPoint2d result(pts[0].x, pts[0].y);
+    for (int i = 0; i < n - 1; ++i) {
+      const double ta = rclcpp::Time(tp.trajectory_points[i].target_time).seconds();
```
willjohnsonk: I think this block could generally use better variable names. I assume these are trajectory A & B but would one of these be the ego? Otherwise a comment may suffice.

yield_plugin/src/yield_plugin.cpp
```diff
@@ -871,51 +858,351 @@ namespace yield_plugin
...
+    int start_idx)
+  {
+    lanelet::BasicPoint2d result(
+      preds.front().predicted_position.position.x,
+      preds.front().predicted_position.position.y);
+    for (int j = start_idx; j < static_cast<int>(preds.size()) - 1; ++j) {
+      const double ta = rclcpp::Time(preds[j].header.stamp).seconds();
+      const double tb = rclcpp::Time(preds[j + 1].header.stamp).seconds();
+      if (ta <= t && t <= tb) {
+        const double s = (tb > ta) ? (t - ta) / (tb - ta) : 0.0;
```
willjohnsonk: Same comment as above re variable naming

yield_plugin/src/yield_plugin_cuda.cu
```diff
@@ -0,0 +1,283 @@
...
+  const float t_lo = fmaxf(t1a, t2a);
+  const float t_hi = fminf(t1b, t2b);
+  if (t_lo >= t_hi) return;
+
+  const float dt1 = t1b - t1a;
+  const float dt2 = t2b - t2a;
+  if (dt1 < 1e-6f || dt2 < 1e-6f) return;
+
+  // Velocities
+  const float vex = (x1b - x1a) / dt1;
```
willjohnsonk: I know this is common shorthand notation in CUDA especially but wondering if it is worth writing out variables for readability. Since it is mostly straightforward math I don't think we have to, just a thought. Other reviewers can add their thoughts

willjohnsonk (changes requested):

yield_plugin/config/parameters.yaml
MishkaMN: This was initially just to prove that CUDA can work with even smaller radius as 6 meter felt big.
However, thinking again, this is good because:
- from ego vehicle's back axle to front bumper is (usually a passenger vehicles like Fusion, Pacifica) 3-4 meters
- detected vehicle's midpoint, which almost always comes from object detection, is like 2-3 meters. 
Therefore, 6 meters is reasonable assuming the worst case scenario which is front bumper to bumper, a head-on collision. So reverting it back with the added comment.

yield_plugin/CMakeLists.txt
MishkaMN: CMAKE_CUDA_ARCHITECTURES was introduced in CMake 3.18.

As a tactical plugin type, no other plugins will be reusing code, so we should be okay. And it was confirmed to work in the docker environment that we use.

yield_plugin/include/yield_plugin/yield_plugin.hpp
MishkaMN: Possibly yeah. If anyone ever reads this convo in the future, I would search for usage of `getLaneletsFromPoint` and see if we can improve it by just searching from the route and their ids instead. However, it should be a different PR and different budget allocation.

yield_plugin/src/yield_plugin_cuda.cu
willjohnsonk: After talking the decision here seems to be keeping the code as-is and adding a readme to the plugin to help clarity.

yield_plugin/src/yield_plugin.cpp
MishkaMN: Yeah these debug logs print for every trajectory points (sometimes even another for loop just to print out the entire trajectory).

I don't know if anyone else ever looked at this code but these were useful during initial coding where trajectory generation needed debugging. But around this time, we used it enough to know it works, so removing these will help not incurring more unexpected latency during debugging. If the user wants to, they can add their specific debug.

yield_plugin/src/yield_plugin.cpp
MishkaMN: added different naming

yield_plugin/src/yield_plugin.cpp
MishkaMN: added different naming

yield_plugin/src/yield_plugin_cuda.cu
MishkaMN: added different naming

yield_plugin/include/yield_plugin/yield_plugin.hpp
```diff
@@ -295,6 +295,22 @@ class YieldPlugin
    */
   std::unordered_map<uint32_t, rclcpp::Time> get_collision_times_concurrently(const carma_planning_msgs::msg::TrajectoryPlan& original_tp, const std::vector<carma_perception_msgs::msg::ExternalObject>& external_objects, double original_tp_max_speed);
 
+private:
+  std::unordered_map<uint32_t, rclcpp::Time> get_collision_times_concurrently_cpu(
```
adev4a, line 313: nit: Can we add a description for these functions as well?

yield_plugin/include/yield_plugin/yield_plugin.hpp
adev4a: Can we add this enhancement to the backlog?

yield_plugin/src/yield_plugin_node.cpp
```diff
@@ -81,6 +83,23 @@ namespace yield_plugin
...
+      const std::vector<yield_plugin::CudaPoint> ego  = {{0.f,0.f,0.f},{1.f,0.f,1.f}};
+      const std::vector<yield_plugin::CudaPoint> obs  = {{1e4f,1e4f,0.f},{1e4f,1e4f,1.f}};
+      cuda_check_all_collisions(ego, obs, {0}, {2}, 2.0f);
+      const double _warm_ms = std::chrono::duration<double, std::milli>(
+        std::chrono::steady_clock::now() - _t0).count();
+      RCLCPP_INFO_STREAM(get_logger(),
+        "CUDA warmup complete (" << _warm_ms << " ms) — first call latency paid at startup");
+    } else {
+      RCLCPP_WARN(get_logger(), "YieldPlugin built with CUDA but no compatible driver found; using CPU collision detection fallback");
+      RCLCPP_WARN(get_logger(), "YieldPlugin collision detection: CPU path (bbox on-route filter + std::async get_collision_time)");
```
adev4a, line 100: nit: This can be INFO as well. Similar to the GPU case.

yield_plugin/README.md
```diff
@@ -0,0 +1,240 @@
...
+                   incoming_mobility_request    ----------+
+```
+
+---
+
+## Two Operating Modes
+
+### 1. Object Avoidance (default)
+
+Triggered whenever cooperative behavior is disabled or the urgency of an active cooperative lanechange request is below `acceptable_urgency`.
```
adev4a: If this is the default behavior can we reword this to being triggered by an object in external objects populated by sensor or v2x? Can we include references to where the cooperative behavior or urgency are set? Is the behavior in these cases a handoff of the planning to different tactical plugins?

yield_plugin/README.md
```diff
@@ -0,0 +1,240 @@
...
+
+**Pipeline:**
+
+1. **Concurrent collision detection** — for every external object in `/external_object_predictions` topic callback, `get_collision_times_concurrently()` is dispatched. There are either CPU or GPU approach. Default is GPU approach, but if there is any CUDA error (GPU driver not accessible etc), it falls back to CPU approach. See "Collision Detection: CPU vs. CUDA" for more.
+
+2. **Earliest collision selection** — the object with the smallest collision timestamp is chosen as the adversary.
```
adev4a: Confusing to read. Can we say the object with the closest time to collision?

yield_plugin/src/yield_plugin_node.cpp
MishkaMN: This should be WARN in my opinion because the intended and better version is the GPU version. If not, one should be notified in a more higher severity.

yield_plugin/include/yield_plugin/yield_plugin.hpp
MishkaMN: Added this improvement: https://usdot-carma.atlassian.net/browse/CAR-6149

yield_plugin/include/yield_plugin/yield_plugin.hpp
MishkaMN: Added

yield_plugin/README.md
MishkaMN: Renamed to that

yield_plugin/README.md
MishkaMN: No this is not a hand-off. I included the clarification in the new longer description. Please take a look at it.

yield_plugin/include/yield_plugin/yield_plugin_cuda.cuh
```diff
@@ -0,0 +1,67 @@
...
+
+namespace yield_plugin
+{
+
+// A single trajectory/prediction sample in flat SoA (structure of array) layout fed to the GPU.
+// Timestamps are normalized (t - reference_t) before upload so float32
+// precision is sufficient for the ~15-second planning horizon.
+struct CudaPoint
+{
+  float x, y;
```
adev4a: nit: I see the intention in keeping position and time separate. But can we add a comment saying something like Position (x,y) in meters for readability

yield_plugin/src/yield_plugin.cpp
```diff
@@ -520,7 +527,9 @@ namespace yield_plugin
     for (size_t i = 0; i < polynomial_coefficients.size(); i++) {
       RCLCPP_DEBUG_STREAM(rclcpp::get_logger("yield_plugin"),"Coefficient " << i << ": " << polynomial_coefficients[i]);
     }
-    const auto smallest_time_step = get_smallest_time_step_of_traj(original_tp);
+    // Cap at 0.1 s: finer steps gave no accuracy benefit but caused O(1/dt) loop blow-up
+    // when upstream planners produced sub-10 ms trajectory timesteps.
+    const double smallest_time_step = std::max(get_smallest_time_step_of_traj(original_tp), 0.1);
```
adev4a, line 532: I think a more complete check would be checking if get_smallest_time_step_of_traj(original_tp) is less than 0.1 before this, looks like it would be invalid if the trajectory is sampled at a lower rate although that should never be case

yield_plugin/src/yield_plugin.cpp
```diff
@@ -633,129 +621,128 @@ namespace yield_plugin
...
-        RCLCPP_DEBUG_STREAM(nh_->get_logger(), "p2b.x: " << p2b.predicted_position.position.x << ", p2b.y: " << p2b.predicted_position.position.y);
+        RCLCPP_DEBUG_STREAM(nh_->get_logger(), "object_seg_start.x: " << object_seg_start.predicted_position.position.x << ", object_seg_start.y: " << object_seg_start.predicted_position.position.y);
+        RCLCPP_DEBUG_STREAM(nh_->get_logger(), "object_seg_end.x: " << object_seg_end.predicted_position.position.x << ", object_seg_end.y: " << object_seg_end.predicted_position.position.y);
 
         // Linearly interpolate positions at a common timestamp for both trajectories
-        double dt = (p2a_t - p1a_t) / (p1b_t - p1a_t);
+        double interp_ratio = (object_seg_start_time - ego_seg_start_time) / (ego_seg_end_time - ego_seg_start_time);
         // if negative extrapolation, skip because car wouldn't go backwards
-        if (dt < 0)
+        if (interp_ratio < 0)
```
adev4a, line 722: Can we add a log print for this case?

yield_plugin/src/yield_plugin.cpp
```diff
@@ -859,63 +846,363 @@ namespace yield_plugin
...
+  }
+
+  std::pair<bool, int> YieldPlugin::find_on_route_in_predictions(
+    const std::vector<carma_perception_msgs::msg::PredictedState>& predictions,
+    int stride, bool zero_speed) const
+  {
+    for (size_t j = 0; j < predictions.size(); j += stride) {
+      const lanelet::BasicPoint2d point(predictions[j].predicted_position.position.x,
+                                        predictions[j].predicted_position.position.y);
+      for (const auto& llt : route_llt_polygons_) {
```
adev4a, line 917: route_llt_polygons_ is currently only updated when a object is detected (in get_earliest_collision_object_and_time), this should instead be updated using latest available route

yield_plugin/src/yield_plugin.cpp
```diff
@@ -859,63 +846,363 @@ namespace yield_plugin
...
+                     interp_ratio * (predictions[j+1].predicted_position.position.y - predictions[j].predicted_position.position.y);
+        break;
+      }
+    }
+    return result;
+  }
+
+  std::pair<bool, int> YieldPlugin::find_on_route_in_predictions(
+    const std::vector<carma_perception_msgs::msg::PredictedState>& predictions,
+    int stride, bool zero_speed) const
```
adev4a: nit: Can we rename variable to object_has_zero_speed

yield_plugin/src/yield_plugin.cpp
```diff
@@ -859,63 +846,363 @@ namespace yield_plugin
...
+  {
+    for (size_t j = 0; j < predictions.size(); j += stride) {
+      const lanelet::BasicPoint2d point(predictions[j].predicted_position.position.x,
+                                        predictions[j].predicted_position.position.y);
+      for (const auto& llt : route_llt_polygons_) {
+        if (boost::geometry::within(point, llt.polygon2d())) {
+          return {true, static_cast<int>(j)};
+        }
+      }
+      if (zero_speed) break;
```
adev4a: Why is zero speed important here? Should ego behavior change for a stationary object?

yield_plugin/src/yield_plugin.cpp
```diff
@@ -859,63 +846,363 @@ namespace yield_plugin
...
+      std::packaged_task<std::optional<rclcpp::Time>()> task(
+        [this, &original_tp, &object, &original_tp_max_speed] {
           return get_collision_time(original_tp, object, original_tp_max_speed);
         });
+      futures[object.id] = task.get_future();
+      threads.emplace_back(std::move(task));
     }
-
-    // Collect results from futures and update collision_times
+    for (auto& t : threads) t.join();
```
adev4a, line 956: It wont let me add a comment on the line below, but do we need this loop? Can we not add these logs to the get_collision_time function?

yield_plugin/src/yield_plugin.cpp
```diff
@@ -859,63 +846,363 @@ namespace yield_plugin
...
     }
 
-    // save route Ids for faster access
+    // save route lanelets for precise on-route checks using polygon containment
+    route_llt_polygons_.clear();
     for (const auto& llt: wm_->getRoute()->shortestPath())
     {
       // TODO: Enhancement https://github.com/usdot-fhwa-stol/carma-platform/issues/2316
       route_llt_ids_.insert(llt.id());
+      route_llt_polygons_.push_back(llt);
```
adev4a: Can we create a route event subscriber for this update instead?

adev4a:
Does this PR close https://github.com/usdot-fhwa-stol/carma-platform/issues/2316 ?

yield_plugin/include/yield_plugin/yield_plugin_cuda.cuh
MishkaMN: Added the comments

yield_plugin/src/yield_plugin.cpp
MishkaMN: if the trajectory is sampled at a lower rate (meaning if the dt is more than 0.1s), we would keep that rate. 
This is the intention of the logic because if it was valid for the upstream tactical_pluginm it should be valid here still.

Pls, lmk if I understood the comment correctly

yield_plugin/src/yield_plugin.cpp
MishkaMN: Added

yield_plugin/src/yield_plugin.cpp
MishkaMN: Good catch. I incorporated it for now using the setRouteCallback. 
If there are many other nodes that use this, it could be part of the world model itself, but probably it is actually few.

yield_plugin/src/yield_plugin.cpp
MishkaMN: renamed

yield_plugin/src/yield_plugin.cpp
MishkaMN: it is just an early exit strategy. If the object has a zero speed, all the remaining predicted points will also be the same, so just break early

MishkaMN:
> Does this PR close #2316 ?

Yes actually! Closed it

yield_plugin/src/yield_plugin.cpp
MishkaMN: you are right. it is an extra loop. removed and printed to the cpu function themselves

yield_plugin/src/yield_plugin.cpp
MishkaMN: added

yield_plugin/config/parameters.yaml
```diff
@@ -21,6 +21,9 @@ time_horizon_until_collision_to_commit_to_stop_in_s: 3.0
 # TODO: Further investigation required for the logic calling this parameter.
 consecutive_clearance_count_for_passed_obstacles_threshold: 20
 # Intervehicle distance that is considered a collision
+# Measured from the rear axle of the ego vehicle to the front of the obstacle
+# A typical passenger vehicle is about 3 meters long, so the worst-case
+# distance from the ego rear axle to the front of an oncoming vehicle is around 6 meters
```
JonSmet: I'm a little confused by this statement: "A typical passenger vehicle is about 3 meters long, so the worst-case distance from the ego rear axle to the front of an oncoming vehicle is around 6 meters" Can you elaborate on this?

If we use 6 meters here, then at that threshold the oncoming vehicle's front bumper will be 3 meters away from the front bumper of the ego vehicle (assuming vehicle lengths of 6 meters).

yield_plugin/README.md
```diff
@@ -0,0 +1,243 @@
+# Yield Plugin
+
+A CARMA Platform tactical plugin that modifies an existing trajectory to safely yield to obstacles and cooperating vehicles. It operates as a ROS 2 lifecycle node (`YieldPluginNode`) wrapping the worker class (`YieldPlugin`).
+
+---
+
+## Role in the Stack
+
+The yield plugin sits downstream of tactical plugins (e.g., inlane_cruising, cooperative_lanechange). It receives a *fully formed* trajectory from the planner and either passes it through unchanged or replaces the timestamps to slow/stop the vehicle before a predicted collision.
```
JonSmet: Nit: Can we change "inlane_cruising" to "inlanecruising_plugin" to match its package/directory name?

yield_plugin/README.md
```diff
@@ -0,0 +1,243 @@
...
+                   incoming_mobility_request    ----------+
+```
+
+---
+
+## Two Operating Modes
+
+### 1. Object Avoidance (default)
+
+Runs whenever the ego trajectory needs to be checked against external objects from `/external_object_predictions` (populated from onboard sensors and/or V2X msgs such as SDSM). This is the default path: it is skipped only when `enable_cooperative_behavior` is `true` (see [yield_config.hpp:37](include/yield_plugin/yield_config.hpp#L37)) *and* the urgency of the latest `MobilityRequest` (`clc_urgency_`, set in `mobilityrequest_cb`, see [yield_plugin.cpp:185](src/yield_plugin.cpp#L185)) exceeds `acceptable_urgency` (see [yield_config.hpp:45](include/yield_plugin/yield_config.hpp#L45)).
```
JonSmet: Since a user would set the configuration parameters (i.e., `enable_cooperative_behavior` and `acceptable_urgency`) in config/parameters.yaml, should we link to that file instead of yield_config.hpp?

yield_plugin/test/test_collision_cuda_benchmark.cpp
```diff
@@ -0,0 +1,628 @@
...
+//     t=0.0 → obs(10,10), ego(10, 0)   dist=10 > 2 m  no
+//     t=1.5 → obs(10,20), ego(10,15)   dist= 5 > 2 m  no
+//     t=3.5 → obs(10,30), ego(10,35)   dist= 5 > 2 m  no   → returns nullopt (MISS)
+//
+//   CUDA kernel: segment (obs 1.5→3.5) × (ego 2→3) overlaps [2.0, 3.0].
+//   Minimising dist²(s) gives s*=0.5 s → t=2.5 s → dist=0 → COLLISION DETECTED.
+
+TEST(CollisionDetectionAccuracy, BetweenTimestampMiss)
+{
+  if (!cuda_is_available()) {
```
JonSmet, line 172: [No change requested] I guess CUDA tests won't be checked during CI, so PRs that affect GPU-based collision avoidance will need to manually state the unit test status.

I can also reach out to Sai to see what actual pricing would look like for using GitHub Action runners that have a lower end GPU.

yield_plugin/config/parameters.yaml
MishkaMN: This was a typo. From rear to rear is what I meant

yield_plugin/README.md
MishkaMN: Renamed

yield_plugin/test/test_collision_cuda_benchmark.cpp
MishkaMN: That'd be great! Ty

yield_plugin/README.md
MishkaMN: Moved the reference to the parameter file

willjohnsonk (dismissed):

yield_plugin/src/yield_plugin.cpp
```diff
@@ -1522,4 +1514,21 @@ namespace yield_plugin
...
+    if (!wm_->getRoute())
+    {
+      RCLCPP_WARN(nh_->get_logger(), "update_route_llt_cache called but route is not available");
+      return;
+    }
+    route_llt_ids_.clear();
+    route_llt_polygons_.clear();
+    for (const auto& llt : wm_->getRoute()->shortestPath())
+    {
+      // TODO: Enhancement https://github.com/usdot-fhwa-stol/carma-platform/issues/2316
```
adev4a: The referenced issue here is closed, can we remove this comment?

adev4a (approved):

JonSmet (approved):

Issue #2648, opened 2026-07-21
Demo map has an issue during turning


MishkaMN:
### **Employee Status** (Select one)

Yes

### Organization Name

Leidos

### How Are You Using or Deploying This Tool?

Majority of the issue happened here was due to https://usdot-carma.atlassian.net/browse/CAR-6158

### Summary

TODO @MishkaMN 

### Version

4.11.0

### Expected Behavior

Should turn fine

### Actual Behavior

Not turning

### Steps to Reproduce the Actual Behavior

Use demo uc2 vip map

PR #2649, opened 2026-07-23
Fix Cooperative_lanechange logic that requires non-closed lanelet before lanechange

MishkaMN:
# PR Details
## Description
As described in the comment and attached powerpoint confluence page here, we previously had a limitation where lanelets just before those involved in the lanechange CANNOT be CLOSED: https://usdot-carma.atlassian.net/wiki/spaces/CRMPLT/pages/2877980673/TFHRC+Non-signalized+Work+Zone+Closed+Lane+Demo,

image

So we had to add a workaround to the map to add a small non-closed, but SOLID boundary lane that shouldn't be lanechanged:

image

As opposed to what the rest of the document says, the fact that plan_delegator adjusts the manaeuver start/end dist to account for front bumper is not the issue. The success of lanechange shouldn't be dictated by lanelets before or after the lanelets of those that are actually involved in the lanechange during routing (in other words, as long as the platform can route, we should be able to generate a trajectory). It is the over-reliance on the lanelet lookup is failing the trajectory generation.

Therefore, this code just replaces that lanelet lookup method and adds a direct geometrical generation of the trajectory since the actual lanelets invovled in the lanechange is fine.

## Related GitHub Issue
No existing issue found.

## Related Jira Key
https://github.com/usdot-fhwa-stol/carma-platform/issues/2648

## Motivation and Context
This issue was also re-emerged as a new problem during VIP demo at TFHRC using carma-system-4.11.0 where the CLC was not able to generate a lanechange because the lanelet before the lanechange did not have any neighbor lanelet (because the turns lanelets are created without any physically shared boundary lanelets). See below:

image

As you can see, the green trajectory cut short because the lanechange started (double white point lines). And that is because the lanelet where the green arrows are don’t have any shared lanelet. This is the same issue as documented in the above page’s attached powerpoint (just same issue happening before the lanelet as opposed to after)

## How Has This Been Tested?
Tested on the white pacifica on the same intersection and same TIM message broadcasted, so we are forced to lanechange right after the turn. Now it generates the trajectory fine:

image

This also fixed the original issue where we needed a small non-closed, solid boundary lanelet before the lanechange. I forcefully blocked the small workaround lanelet we added and blocked that as well using TIM (indicated by the green dots). This still works fine as shown below:

image

So now we don't need the extra small lanelet to lanechange.

## Types of changes

## Checklist:


MishkaMN:
Based off wrong base branch, so closing and recreating at  https://github.com/usdot-fhwa-stol/carma-platform/pull/2650

Issue #2113, opened 2023-06-02
A short/partial trajectory can published by Plan Delegator, resulting in a steering jerk if the vehicle reaches the end of that trajectory


JonSmet:
### Summary

During Emergency Response testing, it was found that a maneuver plan with a future lane change could cause a failure during trajectory planning for that lane change if the end of the lane follow trajectory was not within a configurable distance `starting_downtrack_range` of the start of the lane change trajectory. This issue should be examined further to identify what exactly causes the lane follow trajectory to not end at or near the beginning of the planned lane change trajectory.

Additionally, an approach should be developed at a minimum for CARMA to avoid following a short/partial trajectory that could result in the vehicle reaching the end of that trajectory while travelling at speed, which could result in steering jerk behavior.

It is important to note that AutonomousStuff provided updated SSC calibration parameters for the freightliners to avoid this steering jerk behavior in this scenario, but these updated parameters have not yet been tested as of June 2nd, 2023.

### Version

4.3.0 (Current)

### Expected Behavior

When a lane follow maneuver precedes a lane change maneuver, the generated lane follow trajectory should end at (or within a very short distance, i.e. less than a couple feet?) the beginning of the lane change maneuver.

Additionally, Plan Delegator should be updated to elegantly and safely handle the situation of a partially-generated trajectory that could result in the vehicle reaching the end of that trajectory while travelling at speed. Further discussion is required to determine whether the system should shutdown or attempt to re-plan in this situation.

### Actual Behavior

When a partial trajectory is generated and published due to a future lane change trajectory failing, it can be possible in certain circumstances for that trajectory to be short enough that the vehicle will reach the end of it before generated a new full trajectory. In this case, if the vehicle reaches the end of its trajectory while travelling at speed, a steering jerk can occur.

### Steps to Reproduce the Actual Behavior

1. Start CARMA with cooperative_lanechange's `starting_downtrack_range` parameter set to a large value (i.e. 20.0) to force this issue.
2. Engage CARMA on a route with a lane change
3. Verify that the lane change trajectory fails and the vehicle reaches the end of its partial trajectory. A steering jerk may or may not occur.

PR #2650, opened 2026-07-23
Fix Cooperative_lanechange logic that requires non-closed lanelet before lanechange

MishkaMN:
# PR Details
## Description
As described in the comment and attached powerpoint confluence page here, we previously had a limitation where lanelets just before those involved in the lanechange CANNOT be CLOSED: https://usdot-carma.atlassian.net/wiki/spaces/CRMPLT/pages/2877980673/TFHRC+Non-signalized+Work+Zone+Closed+Lane+Demo,

image

So we had to add a workaround to the map to add a small non-closed, but SOLID boundary lane that shouldn't be lanechanged:

image

As opposed to what the rest of the document says, the fact that plan_delegator adjusts the manaeuver start/end dist to account for front bumper is not the issue. The success of lanechange shouldn't be dictated by lanelets before or after the lanelets of those that are actually involved in the lanechange during routing (in other words, as long as the platform can route, we should be able to generate a trajectory). It is the over-reliance on the lanelet lookup is failing the trajectory generation.

Therefore, this code just replaces that lanelet lookup method and adds a direct geometrical generation of the trajectory (using only iterative before and following lanelets) and since the actual lanelets invovled in the lanechange is fine.

## Related GitHub Issue
No existing issue found.

## Related Jira Key
https://github.com/usdot-fhwa-stol/carma-platform/issues/2648

https://usdot-carma.atlassian.net/browse/CAR-6158

## Motivation and Context
This issue was also re-emerged as a new problem during VIP demo at TFHRC using carma-system-4.11.0 where the CLC was not able to generate a lanechange because the lanelet before the lanechange did not have any neighbor lanelet (because the turns lanelets are created without any physically shared boundary lanelets). See below:

image

As you can see, the green trajectory cut short because the lanechange started (double white point lines). And that is because the lanelet where the green arrows are don’t have any shared lanelet. This is the same issue as documented in the above page’s attached powerpoint (just same issue happening before the lanelet as opposed to after)

## How Has This Been Tested?
Tested on the white pacifica on the same intersection and same TIM message broadcasted, so we are forced to lanechange right after the turn. Now it generates the trajectory fine:

image

This also fixed the original issue where we needed a small non-closed, solid boundary lanelet before the lanechange. I forcefully blocked the small workaround lanelet we added and blocked that as well using TIM (indicated by the green dots). This still works fine as shown below:

image

So now we don't need the extra small lanelet to lanechange.

## Types of changes

## Checklist:


cooperative_lanechange/config/parameters.yaml
```diff
@@ -74,7 +74,7 @@ min_timestep : 0.1
 
 # Double:  The range from the start from which the plugin can start working
 # Units: meters
-starting_downtrack_range : 5.0
+starting_downtrack_range : 50.0
```
MishkaMN, line 77: I increased this because there is really no benefit in blocking the lanechange trajectory generation until the last 5 meters. The vehicle really should be able to generate a trajectory way ahead (it sort of makes debugging confusing). 

I searched the reason we had this parameter in the first place, only reference is here: https://github.com/usdot-fhwa-stol/carma-platform/issues/2113. I no longer encounter such shutdown mentioned in the issue, so it might be already fixed. OR, it could be that another [workaround is already applied in ILC to increase the trajectory when we encountered it during VIP demo](https://github.com/usdot-fhwa-stol/carma-config/blob/develop/demo_uc2_tim_cp/GlobalParamsOverride.yaml#L34C9-L34C31), which [we are fixing this sprint.](https://usdot-carma.atlassian.net/browse/CAR-6143)


Either way, increased this so that the CLC can plan ahead.

MishkaMN: This was also part of the reason why we have a large number here: https://github.com/usdot-fhwa-stol/carma-platform/blob/develop/plan_delegator/config/plan_delegator_params.yaml#L42. That parameter is supposed to notify us the issues when the trajectory generation fails. With this CLC parameter gone and yield_plugin recently being updated, we may be able to reduce the arbitrary number in the plan_delegator down. But I kept the large number until we actually remove this entire starting_downtrack_range  parameter and retest.

MishkaMN: Tracking in this issue: https://github.com/usdot-fhwa-stol/carma-platform/issues/2651

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -316,97 +316,180 @@ namespace basic_autonomy
             return constrained_points;
         }
 
-        std::vector<lanelet::BasicPoint2d> create_lanechange_geometry(lanelet::Id starting_lane_id, lanelet::Id ending_lane_id, double starting_downtrack, double ending_downtrack,
-                                                                   const carma_wm::WorldModelConstPtr &wm, int downsample_ratio, double buffer_ending_downtrack)
+        namespace
```
adev4a: nit: Can we move these three functions to the helper_functions.cpp instead of adding them under a namspace here? And declare them in the header?

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -316,97 +316,180 @@ namespace basic_autonomy
...
-                    is_lanechange_left = true;
-                    shared_boundary_found = true;
-                }
-
-                else if(current_lanelet.rightBound() == ending_lanelet.leftBound()){
-                    RCLCPP_DEBUG_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER), "Lanelet " << std::to_string(current_lanelet.id()) << " shares right boundary with " << std::to_string(ending_lanelet.id()));
-                    shared_boundary_found = true;
-                }
+            // Returns the total 2d arc length of a lanelet's centerline.
+            double lanelet_centerline_length(const lanelet::ConstLanelet& ll)
```
adev4a: nit: I can see the utility in wrapping this to a function for readability, can we rename this to get_lanalet_centerline_length?

adev4a: Also wondering, if this is better suited to the carma_wm?

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -316,97 +316,180 @@ namespace basic_autonomy
...
+            {
+                return carma_wm::geometry::compute_arc_lengths(ll.centerline2d().basicLineString()).back();
+            }
+
+            /**
+             * \brief Builds a centerline covering [pivot - backward_length, pivot + forward_length] by walking the
```
adev4a: nit: Update brief to include argument description

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -316,97 +316,180 @@ namespace basic_autonomy
...
+                    if (previous.empty())
                     {
-                        // Maneuver requires we travel further before completing lane change, but no routable lanelet directly ahead
-                        //In this case we have reached a lanelet which does not have a routable lanelet ahead + isn't adjacent to the lanelet where lane change ends
-                        //A lane change should have already happened at this point
-                        throw(std::invalid_argument("No following lanelets from current lanelet reachable without a lane change, incorrectly chosen end lanelet"));
+                        RCLCPP_WARN_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER),
+                            "create_lanechange_geometry: No routable predecessor lanelet found before lanelet "
+                            << chain.front().id() << " (possibly closed or missing from the map). Using the "
+                            << covered_back << "m of centerline that was reachable going backward.");
```
adev4a: Should this log be elevated to error? Since this is expected to result in no plan

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -316,97 +316,180 @@ namespace basic_autonomy
...
-                    if(current_lanelet.id() == starting_lanelet.id()){
-                        //Looped back to starting lanelet
-                        throw(std::invalid_argument("No lane change in path"));
+                    lanelet::ConstLanelet prev = previous.front();
+                    if (visited.count(prev.id()))
+                    {
+                        RCLCPP_WARN_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER),
+                            "create_lanechange_geometry: Detected a loop in lanelet connectivity before lanelet "
+                            << chain.front().id() << "; stopping centerline extension.");
+                        break;
```
adev4a: In this case can we not recursively keep going back in previous lanelets to complete the plan, in case the immediately prior lanelet is too small to cover the required distance?

adev4a: Elevate log to error if not?

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -316,97 +316,180 @@ namespace basic_autonomy
...
-                    if(wm->getMapRoutingGraph()->left(starting_lane[i])){
-                        curr_end_lanelet = wm->getMapRoutingGraph()->left(starting_lane[i]).get();
+                double covered_fwd = 0.0;
+                while (covered_fwd < forward_length)
+                {
+                    auto following = wm->getMapRoutingGraph()->following(chain.back(), false);
+                    if (following.empty())
+                    {
+                        RCLCPP_WARN_STREAM(rclcpp::get_logger(BASIC_AUTONOMY_LOGGER),
+                            "create_lanechange_geometry: No routable successor lanelet found after lanelet "
```
adev4a: Elevate log to Error

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -316,97 +316,180 @@ namespace basic_autonomy
...
+            lanelet::ConstLanelet starting_lanelet = wm->getMap()->laneletLayer.get(starting_lane_id);
+            lanelet::ConstLanelet ending_lanelet = wm->getMap()->laneletLayer.get(ending_lane_id);
+
+            double lane_change_length = ending_downtrack - starting_downtrack;
+
+            // Build the starting-lane (reference) and ending-lane (target) centerlines independently, each from its
+            // own lanelet's predecessor/successor chain. See build_chain_centerline's doc comment for why this is
+            // done instead of looking up "the lanelet adjacent to X" for every lanelet along the starting lane path.
+            std::vector<lanelet::BasicPoint2d> reference_centerline =
+                build_chain_centerline(wm, starting_lanelet, 0.0, lane_change_length + buffer_ending_downtrack);
```
adev4a, line 336: build_chain_centerline for starting lanelet is currently adding buffer_ending_downtrack to lane change length, the buffer should be the fourth argument

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -316,97 +316,180 @@ namespace basic_autonomy
...
+            std::vector<lanelet::BasicPoint2d> target_lane_centerline =
+                build_chain_centerline(wm, ending_lanelet, lane_change_length, buffer_ending_downtrack);
+
+            // If the map didn't have enough connected/routable lanelets to reach the required length (for example a
+            // closed lanelet blocked further routing), pad the centerlines out with a straight-line extrapolation
+            // instead of failing -- see extrapolate_to_length's doc comment.
+            extrapolate_to_length(reference_centerline, lane_change_length + buffer_ending_downtrack, "starting lanelet " + std::to_string(starting_lane_id) + "'s lane");
```
adev4a, line 343: Can we not fit a spline between the current centerline and the next instead? If not, a straight line assumption should ideally be constrained to a max length

MishkaMN: The problem is under this assumption, we don't have any next centerline. So I am not sure how to handle it

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -316,97 +316,180 @@ namespace basic_autonomy
...
+                                                                       double forward_length)
+            {
+                std::vector<lanelet::ConstLanelet> chain{pivot};
+                std::unordered_set<lanelet::Id> visited{pivot.id()};
 
-                else{
-                    //If there are no following lanelets on route, lanechange should be completing before reaching it
-                    if(wm->getMapRoutingGraph()->following(current_lanelet, false).empty())
+                double covered_back = lanelet_centerline_length(pivot);
+                while (covered_back < backward_length)
```
JonSmet: If this is supposed to cover "pivot - backward length" as stated in the docstring, should `covered_back` be initialized to 0 in the previous line?

basic_autonomy/src/basic_autonomy.cpp
```diff
@@ -316,97 +316,180 @@ namespace basic_autonomy
...
+             * commonly-satisfied requirement, so building each lane's geometry independently from its own chain
+             * (rather than by asking "what's next to lanelet X" for every lanelet in the other lane) is robust to
+             * those gaps.
+             *
+             * If the routing graph runs out of connected lanelets before reaching the requested length (e.g. a
+             * closed lanelet blocks further routing), this logs a warning and returns whatever centerline was
+             * actually reachable instead of throwing -- the caller is responsible for padding the result out if a
+             * minimum length is required (see extrapolate_to_length below).
+             */
+            std::vector<lanelet::BasicPoint2d> build_chain_centerline(const carma_wm::WorldModelConstPtr &wm,
```
JonSmet: Can we also define the arguments in the docstring here and in `extrapolate_to_length()` as well?

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: Moved them to helper functions

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: Renamed and moved it to carma_wm

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: Added argument docs

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: It won't necessarily won't result in no plan. The caller should be the one to handle it.

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: I guess maybe we can? But creating lanechchange geometry from a looped circle doesn't make sense and we should probably avoid it. Lanechange over a loop itself probably less likely to happen, if at all.

We should keep it WARN and not ERROR as it shouldn't stop the trajectory from being generated. It is common/or acceptable condition

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: We should keep it WARN and not ERROR as it shouldn't stop the trajectory from being generated. It is common/or acceptable condition

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: `+` is correct because starting_lanelet is usually the lanelet before the lanechange lanelets reported by the route due to the shift from plan_delegator. 

Also, I added an important doc fix - **the length is measured from the end of pivot for both.** 
So:

```
Suppose according to route, lanechange happens from A2 to B2. 
----------------------------------------------------------------------->>>>>> Direction of travel
Lane A
┌───────────┬────────────┬───────────┐
│    A1     │    A2      │    A3     │
└───────────┴------------┴───────────┘
      ▲S

starting_downtrack (inside A1, not A2! due to plan_delegator)
starting_lanelet = A1, so the `backward_length` is 0.0 for this, 
and `forward_length` is `lanechange_length + buffer`

Lane B
┌───────────┬------------┬───────────┐
│    B1     │    B2      │    B3     │
└───────────┴────────────┴───────────┘
                  ▲E
          ending_downtrack (inside/around B2)
          ending_lanelet = B2, so the `backward_length` is B2 
          (because the dist is measured from end of the pivot lanelet), 
          and `forward_length` is `buffer` which is in B3.
```

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: I fixed the docstring, which correctly indicates that both is accounting from the end of the pivot.

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: Added them

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: Added doc about it

basic_autonomy/include/basic_autonomy/helper_functions.hpp
```diff
@@ -111,5 +111,53 @@ namespace waypoint_generation
     int get_nearest_index_by_downtrack(const std::vector<lanelet::BasicPoint2d>& points, const carma_wm::WorldModelConstPtr& wm,
                                       const carma_planning_msgs::msg::VehicleState& state);
 
+    /**
+     * \brief Builds a centerline covering [pivot - backward_length, pivot + forward_length] by walking the
```
JonSmet: Nit: Can we change "pivot" here to "pivot end point" (or similar)?

MishkaMN: Addressed

basic_autonomy/src/basic_autonomy.cpp
adev4a: Good point.

basic_autonomy/src/basic_autonomy.cpp
MishkaMN: I have updated the doc with the limitation and increased the unit test to 99.9

basic_autonomy/include/basic_autonomy/helper_functions.hpp
```diff
@@ -111,5 +111,53 @@ namespace waypoint_generation
     int get_nearest_index_by_downtrack(const std::vector<lanelet::BasicPoint2d>& points, const carma_wm::WorldModelConstPtr& wm,
                                       const carma_planning_msgs::msg::VehicleState& state);
 
+    /**
+     * \brief Builds a centerline covering [pivot_end_point - backward_length, pivot_end_point + forward_length] by walking the
+     * pivot lanelet's own predecessor/successor chain in the routing graph.
+     *
+     * This deliberately avoids any left()/right()/adjacentLeft()/adjacentRight() lookups: those depend on
+     * lane-adjacency data being recorded for every lanelet visited, which can be missing or stripped for
+     * lanelets that a TrafficControlMessage has closed, or that simply were never linked in the source map.
```
adev4a: nit: Can we make the language here a little more generic to just a closed lanelet (for whatever reason - TCM or otherwise)

basic_autonomy/include/basic_autonomy/helper_functions.hpp
```diff
@@ -111,5 +111,53 @@ namespace waypoint_generation
...
+     * previous()/following() only require that the lanelet be routable, which is a much weaker and more
+     * commonly-satisfied requirement, so building each lane's geometry independently from its own chain
+     * (rather than by asking "what's next to lanelet X" for every lanelet in the other lane) is robust to
+     * those gaps.
+     *
+     * If the routing graph runs out of connected lanelets before reaching the requested length (e.g. a
```
adev4a, line 118: nit: Can we put this on top with the function description before we get into comparison with lookup using left/right/adjacency lookup

cooperative_lanechange/README.md
```diff
@@ -2,4 +2,6 @@
 
 The cooperative_lanechange tactical plugin is responsible for converting high-level lane change maneuvers into detailed lane change trajectories. This plugin generates lane change trajectories for two types of scenarios: (1) "Regular Planning" in which no vehicles are near the host vehicle in the target lane, and (2) "Negotiation" in which communication with a neighboring vehicle is required to open a gap for a safe lane change.
 
-Link to detailed design document on Confluence: [Click Here](https://usdot-carma.atlassian.net/wiki/spaces/CRMPLT/pages/1318715418/Detailed+Design+-+Cooperative+Lanechange)
\ No newline at end of file
+Please see the [small overview documentation](../basic_autonomy/README.md) of the implementation detail in basic_autonomy.
```
adev4a: nit: Can we reword to overview documentation instead of small overview.

adev4a (dismissed):

basic_autonomy/include/basic_autonomy/helper_functions.hpp
adev4a: Also thinking this is described in the readme, which may be the better place to explain design decisions

cooperative_lanechange/README.md
MishkaMN: Reworded

basic_autonomy/include/basic_autonomy/helper_functions.hpp
MishkaMN: Moved up

basic_autonomy/include/basic_autonomy/helper_functions.hpp
MishkaMN: Makes sense. Removed the discussion from here as it is already in the readme

adev4a (approved):

PR #2654, opened 2026-07-31
basic_autonomy logger improvement

MishkaMN:
# PR Details
## Description
Supports https://github.com/usdot-fhwa-stol/carma-utils/pull/260
Now the carma-config is using the new paradigm of log level setting.
Please see above PR for more detail of the changes

## Related Issue
See above

## Motivation and Context
The logging functionality has been long broken for the library code.

## How Has This Been Tested?
Local integration tested

## Types of changes

## Checklist:


.github/workflows/Ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
 
     # Set up a Docker container for the job
     container:
-      image: ${{ needs.determine_docker_org_and_tag.outputs.docker_org }}/autoware.ai:${{ needs.determine_docker_org_and_tag.outputs.docker_tag }}
+      image: ${{ needs.determine_docker_org_and_tag.outputs.docker_org }}/autoware.ai:develop
```
adev4a: Revert change

trajectory_follower_wrapper/src/trajectory_follower_wrapper_node.cpp
```diff
@@ -22,6 +23,7 @@ namespace trajectory_follower_wrapper
   TrajectoryFollowerWrapperNode::TrajectoryFollowerWrapperNode(const rclcpp::NodeOptions &options)
       : carma_guidance_plugins::ControlPlugin(options)
   {
+    basic_autonomy::set_logger(get_logger().get_child("basic_autonomy"));
```
adev4a, line 26: I probably missed this in another PR, why does basic_autonomy logger need to be explicitly set but carma_wm does not?

MishkaMN: Ah, I actually missed it on carma_wm. But basic_autonomy is what we want I think.
The main reason we do this is to enable the logging to be recorded in the rosout so we can access it from the .mcap. It proved to be useful when debugging. Otherwise any static loggers like rclcpp::get_logget("basic_autonomy") doesn't get saved to the rosout. Only node loggers get saved.

I have to apply that change to carma_wm side

MishkaMN: I am working on making the carma_wm logger's logs to be saved in /rosout but it will be a bigger effort, so separated it into a different story: https://usdot-carma.atlassian.net/browse/CAR-6160

.github/workflows/Ci.yml
MishkaMN: Removed, but CI will fail but its okay. We can merged the earlier ones and verify then

adev4a (dismissed):

adev4a (approved):

PR #2653, opened 2026-07-31
carma_wm logger improvement

MishkaMN:
# PR Details
## Description
Supports https://github.com/usdot-fhwa-stol/carma-utils/pull/260
Now the carma-config is using the new paradigm of log level setting.
Please see above PR for more detail of the changes

## Related Issue
See above

## Motivation and Context
The logging functionality has been long broken for the library code.

## How Has This Been Tested?
Local integration tested

## Types of changes

- [X] Defect fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that cause existing functionality to change)

## Checklist:

- [X] I have added any new packages to the sonar-scanner.properties file
- [ ] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
[CARMA Contributing Guide](https://github.com/usdot-fhwa-stol/carma-platform/blob/develop/Contributing.md) 
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

PR #2656, opened 2026-08-05
Refactor yield_plugin to be called by plan_delegator only

MishkaMN:
# PR Details
## Description
Currently, yield_plugin is called in every tactical plugin to verify whether if the generated trajectory is safe or not.
However, usually if an obstacle is present it just changes the trajectory for the earliest plugin so the rest of the plugin's generated trajectory becomes OBE.

Therefore, it is much efficient to just run the safety plugin, the yield_plugin, after all tactical plugin's have generated their trajectory. This PR removes all the yield_plugin calls and moves it to the plan_delegator.

NOTE: some of the plugin's call to yield_plugin fails anyways such as CLC

## Related GitHub Issue
NA

## Related Jira Key
https://usdot-carma.atlassian.net/browse/HASS-673

## Motivation and Context
Preparing for HASS DT WZ use case

## How Has This Been Tested?
Local tested: As you can see the lanechange part is now yielding (turning gray).
image

## Types of changes

## Checklist:


plan_delegator/config/plan_delegator_params.yaml
```diff
@@ -34,6 +34,12 @@ duration_to_signal_before_lane_change: 2.5
 # Configured in VehicleConfigPrams.yaml in carma-config
 # tactical_plugin_service_call_timeout: 100
 
+# Bool: Activates object avoidance logic - plan_delegator calls yield_plugin on the final trajectory
+# before publishing it
+# Units: N/a
+# Configured in VehicleConfigParams.yaml in carma-config
+# enable_object_avoidance: true
```
adev4a: Lets just set this to false by default instead of commenting it out

adev4a:
A unit test is failing but seems to be failing on other PRs as well. Seems like no changes we were affecting that unit test, in this PR

MishkaMN:
> A unit test is failing but seems to be failing on other PRs as well. Seems like no changes we were affecting that unit test, in this PR

Yes the unit test is failing due to some unit test timing out. I have the option to increase it on this PR: https://github.com/usdot-fhwa-stol/carma-platform/pull/2653, but I can include it in this PR as well

plan_delegator/config/plan_delegator_params.yaml
MishkaMN: Actually we want to have it on all the time. So I uncommented, but reworded to clearly state that it is overriden by a different config

adev4a (approved):

MishkaMN:
Currently, `route` unit test is failing which is unrelated. This is due to some timeout issue. Fixing in a different PR:

image

